PR #53, opened 2020-07-27
Detach trait

zzxwill:
detach trait from an application


zzxwill:
Closed it and recreate another PR as my forked repo lost connection after `cloud-native-application/RudrX` was made public.

PR #70, opened 2020-07-31
refactor rudr run to rudr <workload>:run

wonderflow:


api/types/template_types.go
```diff
@@ -14,30 +14,31 @@ See the License for the specific language governing permissions and
...
-// TemplateSpec defines the desired state of Template
+// Template defines the content of a plugin
 type Template struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-	Alias            string                    `json:"alias,omitempty"`
-	Object           unstructured.Unstructured `json:"object,omitempty"`
-	LastCommandParam string                    `json:"lastCommandParam,omitempty"`
-	Parameters       []Parameter               `json:"parameters,omitempty"`
+	Name       string                 `json:"name"`
```
zzxwill, line 28: Is template name the definition name?

api/types/template_types.go
```diff
@@ -14,30 +14,31 @@ See the License for the specific language governing permissions and
...
+	Object     map[string]interface{} `json:"object,omitempty"`
+	Parameters []Parameter            `json:"parameters,omitempty"`
 }
 
+type DefinitionType string
+
+const (
+	TypeWorkload DefinitionType = "workload"
+	TypeTrait    DefinitionType = "trait"
+)
```
zzxwill, line 40: We can add scope here later as we added the spec.extension for ScopeDefinition in OAM kubernetes runtime.

api/types/template_types.go
wonderflow: yes

api/types/template_types.go
wonderflow: yes

PR #76, opened 2020-08-03
implement `rudr status`

zzxwill:
Implement `rudr status <APPLICATION-NAME>` to show the status
of the application, its workloads and traits


pkg/cmd/util/helpers.go
```diff
@@ -149,6 +151,34 @@ func RetrieveApplicationsByName(ctx context.Context, c client.Client, applicatio
 	return applicationMetaList, nil
 }
 
+func RetrieveApplicationStatusByName(ctx context.Context, c client.Client, applicationName string, namespace string) ApplicationStatusMeta {
+	var applicationStatusMeta ApplicationStatusMeta
+	var appConfig corev1alpha2.ApplicationConfiguration
+	if err := c.Get(ctx, client.ObjectKey{Name: applicationName, Namespace: namespace}, &appConfig); err == nil {
```
wonderflow, line 157: We usually check `err!=nil` first. Then all these logic could be written out of this brace

pkg/cmd/util/helpers.go
```diff
@@ -149,6 +151,34 @@ func RetrieveApplicationsByName(ctx context.Context, c client.Client, applicatio
...
+
+		if component, err := GetComponent(ctx, c, componentName, namespace); err == nil {
+			var workload corev1alpha2.WorkloadDefinition
+			json.Unmarshal(component.Spec.Workload.Raw, &workload)
+
+			var trait corev1alpha2.TraitDefinition
+			traitDefinitionList := ListTraitDefinitionsByApplicationConfiguration(appConfig)
+			if len(traitDefinitionList) != 0 {
+				trait = traitDefinitionList[0]
+			}
```
wonderflow, line 169: why only check the first one of slice?

pkg/cmd/status.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+
+	cmdutil "github.com/cloud-native-application/rudrx/pkg/cmd/util"
+
+	"github.com/ghodss/yaml"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewAppStatusCommand(f cmdutil.Factory, c client.Client, ioStreams cmdutil.IOStreams) *cobra.Command {
```
wonderflow, line 14: I think you could try to write a unit test for this function, then you will understand why I suggest use `RunE`

zzxwill: Okay, let me fix workload_run_test.go and trait_bind_test.go first.

pkg/cmd/status.go
```diff
@@ -0,0 +1,51 @@
...
+	ctx := context.Background()
+	cmd := &cobra.Command{
+		Use:     "status",
+		Short:   "get status of an application",
+		Long:    "get status of an application, including its workload and trait",
+		Example: `rudr status <APPLICATION-NAME>`,
+		Run: func(cmd *cobra.Command, args []string) {
+			argsLength := len(args)
+			if argsLength == 0 {
+				cmdutil.PrintErrorMessage("Hint: please specify an application", 1)
```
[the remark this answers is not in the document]
zzxwill: Done.

pkg/cmd/util/helpers.go
```diff
@@ -149,6 +151,34 @@ func RetrieveApplicationsByName(ctx context.Context, c client.Client, applicatio
...
+func RetrieveApplicationStatusByName(ctx context.Context, c client.Client, applicationName string, namespace string) ApplicationStatusMeta {
+	var applicationStatusMeta ApplicationStatusMeta
+	var appConfig corev1alpha2.ApplicationConfiguration
+	if err := c.Get(ctx, client.ObjectKey{Name: applicationName, Namespace: namespace}, &appConfig); err == nil {
+
+		componentName := appConfig.Spec.Components[0].ComponentName
+
+		if component, err := GetComponent(ctx, c, componentName, namespace); err == nil {
+			var workload corev1alpha2.WorkloadDefinition
+			json.Unmarshal(component.Spec.Workload.Raw, &workload)
```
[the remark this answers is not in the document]
zzxwill: Done.

pkg/cmd/status.go
```diff
@@ -0,0 +1,51 @@
...
+	application := cmdutil.RetrieveApplicationStatusByName(ctx, c, appName, namespace)
+
+	// TODO(zzxwill) When application.Trait.Name is "", find a way not to print trait status
+	out, err := yaml.Marshal(application)
+
+	if err != nil {
+		errMsg := fmt.Sprintf("Error: priting workload status hit an issue:%s", err)
+		cmdutil.PrintErrorMessage(errMsg, 1)
+	}
+	fmt.Print(string(out))
```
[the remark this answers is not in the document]
zzxwill: Done.

pkg/cmd/util/helpers.go
zzxwill: Done. Instead I have to write `return` several times, but as you said, the logic seems more clear.

pkg/cmd/util/helpers.go
zzxwill: Fixed, one workload can be attached several traits.

pkg/cmd/status.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func printApplicationStatus(ctx context.Context, c client.Client, ioStreams cmdutil.IOStreams, appName string, namespace string) error {
+	application, err := cmdutil.RetrieveApplicationStatusByName(ctx, c, appName, namespace)
+	if err != nil {
+		return err
+	}
+
+	// TODO(zzxwill) When application.Trait.Name is "", find a way not to print trait status
+	out, err := yaml.Marshal(application)
```
wonderflow, line 49: why don't check error here?

pkg/cmd/util/helpers.go
```diff
@@ -149,6 +151,33 @@ func RetrieveApplicationsByName(ctx context.Context, c client.Client, applicatio
...
 }
 
+func RetrieveApplicationStatusByName(ctx context.Context, c client.Client, applicationName string, namespace string) (ApplicationStatusMeta, error) {
+	var applicationStatusMeta ApplicationStatusMeta
+	var appConfig corev1alpha2.ApplicationConfiguration
+	if err := c.Get(ctx, client.ObjectKey{Name: applicationName, Namespace: namespace}, &appConfig); err != nil {
+		return applicationStatusMeta, err
+	}
+
+	component, err := GetComponent(ctx, c, appConfig.Spec.Components[0].ComponentName, namespace)
```
wonderflow, line 161: We should list component with componentName == applicationName here, instead of the first one of index.

pkg/cmd/util/helpers.go
```diff
@@ -149,6 +151,33 @@ func RetrieveApplicationsByName(ctx context.Context, c client.Client, applicatio
...
+
+	var workload corev1alpha2.WorkloadDefinition
+	err = json.Unmarshal(component.Spec.Workload.Raw, &workload)
+	if err != nil {
+		return applicationStatusMeta, err
+	}
+
+	traitDefinitionList := ListTraitDefinitionsByApplicationConfiguration(appConfig)
+	applicationStatusMeta = ApplicationStatusMeta{
+		Status:   string(appConfig.Status.Conditions[0].Status),
```
wonderflow, line 174: same with above

pkg/cmd/status.go
zzxwill: Done

pkg/cmd/util/helpers.go
zzxwill: In the currently design, there is only one component in the application, so `appConfig.Spec.Components[0].ComponentName` is equal to `applicationName`. But as you said, I should use `applicationName` here instead of `appConfig.Spec.Components[0].ComponentName`. Thanks.

pkg/cmd/util/helpers.go
zzxwill: As for the `status` field, how can I get the status of the appConfig?  I also need to see more cases which have multiple `conditions` field as I couldn't figure out from the struct https://github.com/crossplane/crossplane-runtime/blob/master/apis/core/v1alpha1/condition.go#L76
```
status:
  conditions:
  - lastTransitionTime: "2020-08-04T13:04:24Z"
    message: 'cannot render components: cannot find trait definition "core.oam.dev/v1alpha2"
      "ManualScalerTrait" "manualscaler": TraitDefinition.core.oam.dev "manualscalertraits.core.oam.dev"
      not found'
    reason: Encountered an error during resource reconciliation
    status: "False"
    type: Synced
  dependency: {}
```

PR #86, opened 2020-08-04
init cue dsl function

wonderflow:
use cue template instead of parameters


pkg/utils/system/system.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/cloud-native-application/rudrx/api/types"
+)
+
+const RudrHome = ".rudr"
```
hongchaodeng, line 12: rudr -> oam?

pkg/utils/system/system.go
```diff
@@ -0,0 +1,90 @@
...
+
+func GetHomeDir() (string, error) {
```
hongchaodeng, line 14: GetRudrHomeDir()

pkg/utils/system/system.go
hongchaodeng, line 12: This is not used externally. Should make it private.

pkg/cue/testdata/apps/myapp.yaml
```diff
@@ -0,0 +1,32 @@
...
+        ports:
+          http:
+            service_port: 80
+            container_port: '${echo_server_port}'
+    scopes:
+      - public-scope
+secrets:
+  secret-foo:
+    key1: '{{ config1 }}'
+    key2: '{{ config2 }'
```
hongchaodeng, line 27:
```suggestion
    key2: '{{ config2 }}'
```

pkg/cue/testdata/apps/myapp.yaml
```diff
@@ -0,0 +1,32 @@
+name: myapp
```
hongchaodeng, line 1: Where is this yaml used?

config/samples/workloads/deployment_definition.yaml
```diff
@@ -7,29 +7,31 @@ spec:
...
-        type: string
-        fieldPaths:
-          - "spec.containers[0].image"
-      - name: port
-        short: p
-        required: false
-        type: int
-        fieldPaths:
-          - "spec.containers[0].ports[0].containerPort"
+    template: |
```
hongchaodeng, line 10: It is better to put cue template into a separate file.
We can use cli to specify the .cue file for a definition.

pkg/cue/testdata/apps/myapp.yaml
wonderflow: just example now

config/samples/workloads/deployment_definition.yaml
wonderflow: next PR

pkg/utils/system/system.go
wonderflow: agree

pkg/utils/system/system.go
wonderflow: agree

Issue #68, opened 2020-07-31
[BUG]Workload's complete information cannot be displayed in output fields of traits.


baikai:
## Description
Command "rudr traits" field "APPLIES TO"  should use workload's short name. Otherwise, the workload's complete info  cannot displayed.

## How to reproduce
1 Run command "rudr traits" in terminal
2 Check command output

## Expected behavior
Workload's complete information can be displayed.

## Screenshots
<img width="1170" alt="Screen Shot 2020-07-31 at 11 35 15 AM" src="https://user-images.githubusercontent.com/3839050/88997329-038ad380-d322-11ea-8187-b38a498d614a.png">

## Cluster information
Describe your kubernetes cluster information.
- Kubernetes Version 1.16.6

baikai:
Fixed by commit bcf758edfc2443ad3fd3224d84ec84bb083b9dc5

PR #87, opened 2020-08-05
Fix overlap of workloads in column `APPLIES TO` for command `rudr traits`

baikai:
fix issue #68 by split workload name into multiple lines.


zzxwill:
Will you consider replacing workload to its alias? For example, replace `core.oam.dev/v1alpha2.ContainerizedWorkload` with `containerized`. This can also shorten the length of column `APPLIES TO`.

wonderflow:
what will show? Can you print an example

baikai:
cmd output like below:
Screen Shot 2020-08-05 at 8 01 37 PM

PR #98, opened 2020-08-06
RESTful Draft

ryanzhang-oss:
Added the appConfig and xxxDefinition REST API draft. Let's see if the style is ok

### DIFF
--- a/design/restful-interface
+++ b/design/restful-interface
@@ -1,0 +1,195 @@
+# Vela RESTful API design
+
+Below, I will list the Vela backend RESTful API grouped by each resource.
+
+We will set RESTful version as `APIVersionV1 = "2020-08-15"`
+
+The first version does not use authentication or authorization. We will use go structure to 
+represent request and response body.
+
+We will define some common header keys like "x-oam-client-type", "x-oam-username" but it's not fleshed out yet.
+
+## ApplicationConfiguration related API
+
+### Create 
+**URL** : `/APIVersionV1/appconfig/`
+
+**Method** : `POST`
+
+**Query Parameter** : `None`
+
+**Body** :
+
+```go
+
+type appConfigValue struct {
+    parameter interface{}[] `json:"parameter"` // the parameter list, not sure if they are all string
+    definitionName string `json:"definitionName"` // use to find the definition
+    definitionType string `json:"definitionType"`
+}
+```
+
+**Responses Body** : None
+
+
+### Update 
+**URL** : `/APIVersionV1/appconfig/${appConfigName}`
+
+**Method** : `PUT`
+
+**Query Parameter** : `None`
+
+**Body** :
+
+```go
+
+type appConfigValue struct {
+    parameter interface{}[] `json:"parameter"` // the parameter list, not sure if they are all string
+}
+```
+
+**Responses Body** : None
+
+### Get
+
+**URL** : `/APIVersionV1/appconfig/${appconfigName}`
+
+**Method** : `Get`
+
+**Query Parameter** : `mode` 
+```go
+type mode string
+const fullcontent mode = "fullContent"
+const parameterOnly mode = "parameterOnly"
+const statusOnly mode = "statusOnly"
+```
+**Body** : None
+
+**Responses Body** : 
+```go
+type appConfigValue struct {
+    parameterName string[] `json:"parameterName"` // the parameterName list
+    definition runtime.RawExtension `json:"definition"` // the entire appConfig definition
+    status runtime.RawExtension `json:"status"` // the status only part
+}
+```
+
+### List
+
+**URL** : `/APIVersionV1/appconfig/`
+
+**Method** : `Get`
+
+**Query Parameter** : `None`
+
+**Body** : None
+
+**Responses Body** : 
+```go
+type appConfigNames struct {
+    appConfigName string[] `json:"appConfigName"` // the appconfig name list, no next Token
+}
+```
+
+### Delete
+
+**URL** : `/APIVersionV1/appconfig/${appconfigName}`
+
+**Method** : `Delete`
+
+**Query Parameter** : `dryrun` 
+
+**Body** : None
+
+**Responses Body** : None 
+
+## Definition related API
+
+We will add ${definitionType} in the path to differentiate if it's a workload/trait/scope definition.
+It can be either "workload","trait", or "scope".
+ 
+### Create 
+
+**URL** : `/APIVersionV1/${definitionType}/`
+
+**Method** : `POST`
+
+**Query Parameter** : `None`
+
+**Body** :
+
+```go
+
+type oamDefinition struct {
+    definition runtime.RawExtension `json:"definition"` // the real definition data
+}
+```
+
+### Update 
+
+**URL** : `/APIVersionV1/${definitionType}/${definitionName}`
+
+**Method** : `PUT`
+
+**Query Parameter** : `None`
+
+**Body** :
+
+```go
+
+type oamDefinition struct {
+    definition runtime.RawExtension `json:"definition"` // the real definition data
+}
+```
+
+**Responses Body** : None
+
+### GET
+
+**URL** : `/APIVersionV1/${definitionType}/${definitionName}`
+
+**Method** : `Get`
+
+**Query Parameter** : `mode` 
+```go
+type mode string
+const fullcontent mode = "fullContent"
+const cueTemplate mode = "cueTemplate"
+```
+**Body** : None
+
+**Responses Body** : 
+```go
+type definition struct {
+    definition runtime.RawExtension `json:"definition"` // either the full definition or the cueTempalte
+}
+```
+
+### List
+
+**URL** : `/APIVersionV1/${definitionType}/`
+
+**Method** : `Get`
+
+**Query Parameter** : `None`
+
+**Body** : None
+
+**Responses Body** : 
+```go
+type definitionName struct {
+    definitionNames string[] `json:"definitionNames"` // the definition name list, no next Token
+}
+```
+
+### Delete
+
+**URL** : `/APIVersionV1/${definitionType}/${definitionName}`
+
+**Method** : `Delete`
+
+**Query Parameter** : `dryrun` 
+
+**Body** : None
+
+**Responses Body** : None 
--- a/design/restful-interface
+++ b/design/restful-interface
@@ -1,18 +1,63 @@
 # Vela RESTful API design
 
-Below, I will list the Vela backend RESTful API grouped by each resource.
-
-We will set RESTful version as `APIVersionV1 = "2020-08-15"`
+Below, I list the Vela backend RESTful API grouped by each resource.
 
 The first version does not use authentication or authorization. We will use go structure to 
 represent request and response body.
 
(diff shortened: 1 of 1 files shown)


design/restful-interface
```diff
@@ -0,0 +1,195 @@
+# Vela RESTful API design
+
+Below, I will list the Vela backend RESTful API grouped by each resource.
+
+We will set RESTful version as `APIVersionV1 = "2020-08-15"`
```
wonderflow, line 5: I suggest use `v1` instead of `2020-08-15`

design/restful-interface
```diff
@@ -0,0 +1,195 @@
...
+**Method** : `POST`
+
+**Query Parameter** : `None`
+
+**Body** :
+
+```go
+
+type oamDefinition struct {
+    definition runtime.RawExtension `json:"definition"` // the real definition data
```
wonderflow, line 124:
```suggestion
    definition runtime.RawExtension `json:"definition"` // the real definition data
    definitionType string `json:"type"`
```

design/restful-interface
```diff
@@ -0,0 +1,195 @@
...
+**Method** : `PUT`
+
+**Query Parameter** : `None`
+
+**Body** :
+
+```go
+
+type oamDefinition struct {
+    definition runtime.RawExtension `json:"definition"` // the real definition data
```
wonderflow, line 141:
```suggestion
    definition runtime.RawExtension `json:"definition"` // the real definition data
    definitionType string `json:"type"`
```

design/restful-interface
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+**Responses Body** : None
+
+### GET
+
+**URL** : `/APIVersionV1/${definitionType}/${definitionName}`
```
wonderflow, line 149:
```suggestion
**URL** : `/APIVersionV1/definitions/${definitionName}`
```

Assume definitionName is globally unique

design/restful-interface
wonderflow: In fact if we put the whole Definition CRD here, it will contain definitionType itself:

```
kind: TraitDefinition
spec:
  definitionRef: xx
  extentions:
      xxx
```

design/restful-interface
```diff
@@ -0,0 +1,195 @@
...
+**Responses Body** : 
+```go
+type definition struct {
+    definition runtime.RawExtension `json:"definition"` // either the full definition or the cueTempalte
+}
+```
+
+### List
+
+**URL** : `/APIVersionV1/${definitionType}/`
```
wonderflow, line 170:
```suggestion
**URL** : `/APIVersionV1/definitions?type=<workload/trait>` 
```

default for all types

design/restful-interface
```diff
@@ -0,0 +1,195 @@
...
+**Responses Body** : 
+```go
+type definitionName struct {
+    definitionNames string[] `json:"definitionNames"` // the definition name list, no next Token
+}
+```
+
+### Delete
+
+**URL** : `/APIVersionV1/${definitionType}/${definitionName}`
```
wonderflow, line 187:
```suggestion
**URL** : `/APIVersionV1/definitions/${definitionName}`
```

design/restful-interface
```diff
@@ -0,0 +1,195 @@
...
+
+**Method** : `Delete`
+
+**Query Parameter** : `dryrun` 
+
+**Body** : None
+
+**Responses Body** : None
```
wonderflow, line 195: response `{}` represent OK. If there's error, return :

```
{
  "error": "<reason here>"
}
```

design/restful-interface
ryanzhang-oss: This is the standard RESTful version practice for the version to be a date. I can move it our of the URL though.

design/restful-interface
```diff
@@ -0,0 +1,195 @@
...
+
+The first version does not use authentication or authorization. We will use go structure to 
+represent request and response body.
+
+We will define some common header keys like "x-oam-client-type", "x-oam-username" but it's not fleshed out yet.
+
+## ApplicationConfiguration related API
+
+### Create 
+**URL** : `/APIVersionV1/appconfig/`
```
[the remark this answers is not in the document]
ryanzhang-oss: Create REST doesn't need to include the name if the body already contains the name

design/restful-interface
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+**Responses Body** : None
+
+
+### Update 
+**URL** : `/APIVersionV1/appconfig/${appConfigName}`
+
+**Method** : `PUT`
```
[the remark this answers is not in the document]
ryanzhang-oss: POST URL doesn't have the app name. We can merge them but it is not as clean.

design/restful-interface
ryanzhang-oss: > In fact if we put the whole Definition CRD here, it will contain definitionType itself:
> 
> ```
> kind: TraitDefinition
> spec:
>   definitionRef: xx
>   extentions:
>       xxx
> ```

Yes, that's the plan, it will have the entire definition

design/restful-interface
ryanzhang-oss: I have put the common errors in the doc

ryanzhang-oss:
> 1. plugins(addon) install/remove

My understanding is that those are basically "definition" create/delete

ryanzhang-oss:
> 2\. plugin repository config (update)

Not done yet, my understanding is that we will need "get/list/update" and update is optional if we allow the platform builder just do `kubectl edit`

ryanzhang-oss:
> 3\. env based CRUD

I am not clear on what commands in https://www.yuque.com/wonderflow/lbastq/dgyium lead to that

ryanzhang-oss:
> 4\. system version (read-only)

Is that what the RESTful version or the CLI/Dashboard version?

ryanzhang-oss:
> 5\. we should allow app CRUD by cue format

What does "cue format" mean here? The current App C/R/U all have a query parameter that indicates working on the CUE part only

design/restful-interface
ryanzhang-oss: but that still makes the backend need to do `kubectl get` 3 times

design/restful-interface
ryanzhang-oss: Is there a command like  'rudr trait --help`

design/restful-interface
wonderflow: I never seen version in url path, the examples you give put it in header or parameters not url path. Let move it to header or parameters.

design/restful-interface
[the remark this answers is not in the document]
wonderflow: We'd better add name, so we can align with update.

design/restful-interface
wonderflow: we can have local cache or use parameter for filter, that's not problem.

design/restful-interface
wonderflow: we don't have now

design/restful-interface
```diff
@@ -0,0 +1,299 @@
...
+**Responses Body** : None 
+
+## Definition related API
+
+We will add ${definitionType} in the path to differentiate if it's a workload/trait/scope definition.
+It can be either "workloadDefinition","traitDefinition", or "scopeDefinition".
+ 
+### Create 
+
+**URL** : `/vela.oam.dev/${definitionType}/`
```
wonderflow, line 160: /vela.oam.dev/definitions/${definitionType}/
```
/-> api/apps/<app-name>

 -> api/definitions/<definition-name>?type=<type>
or
 -> api/workloads/<definition-name>


 -> api/
 -> / {s}
 ```

File: design/restful-interface [hunk 11]
```diff
@@ -0,0 +1,299 @@
...
+	definitionName string `json:"definitionName"`
+    definition runtime.RawExtension `json:"definition"` // the definition data
+}
+```
+
+**Responses Body** : None
+
+### GET
+
+**URL** : `/vela.oam.dev/category/${definitionType}/${definitionName}`
```
wonderflow, line 269:
```suggestion
**URL** : `/vela.oam.dev/category/${definitionType}/${definitionName}?env=<name>`
```

design/restful-interface
```diff
@@ -0,0 +1,299 @@
+# Vela RESTful API design
+
+Below, I list the Vela backend RESTful API grouped by each resource.
+
```
wonderflow, line 4:
```
POST /env<name>
Get /env/<name>

POST /apps/<appname>?env=name

```

ryanzhang-oss:
> 5\. we should allow appfile Create/Read/Update

I've folded this in the apps API

design/restful-interface
```diff
@@ -0,0 +1,390 @@
...
+type appConfigValue struct {
+    defintion runtime.RawExtension `json:"defintion"` // the content
+}
+```
+
+**Responses Body** : None
+
+### Get
+
+**URL** : `/api/env/${envName}/apps/${appName}`
```
wonderflow, line 172:
```suggestion
**URL** : `/api/envs/${envName}/apps/${appName}`
```

design/restful-interface
```diff
@@ -0,0 +1,390 @@
...
+```go
+type qureyMode string
+const fullcontent qureyMode = "fullContent"
+const cueTemplate qureyMode = "cueTemplate"
+```
+
+## Environment related API
+Environment is a top level resources in the REST API. 
+### Create 
+**URL** : `/api/env`
```
wonderflow, line 60: I still insist we should be name in create/update api like this:

```
POST /api/envs/<envName>
```

That'll be more RESTful, as we are create one instance of resource

design/restful-interface
```diff
@@ -0,0 +1,390 @@
...
+type env struct {
+    envName   string       `json:"envName"` 
+    nameSpace string       `json:"nameSpace"`    
+}
+```
+
+**Responses Body** : None
+
+### Get
+**URL** : `/api/env/${envName}`
```
wonderflow, line 78:
```suggestion
**URL** : `/api/envs/${envName}`
```

design/restful-interface
```diff
@@ -0,0 +1,390 @@
...
+
+**Method** : `GET`
+
+**Query Parameter** : `None`
+
+**Body** : 
+```go
+type env struct {
+    nameSpace string       `json:"nameSpace"`
```
wonderflow, line 87:
```suggestion
    namespace string       `json:"namespace"`      
```

design/restful-interface
```diff
@@ -0,0 +1,390 @@
...
+    definitionNames string[] `json:"definitionNames"` // the definition name list, no next Token
+}
+```
+
+## Version API
+This is debatable, we are not sure if this is really needed.
+
+### GET
+
+**URL** : `/api/version/${version}`
```
wonderflow, line 376:
```suggestion
**URL** : `/api/version`
```

ryanzhang-oss: normally get has an instance name there, I am thinking of using $latest as the default. Without it, the version itself is a singleton resource. I am

design/restful-interface
ryanzhang-oss: agree

ryanzhang-oss:
> We'd better use json in body instead of go struct, so that front-end engineer can also understand it.

err, manually write Json is not my forte, the go struct already has Json annotations

PR #99, opened 2020-08-06
fix env init don't create namespace

### LINKED ISSUE #82: [BUG] env:init doesn't create namespace
```
✗ rudr env:init env-hello --namespace env-hello
Create env succeed, current env is env-hello

✗ rudr env env-hello
NAME     	NAMESPACE
env-hello	env-hello%           

✗ kubectl get ns | grep env-hello
```

wonderflow:
fixes #82


pkg/cmd/env.go
```diff
@@ -166,6 +170,10 @@ func CreateOrUpdateEnv(ctx context.Context, envArgs *types.EnvMeta, args []strin
 	if err != nil {
 		return err
 	}
+	if err := c.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: envArgs.Namespace}}); err != nil && !apierrors.IsAlreadyExists(err) {
```
zzxwill, line 173: Checking whether the namespace exists before creating is the other option, which one is better?

wonderflow: Haha, I think one rpc request win

PR #122, opened 2020-08-09
add addon:update and refactor addon code

wonderflow:


pkg/cmd/addon.go
```diff
@@ -1,88 +1,108 @@
...
+		NewAddonConfigCommand(ioStream),
+		NewAddonListCommand(ioStream),
+		NewAddonUpdateCommand(ioStream),
 	)
 }
 
 func NewAddonConfigCommand(ioStreams cmdutil.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "addon:config",
+		Use:     "addon:config <reponame> <url>",
```
zzxwill, line 30: If we name every repository url, then we might need to `addon:switch` to switch the currently using repository. The name won't do more help to identify the url as the url itself would.

pkg/cmd/addon.go
```diff
@@ -120,129 +150,33 @@ func NewAddonListCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Comma
...
-				}
-
-				if err != nil {
-					return err
-				}
-				//Check whether the definition is applied
-				var status = "uninstalled"
-				if _, err = cmdutil.GetWorkloadDefinitionByName(ctx, c, namespace, workloadDefinition.Name); err == nil {
-					status = "installed"
-				}
```
zzxwill, line 192: Is it the right way to check the installation status? If yes, I can fulfill the TODO after the merge:)

pkg/cmd/refresh.go
```diff
@@ -36,12 +35,6 @@ func NewRefreshCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Command
 	return cmd
 }
 
-func StatOrCreate(dir string) {
```
zzxwill, line 39: Cool.

pkg/cmd/addon.go
wonderflow: yes, but we should get definition from `.rudr/definitions/` instead of from cluster

pkg/cmd/addon.go
wonderflow: we can list them all, for example:

```
NAME          STATUS
repo1/route   installed
repo2/scaler uninstalled
```

```
vela addon:install repo2/scaler
```

PR #141, opened 2020-08-11
fix addon add and list

### LINKED ISSUE #95: [BUG] addon or capability?
So we need to decide whether to name the extensibility system `addon` or `capability/cap`, i.e.:

```console
$ rudr addon
  config        Set the addon center, default is local (built-in ones).

  ls            $ rudr addon ls -t workload
  show          $ rudr addon show -t workload -n webservice
  install       Install workload/trait to local so it will show up in rudr --help
  uninstall

vs 
 
$ rudr cap
  config        Set the capability center, default is local (built-in ones).

  ls            $ rudr cap ls -t workload
  show          $ rudr cap show -t workload -n webservice
  install       Install workload/trait to local so it will show up in rudr --help
  uninstall
```

### LINKED ISSUE #97: [Discussion] Using github repo as addon/capability might hit `API rate limit exceeded` issue
Per PR https://github.com/cloud-native-application/RudrX/pull/96, I hit an issue as below for a specified IP.
`{"message":"API rate limit exceeded for 47.251.34.248. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}`

The [official doc](https://developer.github.com/v3/#rate-limiting) says.
`For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests.`

### LINKED ISSUE #119: [Feature] more features needed for addon respository
1. allow multiple respository

```
vela addon:config <repoNam> <url>
```

For plugin server, we choose github content API as standard. But we should improve the user experience, we can check the url prefix:

  - prefix `https://api.github.com`: just use the raw API  `https://api.github.com/repos/<owner>/<repo>/contents/<path-to-dir>`
  - prefix `https://github.com`: we can convert the format https://github.com/<owner>/<repo>/tree/<branch>/<path-to-dir> to api format of github
  - prefix `https://github.com`: we should also support the intuitive format  https://github.com/<owner>/<repo>/<path-to-dir>

we can give an example of http server as plugin server, which can also be our mock server for test.

2. update/sync repository

```
vela addon:update <repoName>
```
if repo name not specified, update all repos.

3. install repository

```
vela addon:install <reponame>/<definitionName>

```

* install definition file to cluster,
* install helm chart
* package manager, handle dependency

### LINKED ISSUE #129: [BUG]print some debug message if `vela addon:ls` doesn't list any add-ons
```
➜  vela addon:config zzxwill "https://api.github.com/repos/zzxwill/catalog/contents/repository?ref=plugin"
➜  vela addon:ls
```

### LINKED ISSUE #133: [BUG]`no definition files found, use 'vela refresh' to sync from cluster` should not be thrown until the `system:init` command is executed.
**Describe the bug**

Thrown WARN `no definition files found, use 'vela refresh' to sync from cluster` before init oam-kubernetes-runtime

**Screenshots**

image

**Additional context**

Perhaps executing any command without executing the `system:init` command before directing it to the `system:init` command

wonderflow:
fixes #129 
fixes #97 now we support `--token` to let user pass their token in.
fixes #119 
fixes #133
fixes #95


pkg/cmd/addon.go
```diff
@@ -40,31 +42,33 @@ func NewAddonConfigCommand(ioStreams cmdutil.IOStreams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			config := plugins.RepoConfig{
+			config := &plugins.RepoConfig{
 				Name:    args[0],
-				Address: ConvertURL(args[1]),
+				Address: args[1],
+				Token:   cmd.Flag("token").Value.String(),
```
zzxwill, line 48: github api token?

zzxwill: Yes, it is. I learnt the meaning in the following lines.

pkg/cmd/addon.go
```diff
@@ -40,31 +42,33 @@ func NewAddonConfigCommand(ioStreams cmdutil.IOStreams) *cobra.Command {
...
 			}
 			if !updated {
-				repos = append(repos, config)
+				repos = append(repos, *config)
 			}
 			if err = plugins.StoreRepos(repos); err != nil {
 				return err
 			}
-			ioStreams.Info(fmt.Sprintf("Successfully configured Addon repo: %s", args[0]))
+			ioStreams.Info(fmt.Sprintf("Successfully configured Addon repo: %s, please use 'vela addon:update %s' to sync addons", args[0], args[0]))
```
zzxwill, line 64: Why is it required to manually run `vela addon:update`?
```
$ vela addon:config zzxwill "https://api.github.com/repos/zzxwill/catalog/contents/repository?ref=plugin"
Successfully configured Addon repo: zzxwill, please use 'vela addon:update zzxwill' to sync addons
```

PR #147, opened 2020-08-12
Implement API `api/envs/default/apps/

zzxwill:
implement API `api/envs/default/apps/ and refactor code

### DIFF
--- a/pkg/cmd/dashboard.go
+++ b/pkg/cmd/dashboard.go
@@ -1,4 +1,4 @@
-package main
+package cmd
 
 import (
 	"context"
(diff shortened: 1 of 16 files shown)


cmd/server/main.go
```diff
@@ -1,60 +1,7 @@
 package main
```
wonderflow, line 1: if we have `vela dashboard` command, we don't need this package any more

pkg/cmd/dashboard.go
```diff
@@ -0,0 +1,93 @@
...
+by Cli and called by cmd/server/main.go.
+*/
+func SetupApiServer(kubeClient client.Client) {
+	var logFilePath string
+	var logRetainDate int
+	var logCompress, development bool
+
+	flag.StringVar(&logFilePath, "log-file-path", "", "The log file path.")
+	flag.IntVar(&logRetainDate, "log-retain-date", 7, "The number of days of logs history to retain.")
+	flag.BoolVar(&logCompress, "log-compress", true, "Enable compression on the rotated logs.")
```
wonderflow, line 60: we don't need these arguments for command line, just use stdout

pkg/server/handler/appHandlers.go
```diff
@@ -26,6 +32,14 @@ func GetApps(c *gin.Context) {
 }
 
 func ListApps(c *gin.Context) {
+	kubeClient := c.MustGet("KubeClient")
+	namespace := c.Param("envName")
```
wonderflow, line 36: it's env name,not namespace. Namespace is one information contained in env which can retrieve from envfile by envName.

cmd/server/main.go
zzxwill: Indeed. I still keep it as it can be the access to quickly launch API server. Let me delete it. Thanks.

wonderflow: we can add a Makefile command for quick launch

pkg/cmd/dashboard.go
ryanzhang-oss: this is for the RESTful server logging, won't this be good to have? We don't want to print those logs to the stdout, right?

wonderflow: If this server is deployed in K8s cluster, print to stdout is enough.
If this is used as cli dashboard command, print to sdtout is also fine.

pkg/server/handler/appHandlers.go
```diff
@@ -26,6 +32,14 @@ func GetApps(c *gin.Context) {
...
+	ctx := context.Background()
```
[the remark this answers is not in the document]
zzxwill: Fixed.

pkg/server/handler/appHandlers.go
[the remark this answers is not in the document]
zzxwill: Done.

zzxwill: Fixed, thanks.

zzxwill:
Please ignore failure of build step `Install OAM runtime and prepare WorkloadDefinitions/TraitDefinitions`, I will fix it after https://github.com/cloud-native-application/RudrX/pull/132 is merged. Thanks.

pkg/cmd/dashboard.go
ryanzhang-oss: I don't get it, users don't want to see our server log printing out to their Cli command line, right?

pkg/server/handler/appHandlers.go
```diff
@@ -26,6 +33,21 @@ func GetApps(c *gin.Context) {
 }
 
 func ListApps(c *gin.Context) {
+	kubeClient := c.MustGet("KubeClient")
+	envName := c.Param("envName")
+	envMeta, err := oam.GetEnvByName(envName)
```
ryanzhang-oss, line 38: so looks like the envMeta is stored on the local file system?

zzxwill: Fixed.

zzxwill:
@ryanzhang-oss I like your integration test framework very much. But it will take me some time to have it passed as some fundamental apis like env create, system init needed to be implemented first. 
And will you consider move it to https://github.com/cloud-native-application/RudrX/tree/master/e2e and named a folder `apiserver` to cover all api server test cases?

PR #148, opened 2020-08-12
add integration test for RESTful server

ryanzhang-oss:
Added the example integration test for our Restful server.  Please add at least one test per API there


wonderflow:
@zzxwill  feel free to merge if you approve

PR #162, opened 2020-08-17
    Add vela cli documentation

zzxwill:
This is use for 
1) a first glance of vela command and all sub commands and their usages.
2) dashboard developers to look function and parameters details. We don't need to rewrite Restful API specification for every interface.


wonderflow:
LGTM

PR #168, opened 2020-08-18
Update front-end project

hanxie-crypto:
1. Add code formatting
2. Beautify the code format
3. Interface call


dashboard/mock/getAppList.js
```diff
@@ -4,49 +4,27 @@
 
 export default {
   // 支持值为 Object 和 Array
-  "GET /love/record": [{
-    name: "app-foo",
-    time: '2020-08-11 13:20:45',
-    status: 1,
-    workLoad: {
-      name: 'Delopyment'
+  'GET /api/envs/default/apps': [
```
zzxwill, line 7: Why do we need mocked data here :)

Issue #163, opened 2020-08-17
[BUG] not existed env successfully deleted 


zzxwill:
```
➜  /Users/zhouzhengxi/Programming/golang/src/github.com/cloud-native-application/RudrX git:(master) ✗ go run cmd/vela/main.go env
NAME           	CURRENT	NAMESPACE
abc            	*      	def
default        	       	default
env-application	       	env-application
env-hello      	       	env-hello
env-poc        	       	env-poc
env-trait      	       	env-trait
env-workload   	       	env-workload
➜  /Users/zhouzhengxi/Programming/golang/src/github.com/cloud-native-application/RudrX git:(master) ✗ go run cmd/vela/main.go env:delete XXXX
XXXX deleted
```

mosesyou:
I would work on this.

PR #179, opened 2020-08-19
fix not existed env successfully deleted

mosesyou:
fix: #163 
Delete not existed env would print tips normally now.
```
$ vela env:delete xxxx
xxxx does not exist
```

### DIFF
--- a/pkg/oam/env.go
+++ b/pkg/oam/env.go
@@ -140,7 +140,14 @@ func DeleteEnv(envName string) (string, error) {
 	if err != nil {
 		return message, err
 	}
-	if err = os.RemoveAll(filepath.Join(envdir, envName)); err != nil {
+	envPath := filepath.Join(envdir, envName)
+	if _, err := os.Stat(envPath); err != nil {
+		if os.IsNotExist(err) {
+			err = fmt.Errorf("%s does not exist", envName)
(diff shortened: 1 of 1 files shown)


pkg/oam/env.go
```diff
@@ -140,7 +141,17 @@ func DeleteEnv(envName string) (string, error) {
 	if err != nil {
 		return message, err
 	}
-	if err = os.RemoveAll(filepath.Join(envdir, envName)); err != nil {
+	envPath := filepath.Join(envdir, envName)
+	if _, err := os.Stat(envPath); err != nil {
+		if os.IsNotExist(err) {
+			envExist = false
+			message = envName + " does not exist"
```
wonderflow, line 148: you could directly return this error

mosesyou: if return this error directly,command line will display `Error: stat /Users/admin/.vela/envs/tessdf: no such file or directory`,it's ok?

wonderflow: return  envName + " does not exist"

wonderflow (approved):
LGTM

pkg/oam/env.go
mosesyou: Did it

PR #180, opened 2020-08-19
Implement listing WorkloadDefinitions API

zzxwill:
implemented the API to list workloaddefinitions with
the workloadtype name and all its parameters


wonderflow (approved):
LGTM

PR #188, opened 2020-08-24
fix `cap:center:config` panic

mosesyou:
fix: #186 
check err after `NewCenterClient`
```
$ vela cap:center:config abc http://abc.com
Successfully configured capability center: abc, start to sync from remote
Error: we only support github as repository now
```


wonderflow:
good job

PR #194, opened 2020-08-24
Add Restful API reference

zzxwill:
added Restful API reference for dashboard development
and move cli docs to `documentation/cli`


wonderflow (approved):
LGTM

PR #195, opened 2020-08-24
use unix command style and refactor to new design

wonderflow:


pkg/cmd/workload_run.go
```diff
@@ -5,87 +5,135 @@ import (
...
-		return err
+func AddCompCommands(c types.Args, ioStreams util.IOStreams) *cobra.Command {
+	compCommands := &cobra.Command{
+		Use:                   "comp <commands>",
+		DisableFlagsInUseLine: true,
+		Short:                 "Manage Components",
+		Long:                  "Manage Components",
+		Annotations: map[string]string{
+			types.TagCommandType: types.TypeWorkloads,
+		},
```
zzxwill, line 38: Now it needs to take two steps to create an application?

wonderflow: vela comp run -t deployment myapp

PR #196, opened 2020-08-24
Refine CI build

zzxwill:
remove vela binary build during e2e-test, instead
use the binary after `make build`


wonderflow (approved):
GOOD

PR #201, opened 2020-08-25
add vela comp ls

wonderflow:


pkg/cmd/show.go
```diff
@@ -57,48 +50,125 @@ type Unkown struct {
...
 	if err != nil {
-		return fmt.Errorf("Fetch component %s with Err: %s", componentName, err)
+		return err
 	}
-	if component.Labels == nil {
-		return fmt.Errorf("Can't get workloadDef, please check component %s label \"%s\" is correct.",
-			componentName, types.ComponentWorkloadDefLabel)
+
+	cmd.Printf("About:\n\n")
+	table := uitable.New()
```
zzxwill, line 61: It would be better to only get application detailed information in `showApplication()` and then print it in `uitable`. In this way, the data can reuse by api server. Of course, I can do the job later.

zzxwill (approved):
/lgtm

PR #209, opened 2020-08-26
fix vela app delete

wonderflow:


pkg/cmd/delete.go
```diff
@@ -53,39 +58,43 @@ func NewDeleteCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Command
...
-	o.Component.Namespace = namespace
-	o.AppConfig.Name = args[0]
-	o.AppConfig.Namespace = namespace
-	return nil
-}
-
 func (o *deleteOptions) Delete() error {
-	o.Infof("Deleting AppConfig \"%s\"\n", o.AppConfig.Name)
-	err := o.client.Delete(context.Background(), &o.AppConfig)
+	o.Infof("Deleting AppConfig \"%s\"\n", o.appName)
```
zzxwill, line 72:
```suggestion
	o.Infof("Deleting App \"%s\"\n", o.appName)
```
When creating an app, we call it as `app`, so it might be better to not expose concept `AppConfig`.

zzxwill (approved):
/lgtm

PR #210, opened 2020-08-26
API request parameter check for trait attach

zzxwill:
check parameter `staging` before converting it
to bool and add logs


pkg/server/handler/traitHandler.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/cloud-native-application/rudrx/pkg/server/apis"
 	"github.com/cloud-native-application/rudrx/pkg/server/util"
 	"github.com/gin-gonic/gin"
+	ctrl "sigs.k8s.io/controller-runtime"
```
wonderflow, line 11: do we really need to import this lib only for logs？

zzxwill: Not sure, but this lib is widely used across all API services.

wonderflow (approved):

PR #211, opened 2020-08-26
Development on env

wenxinnnnn:
1.Complete the curd of env.
2.Complete the global env switch.


dashboard/yarn.lock
```diff
@@ -0,0 +1,17632 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
wonderflow, line 2: Can we remove this file?

wenxinnnnn: ok！I will delete yarn.lock , and add to the ignore file

zzxwill:
The whole design is as expected, great.
BTW, it would be better to combine those 13 commits into one or two.

zzxwill:
Closed it as it has been merged in #230

Issue #200, opened 2020-08-25
[Feature]Discussion on Env CLI and env API


wenxinnnnn:
**Is your feature request related to a problem? Please describe.**

In my opinion, env init and evn specify that the namespace should be detached. In Env init, friendly prompts should be repeated.
Moreover, the idempotency of the interface PSOT / API / envs is inconsistent.

**Describe the solution you'd like**

cli:
`vale env init envname namespace nsname`
Envname presence will block and prompt.

`vale env config envname namespcae nsname`
Envname does not exist, will be blocked and prompted

api:
POST /api/envs/                     {"envName":"xxx"}
PUT  /api/envs/:envName      {"namespcae":"xxx"}

The other 1: expand the env describe field, because the env gradually increases, it is not easy to manage by env name alone.
`vale env config envname namespcae nsname describe  xxx `
PUT  /api/envs/:envName      {"namespcae":"xxx","describe":""}

The other 2:  currently get /API/envs does not support paging query.

wonderflow:
I can't very understand this issue, you could describe in Chinese if you don't mind.

wonderflow:
fixed by #215

PR #215, opened 2020-08-27
Optimize env interface and separate responsibilities.

wenxinnnnn:
Refer to #200 .
CLI remains unchanged.
Restful api is split into two interfaces.


e2e/apiserver/apiserver_test.go
```diff
@@ -21,35 +19,39 @@ import (
 )
 
 var (
-	envHelloMeta = types.EnvMeta{
-		Name:      "env-e2e-hello",
+	envHelloMeta = apis.Environment{
```
zzxwill, line 22: What's the purpose of this change?

e2e/commonContext.go
```diff
@@ -215,7 +214,7 @@ var (
 				var r apis.Response
 				err = json.Unmarshal(result, &r)
 				gomega.Expect(http.StatusOK).Should(gomega.Equal(r.Code))
-				output := fmt.Sprintf("Create env succeed, current env is " + envMeta.Name + " namespace is " + envMeta.Namespace)
```
zzxwill, line 218: I think the former expected output is more specific.

pkg/oam/env.go
```diff
@@ -34,11 +35,18 @@ func GetEnvByName(name string) (*types.EnvMeta, error) {
 	return &meta, nil
 }
 
+//Create or update env.
+//If it does not exist, create it and switch to the new env.
+//If it exists, update it and switch to the new env.
 func CreateOrUpdateEnv(ctx context.Context, c client.Client, envName string, namespace string) (error, string) {
```
zzxwill, line 41: - I'm confused with func names  between `CreateEnv` and `UpdateEnv`, and `CreateOrUpdateEnv`.
- API and Cli should depend the same function in oam package, but `CreateOrUpdateEnv` only used by API.

xiaoyuaiheshui:
I'll fix the dashboard to fit this API change.

e2e/commonContext.go
wenxinnnnn: Now the create an env interface does not provide the ability to switch env.

e2e/apiserver/apiserver_test.go
wenxinnnnn: Environment object provides input parameter verification of api interface

pkg/oam/env.go
wenxinnnnn: API and Cli use different user habits, so I suggest that cli still retains the ability to switch while creating

zzxwill: @wonderflow Please also take a look at this design.

e2e/apiserver/apiserver_test.go
zzxwill: types.EnvMeta could not?

e2e/commonContext.go
zzxwill: I got it. @wonderflow is it okay with your initial design?

e2e/apiserver/apiserver_test.go
wenxinnnnn: types.EnvMeta It's domain entity. Environment is similar to dto,in the future, the interface can carry more expressive force

wonderflow (approved):

PR #216, opened 2020-08-27
support scopes in appfile and make health scope as default

wonderflow:


pkg/oam/workload.go
```diff
@@ -65,23 +66,44 @@ func BaseComplete(envName string, workloadName string, appGroup string, flagSet
...
 		case cue.BoolKind:
-			d, _ := strconv.ParseBool(flagValue)
-			workloadData[v.Name] = d
+			workloadData[v.Name], err = flagSet.GetBool(v.Name)
 		case cue.NumberKind, cue.FloatKind:
-			d, _ := strconv.ParseFloat(flagValue, 64)
-			workloadData[v.Name] = d
+			workloadData[v.Name], err = flagSet.GetFloat64(v.Name)
+		}
+		if err != nil {
```
zzxwill, line 90: If v.Type doesn't meet any case, the value `err` is not the one returned by `flagSet.Getxxx()`. This might cause confusions.

zzxwill, line 90: This is cool. The design can be compatible with CLI and API.

zzxwill (approved):
/lgtm

pkg/oam/workload.go
wonderflow: we should make sure only these for type will be stored in template.Prameter

PR #223, opened 2020-08-28
import trait-injector

ryanzhang-oss:
import trait-injector


cmd/core/main.go
```diff
@@ -103,6 +108,22 @@ func main() {
...
+		Log:      ctrl.Log.WithName("controllers").WithName("ServiceBinding"),
+		Scheme:   mgr.GetScheme(),
+		Recorder: mgr.GetEventRecorderFor("servicebinding"),
+	}
+	if err = (tiWebhook).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "ServiceBinding")
+		os.Exit(1)
+	}
+	// +kubebuilder:scaffold:builder
+	go tiWebhook.ServeAdmission()
```
wonderflow, line 125: this leaves two issues here:
1. we'd better align all webhook together https://github.com/cloud-native-application/RudrX/pull/223/files#diff-a280806a9c33d12b7b2842c704ee485eR95-R99
2. we rely on web hook to provide full capability now, we may need to use web hook by default

wonderflow (approved):
Cool

cmd/core/main.go
ryanzhang-oss: The trait-injector itself is a webhook while others just have webhook as validating/mutation so they are not exactly the same, I am not sure if we really need the "trait-injector"

ryanzhang-oss: We better turn on the webhooks though

wonderflow: yes, we need turn on webhooks, if not, the  trait-injector won't work.  This can be low priority work

PR #224, opened 2020-08-28
Add containerized workload part 1

ryanzhang-oss:
Add containerized workload


api/v1alpha1/containerized_types.go
```diff
@@ -0,0 +1,82 @@
...
+	// Resources managed by this workload.
+	Resources []cpv1alpha1.TypedReference `json:"resources,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// Containerized is the Schema for the containerizeds API
+// +kubebuilder:resource:categories={oam}
+// +kubebuilder:subresource:status
+type Containerized struct {
```
wonderflow, line 52: We need to consider give it a new name. What about:

- `Container` workload
- `Pod` workload
- `VelaCore` workload
- `Vela` workload
- `Core` workload

zzxwill:
@ryanzhang-oss Is this workload one of those three workloads to be created for vela? What's the major difference with ContainerizedWorkload? I generated the CRD, compared with `core.oam.dev_containerizedworkloads.yaml` of oam-kubernetes-runtime and noticed some differences:)

api/v1alpha1/containerized_types.go
ryanzhang-oss: @resouer suggested this name and I don't have a strong opinion. I can't seem to think of a better name (including the ones you suggested)

pkg/builtin/workloaddefinition/backend.go
```diff
@@ -0,0 +1,54 @@
...
+  childResourceKinds:
+    - apiVersion: apps/v1
+      kind: Deployment
+    - apiVersion: v1
+      kind: Service
+  extension:
+	defaultTraits:
+        - scale
+        - route
+        - monitor
```
wonderflow, line 22: These three CUE files maybe won't work, but we can fix it in following PRs.

wonderflow (approved):
LGTM, feel free to merge if you feel OK @ryanzhang-oss

PR #233, opened 2020-08-31
Delete an application

wenxinnnnn:


zzxwill (approved):
Great! I din't use the `xxxOptions` struct in precious API implementations, but your solution looks better with less change to the logic of the com functions. 
/lgtm

wonderflow (approved):
Good job

PR #245, opened 2020-09-04
update example & add judgment for comp delete

sunny0826:
In command `vela comp delete frontend -a frontend` , flag `-a` is necessary.

update example and add judgement on flag `-a`

Signed-off-by: guoxudong <sunnydog0826@gmail.com>

### DIFF
--- a/pkg/cmd/delete.go
+++ b/pkg/cmd/delete.go
@@ -66,11 +66,18 @@ func NewCompDeleteCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Comm
 		Annotations: map[string]string{
 			types.TagCommandType: types.TypeApp,
 		},
-		Example: "vela comp delete frontend",
+		Example: "vela comp delete frontend -a frontend",
 	}
 	cmd.SetOut(ioStreams.Out)
 
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		appName, err := cmd.Flags().GetString(App)
+		if err != nil {
+			return err
+		}
+		if appName == "" {
+			return errors.New("must specify name of application, please add flag -a")
+		}
 		newClient, err := client.New(c.Config, client.Options{Scheme: c.Schema})
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


pkg/cmd/delete.go
```diff
@@ -66,11 +66,18 @@ func NewCompDeleteCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Comm
...
 	cmd.SetOut(ioStreams.Out)
 
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		appName, err := cmd.Flags().GetString(App)
+		if err != nil {
+			return err
+		}
+		if appName == "" {
+			return errors.New("must specify name of application, please add flag -a")
+		}
```
wonderflow, line 80: If the appName not specified, we should also possible delete component, for example use compName as AppName

pkg/cmd/delete.go
```diff
@@ -85,10 +92,6 @@ func NewCompDeleteCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Comm
 			return errors.New("must specify name for the app")
```
wonderflow, line 92:
```suggestion
			return errors.New("must specify name for the component")
```

We should change the error message

wonderflow (changes requested):

pkg/cmd/delete.go
sunny0826: In this `delete`  appName  must specified, and the info of cmd is `Delete Component From Application`

```bash
$ bin/vela comp delete nginx
Deleting Component 'nginx' from Application ''
Error: delete appconfig err resource name may not be empty
```

pkg/cmd/delete.go
sunny0826:
```go
if len(args) < 1 {
			return errors.New("must specify name for the app")
		}
```
the `args` get the name of component not app , app name in flag `-a`

pkg/cmd/delete.go
sunny0826:
```go
appName, err := cmd.Flags().GetString(App)
if err != nil {
	return err
}
o.AppName = appName
```

the appname for flag `-a` , If the appName not specified, `err, message := o.DeleteComponent()` will throw err

PR #260, opened 2020-09-11
Clean Up Projects

### LINKED ISSUE #226: [Feature] oam system init install vela core instead of oam-k8s-runtime.
1.  build a helm chart for vela core, along with all the crd, workloaddefinition, traitdefinition, webhooks.
2. follow [the way `osm install` used](https://github.com/openservicemesh/osm/blob/main/Makefile#L49) to build chart into `vela` binary, so we don't need to connect any helm hub for installation.

wonderflow:
1. change go package to oam-dev/kubevela
2. clean up some code format
3. install vela core chart instead of oam-k8s-runtime fixes #226
4. fix release workflow action


zzxwill (approved):
/lgtm

PR #270, opened 2020-09-13
Remove website related files

zzxwill:
Removed website related files and planning moving
the website to oam-dev/kubevela.io


wonderflow (approved):
LGTM thanks

PR #276, opened 2020-09-14
add route trait as vela api gateway

### LINKED ISSUE #175: [Feature]  support api gateway as domain/dns/route trait for web service application

wonderflow:
1. Add a route trait into vela core, this could be used as Api gateway.
2. Use Nginx-Ingress as default backend implementation.

fixes #175

Will add test later, ready for review now.


pkg/controller/v1alpha1/routes/route_controller.go
```diff
@@ -0,0 +1,436 @@
...
+		Spec: corev1.ServiceSpec{
+			Type: corev1.ServiceTypeLoadBalancer,
+		},
+	}
+	// assign selector
+	if routeTrait.Spec.Backend != nil && routeTrait.Spec.Backend.SelectLabels != nil {
+		oamService.Spec.Selector = routeTrait.Spec.Backend.SelectLabels
+	}
+
+	port, labels, err := DiscoverPortLabel(workload)
```
ryanzhang-oss, line 200: this would not work for the core containerizedWorkload or anything that doesn't have the pod.spec as is. I wonder if you want to also look at the childResources?

pkg/controller/v1alpha1/routes/route_controller.go
```diff
@@ -0,0 +1,436 @@
...
+func (r *Reconciler) createSelfsignedIssuer(ctx context.Context, routeTrait *v1alpha1.Route) (string, error) {
+	var selfSigned = "selfsigned"
+	var issuer = certmanager.Issuer{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: certmanager.SchemeGroupVersion.String(),
+			Kind:       certmanager.IssuerKind,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      selfSigned,
+			Namespace: routeTrait.Namespace,
```
ryanzhang-oss, line 276: there is a self signed cert in the vela-system, can you use that?

pkg/controller/v1alpha1/routes/route_controller.go
```diff
@@ -0,0 +1,436 @@
...
+				},
+			},
+		},
+	}
+	var ingressClass = "nginx"
+	ingress.Spec.IngressClassName = &ingressClass
+	ingress.Spec.TLS = []v1beta1.IngressTLS{
+		{
+			Hosts:      []string{routeTrait.Spec.Host},
+			SecretName: routeTrait.Name + "-cert",
```
ryanzhang-oss, line 361: How is this "secreteName" created? Where is the "-cert" suffix introduced?

pkg/controller/v1alpha1/routes/route_controller.go
```diff
@@ -0,0 +1,436 @@
...
+		},
+	}
+	if routeTrait.Spec.DefaultBackend != nil {
+		ingress.Spec.Backend = routeTrait.Spec.DefaultBackend
```
ryanzhang-oss, line 365: Do we need both spec.Backend and Rules.IngressRueValue.Paths.backend in the rout.spec? Can we make them the same?

PROJECT
```diff
@@ -1,3 +1,7 @@
 domain: oam.dev
 repo: github.com/oam-dev/kubevela
+resources:
+- group: standard
+  kind: Route
+  version: v1alpha1
 version: "2"
```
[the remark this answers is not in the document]
wonderflow: It's automatically created by kubebuilder

api/v1alpha1/route_types.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Path is location Path, default for "/"
+	Path string `json:"path,omitempty"`
+
+	// TLS indicate route trait will create SSL secret using cert-manager with specified issuer
+	// If this is nil, route trait will use a selfsigned issuer
+	TLS *TLS `json:"tls,omitempty"`
+
+	// DefaultBackend uses serviceName
+	DefaultBackend *v1beta1.IngressBackend `json:"defaultBackend,omitempty"`
```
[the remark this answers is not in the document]
wonderflow: defaultBackend is important when backend is not available

api/v1alpha1/route_types.go
```diff
@@ -0,0 +1,139 @@
...
+// If Port is not set, the first port discovered will be set.
+// If SelectLabels are not set, all selectorLabels discovered will be set.
+type Backend struct {
+	// Protocol means backend-protocol, HTTP, HTTPS, GRPC, GRPCS, AJP and FCGI, By default uses HTTP
+	Protocol string `json:"protocol,omitempty"`
+	// ReadTimeout used for setting read timeout duration for backend service, the unit is second.
+	ReadTimeout int `json:"readTimeout,omitempty"`
+	// SendTimeout used for setting send timeout duration for backend service, the unit is second.
+	SendTimeout int `json:"sendTimeout,omitempty"`
+	// Port points to backend service port.
```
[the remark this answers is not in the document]
wonderflow: Yes, they are not required, but important

pkg/controller/v1alpha1/routes/route_controller.go
```diff
@@ -0,0 +1,436 @@
...
+	"fmt"
+	"reflect"
+	"strconv"
+
+	"github.com/oam-dev/kubevela/api/v1alpha1"
+	standardv1alpha1 "github.com/oam-dev/kubevela/api/v1alpha1"
+	"github.com/oam-dev/kubevela/pkg/controller/common"
+
+	cpv1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
```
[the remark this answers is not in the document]
wonderflow: I split them by intent, one part is inside kubevela , another is outside

pkg/controller/v1alpha1/routes/route_controller.go
wonderflow: that's not cluster-scope, only same namespace can use. So I create one for in each user's namespace

pkg/controller/v1alpha1/routes/route_controller.go
wonderflow: It'll be created by cert-manager automatically, we only need to specify secret name here.

pkg/controller/v1alpha1/routes/route_controller.go
wonderflow: No we can't, they are different concepts.

* `spec.Backend` is default backend, which will be used only the normal backend won't work.

pkg/controller/v1alpha1/routes/route_controller.go
```diff
@@ -0,0 +1,436 @@
...
+)
+
+const (
+	errApplyNginxIngress = "failed to apply the ingress"
+	errCreateIssuer      = "failed to create cert-manager Issuer"
+)
+
+var (
+	// oamServiceLabel is the pre-defined labels for any serviceMonitor
+	// created by the MetricsTrait,  prometheus operator listens on this
```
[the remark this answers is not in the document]
wonderflow: yes， fixed

pkg/controller/v1alpha1/routes/route_controller.go
wonderflow: yes, I added

config/samples/traits/route.yaml
```diff
@@ -1,42 +1,33 @@
 apiVersion: core.oam.dev/v1alpha2
 kind: TraitDefinition
 metadata:
-  name: ingresses.networking.k8s.io
+  name: routes.standard.oam.dev
   annotations:
-    "definition.oam.dev/apiVersion": "networking.k8s.io/v1beta1"
-    "definition.oam.dev/kind": "Ingress"
+    definition.oam.dev/apiVersion: standard.oam.dev/v1alpha1
+    definition.oam.dev/kind: Route
```
resouer, line 7: Is this annotation auto patched?

Why don't we reuse `definitionRef` (with apiVersion)?

wonderflow: we can't easily get version and kind by definitionRef , so we add annotation here in vela.

zzxwill (approved):
/lgtm

config/samples/traits/route.yaml
resouer: It should be fixed in `definitionRef` eventually, otherwise we should deprecate this field, for example:

```yaml
definitionRef:
    name: routes.standard.oam.dev
    version: v1alpha1
```

or 

```yaml
definitionRef:
  apiVersion: standard.oam.dev/v1alpha1
  kind: Route

PR #281, opened 2020-09-16
vela core install dependencies

ryanzhang-oss:
Vela core now install all the dependencies in a helm chart form on startup. It skips a chart if it detects CRDs in the cluster


wonderflow (approved):
LGTM

charts/vela-core/templates/velaConfig.yaml
```diff
@@ -9,5 +9,14 @@ data:
       "repo": "jetstack",
       "urL": "https://charts.jetstack.io",
       "name": "cert-manager",
-      "version": "v1.0.0"
-    },
\ No newline at end of file
+      "namespace": "cert-manager",
+      "version": "1.0.0"
+    }
+  ingresses.networking.k8s.io: |
```
wonderflow, line 15: please add some readme about how this config get to work

ryanzhang-oss: I will after the whole e2e works, it's still a mess

PR #287, opened 2020-09-17
get cue temp from remote through URI

captainroy-hy:
When it's going to get cue templates of capabilities, use templateURI as higher priority than template strings in definition file. 

BTW, I will complete tests after this PR containing test data is merged.


pkg/plugins/cluster.go
```diff
@@ -100,12 +101,28 @@ func HandleTemplate(in *runtime.RawExtension, name, syncDir string) (types.Capab
...
+		res, err := http.Get(tmp.CueTemplateURI)
+		if err != nil {
+			return types.Capability{}, err
+		}
+		defer res.Body.Close()
+		b, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return types.Capability{}, err
+		}
+		cueTemplate = fmt.Sprintf("%s", b)
```
wonderflow, line 116:
```suggestion
		cueTemplate = string(b)
```

wonderflow, line 116: can you try add some test for this case

wonderflow:
you could use existing file here to do test https://github.com/oam-dev/kubevela/blob/master/vela-templates/web-service.cue @captainroy-hy

captainroy-hy:
> you could use existing file here to do test https://github.com/oam-dev/kubevela/blob/master/vela-templates/web-service.cue @captainroy-hy

Sure, got it.

wonderflow (approved):
Good job

Issue #283, opened 2020-09-16
[BUG]`vela comp run` occurs panic if not set any args


mosesyou:
```
$ vela comp run
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/oam-dev/kubevela/pkg/commands.NewCompRunCommands.func1(0xc0007158c0, 0x40c1920, 0x0, 0x0, 0x0, 0x0)
	/Users/admin/software/mosesyou/kubevela/pkg/commands/comp.go:62 +0x2bb
github.com/spf13/cobra.(*Command).execute(0xc0007158c0, 0x40c1920, 0x0, 0x0, 0xc0007158c0, 0x40c1920)
	/Users/admin/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0xc000244000, 0x16353d891d2c1328, 0x4085660, 0xc00010e058)
	/Users/admin/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/admin/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
	/Users/admin/software/mosesyou/kubevela/cmd/vela/main.go:47 +0x8c
exit status 2
```

wonderflow:
@mosesyou  thanks for reporting, would you like to send a PR?

mosesyou:
sure

wonderflow:
thanks @mosesyou

PR #288, opened 2020-09-17
fix `vela comp run` panic when args empty

mosesyou:
Ref: #283 
`vela comp run` show help info now when args empty.
```
$ vela comp run

Init and Run workloads

Usage:
  vela comp run [args]

Examples:
vela comp run -t <workload-type>

Flags:
  -h, --help          help for run
  -s, --staging       only save changes locally without real update application
  -t, --type string   specify workload type for application

Global Flags:
  -a, --app string   specify application name for component
  -e, --env string   specify env name for application
```


wonderflow (approved):
LGTM

PR #300, opened 2020-09-20
Fix OpenAPIV3Schema Validation Issue

zzxwill:
Temporarily corrects spec.validation.openAPIV3Schema issue, and it would be removed
after this issue was fixed https://github.com/oam-dev/kubevela/issues/284.

### DIFF
--- a/hack/chart/generate.go
+++ b/hack/chart/generate.go
@@ -15,7 +15,11 @@ import (
 func main() {
 	// Path relative to the Makefile where this is invoked.
 	chartPath := filepath.Join("charts", "vela-core")
-	tempChartPath := fixOpenAPIV3SchemaValidationIssue(chartPath)
+	tempChartPath, err := fixOpenAPIV3SchemaValidationIssue(chartPath)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


zzxwill:
Hit issue https://github.com/oam-dev/kubevela/issues/301. Need more time to fix this issue first.

wonderflow (approved):
LGTM

hack/chart/generate.go
```diff
@@ -1,20 +1,112 @@
...
+				if err = ioutil.WriteFile(targetPath, data, os.ModePerm); err != nil {
+					fmt.Fprintln(os.Stderr, "failed to read file:", err)
+					return err
+				}
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return chartPath
```
wonderflow, line 109: if there's an error, we should return it, or the orignal chart will be removed

zzxwill: Every time hits an issue, it returns the original `chartPath`. This won't do anything bad and can the build can continue. I think you are right, an error should throw out and break the build process.

PR #314, opened 2020-09-23
fix bug: handle exception cases in `vela comp status`

captainroy-hy:
Exception cases handled by `vela comp status` currently:
- The component is unhealthy => show diagnosis 
- The component is not referenced by a HealthScope => show status as `Not Diagnosed` and workload status
- The component is unknown to the HealthScope => show status as `Not Diagnosed` and workload status
- The component's workload controller is not working => show status as `Not Diagnosed` and warning info

### DIFF
--- a/pkg/commands/status.go
+++ b/pkg/commands/status.go
@@ -4,13 +4,17 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


wonderflow:
Can you give the sample output for each case?

captainroy-hy:
- case 1
image
- case 2 & 3
image
- case 4
image

captainroy-hy:
change the output of case 2&3 into yaml format
image

wonderflow (approved):

PR #326, opened 2020-09-25
Fix naming in `json` of struct fields

zzxwill:
almost all of names in `json` of struct fileds
should be in `lowerUpper` format


wonderflow (approved):
LGTM， please align with our frontend engineer, this may cause incompatible and makes dashboard broken.

Issue #310, opened 2020-09-23
[Feature] track and show component status changing in `vela init`


captainroy-hy:
Enhance feature inspired by #308  @wonderflow 

**Describe the solution you'd like**

When `vela init xxx` occurs, we need to track and show status changing of the component. 

Proposal: 

The overall strategy as below:
![image](https://user-images.githubusercontent.com/5107364/93953588-befe4000-fd86-11ea-8e97-7078a931a3a5.png)

- The initial status is `Pending`
  - if ( k8sClient can get the workload object ) , `Pending` ==> `Creating`
  - if (appConfig reconcile condition is `true` && k8sClient cannot get the workload object ), stay `Pending`
  - if (appConfig reconcile condition is `false` && k8sClient cannot get the workload object ), `Pending` ==>`InitFailed`
    - then output error msg from appConfig condition
- The 2nd normal status is `Creating`
  - if (workload reconcile condition is `false`), `Creating` ==> `CreateFailed`
    - then output error msg from workload condition
  - if (workload status is empty), stady `Creating`
  - if (workload reconcile condition is `true`), `Creating` ==> `HealthChecking`

When status comes into `HealthChecking`, vela should notify users with output:
> Deployed successfully!
> Start checking the health status ...
> Health checking may complete in minutes, you are optional to exit now and check the health status latter.

IMO, `vela init` is responsible for deploying the component while health checking is the next step after the component deployed. So the user can choose to wait for health checking finish **or** just exit the command and check component status latter.

If user doesn't exit the command, vela continues showing status changing as below:
- The 3rd normal status is `HealthChecking`
  - if (HealthScope indicates the workload is `Healthy`), `HealthChecking` ==> `Healthy`
  - if (HealthScope indicates the workload is `Unhealthy` && the age of the workload >= 120s), `HealthChecking` ==> `Unhealthy`
  - if (HealthScope indicates the workload is `Unhealthy` && the age of the workload < 120s), stay `HealthChecking` 
  - if (no HealthScope referenced ||  HealthScope indicates the workload is `Unknown`), `HealthChecking` ==> `Not Diagnosed`
    - output workload condition msg

> 120s means 2 health check cycles, as current default health check interval is 60s. We can use the buffered time to distinguish whether the workload is truly unhealthy or just still under progressing.

**Describe alternatives you've considered**

**Additional

captainroy-hy:
Demo as below:
[gif show click here](https://gifyu.com/image/4DiN)

[...]

PR #329, opened 2020-09-26
track status changing in vela init

captainroy-hy:
#310 
- track status changing in vela init
- add loop checking in "vela comp status"
- fix e2e-test
- update oam-k8s-runtime dependency to latest tag

Demo as below :

[Gif demo click here](https://gifyu.com/image/4JQe)

image
image
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,7 @@ docker-push:
 e2e-setup:
 	ginkgo version
 	ginkgo -v -r e2e/setup
+	kubectl wait --for=condition=Ready pod -l app.kubernetes.io/name=vela-core,app.kubernetes.io/instance=kubevela -n vela-system --timeout=300s
 	bin/vela dashboard &
 
 e2e-test:
--- a/e2e/commonContext.go
+++ b/e2e/commonContext.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/oam-dev/kubevela/pkg/server/apis"
 	"github.com/oam-dev/kubevela/pkg/server/util"
@@ -190,9 +191,9 @@ var (
 		return ginkgo.Context(context, func() {
 			ginkgo.It("should get status for the component", func() {
 				cli := fmt.Sprintf("vela comp status %s", applicationName)
-				output, err := Exec(cli)
+				output, err := LongTimeExec(cli, 120*time.Second)
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
-				gomega.Expect(output).To(gomega.ContainSubstring(applicationName))
+				gomega.Expect(output).To(gomega.ContainSubstring("Checking health status"))
 				// TODO(zzxwill) need to check workloadType after app status is refined
 			})
 		})
--- a/e2e/commonContext.go
+++ b/e2e/commonContext.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	ctx "context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -8,8 +9,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Netflix/go-expect"
+	corev1alpha2 "github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
 	"github.com/oam-dev/kubevela/pkg/server/apis"
 	"github.com/oam-dev/kubevela/pkg/server/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
@@ -187,9 +191,20 @@ var (
 		})
 	}
 
-	ApplicationCompStatusContext = func(context string, applicationName string, workloadType string) bool {
+	ApplicationCompStatusContext = func(context string, applicationName, workloadType, envName string) bool {
 		return ginkgo.Context(context, func() {
 			ginkgo.It("should get status for the component", func() {
+				ginkgo.By("init new k8s client")
+				k8sclient, err := newK8sClient()
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+				ginkgo.By("check AppConfig reconciled ready")
+				gomega.Eventually(func() int {
+					appConfig := &corev1alpha2.ApplicationConfiguration{}
+					_ = k8sclient.Get(ctx.Background(), client.ObjectKey{Name: applicationName, Namespace: "default"}, appConfig)
+					return len(appConfig.Status.Workloads)
+				}, 120*time.Second, 1*time.Second).ShouldNot(gomega.Equal(0))
+
 				cli := fmt.Sprintf("vela comp status %s", applicationName)
 				output, err := LongTimeExec(cli, 120*time.Second)
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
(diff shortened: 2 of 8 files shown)


e2e/commonContext.go
```diff
@@ -190,9 +191,9 @@ var (
 		return ginkgo.Context(context, func() {
 			ginkgo.It("should get status for the component", func() {
 				cli := fmt.Sprintf("vela comp status %s", applicationName)
-				output, err := Exec(cli)
+				output, err := LongTimeExec(cli, 120*time.Second)
```
zzxwill, line 194: It will need more time to show component status, so you set `LongTimeExec` for `vela comp status`?

pkg/commands/init.go
```diff
@@ -30,6 +34,26 @@ type appInitOptions struct {
...
+type CompStatus string
+
+const (
+	compStatusInitializing    CompStatus = "initializing"
+	compStatusInitFail        CompStatus = "initfail"
+	compStatusInitialized     CompStatus = "initialized"
+	compStatusDeploying       CompStatus = "deploying"
+	compStatusDeployFail      CompStatus = "deployfail"
+	compStatusDeployed        CompStatus = "deployed"
+	compStatusHealthChecking  CompStatus = "healthchecking"
```
zzxwill, line 46: Will `healthChecking` be more proper?

e2e/commonContext.go
captainroy-hy: yes, `vela comp status` needs more time to get precise health status from HealthScope.

pkg/commands/status.go
```diff
@@ -357,34 +331,194 @@ func printPrefix(p string) string {
...
+	if healthStatus == HealthStatusUnhealthy {
+		cTime := appConfig.GetCreationTimestamp()
+		if time.Since(cTime.Time) <= healthCheckBufferTime {
+			return compStatusHealthChecking, HealthStatusUnknown, "", nil
+		}
+	}
+	return compStatusHealthCheckDone, healthStatus, wlhc.Diagnosis, nil
+}
+
+func getAppAndApplicationConfiguration(ctx context.Context, c client.Client, compName, appName string, env *types.EnvMeta) (*application.Application, *v1alpha2.ApplicationConfiguration, error) {
```
zzxwill, line 491: Normally we consider `App` the same as `ApplicationConfiguration`, so I am a little bit confused about the function name.

captainroy-hy: Yes, a wired name... they are the same logically but still refer to different typed structs.

pkg/commands/init.go
captainroy-hy: It seems string doesn't make sense here, I will change them into iota enum..

pkg/commands/init.go
```diff
@@ -30,6 +34,26 @@ type appInitOptions struct {
...
+	compStatusDeployed
+	compStatusHealthChecking
+	compStatusHealthCheckDone
+	compStatusUnknown
+)
+
+var (
+	emojiSucceed = emoji.Sprint(":check_mark_button:")
+	emojiFail    = emoji.Sprint(":cross_mark:")
+	emojiTimeout = emoji.Sprint(":heavy_exclamation_mark:")
```
wonderflow, line 54: I like these emoji 👍  We should use more in the future.

pkg/commands/init.go
```diff
@@ -70,8 +94,68 @@ func NewInitCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Command {
...
+					ioStreams.Info(green.Sprintf("\n%sDeployment Succeed!", emojiSucceed))
+					sDeploy.Stop()
+					break TrackDeployLoop
+				case compStatusDeployFail:
+					ioStreams.Info(red.Sprintf("\n%sDeployment Failed!", emojiFail))
+					ioStreams.Info(red.Sprintf("Reason: %s", failMsg))
+					sDeploy.Stop()
+					return nil
+				}
+			}
```
wonderflow, line 158: Please write functions for these check.

wonderflow: Write a common function like `LoopCheck` for this loop and check logic is better.

pkg/commands/init.go
```diff
@@ -70,8 +94,68 @@ func NewInitCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Command {
...
+
+			sDeploy := spinner.New(spinner.CharSets[14], 100*time.Millisecond,
+				spinner.WithColor("green"),
+				spinner.WithHiddenCursor(true),
+				spinner.WithSuffix(color.New(color.Bold, color.FgGreen).Sprintf(" %s", "Deploying ...")))
+			sDeploy.Start()
+		TrackDeployLoop:
+			for {
+				time.Sleep(2 * time.Second)
```
wonderflow, line 140: The time interval seems to be too long, `1s` or shorter is better.

pkg/commands/status.go
```diff
@@ -357,34 +331,194 @@ func printPrefix(p string) string {
...
+	_, foundStatus, ct, err := getWorkloadInstanceStatusAndCreationTime(ctx, c, env.Namespace, wlRef)
+	if err != nil {
+		return compStatusUnknown, "", err
+	}
+	if foundStatus {
+		//TODO(roywang) temporarily use status to judge workload controller is running
+		// even not every workload has `status` field
+		return compStatusDeployed, "", nil
+	}
+
```
wonderflow, line 419: we should add a `TODO` here, because an OAM deploy including `workload + trait`, but I can only see workload checked here, if route trait not work, the app also won't be able to access.

wonderflow (commented):
This PR is really cool, I hope to see some test cases ,thanks!

pkg/commands/status.go
captainroy-hy: Fixed.

pkg/commands/init.go
captainroy-hy: Fixed. Set interval to `1s`.

pkg/commands/init.go
captainroy-hy: Fixed.

captainroy-hy (commented):
@wonderflow e2e-test for `vela comp status` is blocked because core-controller is not setup in e2e-test. #335

wonderflow:
@captainroy-hy commented there

pkg/commands/status.go
```diff
@@ -212,94 +250,50 @@ func NewCompStatusCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Comm
...
-			break
-		}
-	}
-	if wlStatus.ComponentName == "" {
-		//TODO(roywang) cannot get workload instance
-		//TODO(roywang) if appConfig reconcile condition is false, then output err msg
-		ioStreams.Infof("\nComponent is not created yet.")
+	wlStatus, foundWlStatus := getWorkloadStatusFromAppConfig(appConfig, compName)
+	if !foundWlStatus {
+		ioStreams.Info("\nThe component has not been initialized by oam-core-controller correctly.")
```
wonderflow, line 264: > not been initialized by oam-core-controller correctly

Users won't understand what does it mean, can you give more concrete error information

pkg/commands/status.go
wonderflow: Just naming is `getApp` is good.

wonderflow (approved):
Generally LGTM. Ping @ryanzhang-oss  would you like to review this PR?

pkg/commands/init.go
```diff
@@ -70,8 +70,24 @@ func NewInitCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			o.IOStreams.Info("App Deployed Succeed")
-			//TODO(wonderflow) Wait for app running, and print trait info such as route, domain
```
wonderflow, line 74: by the way, I don't think this PR fixes this TODO

captainroy-hy: Yes, I will put it back.

Makefile
```diff
@@ -75,6 +75,7 @@ docker-push:
 e2e-setup:
 	ginkgo version
 	ginkgo -v -r e2e/setup
+	kubectl wait --for=condition=Ready pod -l app.kubernetes.io/name=vela-core -n vela-system --timeout=300s
```
captainroy-hy, line 78: @wonderflow @ryanzhang-oss  To make sure oam-controller pod is running before other e2e-test start, I employ this approach from oam-kubernetes-runtime project. Otherwise, tests related to interacting with oam-controller may fail or timeout.

wonderflow: Maybe `vela-core` we have changed this name?

captainroy-hy: currently, the value of `app.kubernetes.io/name` is still "vela-core", and `app.kubernetes.io/instance` is "kubevela"
let make it more rigorous as below:
`kubectl wait --for=condition=Ready pod -l app.kubernetes.io/name=vela-core,app.kubernetes.io/instance=kubevela -n vela-system --timeout=300s`

wonderflow (approved):

PR #332, opened 2020-09-27
use return instead of break

silenceper:
use `return` instead of `break` to break out of the outer loop


pkg/commands/logs.go
```diff
@@ -69,7 +69,7 @@ func Run(ctx context.Context, config *stern.Config, ioStreams cmdutil.IOStreams)
 			case str := <-logC:
 				ioStreams.Infonln(str)
 			case <-ctx.Done():
-				break
+				return
```
hongchaodeng, line 72: Good catch!

hongchaodeng (approved):
lgtm

pkg/commands/logs.go
silenceper: Consider adding some golang static inspection tools to github actions.

PR #335, opened 2020-09-28
setup core-controllers before e2e-test

captainroy-hy:
Some e2e tests cannot work correctly for there's no oam-core-controller reconciling resources.
e.g. `vela comp status`  need to get reconciled AppConfig object.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,8 @@ e2e-setup:
 	ginkgo version
 	ginkgo -v -r e2e/setup
 	bin/vela dashboard &
+	bin/vela install &
+	bin/manager &
 
 e2e-test:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -76,6 +76,8 @@ e2e-setup:
 	ginkgo version
 	ginkgo -v -r e2e/setup
 	bin/vela dashboard &
+	bin/vela install &
+	bin/manager &
```
wonderflow, line 80: vela install will run `vela-core`, while `bin/manager` will also run `vela-core`, they may conflict

captainroy-hy:
solved in #329. closed.

PR #352, opened 2020-10-02
Hack the Helm install client and improve the helm chart

ryanzhang-oss:
1. Create a factory to convert a controller client to a kube CLI client so the helm install would work in cluster
2. Move the cert-manager to be a dependency chart
3. Improve the Chart details like randomize some names


api/types/types.go
```diff
@@ -1,8 +1,8 @@
 package types
 
 const (
-	DefaultOAMNS               = "oam-system"
-	DefaultOAMReleaseName      = "vela-core"
+	DefaultOAMNS               = "vela-system"
```
wonderflow, line 4: Why change the namespace

api/types/types.go
```diff
@@ -1,8 +1,8 @@
...
+	DefaultOAMReleaseName      = "kube"
```
wonderflow, line 5: Why name it kube

wonderflow (approved):
Generally LGTM

api/types/types.go
zzxwill, line 5: I prefer the old OAM release as users might `kube` might be mixed up with kube-xxx releases:) And I like the new OAM namespace.

api/types/types.go
ryanzhang-oss: Is vela part of "oam-system"

api/types/types.go
ryanzhang-oss: Isn't the official name "kubeVela"?

ryanzhang-oss:
btw, I tried to move the cert-manager to be a dependency and it works if I just use helm. It doesn't work when I tried to incorporate that into `vela install`

PR #355, opened 2020-10-05
Implement Autoscaler Trait

zzxwill:
Support Cron type and resource usages (cpu) type
scalling


zzxwill:
Currently there are two scalers, KEDA ScaledObject scaler (will generate a HPA scaler) and HPA scaler, working separately, which of metrics should be merge into one HPA scaler.

By investigating how to merge metrics, some internal function of KEDA needs to be hacked, like https://github.com/kedacore/keda/blob/v2/controllers/hpa.go#L43. As we are about to use the KEDA versioned modified by @yanxun, I plan to work with him on the integration on Oct. 9 after the holiday.

samples/scalers/standard_v1alpha2_autoscaler.yaml
```diff
@@ -0,0 +1,27 @@
...
+        duration: 2h                 # required. 持续时长；start hour + duration < 24
+        days: ["Wednesday", "Saturday"]  # optional. In which days the condition will take effect
+        replicas: 5                   # optional.
+        timezone: Asia/Shanghai       # optional. time zone
+    - name: resource-example          # optional. 如果为空，自动生成，需要 mutation/validation
+        enabled: true                   # optional. whether to take effect. default is true
+        type: cpu                       # required. cpu/memory/storage/ephemeral-storage、cron、pps、qps/rps、custom
+        condition:
+          target: 1                     # required. 阈值，扩容临界值  resource --> 85%
+  targetWorkload:   # workloadRef                   # set by users or be auto-patched by outsiders, like OAM
```
resouer, line 23: Why not use `workloadReference`? so you don't need to clarify this in trait definition?

samples/scalers/cron/standard_v1alpha2_autoscaler.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: standard.oam.dev/v1alpha1
```
resouer, line 1: A plain autoscaler CR is not enough to prove this trait works.

Should include a full OAM sample, and the trait should be tested against:

1. (Must have) A abstract workload such as `Containerized` (will be named as `PodScpecableWorkload` soon) in kubevela.
2. (Must have) A plain built-in workload such as `Deployment`.
3. (Optional) A custom workload such as `CloneSet`.

For fast development we didn't require automatic e2e test suites but manual e2e test (i.e. self-proven yamls) is MUST have.

resouer (changes requested):

samples/scalers/standard_v1alpha2_autoscaler.yaml
resouer: Note that I am personally OK to use `targetWorkload`, this could be a chance to show how to define workload reference in workload definition.

resouer:
A slight off topic but important question: how do end users know the **current replicas** of the Component? We can't force users to `kubectl get deploy` obviously.

One option on my head is this trait patch a **current replicas** status on Component, WDYT? /cc @ryanzhang-oss @wonderflow 

In my mind, one of the value the kubevela renders Application/AppFile into Component and AppConfig instead of raw k8s objects is these building block objects is the best place to aggregate status in user facing/platform friendly primitives.

resouer:
> Currently there are two scalers, KEDA ScaledObject scaler (will generate a HPA scaler) and HPA scaler, working separately, which of metrics should be merge into one HPA scaler.

@zzxwill Could you explain why don't we make KEDA as source of truth? Why should user still care about HPA scaler?

api/v1alpha1/autoscaler_types.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// MinReplicas is the maximal replicas
+	// +optional
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// Triggers lists all triggers
+	Triggers []Trigger `json:"triggers"`
+
+	// TargetWorkload specify the workload which is about to be scaled
+	TargetWorkload TargetWorkload `json:"targetWorkload,omitempty"`
```
wonderflow, line 102: Please align with `TypedReference`  as we used in other trait.

api/v1alpha1/autoscaler_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// +optional
+	APIVersion string `json:"apiVersion,omitempty"`
+	// +optional
+	Kind string `json:"kind,omitempty"`
+}
+
+// AutoscalerStatus defines the observed state of Autoscaler
+type AutoscalerStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
```
wonderflow, line 117: please add status for this scaler

go.mod
```diff
@@ -5,50 +5,69 @@ go 1.13
...
-	github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible // https://github.com/kubernetes/client-go/issues/628
+	github.com/Azure/go-autorest => github.com/Azure/go-autorest v14.2.0+incompatible // https://github.com/kubernetes/client-go/issues/628
 	github.com/wercker/stern => github.com/oam-dev/stern v1.13.0-alpha
 	// clint-go had a buggy release, https://github.com/kubernetes/client-go/issues/749
-	k8s.io/client-go => k8s.io/client-go v0.18.6
+	k8s.io/client-go => k8s.io/client-go v0.18.8
+)
+
+// Temporary use the replace due to issue https://github.com/kedacore/keda/discussions/1198
+replace github.com/kedacore/keda => github.com/zzxwill/keda v1.5.1
```
wonderflow, line 65: Like the issue said, use `import github.com/kedacore/keda COMMIT_HASH` will fix

go.mod
```diff
@@ -5,50 +5,69 @@ go 1.13
...
+
+// fix `make` issue
```
wonderflow, line 67: what issue?

go.mod
```diff
@@ -5,50 +5,69 @@ go 1.13
...
+replace github.com/Sirupsen/logrus v1.7.0 => github.com/sirupsen/logrus v1.7.0
+
+replace (
+	github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
+	github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
```
wonderflow, line 72: why need replace?

go.sum
```diff
@@ -2033,36 +2667,53 @@ mvdan.cc/gofumpt v0.0.0-20200709182408-4fd085cb6d5f/go.mod h1:9VQ397fNXEnF84t90W
...
-sigs.k8s.io/yaml v1.1.0 h1:4A07+ZFc2wgJwo8YNlQpr1rVlgUDlxXHhPJciaPY5gs=
 sigs.k8s.io/yaml v1.1.0/go.mod h1:UJmg0vDUVViEyp3mgSv9WPwZCDxu4rQW1olrI1uml+o=
 sigs.k8s.io/yaml v1.2.0 h1:kr/MCeFWJWTwyaHoR9c8EjH9OumOmoF9YGiZd7lFm/Q=
 sigs.k8s.io/yaml v1.2.0/go.mod h1:yfXDCHCao9+ENCvLSE62v9VSji2MKu5jeNfTrofGhJc=
 software.sslmate.com/src/go-pkcs12 v0.0.0-20180114231543-2291e8f0f237/go.mod h1:/xvNRWUqm0+/ZMiF4EX00vrSCMsE4/NHb+Pt3freEeQ=
 sourcegraph.com/sourcegraph/appdash v0.0.0-20190731080439-ebfcffb1b5c0/go.mod h1:hI742Nqp5OhwiqlzhgfbWU4mW4yO10fP+LoT9WOswdU=
 sourcegraph.com/sqs/pbtypes v0.0.0-20180604144634-d3ebe8f20ae4/go.mod h1:ketZ/q3QxT9HOBeFhu6RdvsftgpsbFHBF5Cas6cDKZ0=
+sourcegraph.com/sqs/pbtypes v1.0.0/go.mod h1:3AciMUv4qUuRHRHhOG4TZOB+72GdPVz5k+c648qsFS4=
 vbom.ml/util v0.0.0-20160121211510-db5cfe13f5cc/go.mod h1:so/NYdZXCz+E3ZpW0uAoCj6uzU2+8OWDFv/HxUSs7kI=
+vbom.ml/util v0.0.0-20180919145318-efcd4e0f9787/go.mod h1:so/NYdZXCz+E3ZpW0uAoCj6uzU2+8OWDFv/HxUSs7kI=
```
wonderflow, line 2719: please run `go mod tidy` to remove unused dependency.

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,129 @@
...
+	kedav1alpha1 "github.com/kedacore/keda/api/v1alpha1"
+
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/go-logr/logr"
+	"github.com/oam-dev/kubevela/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	restclient "k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
wonderflow, line 42: nit

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+const (
+	SpecWarningTargetWorkloadNotSet = "Spec.targetWorkload is not set"
```
wonderflow, line 46:
```suggestion
	SpecWarningTargetWorkloadNotSet = "spec.targetWorkload is not set"
```

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,129 @@
...
+	SpecWarningStartAtTimeFormat    = "startAt is not in the right format, which should be like `12:01`"
+)
+
+var (
+	scaledObjectKind       = reflect.TypeOf(kedav1alpha1.ScaledObject{}).Name()
+	scaledObjectAPIVersion = "keda.k8s.io/v1alpha1"
```
wonderflow, line 52: why not use `SchemeGroupVersion` in https://github.com/kedacore/keda/blob/v2/api/v1alpha1/groupversion_info.go#L33 ?

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,129 @@
...
+	if err := r.Get(r.ctx, req.NamespacedName, &scaler); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Autoscaler is deleted")
+		}
+		return ReconcileWaitResult, client.IgnoreNotFound(err)
+	}
+	log.Info("retrieved trait Autoscaler", "APIVersion", scaler.APIVersion, "Kind", scaler.Kind)
+
+	// find ApplicationConfiguration to record the event
+	// comment it as I don't want Autoscaler to know it's in OAM context
```
wonderflow, line 84: why?

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// TODO(zzxwill) compare two scalers and adjust the target replicas @炎寻
+
+	if err := r.scaleByHPA(scaler, namespace, log); err != nil {
+		return ReconcileWaitResult, err
+	}
+
+	if err := r.scaleByKEDA(scaler, namespace, log); err != nil {
+		return ReconcileWaitResult, err
+	}
```
wonderflow, line 100: It's hard to understand why we need to scaleByKEDA after scaleByHPA.

pkg/controller/v1alpha1/autoscaler/keda.go
```diff
@@ -0,0 +1,176 @@
...
+	kedav1alpha1 "github.com/kedacore/keda/api/v1alpha1"
+
+	"errors"
+	"strings"
+
+	"github.com/go-logr/logr"
+	kedatype "github.com/kedacore/keda/pkg/generated/clientset/versioned/typed/keda/v1alpha1"
+	"github.com/oam-dev/kubevela/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
```
wonderflow, line 20: nit

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
wonderflow: As far as I can see in the code, whether it should be scale by HPA or KEDA depends on the trigger's type, so why not separate the logic there?

something like below:

```
switch trigger.type{
case type1:
    scaleByHPA
case typ2:
    scaleByKEDA
}

```

pkg/controller/v1alpha1/autoscaler/keda.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if obj, err := kedaClient.ScaledObjects(namespace).Create(r.ctx, &scaleObj, metav1.CreateOptions{}); err != nil && !apierrors.IsAlreadyExists(err) {
+		log.Error(err, "failed to create KEDA ScaledObj", "ScaledObject", obj)
+		return err
+	}
+
+	//if obj, err := kedaClient.ScaledObjects(namespace).Update(r.ctx, &scaleObj, metav1.UpdateOptions{}); err != nil {
+	//	log.Error(err, "failed to create KEDA ScaledObj", "ScaledObject", obj)
+	//	return err
+	//}
```
wonderflow, line 173: nit

samples/scalers/cron/standard_v1alpha2_autoscaler.yaml
wonderflow: Totally agree with @resouer , we MUST give a full OAM sample instead of the plain autoscaler CR sample.

samples/scalers/resource/standard_v1alpha2_autoscaler.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: example-scaler
+spec:
+  minReplicas: 1                      # optional, Defaults: 1
+  maxReplicas: 4                      # optional, cannot be less that minReplicas
+  triggers:
+    - name: resource-example          # optional. 如果为空，自动生成，需要 mutation/validation
+      enabled: true                   # optional. whether to take effect. default is true
+      type: cpu                       # required. cpu/memory/storage/ephemeral-storage、cron、pps、qps/rps、custom
+      condition:
+        target: 2                     # required. 阈值，扩容临界值  resource --> 85%
```
wonderflow, line 13: Don't mix two languages in explanation

samples/scalers/standard_v1alpha2_autoscaler.yaml
wonderflow: `targetWorkload`  is OK, but I don't like the example emphasis it here, in fact, by using with OAM, users even don't need to understand this field.

wonderflow (changes requested):

samples/scalers/standard_v1alpha2_autoscaler.yaml
zzxwill: At the very beginning, I'd like the name `targetWorkload` to express the meaning of `ownerReferences` in  KEDA, and `scaleTargetRef` in HPA. 
Then I made a mistake. In old version of KEDA, only deployment and job resource are able to be scaled (in [details](https://keda.sh/docs/1.5/concepts/#custom-resources-crd)), I think I have to set `targetWorkload` as the child resource of `workloadRef` resource, like `deployment` of `containerized`. In this situation,  `targetWorkload` and `workloadRef` are not the same. But I found it won't work as the replicas of child resource would be controlled by both `workloadRef` resource and KEDA per my PoC. Only the `workloadRef` resource should be controlled by autoscaler controller. If it's not supported to be scaled, we should focus on solving it, instead of trying to controller its child resource.
As @wonderflow said, in OAM context, users won't need to pay attention to the target workload.

In all, how about we just using `workloadRef` directly? Or there will be two fields in Autoscaler spec https://github.com/oam-dev/kubevela/pull/355/files#diff-3b9f06f31abab8d8f4a1660d1fe3b873R120-R123.

go.mod
zzxwill: https://github.com/docker/distribution/issues/2406

samples/scalers/standard_v1alpha2_autoscaler.yaml
resouer: > But I found it won't work as the replicas of child resource would be controlled by both workloadRef resource and KEDA per my PoC

It's the reason why there's no `replicas` field in `ContainerizedWorkload` and user need to use `manualscaler` trait to control the child resource's replicas.

@ryanzhang-oss We need to change the `replicas` field to `initReplicas` in `PodSpecWorkload`. It should only take effect during first deploy time and autoscaler trait controller will take over later.

@zzxwill for trait side, the approach I think OAM runtime currently take is `workloadRef` always ref the workload instance, and trait controller is responsible for parsing `workloadRef` into child resource required by the trait plugin.

zzxwill: > @ryanzhang-oss Not sure if we want to change `replicas` field to `initReplicas` in `PodSpecWorkload`? It should only take effect during first deploy time and autoscaler trait controller will take over later.

It would be better for the change. Currently in order to stop the replicas of `PodSpecWorkload` reverting to default `spec.replicas`, I set `ownerReference.controller` to be true just for Autoscaler trait, and false for `PodSpecWorkload`.  The scaling is successful, but this leads to error message in `ApplicationConfiguration` log. /cc @ryanzhang-oss 
```
  Warning  cannot apply the deployment  31m                   PodSpecWorkload                            Deployment.apps "component-scaler" is invalid: metadata.ownerReferences: Invalid value: []v1.OwnerReference{v1.OwnerReference{APIVersion:"standard.oam.dev/v1alpha1", Kind:"PodSpecWorkload", Name:"component-scaler", UID:"fbab3f47-7f23-46ab-abdf-98a78513d9d5", Controller:(*bool)(0xc008ed6ab0), BlockOwnerDeletion:(*bool)(0xc008ed6ab1)}, v1.OwnerReference{APIVersion:"standard.oam.dev/v1alpha1", Kind:"Autoscaler", Name:"trait-scaler", UID:"fd168b3b-a70d-4edf-8d81-525af31037c0", Controller:(*bool)(0xc008ed6ab2), BlockOwnerDeletion:(*bool)(0xc008ed6ab3)}}: Only one reference can have Controller set to true. Found "true" in references for component-scaler and trait-scaler
```

> @zzxwill for trait side, the approach I think OAM runtime currently take is `workloadRef` always ref the workload instance, and trait controller is responsible for parsing `workloadRef` into child resource required by the trait plugin.

Done as what you suggested. Thanks.

samples/scalers/cron/standard_v1alpha2_autoscaler.yaml
zzxwill: > 1. (Must have) A abstract workload such as `Containerized` (will be named as `PodSpecWorkload` soon) in kubevela.
> 2. (Must have) A plain built-in workload such as `Deployment`.
> 3. (Optional) A custom workload such as `CloneSet`.

1. supported.  Example manifests in folder `samples/autoscalers/cron/PodSpecWorkload`
2. In progress
3. Pending

go.mod
zzxwill: Add a link in go.mod. Thanks.

go.mod
zzxwill: To fix build issue https://github.com/docker/distribution/issues/2406

samples/scalers/cron/standard_v1alpha2_autoscaler.yaml
zzxwill: 2. Supported Deployment. Example manifests in `samples/autoscalers/cron/deployment`.

But for both `PodSpecWorkload` and `Deployment`, there are warnings in ApplicationConfiguration events as I set ownerRef of workloads (child resource `Deployment` of `PodSpecWorkload`, and `Deployment` itself) to AutoScaler trait. I haven't figured out how to fix it.

- PodSpecWorkload
```
  Warning  cannot apply the deployment  31m                   PodSpecWorkload                            
Deployment.apps "component-scaler" is invalid: metadata.ownerReferences: Invalid value: 
[]v1.OwnerReference{v1.OwnerReference{APIVersion:"standard.oam.dev/v1alpha1", Kind:"PodSpecWorkload", 
Name:"component-scaler", UID:"fbab3f47-7f23-46ab-abdf-98a78513d9d5", Controller:(*bool)(0xc008ed6ab0), 
BlockOwnerDeletion:(*bool)(0xc008ed6ab1)}, v1.OwnerReference{APIVersion:"standard.oam.dev/v1alpha1", 
Kind:"Autoscaler", Name:"trait-scaler", UID:"fd168b3b-a70d-4edf-8d81-525af31037c0", Controller:(*bool)
(0xc008ed6ab2), BlockOwnerDeletion:(*bool)(0xc008ed6ab3)}}: Only one reference can have Controller set to true. 
Found "true" in references for component-scaler and trait-scaler

```

- Deployment
```
cannot apply workload "component-scaler": existing object is not controlled by UID "54cb9f2a-98e3-4f85-8f1d-47522035533e"
```

resouer: You can't let controller A and B control another resource at same time. The trait controller should "take over" in this case. I am not sure if it should "give back" after scaling finished, manual scaler trait or @ryanzhang-oss may have more input here.

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,216 @@
...
+	ownerReference := metav1.OwnerReference{
+		APIVersion:         scaler.APIVersion,
+		Kind:               scaler.Kind,
+		UID:                scaler.GetUID(),
+		Name:               scaler.Name,
+		Controller:         pointer.BoolPtr(true),
+		BlockOwnerDeletion: pointer.BoolPtr(true),
+	}
+
+	// Reference the logic of ManualScalerTrait in OAM Kubernetes Runtime
```
wonderflow, line 121: Don't you feel weird writing this here?

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,216 @@
...
+	for _, res := range resources {
+		resPatch := client.MergeFrom(res.DeepCopyObject())
+		refs := res.GetOwnerReferences()
+		for i, r := range refs {
+			if *r.Controller {
+				refs[i].Controller = pointer.BoolPtr(false)
+			}
+		}
+		refs = append(refs, ownerReference)
+		res.SetOwnerReferences(refs)
```
wonderflow, line 142: why add scaler to ownerRef of childResource?

pkg/controller/v1alpha1/autoscaler/hpa.go
```diff
@@ -0,0 +1,80 @@
...
+					Name:      scalerName,
+					Namespace: namespace,
+					OwnerReferences: []metav1.OwnerReference{
+						{
+							APIVersion:         scaler.APIVersion,
+							Kind:               scaler.Kind,
+							UID:                scaler.GetUID(),
+							Name:               scalerName,
+							Controller:         pointer.BoolPtr(true),
+							BlockOwnerDeletion: pointer.BoolPtr(true),
```
wonderflow, line 55: I see the same logic more than twice, maybe it deserves a function

pkg/controller/v1alpha1/autoscaler/keda.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/crossplane/crossplane-runtime/pkg/event"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/utils/pointer"
+
+	kedav1alpha1 "github.com/kedacore/keda/api/v1alpha1"
+
+	"errors"
+	"strings"
+
```
wonderflow, line 16: nit

pkg/controller/v1alpha1/autoscaler/keda.go
```diff
@@ -0,0 +1,183 @@
...
+			splitDuration := strings.Split(duration, "h")
+			if len(splitDuration) != 2 {
+				log.Error(err, "duration hour is not in the right format, like `12h`.", "duration", duration)
+				return err
+			}
+			if durationHour, err = strconv.Atoi(splitDuration[0]); err != nil {
+				log.Error(err, "duration hour is not in the right format, like `12h`.", "duration", duration)
+				return err
+			}
```
wonderflow, line 83: [`time.ParseDuration`](https://golang.org/pkg/time/#ParseDuration)

pkg/controller/v1alpha1/autoscaler/keda.go
```diff
@@ -0,0 +1,183 @@
...
+	triggers := scaler.Spec.Triggers
+	scalerName := scaler.Name
+	targetWorkload := scaler.Spec.TargetWorkload
+
+	// kedaScalerFlag marks whether KEDA Scaler should be applied
+	var kedaScalerFlag = false
+
+	var kedaTriggers []kedav1alpha1.ScaleTriggers
+	for _, t := range triggers {
+		if t.Type == CronType {
```
wonderflow, line 42: Can we write or **find** a packge for this cron related thing?  I hope to see enough tests for this package, as parse the time\duration\zone like this made me nervous.

pkg/controller/v1alpha1/autoscaler/keda.go
```diff
@@ -0,0 +1,183 @@
...
+			ScaleTargetRef:  &scaleTarget,
+			MinReplicaCount: minReplicas,
+			MaxReplicaCount: maxReplicas,
+			Triggers:        kedaTriggers,
+		},
+	}
+
+	obj, err := kedaClient.ScaledObjects(namespace).Get(r.ctx, scalerName, metav1.GetOptions{})
+	if err != nil {
+		log.Info("KEDA ScaledObj doesn't exist", "ScaledObjectName", scalerName)
```
wonderflow, line 170: Please make sure this error is not exist error.

wonderflow (changes requested):

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	r.ctx = context.Background()
+	r.record = event.NewAPIRecorder(mgr.GetEventRecorderFor("Autoscaler")).
+		WithAnnotations("controller", "Autoscaler")
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.Autoscaler{}).
+		Complete(r)
+}
+
+func (r *AutoscalerReconciler) buildConfig() error {
```
ryanzhang-oss, line 193: Why do we need this in the controller? The config should be passed in through the constructer

pkg/controller/v1alpha1/autoscaler/keda.go
ryanzhang-oss: second that, the following 100 lines of code can be very tricky to test. Please break them up and test them individually at least.

api/v1alpha1/containerized_types.go
```diff
@@ -28,7 +28,7 @@ type PodSpecWorkloadSpec struct {
 	// Replicas is the desired number of replicas of the given podSpec.
 	// These are replicas in the sense that they are instantiations of the same podSpec.
 	// If unspecified, defaults to 1.
-	Replicas *int32 `json:"replicas"`
+	Replicas *int32 `json:"replicas,omitempty"`
```
zzxwill, line 31: @ryanzhang-oss I set this field as optional to fix the issue https://github.com/oam-dev/kubevela/pull/355#discussion_r502791069. Now I won't set Controller to true when setting the workload's `ownerRef` to trait. In this way, the replicas of workload won't be refreshed to the original `spec.replicas` by workload when the filed is not set in manifest. For the sample, please refer to folder `samples/autoscalers/PodSpecWorkload`. Please review it, thanks.

pkg/controller/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// find the resource object to record the event to, default is the parent appConfig.
+	eventObj, err := util.LocateParentAppConfig(r.ctx, r.Client, &scaler)
+	if eventObj == nil {
+		// fallback to workload itself
+		log.Error(err, "Failed to find the parent resource", "Autoscaler", scaler.Name)
+		eventObj = &scaler
+	}
+
+	// Fetch the deployment instance to which the trait refers to
```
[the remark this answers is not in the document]
zzxwill: Updated the comment, resource instance like deployment/statefulset/job are supported, or those workload who implements status sub-resource.

api/v1alpha1/autoscaler_types.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// MinReplicas is the maximal replicas
+	// +optional
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// Triggers lists all triggers
+	Triggers []Trigger `json:"triggers"`
+
+	// TargetWorkload specify the workload which is about to be scaled
+	TargetWorkload TargetWorkload `json:"targetWorkload,omitempty"`
```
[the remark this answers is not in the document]
zzxwill: `workloadReference` is same as that of OAM concept. If the workload could be directly controlled by the trait to scale, the child resource of the workload is `TargetWorkload`.

pkg/controller/v1alpha1/autoscaler/hpa.go
zzxwill: Correct, Thanks. Now `hpa.go` has been deleted.

wonderflow:
I'm working on this PR now.

wonderflow:
closing as we have #438 for the following fixes. Let's move discussion there

PR #356, opened 2020-10-06
rename containerized to podspecworkload

ryanzhang-oss:
Rename `containerized` to `podspecworkload`


zzxwill (approved):
/lgtm

wonderflow:
\cc @captainroy-hy  health scope check should also update since this change, right?

captainroy-hy:
> \cc @captainroy-hy health scope check should also update since this change, right?

That's right, I will update health scope.

PR #366, opened 2020-10-09
add description for workload and trait

### LINKED ISSUE #349: [BUG] Don't show "definition" in vela workloads
`containerizeds.standard.oam.dev` is the schema for "Backend Worker" workload type, not definition, also, it should not be exposed to users of vela.

It should be sth like:

```console
$ vela workloads
NAME         	DESCRIPTION                                                  
backend      	Backend worker without ports exposed          
task         	One-time off task or job
webservice   	Long running service with ports exposed
```

wonderflow:
fixes #349

```
➜  ~ vela workloads
NAME      	DESCRIPTION
backend   	Backend worker without ports exposed
task      	One-time off task or job
webservice	Long running service with ports exposed
➜  ~ vela traits
NAME    	DESCRIPTION                       	APPLIES TO
metrics 	Add metric monitoring for workload
route   	Add a route for workload          	webservice
scale   	Scale replica for workload
```

### DIFF
--- a/api/types/capability.go
+++ b/api/types/capability.go
@@ -46,6 +46,7 @@ type Capability struct {
 	CrdName        string      `json:"crdName,omitempty"`
 	Center         string      `json:"center,omitempty"`
 	Status         string      `json:"status,omitempty"`
+	Description    string      `json:"description,omitempty"`
 
 	//trait only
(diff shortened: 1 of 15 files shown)


zzxwill (approved):
/lgtm

PR #373, opened 2020-10-11
fix check resource exists

### LINKED ISSUE #347: [BUG]When error is returned, it does not necessarily mean that the resource does not exist
https://github.com/oam-dev/kubevela/blob/6ede8601b4025be90b3b3578f913815b26b99276/pkg/commands/util/init.go#L46-L50

https://github.com/oam-dev/kubevela/blob/6ede8601b4025be90b3b3578f913815b26b99276/pkg/commands/util/init.go#L61-L63

When error is returned, it does not necessarily mean that the resource does not exist.

should use `errors.IsNotFound` to check.

silenceper:
fix #347

### DIFF
--- a/pkg/commands/util/init.go
+++ b/pkg/commands/util/init.go
@@ -7,9 +7,9 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
@@ -67,8 +67,8 @@ func NewNamespace(c client.Client, namespace string) error {
 }
 
 // DoesCoreCRDExist check CRD exist
-func DoesCRDExist(cxt context.Context, c client.Client, object runtime.Object) (bool, error) {
-	err := c.List(cxt, object, &client.ListOptions{})
+func DoesCRDExist(cxt context.Context, c client.Client, crdName string) (bool, error) {
+	err := c.Get(cxt, types.NamespacedName{Name:crdName}, &apiextensions.CustomResourceDefinition{})
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return false, nil
--- a/pkg/commands/util/init.go
+++ b/pkg/commands/util/init.go
@@ -7,6 +7,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/commands/util/init.go
```diff
@@ -43,10 +44,18 @@ func GetKubeConfig() string {
 	return filepath.Join(HomeDir(), ".kube", "config")
 }
 
-func IsNamespaceExist(c client.Client, namespace string) bool {
+// DoesNamespaceExist check namespace exist
```
zzxwill, line 47: I prefer to keep the previous logic as it is. `IsNamespaceExist` or `DoesNamespaceExist`, as a very simple function to check whether the namespace is existed, we don't care the error message a lot. IMHO, returning a true or false is better.

silenceper: How to deal with the error situation, such as a network error may occur, choose to ignore it?

pkg/commands/util/init.go
```diff
@@ -43,10 +44,17 @@ func GetKubeConfig() string {
...
```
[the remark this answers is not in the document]
silenceper: no problem.

silenceper:
- add tests
- fix `DoesCRDExist`, It may be more appropriate to check exists based on the crd name(This method is not used)

pkg/commands/util/init_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var (
+	scheme *runtime.Scheme
+)
+
+func init() {
```
wonderflow, line 21: we should not use `init` in test package

silenceper: 😅 I refer to the implementation here and keep it consistent.
I will remove it.

https://github.com/oam-dev/kubevela/blob/64d77656cf5159ecc72493710643de76461c95f7/pkg/controller/dependency/install_test.go#L46-L51

wonderflow: yes, we should remove both of them. Thanks for pointing out.

pkg/controller/dependency/install.go
```diff
@@ -84,9 +84,15 @@ func installHelmChart(client client.Client, chart []byte, log logr.Logger) error
 	log.Info("install helm char", "chart name", helmChart.Name)
 	// create the namespace
 	if helmChart.Namespace != types.DefaultAppNamespace {
-		if len(helmChart.Namespace) > 0 && !cmdutil.IsNamespaceExist(client, helmChart.Namespace) {
-			if err = cmdutil.NewNamespace(client, helmChart.Namespace); err != nil {
-				return errors.Wrap(err, "failed to create the namespace")
+		if len(helmChart.Namespace) > 0 {
+			exist, err := cmdutil.DoesNamespaceExist(client, helmChart.Namespace)
+			if err != nil {
+				return nil
```
wonderflow, line 90:
```suggestion
				return err
```
？

silenceper: sorry, typo

wonderflow (approved):

ryanzhang-oss (approved):
LGTM

pkg/commands/util/init.go
zzxwill: +1 for your change. Thanks.

Issue #371, opened 2020-10-10
[Feature] oam-runtime pod crash when webhook secret is not ready


captainroy-hy:
**Is your feature request related to a problem? Please describe.**

When I work on #292 and e2e-test related to interacting with oam-runtime, each time set up a new vela runtime, the runtime pod always restarts several times before running stably.

As the below log from the runtime containers shows, the runtime startup crash for missing tls cert, because the secret (from cert-manager-webhook) volume is not mounted to the pod yet. The pod falls into `failed` phase from `running` phase. So we cannot use the phase of runtime pod to check it's ready or not. Even though the pod will run into stable finally once the secret volume is ready. 
```
2020-10-10T01:32:38.978Z    INFO    controller-runtime.webhook.webhooks    starting webhook server
2020-10-10T01:32:38.978Z    DEBUG    controller-runtime.manager    non-leader-election runnable finished    {"runnable type": "*webhook.Server"}
2020-10-10T01:32:38.978Z    ERROR    vela-runtime    problem running manager    {"error": "tls: failed to find any PEM data in certificate input"}
github.com/go-logr/zapr.(*zapLogger).Error
    /go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128
main.main
    /workspace/main.go:150
runtime.main
    /usr/local/go/src/runtime/proc.go:203
```

**Describe the solution you'd like**

Possible solution:  Add a looping check on webhook-cert-dir in /cmd/core/main.go. Only continue setuping controller manager when the secret volume is mounted successfully .

ryanzhang-oss:
The main.go is running inside the pod that needs the secret so it's too late for the main.go to check. We need to add some logic in the `vela install`

captainroy-hy:
fixed

PR #379, opened 2020-10-12
add liveness & readiness probes

captainroy-hy:
#371 
- Controller-runtime allows [adding liveness & readiness probes for ctrl-mgr.](https://github.com/kubernetes-sigs/controller-runtime/pull/419) Use probes to help determining the manager is ready to go.
- Add a looping check for webhook secrets volume mounted before start mgr, which can avoid unnecessary pod restart.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ all: build
 
 # Run tests
 test: fmt vet lint
-	go test ./pkg/... -coverprofile cover.out
+	go test ./pkg/... ./cmd/... -coverprofile cover.out
 
 # Build manager binary
 build: fmt vet lint
@@ -78,7 +78,7 @@ docker-push:
 e2e-setup:
 	ginkgo version
 	ginkgo -v -r e2e/setup
-	kubectl wait --for=condition=Ready pod -l app.kubernetes.io/name=vela-core,app.kubernetes.io/instance=kubevela -n vela-system --timeout=600s
+	# kubectl wait --for=condition=Ready pod -l app.kubernetes.io/name=vela-core,app.kubernetes.io/instance=kubevela -n vela-system --timeout=600s
 	bin/vela dashboard &
 
 e2e-test:
--- a/charts/vela-core/templates/kubevela-controller.yaml
+++ b/charts/vela-core/templates/kubevela-controller.yaml
@@ -111,6 +111,7 @@ spec:
             - "--use-webhook=true"
             - "--webhook-port={{ .Values.webhookService.port }}"
             - "--webhook-cert-dir={{ .Values.certificate.mountPath }}"
+            - "--health-addr=:{{ .Values.healthCheck.port }}"
           {{ end }}
           image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
           imagePullPolicy: {{ quote .Values.image.pullPolicy }}
@@ -121,6 +122,21 @@ spec:
             - containerPort: {{ .Values.webhookService.port }}
               name: webhook-server
               protocol: TCP
+            - containerPort: {{ .Values.healthCheck.port }}
+              name: healthz
+              protocol: TCP
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: healthz
+            initialDelaySeconds: 90
+            periodSeconds: 5
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 90
+            periodSeconds: 5
           volumeMounts:
             - mountPath: {{ .Values.certificate.mountPath }}
               name: tls-cert-vol
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -1,10 +1,14 @@
 package main
 
 import (
+	"errors"
 	"flag"
+	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strconv"
+	"time"
 
 	monitoring "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
@@ -24,6 +28,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	velacore "github.com/oam-dev/kubevela/api/v1alpha1"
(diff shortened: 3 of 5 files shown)


captainroy-hy:
As for changes [here](https://github.com/oam-dev/kubevela/blob/c31aa1b57422d66c2a0359bd50b34122180f9de8/charts/vela-core/templates/kubevela-controller.yaml#L128), the e2e-test will certainly fail for it always uses oam-dev/vela-core:latest image...

cmd/core/main.go
```diff
@@ -32,7 +36,10 @@ import (
 	velawebhook "github.com/oam-dev/kubevela/pkg/webhook"
 )
 
-var scheme = runtime.NewScheme()
+var (
+	setupLog = ctrl.Log.WithName("vela-runtime")
```
wonderflow, line 40: let's name it vela-core or kubevela

cmd/core/main.go
```diff
@@ -99,18 +107,22 @@ func main() {
...
-		Scheme:             scheme,
-		MetricsBindAddress: metricsAddr,
-		LeaderElection:     enableLeaderElection,
-		LeaderElectionID:   "vela-runtime",
-		Port:               webhookPort,
-		CertDir:            certDir,
+		Scheme:                 scheme,
+		MetricsBindAddress:     metricsAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "vela-runtime",
```
wonderflow, line 113: setup a const variable please

cmd/core/main.go
```diff
@@ -151,3 +163,45 @@ func main() {
...
+func setupChecks(mgr ctrl.Manager) {
+	setupLog.Info("creating readiness/health check")
+	if err := mgr.AddReadyzCheck("ping", healthz.Ping); err != nil {
+		setupLog.Error(err, "unable to create ready check")
+		os.Exit(1)
+	}
+
+	if err := mgr.AddHealthzCheck("ping", healthz.Ping); err != nil {
+		setupLog.Error(err, "unable to create health check")
+		os.Exit(1)
```
wonderflow, line 177: return error in this function and exit outside will make it clearer

cmd/core/main.go
```diff
@@ -151,3 +163,45 @@ func main() {
...
+	}
+}
+
+// waitSecretVolume waits for webhook secret ready to avoid mgr running crash
+func waitSecretVolume(certDir string) {
+	timeout := 300 * time.Second
```
wonderflow, line 183: I don't think we need so much long time, the k8s itself has mechanism to restart. As I know, this function we added here is just to let the controller won't make the K8s restart backoff take too long. 

I think 60s is enough, right?

cmd/core/main.go
```diff
@@ -151,3 +163,45 @@ func main() {
...
+	interval := 2 * time.Second
+
+	start := time.Now()
+	for {
+		time.Sleep(interval)
+		if time.Since(start) > timeout {
+			setupLog.Error(errors.New("cannot get webhook secret "), "get webhook secret timeout")
+			os.Exit(1)
+		}
+		setupLog.Info("waiting admission webhook secret ... ")
```
wonderflow, line 193: This log is important, we should add some more information such as 'xxx times have checked', 'xxx seconds left to timeout'...

cmd/core/main.go
```diff
@@ -151,3 +163,45 @@ func main() {
...
+func waitSecretVolume(certDir string) {
+	timeout := 300 * time.Second
+	interval := 2 * time.Second
+
+	start := time.Now()
+	for {
+		time.Sleep(interval)
+		if time.Since(start) > timeout {
+			setupLog.Error(errors.New("cannot get webhook secret "), "get webhook secret timeout")
+			os.Exit(1)
```
wonderflow, line 191: exit outside

cmd/core/main.go
```diff
@@ -99,18 +107,22 @@ func main() {
...
+		CertDir:                certDir,
+		HealthProbeBindAddress: healthAddr,
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to create a controller manager")
 		os.Exit(1)
 	}
 
+	setupChecks(mgr)
+	waitSecretVolume(certDir)
```
wonderflow, line 124: please add an e2e test for this, thanks

cmd/core/main.go
captainroy-hy: got it.

cmd/core/main.go
captainroy-hy: Yes, that's right. But maybe 90s is okay. After testing several times, the average time is around 60-70s.

charts/vela-core/templates/kubevela-controller.yaml
```diff
@@ -121,6 +122,21 @@ spec:
...
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: healthz
+            initialDelaySeconds: 90
+            periodSeconds: 5
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
```
ryanzhang-oss, line 137: what does "port: healthz" mean?

cmd/core/main.go
```diff
@@ -151,3 +163,45 @@ func main() {
...
+	}
+
+	if err := mgr.AddHealthzCheck("ping", healthz.Ping); err != nil {
+		setupLog.Error(err, "unable to create health check")
+		os.Exit(1)
+	}
+}
+
+// waitSecretVolume waits for webhook secret ready to avoid mgr running crash
+func waitSecretVolume(certDir string) {
```
ryanzhang-oss, line 182: Can the pod start when the volume doesn't exist?

charts/vela-core/templates/kubevela-controller.yaml
captainroy-hy: I think it's brevity for "health check" as [pkg healthz](https://github.com/kubernetes/apiserver/blob/master/pkg/server/healthz/healthz.go) implies.

cmd/core/main.go
captainroy-hy: ths pod will be stuck in "containercreating" unless the volume is mounted successfully.

cmd/core/main.go
captainroy-hy: fixed

cmd/core/main.go
captainroy-hy: fixed

cmd/core/main.go
captainroy-hy: fixed

ryanzhang-oss (approved):

Makefile
```diff
@@ -78,7 +78,7 @@ docker-push:
 e2e-setup:
 	ginkgo version
 	ginkgo -v -r e2e/setup
-	kubectl wait --for=condition=Ready pod -l app.kubernetes.io/name=vela-core,app.kubernetes.io/instance=kubevela -n vela-system --timeout=600s
+	# kubectl wait --for=condition=Ready pod -l app.kubernetes.io/name=vela-core,app.kubernetes.io/instance=kubevela -n vela-system --timeout=600s
```
wonderflow, line 81: why comment?

captainroy-hy: CI use oam-dev/vela-core:**latest** image in e2e test but the latest image doesn't include changes in this PR. So the commented command will fail for cannot get ready condition pod. Since in #383 , I have disabled all e2e cases related to interacting with runtime controller, so it's okay to comment this command temporarily to let e2e test pass.

wonderflow (approved):

PR #390, opened 2020-10-14
Appfile: Extensible, User-friendly Application Config Format

### LINKED ISSUE #239: [Feature] vela up
Consider a `vela up` command to take `appfile.yaml` as input, similar to `docker-compose up`. 

This also help us to avoid mix usage in `vela app init`.

hongchaodeng:
fix #239 

Vela supports a user-friendly `docker-compose` style config format called `Appfile`. It allows you to define an application's workloads and traits with an opinionated, simplified API interface.

Additionally, the Appfile could be extended with more configurations by adding more capabilities to the OAM system.

For more details, please look at the design proposal in this PR as well.


hack/appfile/definitions/canary.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: canary
+spec:
+  definitionRef:
+    name: canary.standard.oam.dev
+  extension:
+    template: |
```
wonderflow, line 9: Shall we delete these empty file?

testapp/vela.yml
```diff
@@ -0,0 +1,45 @@
+version: "1.0-alpha.1"
+name: testapp
```
wonderflow, line 2: We should setup a `examples` dir for this app

pkg/commands/template.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/oam-dev/kubevela/api/types"
+	cmdutil "github.com/oam-dev/kubevela/pkg/commands/util"
+	mycue "github.com/oam-dev/kubevela/pkg/cue"
+)
+
+func NewTemplateCommand(c types.Args, ioStream cmdutil.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "template",
```
wonderflow, line 13: What does this command used for?

pkg/commands/up.go
```diff
@@ -0,0 +1,172 @@
...
+		o.IO.Infof("app has not been deployed, creating a new deployment...\n")
+	case err == nil:
+		o.IO.Infof("app existed, updating existing deployment...\n")
+	default:
+		return err
+	}
+	return o.apply(ac)
+}
+
+func (o *appfileOptions) apply(ac *v1alpha2.ApplicationConfiguration) error {
```
wonderflow, line 158: We currently have a package named application which is the old format of Appfile `https://github.com/oam-dev/kubevela/blob/master/pkg/application/app.go#L33`

We should merge them, because almost every command will result into change of `application` and finally takes effect there.

We should not setup a totally new way just for `vela up`. This will cause lots of confusion.

pkg/commands/template.go
wonderflow: who is your target user to create a template?

I assume the template is a system level configure, or at least defined in WorkloadDefinition/TraitDefinition level?

pkg/commands/up.go
```diff
@@ -0,0 +1,172 @@
...
+	cmd := exec.Command("kubectl", "apply", "-f", ".vela/deploy.yaml")
```
wonderflow, line 159: We have already implemented `kubectl apply` by using `client-go`, we can reuse that after merge.

https://github.com/oam-dev/kubevela/blob/master/pkg/application/run.go#L15

pkg/commands/up.go
```diff
@@ -0,0 +1,172 @@
...
+	out, err := cmd.CombinedOutput()
+	o.IO.Infof("deploying======\n%s\n", out)
+	if err != nil {
+		return err
+	}
+	o.IO.Infof("app has been deployed %s%s%s\n", emojiRocket, emojiRocket, emojiRocket)
+	o.IO.Infof("\tURL: http://%s/\n", o.Env.Domain)
+	o.IO.Infof("\tPort forward: vela port-forward %s <port>\n", ac.Name)
+	o.IO.Infof("\tSSH: vela exec %s\n", ac.Name)
+	o.IO.Infof("\tLogging: vela log %s\n", ac.Name)
```
wonderflow, line 169: We can't just write like this, it will not work.

wonderflow (changes requested):
Great work! 
The design doc is really good, please align with the existing Application object, thanks!

pkg/commands/template.go
hongchaodeng: I assume all template related stuff will be put under this command.
Currently just to show context def.

hongchaodeng: > I assume the template is a system level configure, or at least defined in WorkloadDefinition/TraitDefinition level?

Can you clarify or be more specific?

pkg/commands/up.go
hongchaodeng: See https://github.com/oam-dev/kubevela/issues/206

design/appfile-design.md
```diff
@@ -0,0 +1,337 @@
...
+  // +vela:cli:enabled=true
+  // +vela:cli:usage=specify commands to run in container
+  // +vela:cli:short=c
+  cmd: [...string]
+}
+```
+
+Here is the takeout:
+* The `parameter` defines the user input fields and is used to render final output with user input values. These fields will be exposed to users in Appfile.
+* The definition `#webservice` is used to tell the name of the template. This name is used to correlate workload and trait types to fields in Appfile.
```
resouer, line 176: Please clarify in the appfile side about how the developer pick workload type.

design/appfile-design.md
```diff
@@ -0,0 +1,337 @@
...
+
+```bash
+vela up
+```
+
+It will build container image, render deployment manifests in yaml, and apply them to the server.
+
+### Extensible Design
+
+The Appfile could be extended with more configurations by adding more capabilities to the OAM system. The config fields in Appfile are strongly correlated to the [capabilities system of OAM](https://github.com/oam-dev/kubevela/blob/master/DESIGN.md#capability-register-and-discovery) – Config fields are registered in the capabilities system and exposed via a [CUE template](https://cuelang.org/).
```
resouer, line 72: Could you please also clarify how to express multiple components in single appfile?

design/appfile-design.md
```diff
@@ -0,0 +1,337 @@
...
+  - [Templating](#templating)
+  - [CLI/UI interoperability](#cliui-interoperability)
+  - [vela up](#vela-up)
+- [Examples](#examples)
+
+## Intro
+
+Vela supports a user-friendly `docker-compose` style config format called `Appfile`. It allows you to define an application's workloads and traits with an opinionated, simplified API interface.
+
+Here's an example to deploy a NodeJS express service:
```
resouer, line 22: Could you also clarify a bit whether this appfile must be put under source code repo of the app or it's flexible (this is related to the `build` section).

resouer (commented):
Some details need to be added. Also ref: https://github.com/oam-dev/kubevela/issues/399

design/appfile-design.md
hongchaodeng: This is clarified in the `type` field below.

design/appfile-design.md
hongchaodeng: This is clarified in `vela up` section

design/appfile-design.md
hongchaodeng: The proposal suggests a `services` field which implicitly implies multi components.
I think adding an example for it will be good enough.

wonderflow (approved):

wonderflow:
Merging it first, and we're going to refactor the rest part together in the following PRs.

PR #421, opened 2020-10-21
fix codecov

hongchaodeng:


wonderflow (approved):

hack/unit_test.sh
```diff
@@ -1,12 +1,6 @@
...
-echo "" > coverage.txt
 
-for d in $(go list ./pkg/... ./cmd/...); do
-  go test -race -coverprofile=profile.out -covermode=atomic $d
-  if [ -f profile.out ]; then
-    cat profile.out >> coverage.txt
-    rm profile.out
-  fi
-done
+go test -race -coverprofile=coverage.txt -covermode=atomic \
```
wonderflow, line 5: BTW, why not move this line to makefile?

hongchaodeng: moved to makefile

Issue #411, opened 2020-10-20
[Feature] support -f argument to specify vela.yml


wonderflow:
**Is your feature request related to a problem? Please describe.**

Currently, `vela up` will only find local dir for file `vela.yml` , we should support `vela up -f <my-path-to-vela.yml>`

PR #422, opened 2020-10-22
add support for `vela up` specify file path

mosesyou:
ref: #411

### DIFF
--- a/pkg/commands/up.go
+++ b/pkg/commands/up.go
@@ -25,6 +25,7 @@ import (
 )
 
 var (
+	appFilePath string
 	emojiRocket = emoji.Sprint(":rocket")
 )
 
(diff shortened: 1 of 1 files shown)


pkg/commands/up.go
```diff
@@ -66,12 +69,25 @@ type appfileOptions struct {
 	Env     *types.EnvMeta
 }
 
-func (o *appfileOptions) Run() error {
+func (o *appfileOptions) Run(cmd *cobra.Command) error {
```
hongchaodeng, line 72: pass in flags instead of cmd?

hongchaodeng (commented):
lgtm after nit

pkg/commands/up.go
mosesyou: use `filePath` instead of cmd now

wonderflow (approved):

PR #423, opened 2020-10-22
fix label not int

wonderflow:


pkg/appfile/service.go
```diff
@@ -97,12 +99,16 @@ func (s Service) RenderService(tm template.Manager, name, ns, image string) (
...
-	traits := []v1alpha2.ComponentTrait{}
-	for k, v := range traitKeys {
-		ts, err := evalTraits(tm.LoadTemplate(k), ctxData, intifyValues(v))
+	traits := make([]v1alpha2.ComponentTrait, 0)
+	for traitType, traitData := range traitKeys {
+		ts, err := evalTraits(tm.LoadTemplate(traitType), ctxData, intifyValues(traitData))
 		if err != nil {
 			return nil, nil, fmt.Errorf("eval traits failed: %w", err)
 		}
+		// one capability corresponds to one trait only
```
hongchaodeng, line 108: Adding the label should be enough to fix the problem.
I don't think we should add such constraint in the code right now. 
It is that we won't tell the story on the tutorials that we agree on.

pkg/appfile/service_test.go
```diff
@@ -0,0 +1 @@
+package appfile
```
hongchaodeng, line 1: delete this empty file?

hongchaodeng (approved):
lgtm

pkg/appfile/service_test.go
wonderflow: I will add tests in the following PRs here.

PR #428, opened 2020-10-23
vela exec

### LINKED ISSUE #171: [Feature] vela exec <appname> entering the process
Useful for debug

captainroy-hy:
to fix #171 
`vela exec` reused core functions of `kubectl exec` but doesn't accept global flags of `kubectl options`,e.g., --namespace, because those args are from vela env.

screenshot
image

image

### DIFF
--- a/pkg/commands/exec.go
+++ b/pkg/commands/exec.go
@@ -1,0 +1,193 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/crossplane/oam-kubernetes-runtime/pkg/oam"
+	"github.com/oam-dev/kubevela/api/types"
+	"github.com/spf13/cobra"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/oam-dev/kubevela/pkg/application"
+	velacmdutil "github.com/oam-dev/kubevela/pkg/commands/util"
+	cmdexec "k8s.io/kubectl/pkg/cmd/exec"
+	k8scmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+const (
+	podRunningTimeoutFlag = "pod-running-timeout"
+	defaultPodExecTimeout = 60 * time.Second
+	defaultStdin          = true
+	defaultTTY            = true
+)
+
+type VelaExecOptions struct {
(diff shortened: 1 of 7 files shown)


hongchaodeng (approved):
lgtm

pkg/commands/exec.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (o *VelaExecOptions) getPodName(compName string) (string, error) {
+	clientSet, err := o.f.KubernetesClientSet()
+	if err != nil {
+		return "", err
+	}
+	podList, err := clientSet.CoreV1().Pods(o.Env.Namespace).List(o.Context, v1.ListOptions{
+		LabelSelector: labels.Set(map[string]string{
+			oam.LabelAppName:      o.App.Name,
+			oam.LabelAppComponent: compName,
```
wonderflow, line 165: I'm not sure if the pod will have these labels?

pkg/commands/exec.go
```diff
@@ -0,0 +1,181 @@
...
+		}).String(),
+	})
+	if err != nil {
+		return "", nil
+	}
+	if podList != nil && len(podList.Items) == 0 {
+		return "", fmt.Errorf("cannot get pods")
+	}
+	// use first pod by default
+	//TODO(roywang) is it necessary to list all pods, just like askComponent?
```
wonderflow, line 175: It's not necessary, but we could check the prefix, it should contain `componentName` as prefix. If no pods found match the prefix, then just return the first one is OK.

wonderflow (approved):
Excellent job. Almost LGTM, you could choose to fix or keep it.

pkg/commands/exec.go
captainroy-hy: It depends on how Workload controller works, e.g., PodSpecWorkload will always pass these labels to PodTemplate 
But we cannot make sure other Workloads work in the same way.
Let me mark TODO here and enhance it in a rigorous manner later.

pkg/commands/exec.go
captainroy-hy: got it.

pkg/commands/exec.go
```diff
@@ -0,0 +1,193 @@
...
+			oam.LabelAppComponent: compName,
+		}).String(),
+	})
+	if err != nil {
+		return "", nil
+	}
+	if podList != nil && len(podList.Items) == 0 {
+		return "", fmt.Errorf("cannot get pods")
+	}
+	for _, p := range podList.Items {
```
zzxwill, line 181: As you choose to return the first Pod whose name matches or doesn't match the component name, Line 181 to 185 seems not necessary.

captainroy-hy: The concern is to return the name-matched pod as a higher priority.

wonderflow (approved):

PR #431, opened 2020-10-23
vela config command

### LINKED ISSUE #424: Add `config` command to manage configurations for applications
Ref https://github.com/oam-dev/kubevela/issues/399

Here's my thoughts: 

## Sub commands

- `config ls`: list all configs
- `config set <name> KEY=VALUE K2=V2`: set KV pairs for a config. If the name doesn't exist create one.
- `config get <name>`: get values of a config
- `config del <name>`: delete a config

## User workflow

1. In Appfile, a special field `config` will be used:
```
services:
  express-server:
    config: <name>
```

2. This will pass in config KVs as in context to templates:
```
context: {
  config: [string]: string
}
```

3. Then workload template can use it:
```
containers:
- name: server
  env: [
    for k, v in context.config {
      name: k
      value: v
    }
  ]
```

hongchaodeng:
fix #424

### DIFF
--- a/pkg/appfile/appfile.go
+++ b/pkg/appfile/appfile.go
@@ -26,12 +26,15 @@ type AppFile struct {
 	UpdateTime time.Time          `json:"updateTime,omitempty"`
 	Services   map[string]Service `json:"services"`
 	Secrets    map[string]string  `json:"secrets"`
+
+	configGetter configGetter
 }
 
 func NewAppFile() *AppFile {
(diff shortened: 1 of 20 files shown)


wonderflow (approved):
LGTM, please add some examples and instructions to tell how to use and make sure this feature won't be broken by other PRs or refactors.

hongchaodeng:
Added test. The critical path has been covered as shown in codecov.

PR #436, opened 2020-10-26
make default appfile as vela.yaml

### LINKED ISSUE #432: [BUG] vela up trying find vela.yml instead of vela.yaml
Version: latest master branch build.

```console
...
    up              	Apply an appfile, by default vela.yml
```

Not sure why we assume this as `.yaml` is the widely adopted preference. See: https://yaml.org/faq.html

hongchaodeng:
fix #432


pkg/appfile/appfile.go
```diff
@@ -17,7 +17,7 @@ var (
 	ErrImageNotDefined = errors.New("image not defined")
 )
 
-const DefaultAppfilePath = "./vela.yml"
+const DefaultAppfilePath = "./vela.yaml"
```
wonderflow, line 20: let's support both of them

wonderflow: check vela.yaml first and then vela.yml

resouer: Let's use `.yaml` only. It's rare and weird for a tool support both.

wonderflow: I'm OK with it。

wonderflow (approved):

PR #437, opened 2020-10-26
refact cmds from comp-centric into app-centric style

### LINKED ISSUE #322: [BUG] all trait command should be app centric
With the introduction of `vela comp`, now a trait needs to attach to a component, instead of an application.
```
✗ vela route -h
Attach route trait to an app

Usage:
  vela route <appname> [args]

Examples:
vela route frontend

Flags:
  -a, --app string      create or add into an existing application group
      --detach          detach trait from component
      --domain string
  -h, --help            help for route
      --issuer string
      --path string
      --port int         (default 443)
  -s, --staging         only save changes locally without real update application

Global Flags:
  -e, --env string   specify env name for application
```

It will cause confusion if an applications is consist of several components.
```
➜  /Users/zhouzhengxi/Programming/golang/src/github.com/zzxwill/kubevela git:(component-api) ✗ vela app show app3
About:

  Name:      	app3
  Created at:	2020-09-23 19:31:30.018492 +0800 CST
  Updated at:	2020-09-24 14:58:58.706684 +0800 CST


Environment:

  Namespace:	default

Components:

  Name    	Type      	Traits
  comp1   	webservice
  web-comp	webservice	route


➜  /Users/zhouzhengxi/Programming/golang/src/github.com/zzxwill/kubevela git:(component-api) ✗ vela scale app3
Error: app3 not exist
```

captainroy-hy:
to fix #322 

### DIFF
--- a/pkg/commands/comp.go
+++ b/pkg/commands/comp.go
@@ -19,6 +19,7 @@ const (
 	App          = "app"
 	WorkloadType = "type"
 	TraitDetach  = "detach"
+	Service      = "svc"
 )
 
 type runOptions oam.RunOptions
--- a/pkg/commands/trait.go
+++ b/pkg/commands/trait.go
@@ -3,9 +3,11 @@ package commands
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/oam-dev/kubevela/api/types"
 	"github.com/oam-dev/kubevela/pkg/application"
+	"github.com/oam-dev/kubevela/pkg/commands/util"
 	cmdutil "github.com/oam-dev/kubevela/pkg/commands/util"
 	"github.com/oam-dev/kubevela/pkg/oam"
 	"github.com/oam-dev/kubevela/pkg/plugins"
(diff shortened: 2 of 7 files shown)


pkg/commands/comp.go
```diff
@@ -19,6 +19,7 @@ const (
 	App          = "app"
 	WorkloadType = "type"
 	TraitDetach  = "detach"
+	Service      = "service"
```
wonderflow, line 22: svc

pkg/commands/trait.go
```diff
@@ -76,8 +78,8 @@ func AddTraitCommands(parentCmd *cobra.Command, c types.Args, ioStreams cmdutil.
 		for _, v := range tmp.Parameters {
 			types.SetFlagBy(pluginCmd.Flags(), v)
 		}
-		pluginCmd.Flags().StringP(App, "a", "", "create or add into an existing application group")
-		pluginCmd.Flags().BoolP(Staging, "s", false, "only save changes locally without real update application")
+		pluginCmd.Flags().StringP(Service, "s", "", "specify one service belonging to the application")
```
wonderflow, line 81: `--svc` is short enough, we don't need to occupy one more flag `s` widely.

captainroy-hy: fixed.

wonderflow (approved):

PR #438, opened 2020-10-26
Implement Autoscaler Trait 

wonderflow:
This is a PR follow up #355

I have squashed all commits in #355 and will add in the following commit, finally, I will squash all of them before merge.


api/v1alpha1/autoscaler_types.go
```diff
@@ -100,7 +69,7 @@ type Trigger struct {
 	Type TriggerType `json:"type"`
 
 	// Condition set the condition when to trigger scaling
-	Condition TriggerCondition `json:"condition"`
+	Condition map[string]string `json:"condition"`
```
zzxwill, line 72: In the first discussion, we intend to explicitly set the condition fields. In the new modification, we can only reply on Autoscaler docs to teach users how to use the trait.

wonderflow: After I digging into the implementation, we have no ideas but set it as `map[string]string`, if not so, we have no extentions.

api/v1alpha1/autoscaler_types.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// MinReplicas is the maximal replicas
+	// +optional
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// Triggers lists all triggers
+	Triggers []Trigger `json:"triggers"`
+
+	// TargetWorkload specify the workload or child workload which is about to be scaled
+	TargetWorkload TargetWorkload `json:"targetWorkload,omitempty"`
```
hongchaodeng, line 89: What's the diff between TargetWorkload and WorkloadReference?

api/v1alpha1/autoscaler_types.go
```diff
@@ -0,0 +1,120 @@
...
+	Name string `json:"name,omitempty"`
+
+	// Enabled marks whether the trigger immediately. Defaults to `true`
+	Enabled bool `json:"enabled,omitempty"`
+
+	// Type allows value in [cpu/memory/storage/ephemeral-storage、cron、pps、qps/rps、custom]
+	Type TriggerType `json:"type"`
+
+	// Condition set the condition when to trigger scaling
+	Condition map[string]string `json:"condition"`
```
hongchaodeng, line 72: omitempty?

api/v1alpha1/autoscaler_types.go
```diff
@@ -0,0 +1,120 @@
...
+	as.Spec.WorkloadReference = reference
+}
+
+// Trigger defines the trigger of Autoscaler
+type Trigger struct {
+	// Name is the trigger name, if not set, it will be automatically generated and make it globally unique
+	Name string `json:"name,omitempty"`
+
+	// Enabled marks whether the trigger immediately. Defaults to `true`
+	Enabled bool `json:"enabled,omitempty"`
```
hongchaodeng, line 66: Why have this field?
If it is not enabled, why not just remote it from trigger list?

api/v1alpha1/autoscaler_types.go
wonderflow: I think it's required, if no condition found, the trigger won't work

api/v1alpha1/autoscaler_types.go
wonderflow: TargetWorkload specify the workload which is going to be scaled,  it could be WorkloadReference or the child resource of it

api/v1alpha1/autoscaler_types.go
wonderflow: removed

hongchaodeng (approved):

PR #439, opened 2020-10-26
update templates

### LINKED ISSUE #418: [Feature] using name of WorkloadDefinition/TraitDefinition as capability name, deprecate the short name
Before we use a short name defined in definition file like below:

```
kind: TraitDefinition
metadata:
  name: routes.standard.oam.dev
spec:
  definitionRef:
    name: routes.standard.oam.dev
  extension:
    template: |
      output: {
       ...
      }
      #route: {
        domain: *"" | string
        issuer: *"" | string
      }
      parameter: #route
```

In this example vela will use `route` defined in template as the trait capability name, so we can run `vela route` to use this capability.

Now, we have decouple definition name with CRD name, so we could write this definition like this:

```
kind: TraitDefinition
metadata:
  name: route
spec:
  definitionRef:
    name: routes.standard.oam.dev
  extension:
    template: |
      output: {
       ...
      }
      #route: {
        domain: *"" | string
        issuer: *"" | string
      }
      parameter: #route
```

So we should use the definition name as the capability name, and deprecate the short name from template.

hongchaodeng:
This PR trying to unify all templates under charts, vela-templates, hack/appfile into one template directory and use one set of definitions.

See hack/vela-templates/README.md for how to update charts templates.

fix #418


pkg/appfile/service.go
```diff
@@ -114,13 +114,14 @@ func (s Service) RenderService(tm template.Manager, name, ns string, cg configGe
 	// render traits
 	traits := make([]v1alpha2.ComponentTrait, 0)
 	for traitType, traitData := range traitKeys {
-		ts, err := evalTraits(tm.LoadTemplate(traitType), ctxData, intifyValues(traitData))
+		defName, rawTmpl := tm.LoadTemplate(traitType)
```
wonderflow, line 117: we should change all the traitType to use definition name instead of magic Variable from template

pkg/appfile/template/manager.go
```diff
@@ -7,7 +7,7 @@ import (
 
 type Manager interface {
 	IsTrait(key string) bool
-	LoadTemplate(key string) string
+	LoadTemplate(key string) (defName, tmpl string)
```
wonderflow, line 10: we don't need to change this function， instead, the key should be definitionname

wonderflow (approved):

PR #472, opened 2020-10-29
fix and clean up charts

hongchaodeng:
currently the charts has following issues:
- cert-manager has manifests in multiple places. Should be combined into one.
  Prefer pure yaml since its helm chart requires `--set installCRDs=true` flag and uninstalling CRD via helm is inconvient.
- fix https://github.com/oam-dev/kubevela/issues/473:
  The bootstrap of vela-core will setup metrics controller. This requires Prometheus Operator CRD. Move related CRD to vela-core chart.
- move prometheus to vela-system namespace. Remove monitoring namespace.
- remove grafana. It's already a dependency of prometheus chart.
- change default image pull policy to IfNotPresent. Should not be Always.


resouer (approved):

resouer:
LGTM

charts/vela-core/templates/velaConfig.yaml
```diff
@@ -4,28 +4,12 @@ metadata:
...
-      "namespace": "monitoring",
-      "version": "6.0.1"
-    }
   servicemonitors.monitoring.coreos.com: |
     {
       "repo": "prometheus-community",
       "urL": "https://prometheus-community.github.io/helm-charts",
       "name": "kube-prometheus-stack",
-      "namespace": "monitoring",
+      "namespace": "vela-system",
```
ryanzhang-oss, line 12: This will break the metrics trait controller. However, we probably don't even need a controller for the metrics trait.

PR #477, opened 2020-10-30
Revise introduction.md

Fei-Guo:
This change revises the introduction.md to better explain this project.

### DIFF
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -57,16 +57,16 @@ Kubernetes capabilities for application management. It is built on top of
 Kubernetes and relieves the pains of both the platform builders and the application
 developers by doing the following:
 
-- Kubevela enforces a single **application** concept and **ALL** the exposed
+- KubeVela enforces a single **application** concept and **ALL** the exposed
   Kubernetes capabilities serve for the applications only.
   It adopts the [Open Application Model](https://github.com/oam-dev/spec) (OAM)
   for its application definition. The conventional Pod and container concepts
   are completely eliminated.
  
-- A Kubevela application is composed of various supported components or
+- A KubeVela application is composed of various supported components or
   [traits](https://github.com/oam-dev/spec/blob/master/introduction.md) and their
   schemas are determined by the platform builders. New capabilities can be added
-  to Kubevela through a CRD registry mechanism. 
+  to KubeVela through a CRD registry mechanism. 
 
 - KubeVela provides a rich set of tools to help platform builders to abstract
   the Kubernetes resource and CRD APIs.
@@ -77,16 +77,16 @@ developers by doing the following:
 
 - Schema changes take effect immediately. Neither recompiliation nor redeployment
   of KubeVela is required. This makes the process of delivering new capabilities
-  using Kubevela extremely simple.
+  using KubeVela extremely simple.
   
-With Kubevela, platform builders now have indefinite flexibilities in designing
+With KubeVela, platform builders now have indefinite flexibilities in designing
 and implementing new capabilities without worrying about what and how to expose
 the new capabilities to the end users. However, the ultimate beneficiaries are
 the application developers:
 - Instead of managing a handful Kubernetes YAML files, only a simple
   docker-compose style [appfile](./docs/developers/devex/appfile.md) is needed
-  to manage an application in Kubevela. 
-- Out of the box, Kubevela provides a CLI tool to simplify the application
+  to manage an application in KubeVela. 
+- Out of the box, KubeVela provides a CLI tool to simplify the application
   management workflow which can be easily integrated with existing CI/CD pipelines.
 
 ## Comparisons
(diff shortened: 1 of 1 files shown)


docs/introduction.md
```diff
@@ -2,69 +2,146 @@
...
-
-KubeVela intends to make shipping applications more enjoyable for developers, and at the same time, provides platform engineers with full extensibility to build their own platforms leveraging the Kubernetes community.
+## Motivation
+
+The modern trends of cloud-native application management have moved towards
+pursuing a simplified developer experience across clouds and infrastructures.
+As the dominating cloud-native platform, Kubernetes, however, has brought
+significant challenges for achieving this goal. Kubernetes has a considerably
```
resouer, line 10: Let's soften the narratives a bit, this part now reads like K8s is doing the wrong way.

Essentially k8s did very well in abstracting away the details of infra and cloud (i.e. bring consistency for application infrastructure), this makes building platforms like KubeVela super easy. But on the other hand, Kubernetes is not the end game, that's why we build KubeVela (a developers facing platform).

docs/introduction.md
```diff
@@ -2,69 +2,146 @@
...
+high learning curve:
+
+- Application developers have to learn quite a few new concepts and resource
```
resouer, line 13: I see there are 3 examples:

1. Low level abstraction - this is unique value for KubeVela, i.e. provides higher level API and tools focus on devs and application.
2. Abstraction leaks - KubeVela can solve this problem by introducing a CRD controller (e.g. Route trait in KubeVela), but they can also be solved in k8s side (e.g. Ingress V2 API).
3. Great extensibility breaks user experience - this is also a unique value for KubeVela, i.e. it defines a framework to "assemble" k8s capabilities into a user facing platform in consistent approach, but this part seems to be more platform builder related?

Another issue I am not sure if we want to give such detailed examples of the learning curve of Kubernetes, since:

1. "Kubernetes is hard" seems to be a consensus today,
2. these examples tend to become `attacking surfaces` and raise low value debate around these examples.

docs/introduction.md
```diff
@@ -2,69 +2,146 @@
...
+  specifications such as pods, ingresses, services, resource quota etc., before
+  deploying their applications.
+
+- The built-in resources sometimes are tied to underline infrastructure.
+  For example, almost all cloud vendors have their own specifications for
+  configuring a volume, making the `PodSpec` different across clouds.
```
resouer, line 19: This section is talking about abstraction leak, though I'd suggest considering Ingress as the example, it leaks every detail of underlying ingress provider to users: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/ . Note that the community is working on Ingress V2 API.

For volumes, the abstraction leak has almost been fixed by PVC/PV.

docs/introduction.md
```diff
@@ -2,69 +2,146 @@
...
-
-
-### Helm, etc.
-
-Helm is a package manager for Kubernetes that provides package, install, and upgrade a set of YAML files for Kubernetes as a unit. 
-
-KubeVela is not a package manager and can't be used to manage or deploy Kubernetes YAML files by default. KubeVela provides a client side tool named Appfile to describe how to build and deploy application in a single file but it's not Kubernetes YAML format. However, KubeVela server side component indeed exposes a main API resource named "Application Configuration" (i.e. the object behind KubeVela Appfile and CLI), it could be packaged and distributed by Helm at ease.
-
-In implementation side, KubeVela heavily relies on Helm to package and manage the third-party plug-ins such as `Prometheus`, etc. 
+KubeVela is a framework to help platform builders to easily expose and extend
```
resouer, line 56: I think we still need to keep "what is KubeVela for developers" section?

Otherwise, the end users will have to be exposed many platform level introduction below (e.g. OAM, CUE etc.)

docs/introduction.md
```diff
@@ -2,69 +2,146 @@
...
+even more constraints in extensibility. They are argurably "hard-coded" PaaS.
+
+Kubernetes based serverless platforms such as Knative, OpenFaaS can be easily
+integrated with Kubevela by registering themself as platform capabilities.
+Even for AWS Lambda, there is an success story to integrate it with Kubevela
+by the tools developed by Crossplane.
+
+### Platform agnostic developer tools
+
+The typical example is [Waypoint](https://github.com/hashicorp/waypoint). It is
```
resouer, line 124: Suggestion:

```
The typical example is [Waypoint](https://github.com/hashicorp/waypoint). It is
a user facing tool which introduces a consistent workflow (i.e., build,
deploy, release) for developers to ship application on top of any platforms. KubeVela can be integrated into Waypoint 
like any other supported platforms. In this case, developers can use the Waypoint
workflow instead of the Kubevela CLI to manage applications.
```

docs/introduction.md
```diff
@@ -2,69 +2,146 @@
...
+YAML files. However, in the server side, Kubevela still relies on Helm to package
+and manage the third-party plugins such as `Prometheus`, etc.
 
 ### Kubernetes
 
-KubeVela is a Kubernetes extension, it's complementary to Kubernetes.
-
-In detail, KubeVela introduced several developer-centric abstractions to Kubernetes such as `Application`, etc and leverages capabilities of Kubernetes as underlying implementation. KubeVela also allows platform engineers to bring in any other capabilities from Kubernetes ecosystem with minimal effort. In nutshell, KubeVela is an highly extensible application management plugin for Kubernetes.
\ No newline at end of file
+KubeVela can be treated as a special Kubernetes distribution for application management.
```
resouer, line 145:
```suggestion
KubeVela can be treated as a special Kubernetes "distribution" that focuses on application management only.
```

docs/introduction.md
Fei-Guo: done. Added in the end of this section.

docs/introduction.md
Fei-Guo: done

resouer:
Generally LGTM! Some other nits but I can send a follow up PR to fix.

Thanks a lot!

Issue #443, opened 2020-10-27
[BUG] Use CUE template as schema for Web Service


resouer:
As next step, we will use CUE template directly as the schema for Web Service workload, and even for many trait.

For example:

```yaml
apiVersion: core.oam.dev/v1alpha2
kind: WorkloadDefinition
metadata:
  name: webservice
spec:
  definitionRef:
    name: apps.deployment
  extension:
    template: |
      outputs: service: {
        apiVersion: "v1"
        kind: "Service"
        metadata:
          name: context.name
        spec: {
          ...
        }
      }
      
      outputs: deployment: {
        apiVersion: "apps/v1"
        kind: "Deployment"
        spec: {
          ...
        }
      }
     #webservice: {
        name: string
        // +usage=specify app image
        // +short=i
        image: string
        // +usage=specify port for container
        // +short=p
        port: *6379 | int
        }
     parameter: #webservice
```

PR #482, opened 2020-10-30
[Feature #443] should report capability not exist instead of file not found

linjiemiao:
Example:
```
version: "1.0-alpha.1"
name: test1
secrets: {}
services:
  hello:
    image: nginx:latest
    name: hello
    port: 80
    route:
      domain: hello.y
      issuer: oam-env-default
    type: webservice1
```

Error: eval service failed: capability webservice1 unexist


hongchaodeng (approved):
lgtm

PR #489, opened 2020-10-31
Fix `charts/vela-core/crds/_.yaml` generation issue

zzxwill:
Fixed the issue of `charts/vela-core/crds/_.yaml` generation
by removing unkonwn struct in pkg/commands/show.go.


wonderflow (approved):
Cool

PR #496, opened 2020-11-02
doc for quick start, exec/logs/port-forward; add render-only to init cmd

hongchaodeng:
ref https://github.com/oam-dev/kubevela/issues/441

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,7 @@ core-uninstall: manifests
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=charts/vela-core/crds
+	./hack/vela-templates/gen_definitions.sh
 
 # Generate code
 generate: controller-gen
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -1,9 +1,129 @@
 # Quick Start
 
-## Deploy Application
+Welcome to KubeVela! In this guide, we'll walk you through how to deploy applications on KubeVela CLI/Appfile.
+
+## Setup
+
+Make sure you have finished and verified installation following [this guide](./install.md).
+
+## 1. Initialize Application Deployment Config
 
 ```console
-$ vela app init
+$ vela init --render-only
+Welcome to use KubeVela CLI! We're going to help you run applications through a couple of questions.
+
+Environment: default, namespace: default
+
+? Do you want to setup a domain for web service:
+? Provide an email for production certification:
+? What would you like to name your application:  testapp
+? Choose workload type for your service:  webservice
+? What would you name this webservice:  testsvc
+? specify app image: crccheck/hello-world
+? specify port for container: 8000
+
+Rendered and written deployment config to vela.yaml
 ```
 
-> TODO Please give an full example of `vela app init` with route
+In current directory, you will have a generated `vela.yaml`:
+
+```yaml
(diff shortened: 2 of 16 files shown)


Makefile
```diff
@@ -139,6 +139,7 @@ core-uninstall: manifests
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=charts/vela-core/crds
+	./hack/vela-templates/gen_definitions.sh
```
wonderflow, line 142: 👍 I was just going to add it in my PR,haha

pkg/appfile/appfile.go
```diff
@@ -21,6 +21,7 @@ var (
 )
 
 const DefaultAppfilePath = "./vela.yaml"
+const DefaultVersion = "1.0-alpha.1"
```
wonderflow, line 24: for a version of format, we usually name it like  `v1alpha1` (k8s) or `2010-09-09`(cloud formation)

pkg/commands/init.go
```diff
@@ -72,6 +75,20 @@ func NewInitCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+			err = ioutil.WriteFile("./vela.yaml", b, 0600)
+			if err != nil {
+				return err
+			}
+			o.IOStreams.Info("\nRendering deployment config to " + color.New(color.FgCyan).Sprint("vela.yaml"))
+
+			if o.renderOnly {
```
wonderflow, line 88: the renderOnly check should move before `oam.BaseRun`(line 73), or it's too late to check.

pkg/appfile/appfile.go
hongchaodeng: v1alpha1 is k8s specific and it is not even semver.
See https://docs.docker.com/compose/:

```
version: "3.8"
services:
```

wonderflow: yes, something like `3.8` is also good for me.

hongchaodeng: Using `v1alpha1` will complicate developers who are not familiar with k8s

wonderflow: what about just `v1`?

pkg/commands/init.go
hongchaodeng: fixed

pkg/appfile/appfile.go
hongchaodeng: It is unwise to declare v1 for now...
This is not something we want to solve before kubecon. I would suggest let's just follow compose spec for now.

hongchaodeng: Since it is alpha we can change that later

hongchaodeng: I removed the `version` field from Appfile. 
This is not something deserved a place right now

docs/quick-start.md
```diff
@@ -1,9 +1,129 @@
...
+
+Applying deploy configs ...
+Checking if app has been deployed...
+app existed, updating existing deployment...
+✅ app has been deployed 🚀🚀🚀
+    Port forward: vela port-forward testapp
+             SSH: vela exec testapp
+         Logging: vela logs testapp
+      App status: vela app status testapp
+  Service status: vela svc status testsvc
```
wonderflow, line 75: Here was changed, only `vela status` is enough

wonderflow (approved):

PR #504, opened 2020-11-03
doc: add app init, config, and capability references

hongchaodeng:
ref https://github.com/oam-dev/kubevela/issues/441

Additionally
- cue fmt all templates
- workload type: change `backend` to `worker`


wonderflow (approved):
LGTM， please squash before merge

PR #505, opened 2020-11-04
update local caps before listing workloads&traits

captainroy-hy:
- for `vela traits` and `vela workloads`, add a bool flag whose default value is true to indicate whether run `vela system update` before listing traits/workloads
- refine the output of `vela system update`: Show added/updated/deleted capabilities if exist any, otherwise(nothing changed), show all existing capabilities.

Demo screenshots:
image
image

add e2e-test

### DIFF
--- a/api/types/capability.go
+++ b/api/types/capability.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"cuelang.org/go/cue"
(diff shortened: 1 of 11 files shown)


wonderflow:
We should reduce the output information, only keep necessary info such as which new capability added and which legacy removed.

wonderflow (approved):

PR #507, opened 2020-11-04
Some revisions for the printed out messages

Fei-Guo:
I focus on the printed out messages in the quick start example. 

I wish to change  
 `specify app image` to `Specify application image`  
 `specify port for container` to `specify application service port` 
as well but this requires a lot of changes since the texts come from the workload definition. Hence, I omit them in this change.


hongchaodeng (approved):
lgtm

PR #518, opened 2020-11-06
Function SyncCapabilityFromCenter Called twices.

linjiemiao:
This function is called one more time.


pkg/oam/capability.go
```diff
@@ -48,7 +48,7 @@ func AddCapabilityCenter(capName, capURL, capToken string) error {
 	if err = plugins.StoreRepos(repos); err != nil {
 		return err
 	}
-	return SyncCapabilityFromCenter(capName, capURL, capToken)
```
wonderflow, line 51: you should remove the one in `NewCapCenterConfigCommand ` instead of here.

wonderflow:
CI was broken by #520, I have fixed.

wonderflow (approved):

PR #525, opened 2020-11-06
only choose OAM app labels for route trait

wonderflow:
The original labels will select all label from podtemplate, this will introduce revision different labels, so the route trait can not route ingress and service to different revisions when they are rolling update. So I narrow the labels, only choose two labels which represent `OAM app config name` and `component name`


zzxwill (approved):
/lgtm

PR #535, opened 2020-11-07
Fix `vela ls` issue

### LINKED ISSUE #534: [BUG] all svc including those deployed without `--staging` flag will appear to be `staging` status
All svc including those deployed without `--staging` flag will appear to be `staging` status in cmd `vela ls
```
vela ls
SERVICE	APP	TYPE      	TRAITS	STATUS 	CREATED-TIME
s1     	a1 	webservice	      	Staging	2020-11-07 21:29:01 +0800 CST
s2     	a2 	webservice	      	Staging	2020-11-07 22:01:00 +0800 CST
s3     	a3 	          	      	Staging	2020-11-07 23:39:20.090882 +0800 CST
```

zzxwill:
By default, all svc deployed without `--staging` will appear
to be `staging` in cmd `vela ls`.
To fix #534

### DIFF
--- a/pkg/commands/ls.go
+++ b/pkg/commands/ls.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"strings"
 
+	"github.com/crossplane/oam-kubernetes-runtime/apis/core/v1alpha2"
+
 	gocmp "github.com/google/go-cmp/cmp"
 	"github.com/gosuri/uitable"
 	"github.com/spf13/cobra"
@@ -70,13 +72,13 @@ func printComponentList(ctx context.Context, c client.Client, appName string, en
 }
 
 func mergeStagingComponents(deployed []apis.ComponentMeta, env *types.EnvMeta, ioStreams cmdutil.IOStreams) []apis.ComponentMeta {
-	apps, err := application.List(env.Name)
+	localApps, err := application.List(env.Name)
 	if err != nil {
 		ioStreams.Error("list application err", err)
 		return deployed
 	}
 	var all []apis.ComponentMeta
-	for _, app := range apps {
+	for _, app := range localApps {
 		comps, appConfig, _, err := app.OAM(env, ioStreams, true)
 		if err != nil {
 			ioStreams.Errorf("convert app %s err %v\n", app.Name, err)
@@ -113,7 +115,11 @@ func mergeStagingComponents(deployed []apis.ComponentMeta, env *types.EnvMeta, i
 					v.Traits[j] = t
 				}
 				aspec.Components[i] = v
+				if compMeta.AppConfig.Spec.Components[i].Traits == nil && len(v.Traits) == 0 {
+					compMeta.AppConfig.Spec.Components[i].Traits = make([]v1alpha2.ComponentTrait, 0)
+				}
 			}
+
 			if !gocmp.Equal(compMeta.Component.Spec, *cspec) || !gocmp.Equal(compMeta.AppConfig.Spec, *aspec) {
 				compMeta.Status = types.StatusStaging
 			}
--- a/pkg/appfile/appfile.go
+++ b/pkg/appfile/appfile.go
@@ -64,12 +64,12 @@ func (app *AppFile) BuildOAM(ns string, io cmdutil.IOStreams, tm template.Manage
 }
 
 // RenderOAM renders Appfile into AppConfig, Components.
-func (app *AppFile) RenderOAM(ns string, io cmdutil.IOStreams, tm template.Manager, slience bool) (
+func (app *AppFile) RenderOAM(ns string, io cmdutil.IOStreams, tm template.Manager, silence bool) (
 	[]*v1alpha2.Component, *v1alpha2.ApplicationConfiguration, []oam.Object, error) {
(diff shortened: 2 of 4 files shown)


pkg/commands/ls.go
```diff
@@ -113,7 +113,14 @@ func mergeStagingComponents(deployed []apis.ComponentMeta, env *types.EnvMeta, i
 					v.Traits[j] = t
 				}
 				aspec.Components[i] = v
+				/* Before any traits is attached to the workload, if ApplicationConfiguration is retrieved through
+				OpenAPI, the .spec.Components[].Traits is nil; if it's resembled by local file, .spec.Components[].Traits
+				is the []v1alpha2.ComponentTrait, 0). Let's set it to be equal before comparing the whole .Spec field.*/
+				if compMeta.AppConfig.Spec.Components[i].Traits == nil && len(v.Traits) == 0 {
+					compMeta.AppConfig.Spec.Components[i].Traits = aspec.Components[i].Traits
+				}
```
zzxwill, line 121: I am afraid the hack way here might not be the best solution. One solution is to change the response of client-go from `ApplicationConfiguration` to make `.spec.Components[].Traits` to be in Golang slice, the other way is to change the type of traits of https://github.com/oam-dev/kubevela/blob/master/pkg/appfile/service.go#L114 which I think the type is reasonable. 
```
traits := make([]v1alpha2.ComponentTrait, 0)
```
I could not find a way out of the dilemma, so here is a solution to the issue #534

hongchaodeng:
@zzxwill 
Can you provide an analysis of the root cause of the issue?

pkg/commands/ls.go
```diff
@@ -113,7 +113,14 @@ func mergeStagingComponents(deployed []apis.ComponentMeta, env *types.EnvMeta, i
 					v.Traits[j] = t
 				}
 				aspec.Components[i] = v
+				/* Before any traits is attached to the workload, if ApplicationConfiguration is retrieved through
```
hongchaodeng, line 116: Why is it retrieved from OpenAPI?

pkg/commands/ls.go
```diff
@@ -113,7 +113,14 @@ func mergeStagingComponents(deployed []apis.ComponentMeta, env *types.EnvMeta, i
...
+				OpenAPI, the .spec.Components[].Traits is nil; if it's resembled by local file, .spec.Components[].Traits
```
hongchaodeng, line 117: What do you mean by "resembled by local file"?

pkg/commands/ls.go
```diff
@@ -113,7 +113,14 @@ func mergeStagingComponents(deployed []apis.ComponentMeta, env *types.EnvMeta, i
...
+				is the []v1alpha2.ComponentTrait, 0). Let's set it to be equal before comparing the whole .Spec field.*/
```
hongchaodeng, line 118: > is the []v1alpha2.ComponentTrait, 0)

What does this mean?

pkg/commands/ls.go
```diff
@@ -113,7 +113,14 @@ func mergeStagingComponents(deployed []apis.ComponentMeta, env *types.EnvMeta, i
...
+				if compMeta.AppConfig.Spec.Components[i].Traits == nil && len(v.Traits) == 0 {
```
hongchaodeng, line 119: Change to following:

```Go
if  len(compMeta.AppConfig.Spec.Components[i].Traits) == 0 {
  compMeta.AppConfig.Spec.Components[i].Traits = nil
}
if len(v.Traits) == 0 {
  v.Traits = nil
}
```

pkg/commands/ls.go
zzxwill: It's retrieved as below https://github.com/oam-dev/kubevela/blob/master/pkg/oam/application.go#L91-L95.
```
var appConfig corev1alpha2.ApplicationConfiguration
		if err := c.Get(ctx, client.ObjectKey{Name: opt.AppName, Namespace: opt.Namespace}, &appConfig); err != nil {
			return appConfigList, err
		}
```

pkg/commands/ls.go
zzxwill: We are comparing appConfig with the one from cached in local file to check whether the status of the application is staging. For details, please refer to https://github.com/oam-dev/kubevela/blob/master/pkg/appfile/service.go#L73.

pkg/commands/ls.go
zzxwill: It should be `make([]v1alpha2.ComponentTrait, 0)`:)

pkg/commands/ls.go
zzxwill: I intended to check both  `compMeta.AppConfig.Spec.Components[i].Traits == nil` and `len(v.Traits) == 0`. If only one is `nil` or `make([]v1alpha2.ComponentTrait, 0`, we don't want to do anything. And they are different which means the status of the application should be `staging`.

is it better?
```
if compMeta.AppConfig.Spec.Components[i].Traits == nil && len(v.Traits) == 0 {
					aspec.Components[i].Traits = nil
				}
```

wonderflow (approved):

Issue #564, opened 2020-11-11
[Feature] Revise the doc with appfile focused


resouer:
Already done:

- [x] Refactor 1: https://github.com/oam-dev/kubevela/pull/558 https://github.com/oam-dev/kubevela/pull/562
- [x] Refactor 2: https://github.com/oam-dev/kubevela/pull/563
- [x] Reference doc: https://github.com/oam-dev/kubevela/pull/579

Remaining task:

- [ ] [Setting Auto-scaling Policy](/en/developers/set-autoscale.md) @zzxwill 
- [x] [Setting Rollout Strategy](/en/developers/set-rollout.md) @wonderflow 
- [x] [Monitoring Application](/en/developers/set-metrics.md) @hongchaodeng 
- [ ] [Fix the help page of CLI](https://github.com/oam-dev/kubevela/issues/590) @hongchaodeng @wonderflow 

Also, below doc need fix, please test it with latest vela and make the doc more readable and neat.

- [ ] [Add Cloud Services](/en/platform-engineers/cloud-resource.md) @szihai

PR #566, opened 2020-11-11
Revise autoscaler doc with appfile focused

zzxwill:
Splitted autoscale docs to be appfile focused,
and add cpu utilization scaling policy for appfile

To implemented #564


docs/en/developers/set-autoscale.md
```diff
@@ -233,3 +124,143 @@ Introduce how to automatically scale workloads by cron in Appfile.
...
+Introduce how to automatically scale workloads by CPU resource utilization.
+
+- Prepare Appfile
+  ```yaml
+  name: testapp
+
+  services:
+    express-server:
+      # this image will be used in both build and deploy steps
+      image: zzxwill/kubevela-appfile-demo:v1
```
resouer, line 138: Can we reuse the previous appfile? So we just update it in this guide.

docs/en/developers/set-autoscale.md
```diff
@@ -233,3 +124,143 @@ Introduce how to automatically scale workloads by cron in Appfile.
...
+
+      build:
+        # Here more runtime specific build templates will be supported, like NodeJS, Go, Python, Ruby.
+        docker:
+          file: Dockerfile
+          context: .
+
+      cmd: ["node", "server.js"]
+      port: 8080
```
resouer, line 147: We do not need to specify CPU limit?

docs/en/developers/set-autoscale.md
zzxwill: Yes, just needed to add field `.autoscale.cron` and field `services.express-server.resources`. Will update this section.

And I was hesitated whether it's proper to specify the image `zzxwill/kubevela-appfile-demo:v1`. I intended to indicate developers should use a personal image here as it's told in https://github.com/oam-dev/kubevela/blob/master/docs/en/developers/devex/appfile.md#2-deploy-app-in-one-command. Or I should repeat the instructions as the appfile.md.

ATTENTION: change the image field in vela.yaml to something you can push to on your host:

> Or you may try the local testing option introduced in the following section.

```yaml
    image: oamdev/testapp:v1
```

docs/en/developers/set-autoscale.md
zzxwill: Oh, sorry. I missed it and I didn't make an e2e test to verify cpu utilization policy in appfile locally for the newly added part. Will cover it and update the doc later. Thanks for pointing it out.

zzxwill: I should. My bad, I wrongly paste the vela.yaml from `examples/testapp` instead of `docs/examples/testapp` which has been fixed #569 , and I didn't do a double check. Thanks for pointing it out.

resouer: Please simply do
```
# Number of CPU units for the service.
cpu: 256
```

not `cpuRequest` ...

resouer: And what is the unit for this cpu limit? This seems not documented anywhere. For example, what does it mean by `cpu: 256` or `cpu: 0.5`? how user should fill this field?

docs/en/developers/set-autoscale.md
resouer: > And I was hesitated whether it's proper to specify the image zzxwill/kubevela-appfile-demo:v1

There's already an existing image `oamdev/testapp:v1`, why don't use it? Also, you don't need to define `build` if the image is already built ....

zzxwill: Developers need write permission to push the image to docker hub per the requirement for building image in appfile.

docs/en/developers/set-autoscale.md
zzxwill: - It might not be explicit to use `cpu` as I want to emphasize it the request for cpu, and maybe later `cpuLimits` would be added. 
```
resources: {
  limits:
    cpu: parameter.cpuRequests
  requests:
    cpu: parameter.cpuRequests
}
```

Besides, it's complicated to explain the cpu unit as the definition here https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L88, while most Kubernetes users will know `cpuRequests` and `cpuLimits`.

Either way, I will add an example to the `usage` part. Thanks.
```
// +usage=CPU core requests for the workload, like `256m`(0.5 core), `1`(1 core)
	// +alias=cpu-requests
	cpuRequests?: string
```

zzxwill: `cpuRequest` 解释还是不应该借助 k8s 的背景，我想想怎么改。
“k8s api resource，是 OAM/Appfile 的 schema provider，不是它的 schema 本身这个关系得明确。”

docs/en/developers/set-autoscale.md
zzxwill: I got it. I should remove field `build` to use the `oamdev/testapp:v1` per the instruction from @wonderflow.

hongchaodeng: I have actually removed build.
Please rebase and check :)

docs/en/developers/set-autoscale.md
resouer: 1. 这个参数名就叫 `cpu`，不要弄那么多花样
2. CPU 单位的解释就叫：`Number of CPU units for the service.` 这里的关键是，你得让用户知道怎么填。用户是填  500m 还是 500， 还是 0.5？这个必须说清楚。

resouer: 也可以直接用 OAM spec 中的定义：https://github.com/oam-dev/spec/blob/master/core/workloads/schema/containerized_workload.md#cpu-count

docs/en/developers/set-autoscale.md
```diff
@@ -233,3 +125,149 @@ Introduce how to automatically scale workloads by cron in Appfile.
...
+  
+      cmd: ["node", "server.js"]
+      port: 8080
+  
+      cpuRequests: "0.01"
+  
+      autoscale:
+        min: 1
+        max: 5
+        cpu: "10"
```
resouer, line 159: 这个不行。这怎么一会儿有引号一会儿没引号的，到底是啥情况？

resouer:
Add please also add reference doc for autoscaler like other traits.

docs/en/developers/set-autoscale.md
zzxwill: > Add please also add reference doc for autoscaler like other traits.

Okay. I tried to automatically generate the docs but i haven't finished it. I will manually add it first. Thanks.

zzxwill:
> Add please also add reference doc for autoscaler like other traits.

Please allow me to add this doc later in my evening. Thanks.

hack/vela-templates/cue/autoscale.cue
```diff
@@ -1,16 +1,18 @@
+import "strconv"
```
zzxwill, line 1: I used CUE builtin package to solve the problem of not using double-quotes for int type parameter. To solve all double-quotes problems, please refer to the discussion https://github.com/oam-dev/kubevela/issues/585.

docs/en/developers/set-autoscale.md
```diff
@@ -1,162 +1,208 @@
...
+  ```yaml
+  name: testapp
+  
+  services:
+    express-server:
+      image: oamdev/testapp:v1
+        
+      cmd: ["node", "server.js"]
+      port: 8080
+      cpu: "0.01"
```
wonderflow, line 119: cpuPercent

zzxwill: It means to set cpu requests/limit, not `autoscaler.cpuPercent`. So it's right. Thanks for reminding me to take a look at it.

hack/vela-templates/cue/autoscale.cue
resouer: As mentioned in #585, we prefer to fix that in Appfile impl. Though I personally not against having this workaround temporarily for autoscaler, please add a comment:
```
// Temporarily use strconv for type converting. This bug will be fixed in kubevela#585 
```

docs/en/developers/set-autoscale.md
```diff
@@ -1,162 +1,208 @@
...
   
+      autoscale:
+        min: 1
+        max: 5
+        cpuPercent: 10
```
resouer, line 124: Why not `cpuPercent: 10%`?

docs/en/developers/set-autoscale.md
resouer, line 119: Is this by intention to give such small cpu quota? why?

charts/vela-core/templates/defwithtemplate/webservice.yaml
```diff
@@ -82,8 +82,7 @@ spec:
       			}
       		}
       	}]
-      	// +usage=CPU core requests for the workload, specify like '0.5', '1'.
-      	// +alias=cpu-requests
-      	cpuRequests?: string
+      	// +usage=Number of CPU units for the service, like `256m` (0.5 CPU core), `1` (1 CPU core)
```
resouer, line 85: This comment is not correct.

docs/en/developers/set-autoscale.md
zzxwill: Yes, by intention. With my benchmark tool, I generated 1000 requests by 200 qps (more qps might lead to crash of the tool ), if more CPU cores is set, the threshold value could not be easily reached. In order to let the example to work, I set 0.01 (10m) cores for the workload.
````
ab -n 10000 -c 200 http://127.0.0.1/
```

charts/vela-core/templates/defwithtemplate/webservice.yaml
zzxwill: Updated.

docs/en/developers/set-autoscale.md
zzxwill: If the workload doesn't run such a simple business, we can set more CPU resource for it. Do I need to change another image for the workload to make `cpu` resource looks better?

wonderflow (approved):

PR #573, opened 2020-11-11
 keep silence when no Capability discovered && clean code

### LINKED ISSUE #565: [Feature] vela appfile should support no template case
if no template specified in Definition, the schema could be arbitrary or align with CRD schema.

In other words, template should be optional.

wonderflow:
1. keep silence when no Capability discovered
2. clean code


wonderflow:
@resouer  Changed the way

resouer (changes requested):
This is totally wrong.

resouer (approved):

PR #586, opened 2020-11-12
go through rollout with kind cluster and fix bugs

wonderflow:
1. add `vela port-forward xxx --route` to use k8s service for port-forward if route trait exist.
2. allow empty domain for K8s cluster which will not have ingress available, empty domain will not create ingress
3. update image in rollout doc
4. update oam-k8s-runtime that health scope will support multiple workload versions.


hongchaodeng:
@wonderflow Please fix UT

hongchaodeng (approved):
lgtm

PR #613, opened 2020-11-16
refactor extend trait feature and doc

wonderflow:
1. teach how to extend trait from writing definition file step by step.
2. support install helm chart from cluster definition , before this PR, the chart can only installed from capability center, so the example in doc won't work.
3. create namespace if not exist when installing chart


hongchaodeng (approved):
lgtm

PR #614, opened 2020-11-16
add more strict golangci check

wonderflow:

This first commit is the original one for add this CI.

Please review if the rule is appropriate. @hongchaodeng 

In the following commits I will fix lint errors one by one.


hongchaodeng (approved):
Excellent! LGTM

Note that there is a github action to run golangci: https://github.com/golangci/golangci-lint-action

PR #616, opened 2020-11-16
Remove duplicated badges

resouer:
Gitter and Slack are not needed, they are highlighted in README.

Also, the Slack channel for kubevela is: https://cloud-native.slack.com/messages#kubevela


hongchaodeng (approved):
lgtm

PR #619, opened 2020-11-17
fix go imports lint complain

wonderflow:

### DIFF
--- a/pkg/commands/cli.go
+++ b/pkg/commands/cli.go
@@ -9,14 +9,15 @@ import (
 	"github.com/oam-dev/kubevela/pkg/utils/common"
 
 	"github.com/gosuri/uitable"
+	"github.com/spf13/cobra"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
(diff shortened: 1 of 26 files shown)


zzxwill (approved):
LGTM

pkg/commands/cli.go
```diff
@@ -9,14 +9,15 @@ import (
...
 	"github.com/gosuri/uitable"
+	"github.com/spf13/cobra"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
 	"github.com/oam-dev/kubevela/api/types"
 	"github.com/oam-dev/kubevela/cmd/vela/fake"
 	"github.com/oam-dev/kubevela/pkg/commands/util"
 	"github.com/oam-dev/kubevela/pkg/utils/system"
 	"github.com/oam-dev/kubevela/version"
```
zzxwill, line 20: Just to make sure, 引用的当前库，是需要分隔开，放到最后吧？

zzxwill:
I spend some time fixing goimport issues in OAM-Kubernetes-Runtime, so it would be better to port this automatic way into runtime repo:)

PR #620, opened 2020-11-17
add description for export const variable and function 

wonderflow:
I found it's too huge to change them in one commit, so I split it into several ones.


hongchaodeng (approved):
lgtm

PR #631, opened 2020-11-19
multi-arch images support

morlay:
* Dockerfile updates to support multi-arch image （support `docker build` too）
* github workflow updates
     * migrate to `ghcr.io` ([see more](https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images)). need to add a new secret `CR_PAT`
     * switch to `docker/login-action` && `docker/build-push-action@v2` for multi arch image build by `buildx`

examples:
https://github.com/users/morlay/packages/container/package/kubevela%2Fvela-core

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,18 +8,38 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
-      - name: Publish to Github Docker Package Registry
-        uses: elgohr/Publish-Docker-Github-Action@2.21
+      - name: Get the version
+        id: get_version
+        run: |
+          VERSION=${GITHUB_REF#refs/tags/}
+          if [[ ${GITHUB_REF} == "refs/heads/master" ]]; then
+            VERSION=latest
+          fi
+          echo ::set-output name=VERSION::${VERSION}
+      - name: Login ghcr.io
+        uses: docker/login-action@v1
         with:
-          name: oam-dev/kubevela/vela-core
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-          registry: docker.pkg.github.com
-          tags: "latest"
-      - name: Publish to Docker Hub Registry
-        uses: elgohr/Publish-Docker-Github-Action@2.21
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.CR_PAT }}
+      - name: Login docker.io
+        uses: docker/login-action@v1
         with:
-          name: oamdev/vela-core
+          registry: docker.io
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-          tags: "latest"
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
+      - uses: docker/build-push-action@v2
+        name: Build & Pushing
+        with:
+          context: .
+          file: Dockerfile
+          labels: |-
+            org.opencontainers.image.source=https://github.com/${{ github.repository }}
+            org.opencontainers.image.revision=${{ github.sha }}
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: |-
+            ghcr.io/${{ github.repository }}/vela-core:${{ steps.get_version.outputs.VERSION }}
+            docker.io/oamdev/vela-core:${{ steps.get_version.outputs.VERSION }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.13 as builder
+FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.13 as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
@@ -15,14 +15,19 @@ COPY api/ api/
 COPY pkg/ pkg/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+ARG TARGETARCH
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} GO111MODULE=on go build -a -o manager-${TARGETARCH} main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-# oamdev/gcr.io-distroless-static:nonroot is syncd from gcr.io/distroless/static:nonroot as somewhere can't reach gcr.io
-FROM oamdev/gcr.io-distroless-static:nonroot
+# Could use `--build-arg=BASE_DISTROLESS=gcr.io/distroless/static:nonroot` to overwrite
+ARG BASE_DISTROLESS
+FROM ${BASE_DISTROLESS:-gcr.io/distroless/static:nonroot}
+
 WORKDIR /
-COPY --from=builder /workspace/manager .
+
+ARG TARGETARCH
+COPY --from=builder /workspace/manager-${TARGETARCH} /manager
 USER nonroot:nonroot
 
 ENTRYPOINT ["/manager"]


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.13 as builder
+FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.13 as builder
```
morlay, line 2: build on same arch of host

Dockerfile
```diff
@@ -15,14 +15,19 @@ COPY api/ api/
 COPY pkg/ pkg/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+ARG TARGETARCH
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} GO111MODULE=on go build -a -o manager-${TARGETARCH} main.go
```
morlay, line 19: equals `for arch in amd64 arm64; do`

Dockerfile
```diff
@@ -15,14 +15,19 @@ COPY api/ api/
...
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-# oamdev/gcr.io-distroless-static:nonroot is syncd from gcr.io/distroless/static:nonroot as somewhere can't reach gcr.io
-FROM oamdev/gcr.io-distroless-static:nonroot
+# Could use `--build-arg=BASE_DISTROLESS=gcr.io/distroless/static:nonroot` to overwrite
+ARG BASE_DISTROLESS
+FROM ${BASE_DISTROLESS:-gcr.io/distroless/static:nonroot}
```
morlay, line 25: official `gcr.io/distroless/static:nonroot` support multi-arch now.
for local build. if can't reach `gcr.io`. 
`--build-arg=BASE_DISTROLESS` to overwrite

could use a fork https://github.com/querycap/distroless.

.github/workflows/docker.yml
```diff
@@ -8,18 +8,38 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
-      - name: Publish to Github Docker Package Registry
-        uses: elgohr/Publish-Docker-Github-Action@2.21
+      - name: Get the version
+        id: get_version
+        run: |
+          VERSION=${GITHUB_REF#refs/tags/}
```
morlay, line 14: consider to add `tags: v*` in this workflow, and remove docker pushing in `release.yaml`

.github/workflows/docker.yml
```diff
@@ -8,18 +8,38 @@ jobs:
...
         with:
-          name: oamdev/vela-core
+          registry: docker.io
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-          tags: "latest"
\ No newline at end of file
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
+      - uses: docker/build-push-action@v2
```
morlay, line 33: see from https://github.com/docker/build-push-action

zzxwill:
Cloud you please explain a little bit of the reason why needs to build into Github CR? For end users, we can use some speed-up service (ie, registry-mirrors) from service providers, like Alibaba Cloud, but for Github CR, I don't know.

morlay:
@zzxwill 

`ghcr.io` is upgrade of `docker.pkg.github.com` 

https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images

But `docker.pkg.github.com` is not support the new format of multi-arch image. and will never .
`docker.pkg.github.com` will shut down someday too.

zzxwill:
@morlay +1. Cool.

morlay:
@zzxwill and have to create a personal token of repo owner https://github.com/settings/tokens
and add new secret `CR_PAT` to repo or group

Dockerfile
zzxwill: Yes, `gcr.io` isn't an option due to network connection issues.

.github/workflows/docker.yml
```diff
@@ -8,18 +8,38 @@ jobs:
...
-      - name: Publish to Docker Hub Registry
-        uses: elgohr/Publish-Docker-Github-Action@2.21
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.CR_PAT }}
+      - name: Login docker.io
+        uses: docker/login-action@v1
         with:
-          name: oamdev/vela-core
+          registry: docker.io
```
zzxwill, line 28: Normally username/password is granted to a specified image in docker.io, why changing the image to a `registry`?

.github/workflows/docker.yml
```diff
@@ -8,18 +8,38 @@ jobs:
...
+          context: .
+          file: Dockerfile
+          labels: |-
+            org.opencontainers.image.source=https://github.com/${{ github.repository }}
+            org.opencontainers.image.revision=${{ github.sha }}
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: |-
+            ghcr.io/${{ github.repository }}/vela-core:${{ steps.get_version.outputs.VERSION }}
+            docker.io/oamdev/vela-core:${{ steps.get_version.outputs.VERSION }}
```
zzxwill, line 45: I noticed two one more tag is added, will both of them be push to the Container Registry respectively? Which one will be used for end-users?

.github/workflows/docker.yml
morlay: Not changes the name.  
it split to steps. 

logins and build & push

morlay: with the `docker/build-push-action@v2`, 

we could push with two tags.

```
ghcr.io/${{ github.repository }}/vela-core:${{ steps.get_version.outputs.VERSION }}
docker.io/oamdev/vela-core:${{ steps.get_version.outputs.VERSION }} 
```

.github/workflows/docker.yml
morlay: both works. it will push to multi registries.

morlay: one more thing, images pushing to ghcr.io will be private by default. 

need to make images public here. https://github.com/orgs/oam-dev/packages

zzxwill: Got it, thanks. I prefer to pushing to only one registry for quick release build as normally `docker push` might be the same time-consuming step. BTW, can those two push be asynchronous?

morlay: synchronous. one after one in tag list. 
so the behavior will be same as old, push to github cr, push to docker.io

asynchronous mode in early version of buildx but will broken tags in same container registry like `oamdev/vela-core:1 oamdev/vela-core:1.1`. they removed it.

morlay: We could remove one of them when updates `release.yml`

.github/workflows/docker.yml
wonderflow: sure, you could do that in another pr

.github/workflows/docker.yml
wonderflow: I think keep both of them is OK, this github action  is  asynchronous with other action such as `e2e test`.

.github/workflows/docker.yml
```diff
@@ -8,18 +8,38 @@ jobs:
...
+      - uses: docker/build-push-action@v2
+        name: Build & Pushing
+        with:
+          context: .
+          file: Dockerfile
+          labels: |-
+            org.opencontainers.image.source=https://github.com/${{ github.repository }}
+            org.opencontainers.image.revision=${{ github.sha }}
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' }}
```
wonderflow, line 42: This action condition has already made sure that pull_request will not trigger this action, will the condition here block normal push?

https://github.com/oam-dev/kubevela/pull/631/files#diff-3f5366f6d6df3ec1179e5efadc6f350bfa88eebf4e2da589b4d94ccb85ae5e94L2-L5

Dockerfile
wonderflow: @morlay  excellent solution！

wonderflow (approved):
Excellent Job!

Generally LGTM with two questions left:

1. Will this push condition block noraml push ? ` ${{ github.event_name != 'pull_request' }}`
2. Can the command `make kind-load`  still work or we need to add some build args?

wonderflow:
secret `CR_PAT` added

morlay:
@wonderflow 

1. `${{ github.event_name != 'pull_request' }}` only block PRs. push on other branches could works well. (but now `docker.yml` already limit for master branch)
2. ARG `TARGETARCH` could be empty

wonderflow:
@morlay Thanks

morlay:
🥳  multi arch image lands.

image


`ghcr.io` need to make image public here https://github.com/orgs/oam-dev/packages.

wonderflow:
Public now

morlay:
@wonderflow 

for release flow. could updates

* add `on.tags: v*` into `docker.yml`
* remove https://github.com/oam-dev/kubevela/blob/master/.github/workflows/release.yml#L95-L109

how do you think?

wonderflow:
@morlay  It's good for me.

morlay:
@wonderflow https://github.com/oam-dev/kubevela/pull/642

Issue #629, opened 2020-11-19
[Feature] faster 'vela workloads / vela traits' command


wonderflow:
- [x] 1. when install capability, should check if helm chart repo is already exist, if so, don't always add the repo

 
![image](https://user-images.githubusercontent.com/2173670/99606280-6674b780-2a44-11eb-8feb-bdaae7da8709.png)

- [ ] 2. add a local cache file to keep all definition hash in a folder, if hash is the same, ignore update check.
- [ ] 3. add a timestamp in the local cache file, if `vela workloads/traits` were executed multiple times in a short period(for example 1min, ignore update)

resouer:
> give a command to list only, don't always check and refresh , such as vela workloads --pure

-1, I don't want to learn there is different between `list` and `list only`.

captainroy-hy:
@wonderflow I would like to help on this issue.

PR #636, opened 2020-11-20
check helm release first before add repo and install chart

wonderflow:
ref #629

This could make `vela workloads/traits` a little faster


hongchaodeng (approved):
lgtm

Issue #630, opened 2020-11-19
[Feature] vela install display timeout but it was already installed


wonderflow:
![image](https://user-images.githubusercontent.com/2173670/99606390-a20f8180-2a44-11eb-94f8-cae135b6ba81.png)

This is because `vela install` including following steps:

1. install `vela-core` chart
2. after the `vela-core` chart pods start running, it will check configmap named `vela-config` to install dependency charts(promethus, cert-manager, flagger, kede)
3. the capability（Workloads and Traits）can only installed successfully after these dependency‘s CRD exists

The timeout usually caused by step 3. But it could self heal automatically after the helm charts were all installed. So it will finally succeed if you execute `vela workloads` again.

We should change the tips of `vela install`

wonderflow:
The tips were added, closing

PR #637, opened 2020-11-20
add tips about vela install capability not ready instead of directly fail

wonderflow:
it could self-heal after background installation succeed

ref #630


hongchaodeng (approved):
lgtm

PR #640, opened 2020-11-21
support contour ingress controller

weibaohui:
support contour ingress controller

### DIFF
--- a/pkg/controller/v1alpha1/routes/ingress/contour_ingress.go
+++ b/pkg/controller/v1alpha1/routes/ingress/contour_ingress.go
@@ -1,0 +1,203 @@
+package ingress
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+
+	standardv1alpha1 "github.com/oam-dev/kubevela/api/v1alpha1"
+
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	certmanager "github.com/wonderflow/cert-manager-api/pkg/apis/certmanager/v1"
+	cmmeta "github.com/wonderflow/cert-manager-api/pkg/apis/meta/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/networking/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/controller/v1alpha1/routes/ingress/contour_ingress.go
```diff
@@ -0,0 +1,203 @@
...
+
+	standardv1alpha1 "github.com/oam-dev/kubevela/api/v1alpha1"
+
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	certmanager "github.com/wonderflow/cert-manager-api/pkg/apis/certmanager/v1"
+	cmmeta "github.com/wonderflow/cert-manager-api/pkg/apis/meta/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/networking/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
```
jman1973, line 18: []()

pkg/controller/v1alpha1/routes/ingress/contour_ingress.go
```diff
@@ -0,0 +1,203 @@
...
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
zzxwill, line 20: Need to reorganize the imports.

weibaohui: Thanks！
but, this is already the result of goimports, or is there any other specific reference alias suggestions?

pkg/controller/v1alpha1/routes/ingress/contour_ingress.go
weibaohui: Thanks！
but, this is already the result of goimports, or is there any other specific reference alias suggestions?

wonderflow (approved):
LGTM， would you like to add some docs about this feature?

For example, tell we now support one more type other then nginx here https://github.com/oam-dev/kubevela/blob/master/docs/en/developers/references/traits/route.md

wonderflow:
@weibaohui Thanks for you contribution

PR #646, opened 2020-11-23
add supported provider type

weibaohui:
add supported provider type

### DIFF
--- a/docs/en/developers/references/traits/route.md
+++ b/docs/en/developers/references/traits/route.md
@@ -29,6 +29,7 @@ Name | Type | Description | Notes
 **Domain** | **string** | specify your host url for this app | [ default to (empty) ]
 **Issuer** | **string** | specify your certificate issue  | [default to no tls]
 **Rules** | [**[]RouteRules**](#routerules) |  | [optional] 
+**Provider** | **string** | ingress controller name ,we support **nginx**,**[contour](#https://github.com/projectcontour/contour)**  | [empty means using nginx-ingress]
 
 
 ### RouteRules


docs/en/developers/references/traits/route.md
```diff
@@ -29,6 +29,7 @@ Name | Type | Description | Notes
 **Domain** | **string** | specify your host url for this app | [ default to (empty) ]
 **Issuer** | **string** | specify your certificate issue  | [default to no tls]
 **Rules** | [**[]RouteRules**](#routerules) |  | [optional] 
+**Provider** | **string** | ingress controller name ,we support **nginx**,**[contour](#https://github.com/projectcontour/contour)**  | [empty means using nginx-ingress]
```
zzxwill, line 32:
```suggestion
**Provider** | **string** | ingress controller name ,we support **nginx**,**[contour](#https://github.com/projectcontour/contour)**  | [default to nginx, which means using nginx-ingress]
```

Let's specify the default value first:)

resouer:
Note that the user of vela doesn't know what is contour or nginx ingress. So there won't be a 'type' field in trait spec.

Instead, the provider of ingress is a platform admin's concern that should be decided by:

1. trait definition, and be configurable by vela install config
2. or, auto discover by whether contour or nginx ingress is available (default provider should be defined in this case)

1. is simplest and 2. is favorable, leave to @wonderflow for final decision

wonderflow:
@resouer yes, I think we could adopt the first solution as a quick fix, that make the provider field configured in traitdefinition template.

wonderflow:
closing as we have automatically generate this doc

PR #663, opened 2020-11-26
Merging oam-runtime into vela-core

wonderflow:
- [x] 1. merge apis,pkg libs
- [x] 2. change imports headers
- [x] 3. merge tests
- [x] 4. merge main.go, including all controllers
- [x] 5. merge scripts, makefiles, charts
- [x] 6. CI checks

---
 docs, design, examples will not merge

PR #664, opened 2020-11-26
fix lint issues 

captainroy-hy:
Most nolint comments are removed and fixed, except:
//nolint:unparam
//nolint:gosec
//nolint:errcheck

### DIFF
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -216,7 +216,7 @@ func waitWebhookSecretVolume(certDir string, timeout, interval time.Duration) er
 				// nolint
 				defer f.Close()
 				// check if dir is empty
-				if _, err := f.Readdir(1); err == io.EOF {
+				if _, err := f.Readdir(1); errors.Is(err, io.EOF) {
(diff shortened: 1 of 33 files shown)


wonderflow:
please hold until #663 merged, thanks, there're too many overlap on file changes.

cmd/core/main.go
```diff
@@ -216,7 +216,7 @@ func waitWebhookSecretVolume(certDir string, timeout, interval time.Duration) er
 				// nolint
 				defer f.Close()
 				// check if dir is empty
-				if _, err := f.Readdir(1); err == io.EOF {
+				if _, err := f.Readdir(1); errors.Is(err, io.EOF) {
```
zzxwill, line 219: Just curious, is using function better?

captainroy-hy: As the linter says, `comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error`

wonderflow (approved):

PR #670, opened 2020-11-28
fix webhook about Trait-Applyto

captainroy-hy:
- fix validating-webhook of appconfig controller cannot recognize workloadDefinitionName
- fix minor lint issues

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply.go
@@ -84,6 +84,7 @@ type workloads struct {
 	dm             discoverymapper.DiscoveryMapper
 }
 
+//nolint:errorlint
 func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error {
 	// they are all in the same namespace
 	var namespace = w[0].Workload.GetNamespace()
@@ -91,6 +92,7 @@ func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus,
 		if !wl.HasDep {
 			err := a.patchingClient.Apply(ctx, wl.Workload, ao...)
 			if err != nil {
+				// TODO(roywang) use errors.As() insteand of type assertion on error
 				if _, ok := err.(*GenerationUnchanged); !ok {
 					// GenerationUnchanged only aborts applying current workload
 					// but not blocks the whole reconciliation through returning an error
@@ -104,6 +106,7 @@ func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus,
 			}
 			t := trait.Object
 			if err := a.updatingClient.Apply(ctx, &trait.Object, ao...); err != nil {
+				// TODO(roywang) use errors.As() insteand of type assertion on error
 				if _, ok := err.(*GenerationUnchanged); !ok {
 					// GenerationUnchanged only aborts applying current trait
 					// but not blocks the whole reconciliation through returning an error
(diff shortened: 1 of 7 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -84,6 +84,7 @@ type workloads struct {
 	dm             discoverymapper.DiscoveryMapper
 }
 
+//nolint:errorlint
```
wonderflow, line 87: which errorlint do we ignore?

pkg/oam/util/test_utils.go
```diff
@@ -102,6 +103,7 @@ func (matcher ErrorMatcher) FailureMessage(actual interface{}) (message string)
 }
 
 // NegatedFailureMessage builds an error message.
+//nolint:errorlint
```
wonderflow, line 106: why?

pkg/webhook/core.oam.dev/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -187,14 +194,14 @@ func ValidateTraitAppliableToWorkloadFn(_ context.Context, v ValidatingAppConfig
 				if strings.HasPrefix(applyTo, "*.") && workloadGroup == applyTo[2:] {
 					continue ValidateApplyTo
 				}
-				if workloadType == applyTo ||
-					workloadDefRefName == applyTo {
+				if workloadDefRefName == applyTo ||
+					workloadDefName == applyTo {
```
wonderflow, line 198: we should rename them and add explaination here:
```
// according to OAM convention, Spec.Reference.Name in workloadDefinition is CRD name
var crdName = workloadDefRefName
// according to OAM convention, name of workloadDefinition is the workload type.
var workloadTypeName = workloadDefName
```

pkg/oam/util/test_utils.go
captainroy-hy: Linter warns that it should use errors.As() instead of type assertion for it will fail on wrapped error. But if I use `errors.As()` to refactor the func, it may result in unpredictable issues since it's a util func. So I just comment nolint for now. 
BTW, the same issues happen in the project several times, I would like to fix these `nolint:errorlint` latter.

pkg/webhook/core.oam.dev/v1alpha2/applicationconfiguration/validating_handler.go
captainroy-hy: got it.

pkg/webhook/core.oam.dev/v1alpha2/applicationconfiguration/validating_handler_test.go
```diff
@@ -332,37 +347,28 @@ func TestValidateTraitAppliableToWorkloadFn(t *testing.T) {
...
 							}},
 						},
 					},
 				},
 			},
 			want: []error{fmt.Errorf(errFmtUnappliableTrait,
-				"example.com/v1, Kind=TestTrait", "unknown.group/v1, Kind=TestWorkload1", "example-comp",
-				[]string{"example.com", "TestWorkload"})},
+				"TestTrait", "TestWorkload", "example-comp",
+				[]string{"example.com", "TestWorkload2"})},
```
wonderflow, line 371: I don't see any invalid cases here, please add some cases that don't match applyTo rule and will be blocked by webhook.

pkg/oam/util/test_utils.go
wonderflow: I see, thanks for your explain, please add TODO and explain in it,thanks!

pkg/webhook/core.oam.dev/v1alpha2/applicationconfiguration/validating_handler_test.go
captainroy-hy: Sure, I will complete unit tests.

wonderflow (approved):

PR #684, opened 2020-12-01
add svc flag for exec command.

majian159:
add svc flag for exec command.


wonderflow (approved):
Can you please add a test for this case

majian159:
> Can you please add a test for this case

unit test created

Issue #658, opened 2020-11-25
[Feature] Add controller startup parameters to allow disbale inner workload/trait controller


wonderflow:
By default, vela-core will start all inner controllers. We could add some prameters such as:

```
./vela-core --disable-cap "route,metrics"
```

To disable these inner controller to run.

Note: we should also distinguish them in webhook part.

zzxwill:
For what purpose were some builtin capabilities be disabled?

wonderflow:
@zzxwill  we should allow users can only use oam-framework, especially for user who are migrated from oam-runtime

PR #687, opened 2020-12-01
Support disable installation of builtin capabilities

zzxwill:
Added vela core bootstrap flag `--enable-caps` to support no
installation for builtin capabilities

Feature #658

### DIFF
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -4,8 +4,8 @@
 
 replicaCount: 1
 useWebhook: true
-# By default, enable all builtin capabilities
-enableCaps: all
+# By default, don't disable any builtin capabilities
+disableCaps: ""
 image:
   repository: oamdev/vela-core
   tag: latest
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -4,6 +4,8 @@
 
 replicaCount: 1
 useWebhook: true
+# By default, enable all builtin capabilities
+enableCaps: all
 image:
   repository: oamdev/vela-core
   tag: latest
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -9,7 +9,6 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strconv"
-	"strings"
 	"syscall"
 	"time"
 
@@ -38,6 +37,7 @@ import (
 	oamcontroller "github.com/oam-dev/kubevela/pkg/controller/core.oam.dev"
 	oamv1alpha2 "github.com/oam-dev/kubevela/pkg/controller/core.oam.dev/v1alpha2"
 	"github.com/oam-dev/kubevela/pkg/controller/dependency"
+	"github.com/oam-dev/kubevela/pkg/controller/utils"
 	velawebhook "github.com/oam-dev/kubevela/pkg/webhook"
 	oamwebhook "github.com/oam-dev/kubevela/pkg/webhook/core.oam.dev/v1alpha2"
 )
@@ -74,7 +74,7 @@ func main() {
 	var useWebhook, useTraitInjector bool
 	var controllerArgs oamcontroller.Args
 	var healthAddr string
-	var enableCaps string
+	var disableCaps string
 
 	flag.BoolVar(&useWebhook, "use-webhook", false, "Enable Admission Webhook")
 	flag.BoolVar(&useTraitInjector, "use-trait-injector", false, "Enable TraitInjector")
@@ -93,7 +93,7 @@ func main() {
 	flag.StringVar(&healthAddr, "health-addr", ":9440", "The address the health endpoint binds to.")
 	flag.BoolVar(&controllerArgs.ApplyOnceOnly, "apply-once-only", false,
 		"For the purpose of some production environment that workload or trait should not be affected if no spec change")
-	flag.StringVar(&enableCaps, "enable-caps", "all", "To be enabled builtin capability list.")
+	flag.StringVar(&disableCaps, "disable-caps", "", "To be disabled builtin capability list.")
 	flag.Parse()
 
 	// setup logging
@@ -141,14 +141,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	var capabilities []string
-	switch enableCaps {
-	case "none":
-		capabilities = []string{}
-	case "all":
-		capabilities = []string{"metrics", "podspecworkload", "route", "autoscale"}
-	default:
-		capabilities = strings.Split(enableCaps, ",")
+	if err := utils.CheckDisabledCapabilities(disableCaps); err != nil {
+		setupLog.Error(err, "unable to get enabled capabilities")
+		os.Exit(1)
 	}
 
 	if useWebhook {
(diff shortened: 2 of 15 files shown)


zzxwill:
@wonderflow I changed flag `--disable-cap` to `--enable-caps` at last. 
Reasons:
- Using value `none` can also support disabling all builtin controllers. (By default, value `all` will support enabling all capabilities.)
- It will be easier to set the value for the flag if platform builders want to enable one or more builtin capabilities one day.
For example, there are four capabilities `"metrics", "podspecworkload", "route", "autoscale"`, if builders want to disable `metrics`, we have to set `--disable-caps=podspecworkload,route,autoscale`. But now, we can just set `--enable-caps=metrics`. It seems straighter.

charts/vela-core/values.yaml
```diff
@@ -4,6 +4,8 @@
 
 replicaCount: 1
 useWebhook: true
+# By default, enable all builtin capabilities
+enableCaps: all
```
wonderflow, line 8: 一方面为了兼容性，另一方面我们也倾向于别人安装所有内置能力，应该反着来，参数是 disableCaps，默认为空不disable

wonderflow: 参数对应的文档也写一下哈，包括其他参数，可以另开一个issue

cmd/core/main.go
```diff
@@ -138,13 +141,23 @@ func main() {
 		os.Exit(1)
 	}
 
+	var capabilities []string
+	switch enableCaps {
+	case "none":
```
ryanzhang-oss, line 146: how do users know those values?

zzxwill: They don't know, I will document the values.

Approximately, they are the `TYPE` from `vela workloads/traits`.
```
TYPE       	CATEGORY	DESCRIPTION
+task      	workload	One-off task to run a piece of code or script to completion
+webservice	workload	Long-running scalable service with stable endpoint to receive external traffic
+worker    	workload	Long-running scalable backend worker without network endpoint
+autoscale 	trait   	Automatically scale the app following certain triggers or metrics
+metrics   	trait   	Configure metrics targets to be monitored for the app
+rollout   	trait   	Configure canary deployment strategy to release the app
+route     	trait   	Configure route policy to the app
+scaler    	trait   	Manually scale the app
```

But in fact we only support "metrics", "route", "autoscale" as others don't have backend controllers. Besides, the values include "podspecworkload".

zzxwill: It's not friendly for platform builders. Fortunately, most of time, just setting the flag to `all` is enough to make Vela compatible with previous OAM Kubernetes Runtime.

cmd/core/main.go
```diff
@@ -12,6 +12,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/oam-dev/kubevela/pkg/controller/utils"
```
wonderflow, line 15: nit

charts/vela-core/values.yaml
```diff
@@ -4,6 +4,8 @@
 
 replicaCount: 1
 useWebhook: true
+# By default, don't disable any builtin capabilities
+disableCaps: none
```
wonderflow, line 8:
```suggestion
disableCaps: ''
```

不要 none， 空就代表没有。甚至不要这个参数

pkg/controller/setup.go
```diff
@@ -17,22 +17,35 @@ limitations under the License.
 package controller
 
 import (
+	mapset "github.com/deckarep/golang-set"
 	ctrl "sigs.k8s.io/controller-runtime"
 
+	"github.com/oam-dev/kubevela/pkg/controller/common"
```
wonderflow, line 23: nit

charts/vela-core/values.yaml
zzxwill: 这是 values.yaml，要么写、要么不写，不存在不要。因为 build 出来的 chart，不推荐用户去删减 values.yaml。

文档里这里会有三类合法值
- ""/none
- all
- "autoscale,route"

改正 “” 我接受的。谢谢。

cmd/core/main.go
```diff
@@ -138,13 +141,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	capabilities, err := utils.GetEnabledCapabilities(disableCaps)
+	if err != nil {
+		setupLog.Error(err, err.Error())
```
wonderflow, line 146:
```suggestion
		setupLog.Error(err, "unable to get enabled capabilities")
```

两个一样的error message没必要输出2次

pkg/controller/utils/utils.go
```diff
@@ -106,3 +109,26 @@ func SelectOAMAppLabelsWithoutRevision(labels map[string]string) map[string]stri
...
+		var disableSlice []interface{}
+		for _, c := range strings.Split(disableCaps, ",") {
+			if !allBuiltinCapabilities.Contains(c) {
+				return nil, errors.New("disable caps list is invalid")
+			}
+			disableSlice = append(disableSlice, c)
+		}
+		capabilities = allBuiltinCapabilities.Difference(mapset.NewSetFromSlice(disableSlice))
+	}
+	return capabilities, nil
```
wonderflow, line 133: 这里我有个疑问，以后我们新写的workload和trait 代码，用什么机制保证不会忘了加到这个Set里？

1. 默认情况下，我希望我们不需要额外的机制，就能启动所有内置的Controller
2. 如果disable的是all，我希望我们也不需要额外的机制，就能只启动oam-runtime原来的Controller。

现在这个函数似乎很容易导致两个默认情况出问题？

zzxwill: Ryan 也问了类似的问题，我们目前的机制是这样的。
```
# pkg/controller/setup.go
// Setup workload controllers.
func Setup(mgr ctrl.Manager) error {
	for _, setup := range []func(ctrl.Manager) error{
		metrics.Setup, podspecworkload.Setup, routes.Setup,
		autoscalers.Setup,
	} {
		if err := setup(mgr); err != nil {
			return err
		}
	}
	return nil
}
```
通过 controller 包名注册 Setup，这里有几个问题：
- 这里的包名跟 definition name 不一定相等
- Golang 的反射机制貌似不支持 package string 转换为包，然后调用函数的能力
- 一个 capability 是否有后端 controller，也没办法确定性的判断出来

所以，我没想到合适的办法。

pkg/controller/utils/utils.go
```diff
@@ -106,3 +114,29 @@ func SelectOAMAppLabelsWithoutRevision(labels map[string]string) map[string]stri
...
+func CheckDisabledCapabilities(disableCaps string) error {
+	switch disableCaps {
+	case common.DisableNoneCaps:
+		return nil
+	case common.DisableAllCaps:
+		return nil
+	default:
+		for _, c := range strings.Split(disableCaps, ",") {
+			if !allBuiltinCapabilities.Contains(c) {
+				return errors.New("disable caps list is invalid")
```
wonderflow, line 128:
```suggestion
				return fmt.Errorf("%s in disable caps list is not built-in", c)
```

pkg/webhook/register.go
```diff
@@ -4,6 +4,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
+	"github.com/oam-dev/kubevela/pkg/controller/utils"
+
+	"github.com/oam-dev/kubevela/pkg/controller/common"
+
```
wonderflow, line 10: nit

wonderflow (approved):
LGTM

PR #689, opened 2020-12-01
Move GetConfig into specific commands

horis233:
Fix https://github.com/oam-dev/kubevela/issues/655

This PR move GetConfig into specific vela commands to make sure other commands can work without kubeconfig.

During updating, I found some commands that don't need the controller-runtime client arguments. Thus, I removed the controller-runtime client argument from their function parameters.

Moreover, I found some imports don't follow the `goimports` formatting, I sort them up as well.

### DIFF
--- a/pkg/commands/capability.go
+++ b/pkg/commands/capability.go
@@ -9,10 +9,9 @@ import (
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/oam-dev/kubevela/pkg/oam/discoverymapper"
-
 	"github.com/oam-dev/kubevela/apis/types"
 	cmdutil "github.com/oam-dev/kubevela/pkg/commands/util"
+	"github.com/oam-dev/kubevela/pkg/oam/discoverymapper"
 	"github.com/oam-dev/kubevela/pkg/serverlib"
 )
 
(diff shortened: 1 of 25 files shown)


wonderflow:
@horis233  Thanks very much for your contribution, can you help improve the code test, make sure the CI won't fail for that.

pkg/commands/capability.go
```diff
@@ -22,12 +22,20 @@ func CapabilityCommandGroup(c types.Args, ioStream cmdutil.IOStreams) *cobra.Com
...
 		Short: "Manage capability centers and installing/uninstalling capabilities",
 		Long:  "Manage capability centers and installing/uninstalling capabilities",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			restConf, err := config.GetConfig()
+			if err != nil {
+				return err
+			}
+			c.Config = restConf
+			return nil
+		},
```
wonderflow, line 32: I  see lot's of places has the same function, do you mind to write a common one?

horis233: @wonderflow Thanks for your comments, I create a common function in https://github.com/oam-dev/kubevela/blob/5e4f270b1e50e16b0a670ee6b3ac4d94bbc90e86/apis/types/args.go#L16-L23

wonderflow (approved):
Thanks! LGTM after CI pass

PR #690, opened 2020-12-02
add a tip for make diff fail

wonderflow:


zzxwill (approved):
LGTM.

This prompt is helpful. `make diff` output won't do any help for debugging the failure some times, but `make reviewable` will. Thanks.

PR #702, opened 2020-12-03
Remove automatically prefixed "[BUG]" in the issue title

zzxwill:
Remove "[BUG]" from the default issue as maintainers can
use labeling mechanism to tell whether it is


wonderflow (approved):
GOOD

Issue #698, opened 2020-12-02
[Feature] Using Terraform HCL as IaC module in KubeVela


zzxwill:
**Background**

Questions from @hongchaodeng 
1. Why Terraform?
Terraform is the most famous tool to deploy infrastructure resources, and even Kubernetes resources.

2. Why should KubeVela integrate Terraform?
KubeVela is open, and can interacts with Open Source community with low efforts.
KubeVela, as an App Platform, which, in most of time, consumes backend services, which could be easily applied by Terraform.

3. Who will use Terraform in KubeVela? How?
Platform builders who adopt KubeVela, but at the same time, need backend services which they generated by Terraform before.

4. What's the difference between using Terraform directly and using it in KubeVela?
For better consistency experience around appfile.

5. Why don't we regard KubeVela as a Terraform Provider?
To become a custom Terraform builder, 
1). KubeVela has to follow the specification on creating a custom Terraform builder https://learn.hashicorp.com/collections/terraform/providers
2). More importantly, KubeVela isn't the platform to product resources, Service/Application/Operation. They are not independent, or just have simple dependent relationship on each other. Writing CRUD for each object to be deployable by Terraform could not grant Operations could apply on Services in Application.

**Requirement**
1. Empower KubeVela to deploy infrastructure services, like database; and maybe Kubernetes resource by [Kubernetes provider](https://learn.hashicorp.com/tutorials/terraform/kubernetes-provider).
2. Provide a mechanism for applications to consume the services.

**Describe the solution you'd like**

Make it possible to use .tf modules in OAM definitions for cloud resources provision.

resouer:
Please see this comment: https://github.com/oam-dev/kubevela/pull/708#issuecomment-738247267, we need a design and tech review before implementation.

zzxwill:
> Please see this comment: [#708 (comment)](https://github.com/oam-dev/kubevela/pull/708#issuecomment-738247267), we need a design and tech review before implementation.

Sure. I just spent 40 minutes to have a quick PoC, try to extend vela up,  and hear feedback. Will consider the comments from you and Hongchao, then make a solid design. Thanks.

wonderflow:
Let's do a terraform demo like this:
```
apiVersion: core.oam.dev/v1alpha2
kind: WorkloadDefinition
metadata:
  name: ossbucket
  annotations:
    definition.oam.dev/description: "OSS Bucket"
spec:
  definitionRef:
    name: "osss.dummy"
  extension:
[...]

PR #708, opened 2020-12-03
Integrate Terraform into KubeVela

zzxwill:
Integrated Terraform into KubeVela to enable it to deploy
infrastructure resources by `vela up`

#698

### DIFF
--- a/pkg/appfile/addon.go
+++ b/pkg/appfile/addon.go
@@ -1,0 +1,49 @@
+package appfile
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/oam-dev/kubevela/pkg/utils/common"
+)
+
+// Terraform is an addon
+const Terraform = "terraform"
+
+// Addon type, like `terraform`
+type Addon map[string]interface{}
+
+// DeployAddon deploys addon resources
+func DeployAddon(addons map[string]Addon) error {
+	for k, v := range addons {
+		switch k {
+		case Terraform:
+			fmt.Printf(k, v)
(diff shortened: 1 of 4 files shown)


zzxwill:
I did a quick PoC to integrate Terraform in this draft PR and write a doc on how to deploy Alibaba Cloud OSS resource with the power of Terraform by `vela up` at https://github.com/zzxwill/catalog/blob/terraform/applications/terraform/README.md.

image

resouer:
/hold

Simply execute .tf by Appfile is really not what we want.

First of all, the Terraform module should be registered in the same way of CUE modules, i.e. via definitions. And then, a short term solution is execute the .tf to provision the cloud resources, then output the connection information to a secret/config so the services in Appfile could consume it.

The long term solution, which I think is quite interesting, is we fully decouple the model layer from Kubernetes. So people can register any IaC module with definitions, and then Appfile (OAM user interface object) will become truly platform agnostic. Please also see: https://github.com/aws/aws-proton-public-roadmap/issues/1 /cc @rynowak since we've talked this in slack.

/cc @hongchaodeng ,please help to review if the short term goal make sense, and if yes, help @zzxwill to deliver a design before implementation.

ryanzhang-oss:
just curious, where is the "teraform" command line binary installed? IIRC, the change basically invokes a command line "teraform xxx" during the `vela up` execution.

wonderflow:
To make it clear and avoid confuse others. I have to say, we will never integrate terraform as a raw file in APPFILE like this.

resouer:
Let's close this one and move to design phase.

PR #715, opened 2020-12-04
fix task(k8s job) definition restartpolicy

912988434:
fix task(k8s job) definition restartpolicy


charts/vela-core/templates/defwithtemplate/task.yaml
```diff
@@ -35,7 +36,10 @@ spec:
       	// +usage=Which image would you like to use for your service
       	// +short=i
       	image: string
-      
+       
+        // +usage=Restart policy for all containers within the pod
+        policy: string 
+
```
wonderflow, line 42: good job, you could try `make reviewable` to fix the format

PR #717, opened 2020-12-04
build version into vela-core and lock helm version on release

### LINKED ISSUE #531: print build time or git hash at start of vela core server
This helps to pinpoint user errors.
For example: https://github.com/oam-dev/kubevela/issues/529#issuecomment-723361052

### LINKED ISSUE #712: Vela Release should lock the image version in chart values.yaml
refer #710

wonderflow:
1. This PR will use `sed` to replace image tag from `latest` to release version, so our chart will lock in the fixed image version. fix #712
2. Build vela-core image with version and git revision, so we can get the specified version for debug. fix #531

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,6 +19,11 @@ jobs:
             VERSION=latest
           fi
           echo ::set-output name=VERSION::${VERSION}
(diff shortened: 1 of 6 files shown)


hongchaodeng (approved):
lgtm

PR #723, opened 2020-12-04
fix task( k8s job ) restartpolicy default Always

912988434:
Update the restartpolicy default value Always-->Never。
Task creates job RestartPolicy by default with a value of Always, which obviously does not work. I define a parameter policy, which defaults to Never.

### DIFF
--- a/charts/vela-core/templates/defwithtemplate/task.yaml
+++ b/charts/vela-core/templates/defwithtemplate/task.yaml
@@ -16,7 +16,7 @@ spec:
       		parallelism: parameter.count
       		completions: parameter.count
       		template: spec: {
-      			restartPolicy: parameter.policy
+      			restartPolicy: parameter.restartPolicy
       			containers: [{
       				name:  context.name
       				image: parameter.image
@@ -38,7 +38,7 @@ spec:
       	image: string
       
       	// +usage=Define job restart policy, the value should be Never or OnFailure. By default, it's Never.
-      	policy: *"Never" | string
+      	restartPolicy: *"Never" | string
       
       	// +usage=Commands to run in the container
       	cmd?: [...string]
--- a/charts/vela-core/templates/defwithtemplate/task.yaml
+++ b/charts/vela-core/templates/defwithtemplate/task.yaml
@@ -37,8 +37,9 @@ spec:
       	// +short=i
       	image: string
       
-        // +usage=Define job restartpolicy policy, the value should be Never or OnFailur,,is not Always
-        policy: *"Never"|string
+	// +usage=Define job restart policy, the value should be Never or OnFailure. By default, it's Never.
+	policy: *"Never" | string
+
       	// +usage=Commands to run in the container
       	cmd?: [...string]
       }
--- a/charts/vela-core/templates/defwithtemplate/task.yaml
+++ b/charts/vela-core/templates/defwithtemplate/task.yaml
@@ -16,7 +16,7 @@ spec:
       		parallelism: parameter.count
       		completions: parameter.count
       		template: spec: {
-      			restartPolicy: parameter.restartPolicy
+      			restartPolicy: parameter.restart
       			containers: [{
       				name:  context.name
       				image: parameter.image
@@ -38,7 +38,7 @@ spec:
       	image: string
       
       	// +usage=Define job restart policy, the value should be Never or OnFailure. By default, it's Never.
-      	restartPolicy: *"Never" | string
+      	restart: *"Never" | string
       
       	// +usage=Commands to run in the container
       	cmd?: [...string]
--- a/charts/vela-core/templates/defwithtemplate/task.yaml
+++ b/charts/vela-core/templates/defwithtemplate/task.yaml
@@ -16,6 +16,7 @@ spec:
       		parallelism: parameter.count
       		completions: parameter.count
       		template: spec: {
+                        restartPolicy: parameter.policy
       			containers: [{
       				name:  context.name
       				image: parameter.image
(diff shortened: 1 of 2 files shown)


wonderflow:
please refer to https://github.com/oam-dev/kubevela/pull/713/files

You should change `hack/vela-templates/cue/task.cue`.

Make sure it's formatted by execute `make reviewable`

charts/vela-core/templates/defwithtemplate/task.yaml
```diff
@@ -36,6 +37,8 @@ spec:
       	// +short=i
       	image: string
       
+        // +usage=Define job restartpolicy policy, the value should be Never or OnFailur,,is not Always
```
wonderflow, line 40:
```suggestion
        // +usage=Define job restart policy, the value should be Never or OnFailure. By default, it's Never.
```

charts/vela-core/templates/defwithtemplate/task.yaml
```diff
@@ -36,6 +37,8 @@ spec:
...
+        policy: *"Never"|string
```
resouer, line 41: Let's name this parameter 'restart'.

Similar to: https://docs.docker.com/config/containers/start-containers-automatically/

charts/vela-core/templates/defwithtemplate/task.yaml
```diff
@@ -16,6 +16,7 @@ spec:
       		parallelism: parameter.count
       		completions: parameter.count
       		template: spec: {
+                        restartPolicy: parameter.policy
```
ryanzhang-oss, line 19: The CUE parameter doesn't have a policy field, where is this referring to?  You can just hardcode it to be "Never" here

ryanzhang-oss (changes requested):
The CUE parameter doesn't have a policy field, where is this referring to? Please add the parameter or you can just hardcode it to be "Never".

charts/vela-core/templates/defwithtemplate/task.yaml
912988434: sorry,task.cue has been updated

charts/vela-core/templates/defwithtemplate/task.yaml
912988434: I feel that following the naming in K8S can make it easier to understand this parameter

resouer: On the contrary, 'policy' doesn't make any sense from user's perspective. It could be rollout policy, update policy etc.

We either name it as 'restart' or 'restartPolicy' for user parameter.

912988434: Yes, restartpolicy seems more appropriate as a parameter

resouer: I'd suggest `restart` for short ^^

charts/vela-core/templates/defwithtemplate/task.yaml
```diff
@@ -36,6 +37,9 @@ spec:
       	// +short=i
       	image: string
       
+      	// +usage=Define job restart policy, the value should be Never or OnFailure. By default, it's Never.
+      	restartPolicy: *"Never" | string
```
wonderflow, line 41:
```suggestion
      	restart: *"Never" | string
```

Let's name it `restart` for short

charts/vela-core/templates/defwithtemplate/task.yaml
```diff
@@ -16,6 +16,7 @@ spec:
       		parallelism: parameter.count
       		completions: parameter.count
       		template: spec: {
+      			restartPolicy: parameter.restartPolicy
```
wonderflow, line 19:
```suggestion
      			restartPolicy: parameter.restart
```

hack/vela-templates/cue/task.cue
```diff
@@ -5,6 +5,7 @@ output: {
 		parallelism: parameter.count
 		completions: parameter.count
 		template: spec: {
+			restartPolicy: parameter.restartPolicy
```
wonderflow, line 8:
```suggestion
			restartPolicy: parameter.restart
```

hack/vela-templates/cue/task.cue
```diff
@@ -25,6 +26,9 @@ parameter: {
 	// +short=i
 	image: string
 
+	// +usage=Define job restart policy, the value should be Never or OnFailure. By default, it's Never.
+	restartPolicy: *"Never" | string
```
wonderflow, line 30:
```suggestion
	restart: *"Never" | string
```

charts/vela-core/templates/defwithtemplate/task.yaml
912988434: Ok, long parameters are a bit cumbersome

charts/vela-core/templates/defwithtemplate/task.yaml
912988434: Ok, long parameters are a bit cumbersome

wonderflow (approved):

wonderflow:
ping @ryanzhang-oss

charts/vela-core/templates/defwithtemplate/task.yaml
```diff
@@ -36,6 +37,9 @@ spec:
       	// +short=i
       	image: string
       
+      	// +usage=Define job restart policy, the value should be Never or OnFailure. By default, it's Never.
```
ryanzhang-oss, line 40:
```suggestion
      	// +usage=Define the job restart policy, the value can only be Never or OnFailure. By default, it's Never.
```

ryanzhang-oss (approved):

hack/vela-templates/cue/task.cue
```diff
@@ -25,6 +26,9 @@ parameter: {
 	// +short=i
 	image: string
 
+	// +usage=Define job restart policy, the value should be Never or OnFailure. By default, it's Never.
```
wonderflow, line 29:
```suggestion
	// +usage=Define the job restart policy, the value can only be Never or OnFailure. By default, it's Never.
```

PR #724, opened 2020-12-04
WIP: add fields for OAM AppConfig and Component to support CUE server side extension

wonderflow:
- [x] 1. Add fields for OAM AppConfig and Component. 
- [ ] 2. Add `template`fields for Definition.
- [ ] 3. Add webhook validation for both fields.

The name is now aligned to OAM Spec. 

https://github.com/oam-dev/spec/blob/master/3.component.md#examples 

```
kind: Component
metadata:
  name: frontend
spec:
  type: Server # <--- the type of the workload
  settings: # <---- the settings of the workload
    osType: linux
    containers:
    - name: my-cool-workload
      image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
      resources:
      ....      
```

https://github.com/oam-dev/spec/blob/master/7.application_configuration.md#example

```
kind: ApplicationConfiguration
spec:
  components:
      ...
      traits:
        - name: manualscaler
          properties:
            replicaCount: 2
```

But maybe there's a little weird that we using `name/properties` in trait while using `type/settings` in component.

 @resouer @hongchaodeng @ryanzhang-oss  What do you think? We will make final decision here. Once confirmed we will not change as we must keep consistency and compatibility for it.

### DIFF
--- a/apis/core.oam.dev/v1alpha2/core_types.go
+++ b/apis/core.oam.dev/v1alpha2/core_types.go
@@ -215,10 +215,21 @@ type ComponentSpec struct {
 	// A Workload that will be created for each ApplicationConfiguration that
 	// includes this Component. Workload is an instance of a workloadDefinition.
 	// We either use the GVK info or a special "type" field in the workload to associate
-	// the content of the workload with its workloadDefinition
+	// the content of the workload with its workloadDefinition.
+	// The 'workload' field is mutually exclusive with the 'type' and 'settings' field.
 	// +kubebuilder:validation:EmbeddedResource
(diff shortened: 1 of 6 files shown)


resouer:
The spec is the final decision as long as the absence is in component/appconfig layer.

wonderflow:
@resouer Can you help explain the naming reason here once more?

resouer:
@wonderflow it's fully inherit from v1alpha1, and the reason for v1alpha1 is trait is a 'type', not instance.

wonderflow:
closing as we won't support this way recently. We will just use application object.

PR #725, opened 2020-12-04
fix long text port forword

Cweiping:
cut `port-forword` Usage text, add Example text
Close: https://github.com/oam-dev/kubevela/issues/696


wonderflow (approved):
Cool! LGTM

PR #728, opened 2020-12-05
fix cmd uitable wrap issue

captainroy-hy:
- Issue: too much content in one table cell makes the table mess.

Before:
image
Fix: specify `MaxColWidth=80 and Wrap=true` to uitable 
After:
image

- move funcs/vars used for printing into one go file.

### DIFF
--- a/pkg/commands/config.go
+++ b/pkg/commands/config.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"strings"
 
-	"github.com/gosuri/uitable"
 	"github.com/spf13/cobra"
 
 	"github.com/oam-dev/kubevela/apis/types"
@@ -75,8 +74,7 @@ func ListConfigs(ioStreams cmdutil.IOStreams, cmd *cobra.Command) error {
 	if err != nil {
 		return err
 	}
-	table := uitable.New()
-	table.MaxColWidth = 60
+	table := newUITable()
 	table.AddRow("NAME")
 	cfgList, err := listConfigs(d)
 	if err != nil {
--- a/pkg/commands/print.go
+++ b/pkg/commands/print.go
@@ -1,0 +1,48 @@
+package commands
+
+import (
+	"time"
+
+	"github.com/briandowns/spinner"
+	"github.com/fatih/color"
+	"github.com/gosuri/uitable"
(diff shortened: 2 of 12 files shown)


wonderflow:
Why the first one has `category` but he second don't?

captainroy-hy:
> Why the first one has `category` but he second don't?

The first one is from `vela install`, the second one is `vela workloads`, sorry for confusing screenshots.

pkg/commands/config.go
```diff
@@ -75,7 +74,7 @@ func ListConfigs(ioStreams cmdutil.IOStreams, cmd *cobra.Command) error {
 	if err != nil {
 		return err
 	}
-	table := uitable.New()
+	table := newUITable()
 	table.MaxColWidth = 60
```
wonderflow, line 78: some is 60, while you defined 80. Can you try to verify all of these commands whether they should use 60 or 80. I prefer 80, but I'm afraid some of them defined to 60 because there're too many columns.

captainroy-hy: got it, I also prefer 80 and will check all.

pkg/commands/print.go
```diff
@@ -0,0 +1,48 @@
...
+var (
+	emojiSucceed   = emoji.Sprint(":check_mark_button:")
+	emojiFail      = emoji.Sprint(":cross_mark:")
+	emojiLightBulb = emoji.Sprint(":light_bulb:")
+)
+
+// newUITable creates a new table with fixed MaxColWidth
+func newUITable() *uitable.Table {
+	t := uitable.New()
+	t.MaxColWidth = 80
```
zzxwill, line 30: In most lines of setting MaxColWidth, the value is 60. Why do you change the it to 80? To make the display more readable for `vela traits/workloads`?

captainroy-hy: Yes, currently capability description contains much more content to display.

zzxwill: It might be better to set the default MaxColWidth to 60, and only set it to 80 for `vela workloads/traits`. These lines can be removed.

image

zzxwill (approved):
LGTM

wonderflow (approved):

PR #742, opened 2020-12-08
Extend application's capabilities by trait definition

leejanee:
Expand the application's capabilities in the following two ways

1. Patch the workload CR by describing the patch content in the definition
2. Create association CR for workload by describing the output object in the definition

- [x] _samples in config/samples/vela-server_


config/samples/vela-server/Demo.md
```diff
@@ -0,0 +1,145 @@
+# Vela Server
```
hongchaodeng, line 1: This shouldn't be called `vela server`. Probably `application`?

config/samples/vela-server/template.yaml
```diff
@@ -59,18 +59,79 @@ spec:
...
+    name: "null"
   extension:
     template: |-
-      output: {
-      	apiVersion: "core.oam.dev/v1alpha2"
-      	kind:       "ManualScalerTrait"
-      	spec: {
-      		replicaCount: parameter.replicas
-      	}
+      patch: {
```
hongchaodeng, line 65: Could you add some doc?
I couldn't find any doc describing what patch does.

hongchaodeng:
Why just update existing templates and CLI to use server side CUE?

hongchaodeng:
Why not just update existing templates and CLI to use server side CUE?

config/samples/vela-server/Demo.md
leejanee: Agree, I will edit it

wonderflow (approved):
LGTM，I think we can fix in the following PRs. @hongchaodeng

hongchaodeng:
Sounds great! @wonderflow

PR #743, opened 2020-12-08
faster refresh capabilities

### LINKED ISSUE #629: [Feature] faster 'vela workloads / vela traits' command

captainroy-hy:
to fix #629 


pkg/commands/refresh.go
```diff
@@ -10,6 +16,8 @@ import (
 	cmdutil "github.com/oam-dev/kubevela/pkg/commands/util"
 	"github.com/oam-dev/kubevela/pkg/plugins"
 	"github.com/oam-dev/kubevela/pkg/utils/system"
+
+	hashstructure "github.com/mitchellh/hashstructure/v2"
 )
 
 type refreshStatus string
```
zzxwill, line 23: 我见到比较多的 type name 首字母是大写的，请确认一下：）

pkg/commands/refresh.go
```diff
@@ -164,5 +213,78 @@ func refreshResultReport(newCaps, oldCaps []types.Capability) map[refreshStatus]
...
 }
+
+// checkRefreshInterval checks whether refresh caps from cluster or not
+// if a short time passed since last refresh, return true to use cached capabilities
+// else return false to refresh
+func checkRefreshInterval(capDir string, maxRefreshInterval time.Duration) (bool, error) {
+	timestamp := strconv.FormatInt(time.Now().Unix(), 10)
+	tmpDir := filepath.Join(capDir, ".tmp")
+	timeFilePath := filepath.Join(tmpDir, ".lastRefresh")
+	exist, _ := system.CreateIfNotExist(tmpDir)
```
zzxwill, line 227: Should we delete the dir later?

pkg/controller/core.oam.dev/v1alpha2/core/scopes/healthscope/healthscope.go
```diff
@@ -397,6 +397,7 @@ func (p PeerHealthConditions) MergePeerWorkloadsConditions(basic *WorkloadHealth
 	// copy to keep idempotent
 	peerHCs := make(PeerHealthConditions, len(p))
 	copy(peerHCs, p)
+	//nolint:makezero
```
zzxwill, line 400: 我懂了 makezero 的含义了，为了更明确，要不要换一个名字，比如 `slice-make-zero-length`?

pkg/commands/refresh_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"os"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestCheckRefreshInterval(t *testing.T) {
```
wonderflow, line 12: please don't use ginkgo for pure unit-test. It's hard to maintain. We prefer ginkgo only when the test need K8s environment.

We prefer table driven test for normal unit-test, you can refer to https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

pkg/commands/refresh.go
```diff
@@ -21,14 +29,27 @@ const (
 	deleted   refreshStatus = "Deleted"
 )
 
+const (
+	refreshInterval = time.Minute
```
wonderflow, line 33: please make the default time to be 5min

pkg/commands/refresh.go
```diff
@@ -10,6 +16,8 @@ import (
 	cmdutil "github.com/oam-dev/kubevela/pkg/commands/util"
 	"github.com/oam-dev/kubevela/pkg/plugins"
 	"github.com/oam-dev/kubevela/pkg/utils/system"
+
+	hashstructure "github.com/mitchellh/hashstructure/v2"
```
wonderflow, line 20: nit

pkg/commands/refresh.go
```diff
@@ -152,6 +194,13 @@ func refreshResultReport(newCaps, oldCaps []types.Capability) map[refreshStatus]
...
 			if oldCap.Name == newCap.Name {
 				found = true
+				if h, ok := cachedHash[newCap.Name]; ok {
+					// use cached hash to determine whether the cap is changed
+					if h == newHash[newCap.Name] {
+						report[unchanged] = append(report[unchanged], newCap)
+						break
+					}
+				}
 				if types.EqualCapability(oldCap, newCap) {
```
wonderflow, line 204: do we still need this when we have already use cachedHash ?

pkg/controller/core.oam.dev/v1alpha2/core/scopes/healthscope/healthscope.go
captainroy-hy: yes... it's an ambiguous name, but makezero is the name of the linter, and I have already commented it in [.golangci-lint.yml](https://github.com/oam-dev/kubevela/blob/e060a6ba463d322c50a8f8a2d476675b7a39e37b/.golangci.yml#L107)

pkg/commands/refresh.go
captainroy-hy: got it, shall we add a flag to enforce refresh, just for dev convenience?

pkg/commands/refresh_test.go
captainroy-hy: got it.

pkg/commands/refresh.go
captainroy-hy: Yes, I think we can remove it. It seems unlikely to delete the cache file by someone.

pkg/commands/refresh.go
wonderflow: > shall we add a flag to enforce refresh, just for dev convenience?

Sure, but users may not be aware of that. That could be very useful for development and demo.

pkg/commands/refresh.go
captainroy-hy: After reconsideration, I think it's okay to keep this part in case of cached hash is missing.

pkg/commands/refresh.go
captainroy-hy: Since it's in the dir `.vela/capabilities`, I think it's okay to be deleted along with `.vela/capabilities` if necessary.

wonderflow (approved):
lgtm

wonderflow:
please resolve conflicts

PR #745, opened 2020-12-09
make Vela work with ACK and clean up imports and scripts

ryanzhang-oss:
- minor changes to make vela work on ACK
- minor changes to clean up Go imports order  (we should enforce that during CR, it should always be no more than 3 sections with the GO lib the first and the home import the last)
- minor change to add a comment on the auto-generated definition files


pkg/commands/util/factory.go
```diff
@@ -23,7 +23,6 @@ var defaultCacheDir = filepath.Join(homedir.HomeDir(), ".kube", "http-cache")
 var _ genericclioptions.RESTClientGetter = &restConfigGetter{}
 
 // NewRestConfigGetter create config for helm client.
-// TODO(wonderflow): we should fix this hardcode client build,with restConfig as parameter.
```
wonderflow, line 26: why delete?

wonderflow (approved):

PR #765, opened 2020-12-10
Generate restful api based on Swagger

zzxwill:
As OpenAPI has to be quickly developped to support front-end developement, so
did the restufl api docs. Based on swagger, generate the docs automatically


pkg/server/util/api.go
```diff
@@ -13,7 +13,7 @@ import (
 var DefaultDashboardPort = ":38081"
 
 // DefaultAPIServerPort refers to the default port number of APIServer
-const DefaultAPIServerPort = ":8081"
+const DefaultAPIServerPort = ":38081"
```
hongchaodeng, line 16: Why change this?

If you want to change the port, also change the `dev`config in proxy.ts .

Please also add a comment here.

zzxwill: New version of Alilang takes the port at system start. Okay. changed the `dev` config. Thanks.

hongchaodeng (approved):
lgtm

Issue #749, opened 2020-12-09
[Feature] Support Definition Reference be Optional


wonderflow:
We're supporting appliction object #742, which could let a definition to be a patch.

![image](https://user-images.githubusercontent.com/2173670/101584206-16bc5700-3a18-11eb-8f8a-9ea22fb8db15.png)

This means, the definition will not have any CRD operator behind it, so we should support the definition reference to be optional.

In the picture, we're using "null", instead of that, we should leave the whole section empty.

```
apiVersion: core.oam.dev/v1alpha2
kind: TraitDefinition
metadata:
  annotations:
    definition.oam.dev/description: "add sidecar to the app"
  name: sidecar
spec:
  appliesToWorkloads:
    - webservice
    - worker
  extension:
    template: |-
      _containers: context.input.spec.template.spec.containers+[parameter]
      patch: {
         spec: template: spec: containers: _containers
      }
      parameter: {
         name: string
         image: string
         command?: [...string]
      }
```

wonderflow:
@leejanee  This issue must be fixed before #742 can be merge.

zzxwill:
https://github.com/oam-dev/kubevela/pull/742 is merged. Will it also apply to WorkloadDefinition?

captainroy-hy:
related WIP issue #907

zzxwill:
> related WIP issue #907

Got it. Besides TraitDefintion, will it also allpy to WorkloadDefinition?

wonderflow:
It's duplicated with #907  I'll close it and let's discuss more in that issue.

PR #768, opened 2020-12-11
Make the definitionRef field of traitDef optional

leejanee:
- [x] Support Definition Reference be Optional #749

Modified TraitDefinition CRD, and removed the definitionRef field from required list


leejanee:
The patch type traitDefinition does not need the definitionRef field, and only the application will use it.

ryanzhang-oss:
Is this PR still valid?

wonderflow:
closing as it's already in the code base.

Issue #661, opened 2020-11-25
[BUG] missing webhooks for AppConfig&Comp in vela-core chart


captainroy-hy:
Below 4 webhooks are not configurated in [vela-core chart](https://github.com/oam-dev/kubevela/blob/master/charts/vela-core/templates/webhook.yaml).
```yaml
kind: ValidatingWebhookConfiguration
webhooks:
  - name: "validate.applicationconfigurations.core.oam.dev"
...
  - name: "validate.component.core.oam.dev"
...
---
kind: MutatingWebhookConfiguration
  - name: "mutate.applicationconfigurations.core.oam.dev"
...
  - name: "mutate.component.core.oam.dev"
...
```

wonderflow:
fixed by #663

PR #774, opened 2020-12-15
implement conflictsWtih feature in trait definition

### LINKED ISSUE #553: `conflictsWtih` feature in trait definition.
This is important to maintain the rapid growing traits in KubeVela.

captainroy-hy:
to fix #553 

- add unit tests

### DIFF
--- a/pkg/webhook/core.oam.dev/v1alpha2/applicationconfiguration/validating_handler.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/applicationconfiguration/validating_handler.go
@@ -11,6 +11,7 @@ import (
 	"github.com/oam-dev/kubevela/pkg/oam/discoverymapper"
 
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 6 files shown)


pkg/webhook/core.oam.dev/v1alpha2/applicationconfiguration/validating_handler.go
```diff
@@ -210,6 +215,74 @@ func ValidateTraitAppliableToWorkloadFn(_ context.Context, v ValidatingAppConfig
...
+		}
+
+		for rulesOwner, rules := range allConflictRules {
+			if len(rules) == 0 {
+				// empty rules means this trait can work with any other ones
+				continue
+			}
+			noConflictWithAll := false
+			for _, rule := range rules {
+				if rule == "*" {
```
wonderflow, line 238: if we write a `*` as conflict rule, this might mean the trait is conflict to all other traits, right?

captainroy-hy: Yes, that makes sense. I originally treated "*" as an explicit declaration of empty rule set. Let me remove it.

wonderflow (approved):
Generally LGTM， ping @ryanzhang-oss  please review.

@captainroy-hy   It's fine that there's only unit-test. But I hope you have already tested in real case use our current demo.

captainroy-hy:
> @captainroy-hy It's fine that there's only unit-test. But I hope you have already tested in real case use our current demo.

Yes, I have already tested it by deploying the changed webhook into vela cluster manually.
But it seems AppConfig/Component webhooks has not yet enabled as mentioned in #661 ?

wonderflow:
@captainroy-hy  You could raise another PR to enable them by default.

hongchaodeng (approved):
lgtm

PR #776, opened 2020-12-16
support mutli storage resources,define storage interface and implemen…

Cweiping:
Fix Issue:https://github.com/oam-dev/kubevela/issues/751


Cweiping:
I already change the access to the application list and show. I look forward to your comments. 
After confirming the common interface, I will carry out the development of other storage and the unit test of the storage common interface.

wonderflow (commented):
Good Job!  I hope we can increase the unit test coverage and write test for each driver

pkg/storage/storage.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Storage is common storage client，use it to get app and others resource
+type Storage struct {
+	driver.Driver
+}
+
+// NewStorage form driver type
+func NewStorage(driverName string) *Storage {
+	// TODO remove driverName param ,should use environment get it
+	// FIXME use env to get user storageDriver
```
wonderflow, line 15: using driverName as param is good for me.

Cweiping: OK, why do you think DriverName is more friendly?

wonderflow: we can make it as system config using ways more than from ENV. I don't want we rely on the only way "from ENV"

Cweiping: Got it.

pkg/storage/storage.go
```diff
@@ -0,0 +1,45 @@
+package storage
```
LeoLiuYan, line 1: I think the `storage` package should in appfile dir which is `pkg/appfile` rather than in pkg dir since it service for appfile.

pkg/application/app.go
```diff
@@ -21,10 +15,16 @@ import (
...
-	"github.com/oam-dev/kubevela/pkg/utils/env"
-	"github.com/oam-dev/kubevela/pkg/utils/system"
+	"github.com/oam-dev/kubevela/pkg/storage"
+	"github.com/oam-dev/kubevela/pkg/storage/driver"
 )
 
+var store *storage.Storage
+
+func init() {
+	store = storage.NewStorage("")
```
LeoLiuYan, line 25: The `storage` variable initialization is not well placed in the `init()` function.

wonderflow: yes， please declare it explicitly

wonderflow (commented):
you could run `make reviewable` to check what is missing.

Cweiping:
> you could run `make reviewable` to check what is missing.

It can work locally.
 0.0

pkg/storage/storage.go
Cweiping: Done

pkg/application/app.go
Cweiping: How should I get the user-specified storage type?

wonderflow:
@Cweiping  It works and you should find something missing in the git commit. try `git status`to check

pkg/appfile/storage/driver/driver.go
```diff
@@ -0,0 +1,26 @@
...
+	// Delete application
+	Delete(envName, appName string) error
+	// Get application
+	Get(envName, appName string) (*RespApplication, error)
+	// Name of storage driver
+	Name() string
+}
+
+// RespApplication will get  application from Storage client
+type RespApplication struct {
```
wonderflow, line 23: this is exactly the same with `Application`, why don't use that?

wonderflow (commented):
Generally LGTM, ping @zzxwill @hongchaodeng  please review

pkg/application/app.go
wonderflow: We could provide multiple ways, use the default package variable is OK.

We could provide more ways like:

1. system environment 
2. system configfile
3. startup arguments

pkg/appfile/storage/driver/driver.go
Cweiping: Repeat import

wonderflow: Why not move `Application` out?

pkg/appfile/storage/driver/driver.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/oam-dev/kubevela/pkg/appfile"
+	"github.com/oam-dev/kubevela/pkg/appfile/template"
+)
+
+// Driver is mutli implement interface
+type Driver interface {
+	// List applications
+	List(envName string) ([]*RespApplication, error)
+	// Save application
+	Save(app *RespApplication, envName string) error
```
zzxwill, line 13: Does `Save` means creating an application?

pkg/appfile/storage/driver/local.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/oam-dev/kubevela/pkg/utils/env"
+	"github.com/oam-dev/kubevela/pkg/utils/system"
+)
+
+// LocalDriverName is local storage driver name
+const LocalDriverName = "Local"
+
+// Local Storage
+type Local struct {
+}
```
zzxwill, line 24: Are you going to add fields in struct `Local`, or just by-design?

pkg/appfile/storage/driver/local.go
```diff
@@ -0,0 +1,131 @@
...
+// Name  is local storage driver name
+func (l *Local) Name() string {
+	return LocalDriverName
+}
+
+// List applications from local storage
+func (l *Local) List(envName string) ([]*RespApplication, error) {
+	appDir, err := getApplicationDir(envName)
+	if err != nil {
+		return nil, fmt.Errorf("get app dir from env %s err %w", envName, err)
```
zzxwill, line 40:
```suggestion
		return nil, fmt.Errorf("Getting application directory from env %s failed, error: %w", envName, err)
```
Or a complete error message to help users to understand the issue, like `Application directory could not be retrieved from environment...`, `Getting application directory from environment xxx hit an issue:`

pkg/appfile/storage/storage.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/oam-dev/kubevela/pkg/appfile/storage/driver"
+)
+
+// Store application store client
+var Store *Storage
+
+func init() {
+	Store = NewStorage("")
```
zzxwill, line 11: Using the default driver name `Local` instead of ""?

pkg/appfile/storage/driver/driver.go
linfan: Better to avoid this duplicate type

pkg/appfile/storage/driver/driver.go
Cweiping: No, just save to app dir.
https://github.com/oam-dev/kubevela/blob/master/pkg/commands/up.go#L206

pkg/appfile/storage/driver/local.go
Cweiping: Each storage type will have a struct and implement `type Driver interface`.
The local storage type has no fields to add. To implement the general driver interface.
We just need to use the interface of `type Storage struct`.

pkg/appfile/storage/storage.go
Cweiping: Done

pkg/appfile/storage/driver/local.go
Cweiping: Done

pkg/appfile/storage/driver/driver.go
Cweiping: Because I don't know where it should move. Can you give me some advice? 0.0

https://github.com/oam-dev/kubevela/blob/master/pkg/application/app.go#L29

wonderflow: You can just put it here if you don't know

Cweiping: I try two ways to resolve it：
1. [move application/Application to storage/driver pkg ](https://github.com/oam-dev/kubevela/pull/776/commits/a4b52935221c6fb3088bc44f8baf8d39864b1fec)
2. [convert storage/driver.RespApplication to application/Application](https://github.com/Cweiping/kubevela/commit/de82405cb366456be13d0368512a1ffb5473332f)：
 ```Go
type Application struct {
	driver.RespApplication
}
```



Waiting for your comments.

File: pkg/serverlib/workload.go [hunk 9]
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/spf13/pflag"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/oam-dev/kubevela/pkg/appfile/storage/driver"
+
```
wonderflow, line 14: please format the import ，and please clean all same issue

wonderflow (approved):
Generally LGTM, thanks for your contribution！

pkg/serverlib/workload.go
Cweiping: Done

PR #811, opened 2020-12-24
fix unstable integration test

captainroy-hy:
With webhooks of AppConfig enabled when creating an AppConfig, it's necessary to make sure WorkloadDefintion/TraitDefintion/Component are created successfully, otherwise the validating-webhook would fail.

### DIFF
--- a/test/integration/appconfig_test.go
+++ b/test/integration/appconfig_test.go
@@ -159,7 +159,13 @@ func TestAppConfigController(t *testing.T) {
 						},
 					}))
 
-				if err := c.Create(context.Background(), ac); err != nil {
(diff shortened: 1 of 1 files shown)


test/integration/appconfig_test.go
```diff
@@ -159,7 +159,13 @@ func TestAppConfigController(t *testing.T) {
 						},
 					}))
 
-				if err := c.Create(context.Background(), ac); err != nil {
+				if err := waitFor(context.Background(), 3*time.Second, func() (bool, error) {
```
zzxwill, line 162: Just curious whether 3 seconds is enough?

zzxwill (approved):
LGTM

wonderflow (approved):

PR #815, opened 2020-12-25
add code lint check and fix code style

majian159:


dashboard/src/pages/System/Environment/index.tsx
```diff
@@ -1,5 +1,3 @@
-import './index.less';
```
zzxwill, line 1: 多余的？

majian159: yeah, index.less is empty

captainroy-hy:
@majian159 failed e2e test will be fixed in #811 , pls rebase after it's merged.

zzxwill (approved):
LGTM

PR #824, opened 2020-12-25
Add "description" field for each workload from ListWorkloads API

zzxwill:
Workload item doesn't have descriptin property, added it to help display
the description in capability list on Dashboard

Fix #821

### DIFF
--- a/pkg/server/apis/types.go
+++ b/pkg/server/apis/types.go
@@ -47,9 +47,9 @@ type WorkloadRunBody struct {
 
 // WorkloadMeta store workload metadata for dashboard restful API server
 type WorkloadMeta struct {
-	Name       string            `json:"name"`
-	Parameters []types.Parameter `json:"parameters,omitempty"`
-	AppliesTo  []string          `json:"appliesTo,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/server/apis/types.go
```diff
@@ -47,9 +47,9 @@ type WorkloadRunBody struct {
 
 // WorkloadMeta store workload metadata for dashboard restful API server
 type WorkloadMeta struct {
-	Name       string            `json:"name"`
-	Parameters []types.Parameter `json:"parameters,omitempty"`
-	AppliesTo  []string          `json:"appliesTo,omitempty"`
```
wonderflow, line 52: why delete appliesTo?

wonderflow: BTW, it's better to have a unit test for it. Now, even we delete some field(which should be a break change) that the CI won't fail.

zzxwill: Workload should not have `AppliesTo` field. And I have confirmed all callers of this struct and it's only called by `pkg/server/workloadHandler.go`.

wonderflow (approved):

PR #838, opened 2020-12-29
implement ApplyOnceOnlyForce

captainroy-hy:
- add unit test
- update doc

### DIFF
--- a/apis/core.oam.dev/v1alpha2/core_types.go
+++ b/apis/core.oam.dev/v1alpha2/core_types.go
@@ -439,6 +439,13 @@ type WorkloadStatus struct {
 	// ComponentRevisionName of current component
 	ComponentRevisionName string `json:"componentRevisionName,omitempty"`
 
+	// ObservedGeneration indicates the generation observed by the appconfig controller.
+	// The same field is also recorded in the annotations of workloads.
+	// A workload is possible to be deleted from cluster after created.
+	// This field is useful to track the observed generation of workloads after they are
+	// deleted.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
 	// Reference to a workload created by an ApplicationConfiguration.
 	Reference runtimev1alpha1.TypedReference `json:"workloadRef,omitempty"`
 
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -4,6 +4,8 @@
 
 replicaCount: 1
 installCertManager: false
+# Valid applyOnceOnly values: true/false/on/off/force
+applyOnceOnly: "off"
 useWebhook: true
 # By default, don't disable any builtin capabilities
 disableCaps: ""
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -9,6 +9,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
@@ -79,6 +80,7 @@ func main() {
 	var disableCaps string
 	var storageDriver string
 	var syncPeriod time.Duration
+	var applyOnceOnly string
 
 	flag.BoolVar(&useWebhook, "use-webhook", false, "Enable Admission Webhook")
 	flag.BoolVar(&useTraitInjector, "use-trait-injector", false, "Enable TraitInjector")
@@ -95,8 +97,8 @@ func main() {
 	flag.IntVar(&controllerArgs.RevisionLimit, "revision-limit", 50,
 		"RevisionLimit is the maximum number of revisions that will be maintained. The default value is 50.")
 	flag.StringVar(&healthAddr, "health-addr", ":9440", "The address the health endpoint binds to.")
-	flag.BoolVar(&controllerArgs.ApplyOnceOnly, "apply-once-only", false,
-		"For the purpose of some production environment that workload or trait should not be affected if no spec change")
+	flag.StringVar(&applyOnceOnly, "apply-once-only", "false",
+		"For the purpose of some production environment that workload or trait should not be affected if no spec change, available options: on, off, force.")
 	flag.StringVar(&controllerArgs.CustomRevisionHookURL, "custom-revision-hook-url", "",
 		"custom-revision-hook-url is a webhook url which will let KubeVela core to call with applicationConfiguration and component info and return a customized component revision")
 	flag.StringVar(&disableCaps, "disable-caps", "", "To be disabled builtin capability list.")
@@ -164,6 +166,23 @@ func main() {
 		}
 	}
 
+	switch strings.ToLower(applyOnceOnly) {
+	case "", "false", string(oamcontroller.ApplyOnceOnlyOff):
+		controllerArgs.ApplyMode = oamcontroller.ApplyOnceOnlyOff
+		setupLog.Info("ApplyOnceOnly is disabled")
+	case "true", string(oamcontroller.ApplyOnceOnlyOn):
+		controllerArgs.ApplyMode = oamcontroller.ApplyOnceOnlyOn
+		setupLog.Info("ApplyOnceOnly is enabled, that means workload or trait only apply once if no spec change even they are changed by others")
+	case string(oamcontroller.ApplyOnceOnlyForce):
+		controllerArgs.ApplyMode = oamcontroller.ApplyOnceOnlyForce
+		setupLog.Info("ApplyOnceOnlyForce is enabled, that means workload or trait only apply once if no spec change even they are changed or deleted by others")
+	default:
+		setupLog.Error(fmt.Errorf("invalid apply-once-only value: %s", applyOnceOnly),
+			"unable to setup the vela core controller",
+			"valid apply-once-only value:", "on/off/force, by default it's off")
+		os.Exit(1)
+	}
+
 	if err = oamv1alpha2.Setup(mgr, controllerArgs, logging.NewLogrLogger(setupLog)); err != nil {
 		setupLog.Error(err, "unable to setup the oam core controller")
 		os.Exit(1)
@@ -203,9 +222,6 @@ func main() {
 
 	setupLog.Info("starting the vela controller manager")
 
-	if controllerArgs.ApplyOnceOnly {
-		setupLog.Info("applyOnceOnly is enabled that means workload or trait only apply once if no spec change even they are changed by others")
-	}
 	if err := mgr.Start(makeSignalHandler()); err != nil {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
--- a/design/vela-core/apply-once-only.md
+++ b/design/vela-core/apply-once-only.md
@@ -127,3 +127,28 @@ Since discrepancy is found, vela-core controller will apply(update) the Deployme
 Thus, the changes we made to the Deployment before will also be eliminated.
 
 The same mechanism also works for Trait as well as Workload.
+
+### Apply Once Only Force
+
+Based on the same mechanism as `apply-once-only`, `apply-once-only-force` allows to skip re-creating a workload or trait that has already been DELETED from the cluster if its spec is not changed. 
+It's regarded as a stronger case of `apply-once-only`.
+
+## Usage
+
+Three available options are provided to a vela-core runtime setup flag named `apply-one-only`, referring to three modes:
+
+- off - `apply-once-only` is disabeld, this is the default option
+- on - `apply-once-only` is enabled
+- force - `apply-once-only-force` is enabled
+
+You can set it through `helm` chart value `applyOnceOnly` which is "off" by default if omitted, for example
+
+```shell
+helm install -n vela-system kubevela ./charts/vela-core --set applyOnceOnly=on
+```
+or
+```
+helm install -n vela-system kubevela ./charts/vela-core --set applyOnceOnly=force
+```
+
+
--- a/pkg/controller/core.oam.dev/oamruntime_controller.go
+++ b/pkg/controller/core.oam.dev/oamruntime_controller.go
@@ -16,15 +16,33 @@ limitations under the License.
 
 package core_oam_dev
 
+// ApplyOnceOnlyMode enumerates ApplyOnceOnly modes.
+type ApplyOnceOnlyMode string
+
+const (
+	// ApplyOnceOnlyOff indicates workloads and traits should always be affected.
+	// It means ApplyOnceOnly is disabled.
+	ApplyOnceOnlyOff ApplyOnceOnlyMode = "off"
+
+	// ApplyOnceOnlyOn indicates workloads and traits should not be affected
+	// if no spec change is made in the ApplicationConfiguration.
+	ApplyOnceOnlyOn = "on"
+
+	// ApplyOnceOnlyForce is a more strong case for ApplyOnceOnly, the workload
+	// and traits won't be affected if no spec change is made in the ApplicationConfiguration,
+	// even if the workload or trait has been deleted from cluster.
+	ApplyOnceOnlyForce = "force"
+)
+
 // Args args used by controller
 type Args struct {
 	// RevisionLimit is the maximum number of revisions that will be maintained.
 	// The default value is 50.
 	RevisionLimit int
 
-	// ApplyOnceOnly indicates whether workloads and traits should be
+	// ApplyMode indicates whether workloads and traits should be
 	// affected if no spec change is made in the ApplicationConfiguration.
-	ApplyOnceOnly bool
+	ApplyMode ApplyOnceOnlyMode
 
 	// CustomRevisionHookURL is a webhook which will let oam-runtime to call with AC+Component info
 	// The webhook server will return a customized component revision for oam-runtime
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
@@ -102,22 +102,22 @@ func Setup(mgr ctrl.Manager, args core.Args, l logging.Logger) error {
 		Complete(NewReconciler(mgr, dm,
 			WithLogger(l.WithValues("controller", name)),
 			WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
-			WithApplyOnceOnly(args.ApplyOnceOnly)))
+			WithApplyOnceOnlyMode(args.ApplyMode)))
 }
 
 // An OAMApplicationReconciler reconciles OAM ApplicationConfigurations by rendering and
 // instantiating their Components and Traits.
 type OAMApplicationReconciler struct {
-	client        client.Client
-	components    ComponentRenderer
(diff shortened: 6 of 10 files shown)


pkg/controller/core.oam.dev/oamruntime_controller.go
```diff
@@ -25,4 +25,9 @@ type Args struct {
 	// ApplyOnceOnly indicates whether workloads and traits should be
 	// affected if no spec change is made in the ApplicationConfiguration.
 	ApplyOnceOnly bool
+
+	// ApplyOnceOnlyForce indicates whether workloads should be
+	// affected if no spec change is made in the ApplicationConfiguration
+	// even if the workload has been deleted.
```
wonderflow, line 31:
```suggestion
	// ApplyOnceOnlyForce is more strong case for applyOnceOnly, the workload won't be affected(re-created) even if the workload has been deleted.
```

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -101,22 +101,24 @@ func Setup(mgr ctrl.Manager, args core.Args, l logging.Logger) error {
...
+	components         ComponentRenderer
+	workloads          WorkloadApplicator
+	gc                 GarbageCollector
+	scheme             *runtime.Scheme
+	log                logging.Logger
+	record             event.Recorder
+	preHooks           map[string]ControllerHooks
+	postHooks          map[string]ControllerHooks
+	applyOnceOnly      bool
+	applyOnceOnlyForce bool
```
wonderflow, line 121:
```
type ApplyOnceOnly struct{
  Force bool
}
```

```suggestion
	applyOnceOnly      *ApplyOnceOnly
```

If the applyOnceOnly is nil, no apply-once-only enabled. If it's not nil, check if force flag is enabled.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -298,17 +309,25 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
 	r.record.Event(ac, event.Normal(reasonRenderComponents, "Successfully rendered components", "workloads", strconv.Itoa(len(workloads))))
 
 	applyOpts := []resource.ApplyOption{resource.MustBeControllableBy(ac.GetUID())}
-	if r.applyOnceOnly {
+	applyingWorkloads := make([]Workload, len(workloads))
+	copy(applyingWorkloads, workloads)
+	if r.applyOnceOnlyForce {
+		applyingWorkloads = filterAppliedWorkloads(ac, applyingWorkloads)
```
wonderflow, line 315: I hope we can still use applyOpts for this? Can't we?

wonderflow: If you filter out, the traits of these workload will not work.

wonderflow (changes requested):
After talking with @captainroy-hy  that crossplane-runtime can't work if workload not exist https://github.com/crossplane/crossplane-runtime/blob/479dbc8a456086622ce1afab66aad11ba6c8f78d/pkg/resource/api.go#L138

So we should need to refactor that part first before this PR.

pkg/controller/core.oam.dev/oamruntime_controller.go
ryanzhang-oss: Do we really need two so similar flags?

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -298,17 +309,25 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
 	r.record.Event(ac, event.Normal(reasonRenderComponents, "Successfully rendered components", "workloads", strconv.Itoa(len(workloads))))
 
 	applyOpts := []resource.ApplyOption{resource.MustBeControllableBy(ac.GetUID())}
-	if r.applyOnceOnly {
+	applyingWorkloads := make([]Workload, len(workloads))
+	copy(applyingWorkloads, workloads)
+	if r.applyOnceOnlyForce {
```
ryanzhang-oss, line 314: the apply function takes the option, we can do the filter there

captainroy-hy: Yes, I will refine this PR after #848

cmd/core/main.go
```diff
@@ -93,8 +94,18 @@ func main() {
...
 		"For the purpose of some production environment that workload or trait should not be affected if no spec change")
+	flag.BoolVar(&applyOnceOnlyForce, "apply-once-only-force", false,
+		"Stronger case for applyOnceOnly, it will override the value of flag 'ApplyOnceOnly' ")
+	controllerArgs.ApplyMode = oamcontroller.ApplyOnceOnlyOff
+	if applyOnceOnlyForce {
+		controllerArgs.ApplyMode = oamcontroller.ApplyOnceOnlyForce
+	} else if applyOnceOnly {
+		controllerArgs.ApplyMode = oamcontroller.ApplyOnceOnlyOn
+	}
+
```
captainroy-hy, line 108: To keep backward compatible, apply-once-only flag is retained and a new but similar flag apply-once-only-force is added. If it's okay to remove apply-once-only boolean flag, I would like to just use one string flag(off/on/force) instead of two.

wonderflow: yes, use `off/on/force` and update the design doc.

wonderflow: you can also make the compatibility:

apply-once-only is a string

`apply-once-only="true" or "on"  ` means enable it.
`apply-once-only="" or "false" or "off"  ` means disable it.
`apply-once-only="force"  means enable force apply only once.

In other case, please report an error on start.

captainroy-hy: Got it. Good idead!

wonderflow: yes, true/false in yaml can be easily convert to string

design/vela-core/apply-once-only.md
```diff
@@ -127,3 +127,7 @@ Since discrepancy is found, vela-core controller will apply(update) the Deployme
 Thus, the changes we made to the Deployment before will also be eliminated.
 
 The same mechanism also works for Trait as well as Workload.
+
+## Apply Once Only Force
+
+Based on the same mechanism as `apply-once-only`, `apply-once-only-force` allows to skip creating a workload that has already been DELETED from the cluster if its spec is not changed. `apply-once-only-force` is disabled by default.
```
wonderflow, line 133: please describe how to use it here.

apis/core.oam.dev/v1alpha2/core_types.go
```diff
@@ -439,6 +439,9 @@ type WorkloadStatus struct {
 	// ComponentRevisionName of current component
 	ComponentRevisionName string `json:"componentRevisionName,omitempty"`
 
+	// ObservedGeneration indicates the generation observed by the appconfig controller.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
wonderflow, line 443: why do we need to add now but not used in apply-once-only before?

pkg/controller/core.oam.dev/oamruntime_controller.go
```diff
@@ -16,15 +16,32 @@ limitations under the License.
...
+	// It means ApplyOnceOnly is disabled.
+	ApplyOnceOnlyOff ApplyOnceOnlyMode = "off"
+
+	// ApplyOnceOnlyOn indicates workloads and traits should not be affected
+	// if no spec change is made in the ApplicationConfiguration.
+	ApplyOnceOnlyOn = "on"
+
+	// ApplyOnceOnlyForce is a more strong case for ApplyOnceOnly, the workload
+	// and traits won't be affected(re-created) even if the workload has been deleted.
+	ApplyOnceOnlyForce = "force"
```
wonderflow, line 33: please descirbe these usage(on/off/force) into the design doc

cmd/core/main.go
```diff
@@ -164,6 +166,23 @@ func main() {
...
+	case "true", string(oamcontroller.ApplyOnceOnlyOn):
+		controllerArgs.ApplyMode = oamcontroller.ApplyOnceOnlyOn
+		setupLog.Info("ApplyOnceOnly is enabled, that means workload or trait only apply once if no spec change even they are changed by others")
+	case string(oamcontroller.ApplyOnceOnlyForce):
+		controllerArgs.ApplyMode = oamcontroller.ApplyOnceOnlyForce
+		setupLog.Info("ApplyOnceOnlyForce is enabled, that means workload or trait only apply once if no spec change even they are changed or deleted by others")
+	default:
+		setupLog.Error(fmt.Errorf("invalid apply-once-only value: %s", applyOnceOnly),
+			"unable to setup the vela core controller",
+			"valid apply-once-only value:", "true/false/on/off/force ")
```
wonderflow, line 182:
```suggestion
			"valid apply-once-only value:", "on/off/force ")
```

true/false is for compatibility, don't need to tell

cmd/core/main.go
```diff
@@ -95,7 +97,7 @@ func main() {
 	flag.IntVar(&controllerArgs.RevisionLimit, "revision-limit", 50,
 		"RevisionLimit is the maximum number of revisions that will be maintained. The default value is 50.")
 	flag.StringVar(&healthAddr, "health-addr", ":9440", "The address the health endpoint binds to.")
-	flag.BoolVar(&controllerArgs.ApplyOnceOnly, "apply-once-only", false,
+	flag.StringVar(&applyOnceOnly, "apply-once-only", "false",
 		"For the purpose of some production environment that workload or trait should not be affected if no spec change")
```
wonderflow, line 101:
```suggestion
		"For the purpose of some production environment that workload or trait should not be affected if no spec change, available options: on, off, force.")
```

charts/vela-core/values.yaml
```diff
@@ -4,6 +4,8 @@
 
 replicaCount: 1
 installCertManager: false
+# Valid applyOnceOnly value: true/false/on/off/force
+applyOnceOnly: false
```
wonderflow, line 8:
```suggestion
applyOnceOnly: off
```

charts/vela-core/values.yaml
```diff
@@ -4,6 +4,8 @@
 
 replicaCount: 1
 installCertManager: false
+# Valid applyOnceOnly value: true/false/on/off/force
```
wonderflow, line 7:
```suggestion
# Valid applyOnceOnly values: true/false/on/off/force
```

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -576,36 +574,80 @@ func (e *GenerationUnchanged) Error() string {
 		"Please ignore this error in other logic.")
 }
 
-func applyOnceOnly() apply.ApplyOption {
-	return func(ctx context.Context, current, desired runtime.Object) error {
-		if current == nil {
+// applyOnceOnly is an ApplyOption that controls the applying mechanism for workload and trait.
+// More detail refers to ApplyOnceOnlyMode
```
wonderflow, line 578:
```suggestion
// More detail refers to the ApplyOnceOnlyMode type annotation
```

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -576,36 +574,80 @@ func (e *GenerationUnchanged) Error() string {
...
-		if dLabels[oam.LabelOAMResourceType] == oam.ResourceTypeWorkload ||
-			dLabels[oam.LabelOAMResourceType] == oam.ResourceTypeTrait {
-			// check whether spec changes occur on the workload or trait,
-			// according to annotations and lables
-			if c.GetAnnotations()[oam.AnnotationAppGeneration] !=
-				d.GetAnnotations()[oam.AnnotationAppGeneration] {
+		dLabels := d.GetLabels()
+		if dLabels[oam.LabelOAMResourceType] != oam.ResourceTypeWorkload &&
+			dLabels[oam.LabelOAMResourceType] != oam.ResourceTypeTrait {
+			// if target resource is not a workload nor trait
```
wonderflow, line 593: You should give some example here, in which case will a resource applied that is not workload and trait

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -576,36 +574,80 @@ func (e *GenerationUnchanged) Error() string {
...
+			// skip this option
+			return nil
+		}
+
+		// the resource doesn't exist (maybe not created before, or created but deleted by others)
+		if existing == nil {
+			if mode != core.ApplyOnceOnlyForce {
+				// under non-force mode, always create the resource if not exist.
```
wonderflow, line 601:
```suggestion
				// non-force mode will always create the resource if not exist.
```

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -576,36 +574,80 @@ func (e *GenerationUnchanged) Error() string {
...
+				if !createdBefore {
+					for _, t := range w.Traits {
+						if t.Reference.GetObjectKind().GroupVersionKind() == desired.GetObjectKind().GroupVersionKind() &&
+							t.Reference.Name == d.GetName() {
+							// a recorded trait matched target resource
+							createdBefore = true
+						}
+					}
+				}
+				if createdBefore {
```
wonderflow, line 621: use if else in this case

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -576,36 +574,80 @@ func (e *GenerationUnchanged) Error() string {
...
+					// appconfig status recorded a matched reference, that means the resource is created before
```
wonderflow, line 622:
```suggestion
					// appconfig status recorded the resource version applied, only obeservedGeneration and componentRevisionName are both equal can mean the resource is created before
```

apis/core.oam.dev/v1alpha2/core_types.go
captainroy-hy: apply-once-only determines a generation change according to information saved in labels and annotation. But in apply-once-apply-force, existing obj may be nil, we cannot get information to compare, so we need observedGeneration saved in ac.status.workloads.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
captainroy-hy: Line 617 may change `createdBefore` to true, "if else" will miss that.

wonderflow: I'm wrong, it can't be else

wonderflow: please add an annotation cause it's confusing.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -576,36 +574,80 @@ func (e *GenerationUnchanged) Error() string {
...
+
+		// the resource already exists
+		e, _ := existing.(metav1.Object)
+		if e == nil {
+			return errors.Errorf("cannot access metadata of existing object: %q",
+				existing.GetObjectKind().GroupVersionKind())
+		}
+		eLabels := e.GetLabels()
+		if (e.GetAnnotations()[oam.AnnotationAppGeneration] != d.GetAnnotations()[oam.AnnotationAppGeneration]) ||
+			(eLabels[oam.LabelAppComponentRevision] != dLabels[oam.LabelAppComponentRevision]) {
```
wonderflow, line 645: add an annotation here like L622

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/suite_test.go
```diff
@@ -150,7 +151,9 @@ var _ = BeforeSuite(func(done Done) {
 	}, time.Second*30, time.Millisecond*500).Should(BeNil())
 	Expect(mapping.Resource.Resource).Should(Equal("foo"))
 
-	reconciler = NewReconciler(mgr, dm, WithLogger(logging.NewLogrLogger(ctrl.Log.WithName("suit-test-appconfig"))))
+	reconciler = NewReconciler(mgr, dm,
+		WithLogger(logging.NewLogrLogger(ctrl.Log.WithName("suit-test-appconfig"))),
+		WithApplyOnceOnlyMode(core_oam_dev.ApplyOnceOnlyOff))
```
wonderflow, line 156: make it as default in NewReconciler, so you don't need to add this line

wonderflow (commented):
Good Job! LGTM after nits

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/suite_test.go
captainroy-hy: got it

apis/core.oam.dev/v1alpha2/core_types.go
wonderflow: good , why not explain it in the annotation

captainroy-hy: sure, I will add more explanation in detail.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
captainroy-hy: I will remove this logic because currently, we use this function only for workloads and traits

wonderflow: Maybe scope will use it？

captainroy-hy: Scope still uses raw client Update but not 3-way-diff Apply.

wonderflow: yes, maybe we will use it in the future, so I think we should keep the code but add an annotation to tell why( even the reason is some kinds of possiblilty)

captainroy-hy: Got it.

wonderflow (approved):
LGTM

PR #844, opened 2020-12-30
Fix `each child in a list should have a unique "key" prop.'

sunny0826:
Fix the problem discussed in https://github.com/oam-dev/kubevela/pull/834#discussion_r549551292

### DIFF
--- a/dashboard/src/pages/Capability/Components/ShowComponent/index.tsx
+++ b/dashboard/src/pages/Capability/Components/ShowComponent/index.tsx
@@ -73,7 +73,7 @@ export default ({ name, parameters }: ShowParameters) => {
             responsive: ['md'],
             render: (text, row) => [
               <Space size="small" align="center">
-                {!row.default ? undefined : <Text code>{text}</Text>}
+                {!row.default ? undefined : <Text code>{text.toString()}</Text>}
               </Space>,
             ],
           },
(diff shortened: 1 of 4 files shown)


zzxwill (approved):
LGTM

dashboard/src/pages/Capability/Components/ShowComponent/index.tsx
```diff
@@ -73,7 +73,7 @@ export default ({ name, parameters }: ShowParameters) => {
             responsive: ['md'],
             render: (text, row) => [
```
majian159, line 74: Should not return an array, should return a element

dashboard/src/pages/Capability/Components/ShowComponent/index.tsx
```diff
@@ -73,7 +73,7 @@ export default ({ name, parameters }: ShowParameters) => {
...
               <Space size="small" align="center">
-                {!row.default ? undefined : <Text code>{text}</Text>}
+                {!row.default ? undefined : <Text code>{text.toString()}</Text>}
               </Space>,
             ],
```
majian159, line 78: Should not return an array, should return a element

majian159 (changes requested):
`Table` `render` Should not return an array, should return a element.

dashboard/src/pages/Capability/Components/ShowComponent/index.tsx
sunny0826: fixed

sunny0826 (commented):
fixed

hongchaodeng (commented):
lgtm

PR #850, opened 2020-12-31
support env and config storage driver

Cweiping:
Issue: https://github.com/oam-dev/kubevela/issues/751

### DIFF
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -33,11 +33,13 @@ import (
 
 	oamcore "github.com/oam-dev/kubevela/apis/core.oam.dev"
 	velacore "github.com/oam-dev/kubevela/apis/standard.oam.dev/v1alpha1"
+	"github.com/oam-dev/kubevela/pkg/appfile/storage/driver"
 	velacontroller "github.com/oam-dev/kubevela/pkg/controller"
 	oamcontroller "github.com/oam-dev/kubevela/pkg/controller/core.oam.dev"
 	oamv1alpha2 "github.com/oam-dev/kubevela/pkg/controller/core.oam.dev/v1alpha2"
 	"github.com/oam-dev/kubevela/pkg/controller/dependency"
 	"github.com/oam-dev/kubevela/pkg/controller/utils"
+	"github.com/oam-dev/kubevela/pkg/utils/system"
 	oamwebhook "github.com/oam-dev/kubevela/pkg/webhook/core.oam.dev"
 	velawebhook "github.com/oam-dev/kubevela/pkg/webhook/standard.oam.dev"
 	"github.com/oam-dev/kubevela/version"
@@ -76,6 +78,7 @@ func main() {
 	var controllerArgs oamcontroller.Args
 	var healthAddr string
 	var disableCaps string
+	var storageDriver string
 
 	flag.BoolVar(&useWebhook, "use-webhook", false, "Enable Admission Webhook")
 	flag.BoolVar(&useTraitInjector, "use-trait-injector", false, "Enable TraitInjector")
@@ -94,7 +97,10 @@ func main() {
 	flag.StringVar(&healthAddr, "health-addr", ":9440", "The address the health endpoint binds to.")
 	flag.BoolVar(&controllerArgs.ApplyOnceOnly, "apply-once-only", false,
 		"For the purpose of some production environment that workload or trait should not be affected if no spec change")
+	flag.StringVar(&controllerArgs.CustomRevisionHookURL, "custom-revision-hook-url", "",
+		"custom-revision-hook-url is a webhook url which will let KubeVela core to call with applicationConfiguration and component info and return a customized component revision")
 	flag.StringVar(&disableCaps, "disable-caps", "", "To be disabled builtin capability list.")
+	flag.StringVar(&storageDriver, "storage-driver", driver.LocalDriverName, "Application file save to the storage driver")
 	flag.Parse()
 
 	// setup logging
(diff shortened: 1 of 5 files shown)


wonderflow (changes requested):
I don't think the driver can be an option defined in `vela env`. It should be a system level option. That once kubevela deployed somewhere, the driver is specified and can't be changed.

Cweiping:
> I don't think the driver can be an option defined in `vela env`. It should be a system level option( implemented by SYSTEM ENV or Start args options). That once kubevela deployed somewhere, the driver is specified and can't be changed.

Please review again.

pkg/appfile/storage/storage.go
```diff
@@ -1,37 +1,44 @@
...
+	_ = NewStorage(driver.LocalDriverName)
+}
+
 // NewStorage form driver type
-func NewStorage(driverName string) *Storage {
-	// TODO remove driverName param ,should use environment get it
-	// FIXME use env to get user storageDriver
+// 1. get system environment from STORAGEDRIVER
+// 2. startup arguments from storage-driver
+func NewStorage(driverName string) error {
```
ryanzhang-oss, line 26: Why not return the Storage just like the original function? Assigning a global variable in the function can lead to overriding the variable unintentionally if the NewStorage function is called multiple times.

pkg/appfile/storage/storage.go
```diff
@@ -1,37 +1,44 @@
...
-}
-
 // Storage is common storage client，use it to get app and others resource
 type Storage struct {
 	driver.Driver
 }
 
+func init() {
+	// default use Local Storage driver
+	_ = NewStorage(driver.LocalDriverName)
```
ryanzhang-oss, line 20: why ignore the error? This might not work in the controller as an init() function.

pkg/appfile/storage/storage.go
```diff
@@ -1,37 +1,44 @@
...
+
 // NewStorage form driver type
-func NewStorage(driverName string) *Storage {
-	// TODO remove driverName param ,should use environment get it
-	// FIXME use env to get user storageDriver
+// 1. get system environment from STORAGEDRIVER
+// 2. startup arguments from storage-driver
+func NewStorage(driverName string) error {
+	// first use system environment
+	if storageDriver := os.Getenv("STORAGEDRIVER"); len(storageDriver) != 0 {
```
ryanzhang-oss, line 28: We better use one place to consolidate the ENV and parameters. This will silently override the user input in the flag.

cmd/core/main.go
```diff
@@ -172,6 +175,10 @@ func main() {
 		setupLog.Error(err, "unable to setup the vela core controller")
 		os.Exit(1)
 	}
+	if err = storage.NewStorage(storageDriver); err != nil {
```
ryanzhang-oss, line 178: How does local drive work in a controller? On the flip side, where is app/appFile used on the server side?

ryanzhang-oss (changes requested):

pkg/appfile/storage/storage.go
Cweiping: We need to support the storage mode specified by the `--storage-driver` flag, so we can't get parameters directly in init.

It would be nice if `NewStorage ` could be made private.

cmd/core/main.go
Cweiping: 1. We can start the environment variable configuration in two ways(Default: Local):
   a) first use system environment
   b) startup arguments from storage-driver
2.  If local storage is used, it will still be stored under ` .vala`. It should also support customization and split with CLI.
3. It is preferred to use [it](https://github.com/oam-dev/kubevela/blob/master/pkg/application/app.go) to complete the application action, and it is not recommended to use `appfile/storage` pkg.

pkg/appfile/storage/storage.go
Cweiping: OK, I move it to `/cmd/core/main.go`.

cmd/core/main.go
```diff
@@ -95,6 +96,7 @@ func main() {
 	flag.BoolVar(&controllerArgs.ApplyOnceOnly, "apply-once-only", false,
 		"For the purpose of some production environment that workload or trait should not be affected if no spec change")
 	flag.StringVar(&disableCaps, "disable-caps", "", "To be disabled builtin capability list.")
+	flag.StringVar(&storageDriver, "storage-driver", "Local", "Application file save to the storage driver")
```
wonderflow, line 99: please use the const variable `driver.LocalDriverName` here instead of `Local`

cmd/core/main.go
```diff
@@ -172,6 +174,15 @@ func main() {
 		setupLog.Error(err, "unable to setup the vela core controller")
 		os.Exit(1)
 	}
+	if driver := os.Getenv("STORAGEDRIVER"); len(driver) == 0 {
```
wonderflow, line 177: please use a const variable for `STORAGEDRIVER `

wonderflow (approved):
generally LGTM, ping @ryanzhang-oss

cmd/core/main.go
Cweiping: Done!

cmd/core/main.go
Cweiping: Done

wonderflow:
@Cweiping  please fix conflicts, thanks!

ryanzhang-oss (approved):

ryanzhang-oss:
LGTM, please resolve the conflict

Cweiping:
@wonderflow @ryanzhang-oss  
Done.

PR #853, opened 2020-12-31
add test for component revision hook

wonderflow:


cmd/core/main.go
```diff
@@ -94,6 +94,8 @@ func main() {
 	flag.StringVar(&healthAddr, "health-addr", ":9440", "The address the health endpoint binds to.")
 	flag.BoolVar(&controllerArgs.ApplyOnceOnly, "apply-once-only", false,
 		"For the purpose of some production environment that workload or trait should not be affected if no spec change")
+	flag.StringVar(&controllerArgs.CustomRevisionHookURL, "custom-revision-hook-url", "",
+		"custom-revision-hook-url is a webhook url which will let oam-runtime to call with AC+Component info and return a customized component revision")
```
zzxwill, line 98: How about we naming `oam-runtime` to `KubeVela core`, `AC+Component` to `ApplicationConfiguration and Component`?

wonderflow: yes, good catch, thanks

wonderflow, line 98:
```suggestion
		"custom-revision-hook-url is a webhook url which will let KubeVela core to call with applicationConfiguration and component info and return a customized component revision")
```

Diddaa (approved):

Issue #849, opened 2020-12-30
[Feature] IngressClass  support for route


Incubator4th:
Environment: RKE 1.19.2 with traefik v2.2 as ingress controller

Well,in the process of referring to the installation guide, some problems were encountered.

Which is the same issue with #632 

After a series of problem diagnosis,I found that the route trait create ingress with annotations `kubernetes.io/ingress.class=nginx`,but I used traefik, this is why this service is pending and route got timemout.

Later, after researching, I understood how resources are called by chain.

`Vela.yml route ->  route of TraitDefinition.core.oam.dev/v1alpha2 ->  Route.standard.oam.dev/v1alpha1`

Well I know that I can define my own parameter using TraitDefinition for another ingress or others.But problem is in offical route traitdefinition,it should support ingressclass to set which ingress controller shound use.

Such as:

```yaml
services:
  web:
    route:
      ingressclass: nginx
```

and then auto add annotation `kubernetes.io/ingress.class`

wonderflow:
Thanks for your advice. I totally agree the `ingressclass` could be an option(with default value 'nginx'), would you like to send a PR for it? @Incubator4th

Incubator4th:
> Thanks for your advice. I totally agree the `ingressclass` could be an option(with default value 'nginx'), would you like to send a PR for it? @Incubator4th

Thanks for appreciation.It's interest to modify by myself. 

Before doing that,in my understanding, add parameter ingressclass requires modify crd define yaml , main controller and trais crd value. It would has a chain change.And I would like to try it.

By the way,on the one hand this project kubevela aims to let  developers to understand kubernetes less.
But on the other hand, inevitably,some services need tcp or udp port,and what follows is the need to define service type like clusterIP.

I'm not sure if this is against the purpose of the project,cause of just like docker-compose have port mapping.

Can ask about the development route in this regard ?

wonderflow:
Hi, @Incubator4th  really good thoughts! Thanks.

> kubevela aims to let developers to understand kubernetes less.

Yes, so we use WorkloadDefinition/TraitDefinition along with cuelang template to help reduce the complexity. 

>  inevitably,some services need tcp or udp port,and what follows is the need to define service type like clusterIP.

Yes. This requires the platform to have enough extensibility. With the help of kubevela, you can easily change the Workload/TraitDefinition of define a new one to meet the requirement.
 
>  Can ask about the development route in this regard ?

[...]

PR #856, opened 2020-12-31
Add provider parameter to route trait

Incubator4th:
#849  add provider to route,and default with nginx


Incubator4th:
@wonderflow I wonder why git diff workflow failed by only change cue template

PR #857, opened 2020-12-31
using 3-way-merge-patch instead of "json merge patch for workload" and "update for trait"

### LINKED ISSUE #848: [Feature] Refactor and enhance resource applicator

Since these two points ([1](https://github.com/crossplane/crossplane-runtime/blob/479dbc8a456086622ce1afab66aad11ba6c8f78d/pkg/resource/api.go#L138), [2](https://github.com/crossplane/crossplane-runtime/blob/479dbc8a456086622ce1afab66aad11ba6c8f78d/pkg/resource/api.go#L151)) in crossplane-runtime, it's necessary to refacotr the applicator to enhance ApplyOption and Patch option.

Based on crossplane-runtime interfaces and patching mechanism of `kubectl edit`, refactor applicators.

**Additional context**

captainroy-hy:
to fix #848 
- [x] implement util/applicator
New applicator makes a patch through computing a three way diff in client side, just like what `kubectl apply` do under the hood.
- [x] replace applicator used in oam controllers
- [x] unit test, e2e-test
- [x] add a new annotation  `app.oam.dev/last-applied-configuration` for every workload/trait created by appconfig controller, this will instore the configuration for calculating the 3-way-merge-patch-diff.

---

The workload/trait update logic will changed:

## workload

the old json-merge patch will be replaced by using three-way-merge patch.

The reason why abandon json-merge patch is that, it cannot remove a field through unsettling value in the patched manifest. 

### Before

For example, apply below deployment as a workload. json-merge patch cannot remove `minReadySeconds` field through applying a modifiied manifest with `minReadySeconds` omitted .
```yaml
# original workload manifest
apiVersion: apps/v1
kind: Deployment
...
spec:
  minReadySeconds: 60
  replicas: 3
  template:
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
---
# modified workload manifest
apiVersion: apps/v1
kind: Deployment
...
spec:
  # minReadySeconds: 60 <=== unset to remove field
  replicas: 3
  template:
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
---
# result 
apiVersion: apps/v1
kind: Deployment
...
spec:
  minReadySeconds: 60 # <=== not removed
  replicas: 3
  template:
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
```
### After

By computing a 3-way diff, we can get a patch aware of the field set in last-applied manifest is omitted in the new modified manifest, namely, users wanna remove this field. 
And an annotation, `app.oam.dev/last-applied-configuration`, is used to record last-applied-state of the resource for further use in computing 3-way diff next time.

```yaml
# result 
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations: # v=== record last-applied-state
    app.oam.dev/last-applied-configuration: | 
    {"apiVersion":"apps/v1","kind":"Deployment","metadata":{"name":"nginx-deployment","labels":{"app":"nginx"}},"spec":{"replicas":3,"selector":{"matchLabels":{"app":"nginx"}},"template":{"metadata":{"labels":{"app":"nginx"}},"spec":{"containers":[{"name":"nginx","image":"nginx:1.14.2"}]}}}}
...
spec:
  # minReadySeconds: 60 <=== removed successfully
  replicas: 3
  template:
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
```
---

## trait

The reasons why abandon update 

 - update always eliminates all fields set by others (e.g., trait controllers)
 - if trait contains immutable field (e.g., k8s Service), update fails

### Before
For example, apply below Service as a trait.
```yaml
# original trait manifest
apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  selector:
    app: myweb
  ports:
    - protocol: TCP
      port: 80
---
# after applying
apiVersion: v1
kind: Service
...
spec:
  clusterIP: 172.21.7.149 # <=== immutable field
  ports:
  - port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: myweb
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
---
# update with original manifest fails
# reconciling also fails for cannot applying trait
```
Additonally, if a trait has no immutable field, update will eliminate all fields set by others.
```yaml
# original trait manifest
kind: Bar
spec:
    f1: v1
# someone add a new field to it
kind: Bar
spec:
    f1: v1
    f2: v2 # <=== newly set field
# after reconciling AppConfig
kind: Bar
spec:
    f1: v1
    # f2: v2 <=== removed
```
But in practice, we expect to keep these changes.

### After

the update will be replaced by using three-way-merge patch.

It works in the same way as workloads. We use annotation, `app.oam.dev/last-applied-configuration`, to record last-applied manifest.

- Because 3-way diff will ignore the fields not touched in last-applied-state, the immutable fields will not be involved into patch data.
- Because 3-way diff also will ignore the fields set by others (others are not supposed to modify the field in original manifest), the changes made by others will be retained. 
    
### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply.go
@@ -22,7 +22,6 @@ import (
 	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
-	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -32,6 +31,7 @@ import (
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
 	"github.com/oam-dev/kubevela/pkg/oam/discoverymapper"
 	"github.com/oam-dev/kubevela/pkg/oam/util"
+	"github.com/oam-dev/kubevela/pkg/utils/apply"
 )
 
 // Reconcile error strings.
@@ -52,7 +52,7 @@ const (
 // A WorkloadApplicator creates or updates or finalizes workloads and their traits.
 type WorkloadApplicator interface {
 	// Apply a workload and its traits.
-	Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error
+	Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...apply.ApplyOption) error
 
 	// Finalize implements pre-delete hooks on workloads
 	Finalize(ctx context.Context, ac *v1alpha2.ApplicationConfiguration) error
@@ -60,12 +60,14 @@ type WorkloadApplicator interface {
 
 // A WorkloadApplyFns creates or updates or finalizes workloads and their traits.
 type WorkloadApplyFns struct {
-	ApplyFn    func(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error
+	ApplyFn    func(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...apply.ApplyOption) error
 	FinalizeFn func(ctx context.Context, ac *v1alpha2.ApplicationConfiguration) error
 }
 
-// Apply a workload and its traits.
-func (fn WorkloadApplyFns) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error {
+// Apply a workload and its traits. It employes the same mechanism as `kubectl apply`, that is, for each resource being applied,
+// computing a three-way diff merge in client side based on its current state, modified stated and last-applied-state which is
+// tracked through an specific annotaion. If the resource doesn't exist before, Apply will create it.
+func (fn WorkloadApplyFns) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...apply.ApplyOption) error {
 	return fn.ApplyFn(ctx, status, w, ao...)
 }
 
@@ -75,24 +77,18 @@ func (fn WorkloadApplyFns) Finalize(ctx context.Context, ac *v1alpha2.Applicatio
 }
 
 type workloads struct {
-	// use patching-apply for creating/updating Workload
-	patchingClient resource.Applicator
-	// use updating-apply for creating/updating Trait
-	updatingClient resource.Applicator
-	rawClient      client.Client
-	dm             discoverymapper.DiscoveryMapper
+	applicator apply.Applicator
+	rawClient  client.Client
+	dm         discoverymapper.DiscoveryMapper
 }
 
-//nolint:errorlint
-func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error {
+func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...apply.ApplyOption) error {
 	// they are all in the same namespace
 	var namespace = w[0].Workload.GetNamespace()
 	for _, wl := range w {
 		if !wl.HasDep {
-			err := a.patchingClient.Apply(ctx, wl.Workload, ao...)
-			if err != nil {
-				// TODO(roywang) use errors.As() instead of type assertion on error
-				if _, ok := err.(*GenerationUnchanged); !ok {
+			if err := a.applicator.Apply(ctx, wl.Workload, ao...); err != nil {
+				if !errors.Is(err, &GenerationUnchanged{}) {
 					// GenerationUnchanged only aborts applying current workload
 					// but not blocks the whole reconciliation through returning an error
 					return errors.Wrapf(err, errFmtApplyWorkload, wl.Workload.GetName())
@@ -104,9 +100,8 @@ func (a *workloads) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus,
 				continue
 			}
 			t := trait.Object
-			if err := a.updatingClient.Apply(ctx, &trait.Object, ao...); err != nil {
-				// TODO(roywang) use errors.As() instead of type assertion on error
-				if _, ok := err.(*GenerationUnchanged); !ok {
+			if err := a.applicator.Apply(ctx, &trait.Object, ao...); err != nil {
+				if !errors.Is(err, &GenerationUnchanged{}) {
 					// GenerationUnchanged only aborts applying current trait
 					// but not blocks the whole reconciliation through returning an error
 					return errors.Wrapf(err, errFmtApplyTrait, t.GetAPIVersion(), t.GetKind(), t.GetName())
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
-	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/crossplane/crossplane-runtime/pkg/test"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
@@ -37,7 +36,16 @@ import (
 
 	"github.com/oam-dev/kubevela/pkg/oam/mock"
 	"github.com/oam-dev/kubevela/pkg/oam/util"
+	"github.com/oam-dev/kubevela/pkg/utils/apply"
 )
+
+// ApplyFn mocks apply.Applicator for test convenience.
+type ApplyFn func(context.Context, runtime.Object, ...apply.ApplyOption) error
+
+// Apply implements apply.Applicator
+func (fn ApplyFn) Apply(ctx context.Context, o runtime.Object, ao ...apply.ApplyOption) error {
+	return fn(ctx, o, ao...)
+}
 
 func TestApplyWorkloads(t *testing.T) {
 	errBoom := errors.New("boom")
@@ -124,23 +132,21 @@ func TestApplyWorkloads(t *testing.T) {
 	}
 
 	cases := map[string]struct {
-		reason         string
-		client         resource.Applicator
-		updatingClient resource.Applicator
-		rawClient      client.Client
-		args           args
-		want           error
+		reason     string
+		applicator apply.Applicator
+		rawClient  client.Client
+		args       args
+		want       error
 	}{
 		"ApplyWorkloadError": {
 			reason: "Errors applying a workload should be reflected as a status condition",
-			client: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error {
+			applicator: ApplyFn(func(_ context.Context, o runtime.Object, _ ...apply.ApplyOption) error {
 				if w, ok := o.(*unstructured.Unstructured); ok && w.GetUID() == workload.GetUID() {
 					return errBoom
 				}
 				return nil
 			}),
-			updatingClient: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
-			rawClient:      nil,
+			rawClient: nil,
 			args: args{
 				w:  []Workload{{Workload: workload, Traits: []*Trait{{Object: *trait}}}},
 				ws: []v1alpha2.WorkloadStatus{}},
@@ -148,8 +154,7 @@ func TestApplyWorkloads(t *testing.T) {
 		},
 		"ApplyTraitError": {
 			reason: "Errors applying a trait should be reflected as a status condition",
-			client: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
-			updatingClient: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error {
+			applicator: ApplyFn(func(_ context.Context, o runtime.Object, _ ...apply.ApplyOption) error {
 				if t, ok := o.(*unstructured.Unstructured); ok && t.GetUID() == trait.GetUID() {
 					return errBoom
 				}
@@ -163,8 +168,7 @@ func TestApplyWorkloads(t *testing.T) {
 		},
 		"Success": {
 			reason: "Applied workloads and traits should be returned as a set of UIDs.",
-			client: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
-			updatingClient: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error {
+			applicator: ApplyFn(func(_ context.Context, o runtime.Object, _ ...apply.ApplyOption) error {
 				if o.GetObjectKind().GroupVersionKind().Kind == trait.GetKind() {
 					// check that the trait should not have a workload ref since we didn't return a special traitDefinition
 					obj, _ := util.Object2Map(o)
@@ -181,9 +185,8 @@ func TestApplyWorkloads(t *testing.T) {
 			},
 		},
 		"SuccessWithScope": {
-			reason:         "Applied workloads refs to scopes.",
-			client:         resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
-			updatingClient: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
+			reason:     "Applied workloads refs to scopes.",
+			applicator: ApplyFn(func(_ context.Context, o runtime.Object, _ ...apply.ApplyOption) error { return nil }),
 			rawClient: &test.MockClient{
 				MockGet: func(_ context.Context, key client.ObjectKey, obj runtime.Object) error {
 					if scopeDef, ok := obj.(*v1alpha2.ScopeDefinition); ok {
@@ -223,9 +226,8 @@ func TestApplyWorkloads(t *testing.T) {
 			},
 		},
 		"SuccessWithScopeNoOp": {
-			reason:         "Scope already has workloadRef.",
-			client:         resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
-			updatingClient: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
+			reason:     "Scope already has workloadRef.",
+			applicator: ApplyFn(func(_ context.Context, o runtime.Object, _ ...apply.ApplyOption) error { return nil }),
 			rawClient: &test.MockClient{
 				MockGet: func(_ context.Context, key client.ObjectKey, obj runtime.Object) error {
 					if scopeDef, ok := obj.(*v1alpha2.ScopeDefinition); ok {
@@ -265,9 +267,8 @@ func TestApplyWorkloads(t *testing.T) {
 			},
 		},
 		"SuccessRemoving": {
-			reason:         "Removes workload refs from scopes.",
-			client:         resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
-			updatingClient: resource.ApplyFn(func(_ context.Context, o runtime.Object, _ ...resource.ApplyOption) error { return nil }),
+			reason:     "Removes workload refs from scopes.",
+			applicator: ApplyFn(func(_ context.Context, o runtime.Object, _ ...apply.ApplyOption) error { return nil }),
 			rawClient: &test.MockClient{
 				MockGet: func(_ context.Context, key client.ObjectKey, obj runtime.Object) error {
 					if key.Name == scope.GetName() {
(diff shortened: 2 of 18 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration_without_traitdefinition_test.go
```diff
@@ -83,9 +83,6 @@ spec:
       containers:
         - name: nginx
           image: nginx:1.9.4
-          ports:
-            - containerPort: 80
-              name: nginx
```
wonderflow, line 88: why remove it?

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go
```diff
@@ -124,32 +124,32 @@ func TestApplyWorkloads(t *testing.T) {
...
 	cases := map[string]struct {
-		reason         string
-		client         resource.Applicator
-		updatingClient resource.Applicator
-		rawClient      client.Client
-		args           args
-		want           error
+		reason             string
+		patchingApplicator applicator.PatchingApplicator
+		updatingApplicator applicator.UpdatingApplicator
```
wonderflow, line 129: why we still need two? I can only see `applicator.Applicator` above.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration_without_traitdefinition_test.go
captainroy-hy: server-side apply complaints for missing protocol field in ports. and it seems ports field has no impact on this case.

pkg/utils/applicator/applicator.go
```diff
@@ -0,0 +1,220 @@
...
+		goto ClientSidePatch
+	}
+	if err != nil {
+		return errors.Wrap(err, "cannot patch object")
+	}
+	return nil
+ClientSidePatch:
+	// fall back to client-side patch if server-side not work
+	patch = client.Merge
+	return errors.Wrap(a.c.Patch(ctx, desired, patch), "cannot patch object")
```
wonderflow, line 132: Usually We don't like `goto` in such case, you can write it in one line instead of using `goto` if you feel redundant.

```
return errors.Wrap(a.c.Patch(ctx, desired, client.Merge), "cannot patch object")
```

test/integration/appconfig_test.go
```diff
@@ -101,12 +101,6 @@ func TestAppConfigController(t *testing.T) {
 									Name: envVars[2],
 								},
 							},
-							Ports: []v1alpha2.ContainerPort{
-								{
-									Name: "http",
-									Port: 8080,
-								},
-							},
```
wonderflow, line 109: why deleted?

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go
captainroy-hy: just to change the existing codes as little as possible

pkg/utils/applicator/applicator_test.go
```diff
@@ -0,0 +1,440 @@
...
+
+func (t *testObject) GetObjectKind() schema.ObjectKind {
+	return schema.EmptyObjectKind
+}
+
+type testNoMetaObject struct {
+	runtime.Object
+}
+
+func TestAPIApplicatorPatchingApply(t *testing.T) {
```
wonderflow, line 40: I think a mocked client is not enough, I hope to see real case using ginkgo(suit test) that can interact with K8s apiserver.

The tese cases should contain:

1. Deployment as Component(workload) , do strategyMerge patch。
    - normal field patch(add/update/delete)
    - array field patch(add/update/delete)
2. CRD as Component(workload) , do jsonMerge patch。
       - normal field patch(add/update/delete)
       - array field patch(add/update/delete)

captainroy-hy: Sure, I will cover these cases.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration_without_traitdefinition_test.go
ryanzhang-oss: I think this has something to do with CRD and K8s version

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go
ryanzhang-oss: Is it easy to merge these two later?

captainroy-hy: you mean merge like [this](https://github.com/oam-dev/kubevela/blob/7b18ef840b4889159cbef2f3a62ab51a2f1bbb38/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go#L485) ?

ryanzhang-oss: no, I mean merge into one applicator

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration_without_traitdefinition_test.go
captainroy-hy: The reason why SSA complaints missing protocol field is, start from k8s v1.16, new Go struct markers, `//+listType` and `//+listMapKey`, are introduced. For `type ContainerPort struct`, [SSA requires `containerPort` and `protocol` as associative list's map key](https://github.com/kubernetes/kubernetes/blob/5e22f7feadeb97aa8fad9330c936104479ad56d6/staging/src/k8s.io/api/core/v1/types.go#L2236-L2238). Even though `protocol` field has default value "TCP", but SSA works before defaulting system assign default value. So if `protocol` is missed, SSA will fail.

I will revert this deletion and complete them.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go
```diff
@@ -328,8 +321,15 @@ func TestApplyWorkloads(t *testing.T) {
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
 			mapper := mock.NewMockDiscoveryMapper()
+			testApplyClient := struct {
+				apply.PatchingApplicator
+				apply.UpdatingApplicator
```
wonderflow, line 326: why there still need updatingApplicator?

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go
```diff
@@ -328,8 +321,15 @@ func TestApplyWorkloads(t *testing.T) {
...
+			}{
+				tc.patchingApplicator,
+				nil,
```
wonderflow, line 329: if the updatingApplicator is nil, can we remove it?

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_test.go
```diff
@@ -475,7 +475,14 @@ func TestFinalizeWorkloadScopes(t *testing.T) {
 		t.Run(tc.caseName, func(t *testing.T) {
 			acTest := ac
 			mapper := mock.NewMockDiscoveryMapper()
-			w := workloads{patchingClient: tc.client, rawClient: tc.rawClient, dm: mapper}
+			testApplyClient := struct {
+				apply.UpdatingApplicator
```
wonderflow, line 479: the same question here

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_trait_test.go
```diff
@@ -203,15 +191,17 @@ spec:
...
 
 	AfterEach(func() {
-		logf.Log.Info("delete all cluster-scoped resources")
-		By("Delete CRD used as fake trait")
-		Expect(k8sClient.Delete(context.Background(), &crd)).Should(BeNil())
-		By("Delete fake TraitDefinition ")
-		Expect(k8sClient.Delete(context.Background(), &traitDef)).Should(BeNil())
+		Expect(k8sClient.DeleteAllOf(ctx, &appConfig, client.InNamespace(namespace))).Should(Succeed())
+		Expect(k8sClient.DeleteAllOf(ctx, &cw, client.InNamespace(namespace))).Should(Succeed())
+		Expect(k8sClient.DeleteAllOf(ctx, &component, client.InNamespace(namespace))).Should(Succeed())
```
wonderflow, line 196: it's smart to use DeleteAllOf  👍

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_trait_test.go
```diff
@@ -280,11 +270,57 @@ spec:
...
+			By("Reconcile")
+			reconcileRetry(reconciler, req)
+
+			By("Check the change is retained")
+			changedTrait = unstructured.Unstructured{}
+			changedTrait.SetAPIVersion("example.com/v1")
+			changedTrait.SetKind("Bar")
+			Expect(k8sClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: traitName}, &changedTrait)).Should(Succeed())
+			v, _, _ = unstructured.NestedString(changedTrait.UnstructuredContent(), "spec", "added")
+			Expect(v).Should(Equal("bar"))
```
wonderflow, line 323: It makes sure that normal reconcile won't change the added spec.

Can you add one more action below:

1. update the trait in AppConfig, add a new field, and delete an old field added by AppConfig before.
2.  reconcile it
3. check the third-party added field is still there, and check the new field added, the old field deleted as expected.

pkg/oam/labels.go
```diff
@@ -44,4 +44,8 @@ const (
 const (
 	// AnnotationAppGeneration records the generation of AppConfig
 	AnnotationAppGeneration = "app.oam.dev/generation"
+
+	// AnnotationLastAppliedConfig records the previous configuration of a
+	// resource for use in a three way diff during a patching apply
+	AnnotationLastAppliedConfig = "oam.dev/last-applied-configuration"
```
wonderflow, line 50:
```suggestion
	AnnotationLastAppliedConfig = "app.oam.dev/last-applied-configuration"
```

pkg/utils/apply/apply.go
```diff
@@ -0,0 +1,205 @@
...
+type Object interface {
+	metav1.Object
+	runtime.Object
+}
+
+// Applicator provides two approaches to applies changes to an object,
+// patching and updating.
+type Applicator interface {
+	PatchingApplicator
+	UpdatingApplicator
```
wonderflow, line 26: Will we have scenario to use UpdatingApplicator? If not, I suggest we remove it.

pkg/utils/apply/apply.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+// An Object is a Kubernetes object.
+type Object interface {
+	metav1.Object
+	runtime.Object
+}
+
+// Applicator provides two approaches to applies changes to an object,
+// patching and updating.
```
wonderflow, line 23: We can provide only PatchingApplicator here, then it's clear.

If we provide both patch and update. We must explain more details for the difference of these two approaches. Or it may cause lot's of confusion.

pkg/utils/apply/apply.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		return nil, errors.Wrap(c.Create(ctx, desired), "cannot create object")
+	}
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get object")
+	}
+	return existing, nil
+}
+
+func applyApplyOptions(ctx context.Context, existing, desired runtime.Object, aos []ApplyOption) error {
```
wonderflow, line 174: I suggest naming it `runApplyOptions`  or `executeApplyOptions`

pkg/utils/apply/apply.go
```diff
@@ -0,0 +1,205 @@
...
+
+// createOrGetExisting will create the object if it does not exist
+// or return the existing object if it already exists
+func createOrGetExisting(ctx context.Context, c client.Client, desired runtime.Object, ao ...ApplyOption) (runtime.Object, error) {
+	m, ok := desired.(Object)
+	if !ok {
+		return nil, errors.New("cannot access object metadata")
+	}
+
+	if m.GetName() == "" && m.GetGenerateName() == "" {
```
wonderflow, line 144: can we create the object successfully if this case exist? Maybe k8s apiserver will complain like `resource without a name can't be created`

pkg/utils/apply/apply.go
```diff
@@ -0,0 +1,205 @@
...
+	err := c.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, existing)
+	if kerrors.IsNotFound(err) {
+		// apply ApplyOption even though the object does not exist
+		if err := applyApplyOptions(ctx, nil, desired, ao); err != nil {
+			return nil, err
+		}
+		if err := addLastAppliedStateAnnotation(desired); err != nil {
+			return nil, err
+		}
+		return nil, errors.Wrap(c.Create(ctx, desired), "cannot create object")
```
wonderflow, line 166: BTW, I noticed the code here is exactly the same with L146~L152, if we really need both, we can use an anonymous function for this case.

pkg/utils/apply/scheme.go
```diff
@@ -0,0 +1,64 @@
...
+	rbacv1beta1.AddToScheme,
+	rbacv1alpha1.AddToScheme,
+	schedulingv1alpha1.AddToScheme,
+	storagev1beta1.AddToScheme,
+	storagev1.AddToScheme,
+}
+
+func init() {
+	_ = builder.AddToScheme(scheme)
+}
```
wonderflow, line 64: You don't need to manually add the built-in scheme here. client-go has already done it for you:

Check `k8s.io/client-go/kubernetes/scheme`

pkg/utils/apply/apply_resource_test.go
```diff
@@ -0,0 +1,196 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
+
+	oamstd "github.com/oam-dev/kubevela/apis/standard.oam.dev/v1alpha1"
+	oamutil "github.com/oam-dev/kubevela/pkg/oam/util"
+)
+
+var _ = Describe("Test Server-side Applying", func() {
```
wonderflow, line 19: do we still have server-side apply in this PR?

pkg/utils/apply/patch.go
```diff
@@ -0,0 +1,160 @@
...
+
+var metadataAccessor = meta.NewAccessor()
+
+// threeWayMergePatch creates a patch by computing a three way diff.
+func threeWayMergePatch(currentObj, modifiedObj runtime.Object) (client.Patch, error) {
+	current, err := json.Marshal(currentObj)
+	if err != nil {
+		return nil, err
+	}
+	original, err := getOriginalConfiguration(currentObj)
```
wonderflow, line 25: what will happen if the original is nil?

pkg/utils/apply/apply.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+type creatorFn func(context.Context, client.Client, runtime.Object, ...ApplyOption) (runtime.Object, error)
+
+func (fn creatorFn) createOrGetExisting(ctx context.Context, c client.Client, o runtime.Object, ao ...ApplyOption) (runtime.Object, error) {
+	return fn(ctx, c, o, ao...)
+}
+
+type patcher interface {
+	calculatePatch(c, m runtime.Object) (client.Patch, error)
```
wonderflow, line 81: the name should be `threeWayMergePatch`.  The `three way merge patch` is a famous name in K8s.

wonderflow (commented):
This PR is really great! Thanks @captainroy-hy 

I think this PR is also very important, it's very core part, we need to review carefully. \cc @ryanzhang-oss

pkg/utils/apply/patch.go
captainroy-hy: if original is nil, three-way-merge (both StrategicMerge and JSONMerge) will use "{}" as the original state. It would not result in a panic or failure.

pkg/utils/apply/apply.go
captainroy-hy: got it.

pkg/utils/apply/apply_resource_test.go
captainroy-hy: no server-side apply anymore, let me correct the case name.

pkg/utils/apply/scheme.go
captainroy-hy: Yes, I tried to use client-go scheme directly. But later I found not all types registered in client-go scheme are supported by strategic-merge as kubectl shown. so I just pick up the supported types from client-go scheme.

And another reason why not use client-go scheme is client-go/kubernetes/scheme.Scheme is a pkg level pointer variable.  It's possible to inject other types into it accidently, e.g., the test in /test/integration injects vela scheme into client-go scheme.Scheme indirectly.

pkg/utils/apply/apply.go
captainroy-hy: Totally agreed. I think we don't need updatingApply anymore. Let me remove it.

pkg/utils/apply/apply.go
captainroy-hy: We can remove UpdatingApplicator. Even there's scenario where need updating-apply, someone can just use raw client.Update() directly.

pkg/utils/apply/apply.go
captainroy-hy: it's used for creating objects with only generateName. I made a mistake here. it should be 
`m.GetGenerateName() != "" ` I will fix it.

pkg/utils/apply/apply.go
captainroy-hy: got it.

pkg/utils/apply/scheme.go
wonderflow: Does kubectl client code use their own?

pkg/utils/apply/apply.go
wonderflow: > someone can just use raw client.Update() directly.

Good idea!

pkg/utils/apply/scheme.go
captainroy-hy: Yes, it uses [this scheme](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/scheme/install.go) to determine strategicmerge or jsonmerge.

wonderflow: Can we use the same package with the kubectl's?

captainroy-hy: I checked again. Kubectl also use the scheme of client-go. I make a copy of client-go scheme to use now, instead of using the scheme of client-go directly.

ryanzhang-oss (commented):
LGTM overall. However, please note that this would only work if the fields we modify in the AC (apply) are not touched by the user controller, otherwise, it will result in a conflict.

pkg/utils/apply/patch.go
```diff
@@ -0,0 +1,166 @@
...
+	// do not include an empty map
+	if len(annots) == 0 {
+		_ = metadataAccessor.SetAnnotations(obj, nil)
+	}
+
+	var modified []byte
+	modified, err = json.Marshal(obj)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
captainroy-hy: line 116 need marshalled "obj without last-applied-config annotation" to set annotation. And 120-123 need to marshal obj with the annotation. So current logic may be okay.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration_without_traitdefinition_test.go
```diff
@@ -85,6 +85,7 @@ spec:
           image: nginx:1.9.4
           ports:
             - containerPort: 80
+              protocol: TCP
```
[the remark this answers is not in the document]
captainroy-hy: I will revert this change. It's used for testing server-side apply, but now we don't need it anymore.

wonderflow:
related: https://github.com/crossplane/oam-kubernetes-runtime/issues/272

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply.go
```diff
@@ -52,20 +52,20 @@ const (
...
 // A WorkloadApplyFns creates or updates or finalizes workloads and their traits.
 type WorkloadApplyFns struct {
-	ApplyFn    func(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error
+	ApplyFn    func(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...apply.ApplyOption) error
 	FinalizeFn func(ctx context.Context, ac *v1alpha2.ApplicationConfiguration) error
 }
 
 // Apply a workload and its traits.
-func (fn WorkloadApplyFns) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...resource.ApplyOption) error {
+func (fn WorkloadApplyFns) Apply(ctx context.Context, status []v1alpha2.WorkloadStatus, w []Workload, ao ...apply.ApplyOption) error {
```
hongchaodeng, line 68: Can you add more comments how this is implemented?

captainroy-hy: Sure, I will add more comments.

hongchaodeng (approved):
The code should work. LGTM and great job!
please add more comments to explain the implementation more. This looks hard.

pkg/utils/apply/patch.go
```diff
@@ -0,0 +1,166 @@
...
+	case err != nil:
+		return nil, err
+	default:
+		// use StrategicMergePatch for K8s built-in resources
+		patchType = types.StrategicMergePatchType
+		lookupPatchMeta, err = strategicpatch.NewPatchMetaFromStruct(versionedObject)
+		if err != nil {
+			return nil, err
+		}
+		patchData, err = strategicpatch.CreateThreeWayMergePatch(original, modified, current, lookupPatchMeta, true)
```
allenhaozi, line 67: elegant way,  I use a little complex way that add preApply function pre-patch the field managed by trait when workload use 
```
ao = append(ao, client.ForceOwnership)
a.Patch(ctx, desired, client.Apply, ao...) 
```
patch the resource

wonderflow:
Hi， @captainroy-hy  Great PR!  Before merge, I think we should finish the following actions:

1. the PR description should be more clear, tell us what is the expected behaviors before and after this PR. (it's better if we make it a small design doc)
2. check will this PR fix  https://github.com/oam-dev/kubevela/issues/876 
3. make sure the diff code coverage  larger than 70% and we need send it back to oam-k8s-runtime.

captainroy-hy:
@wonderflow 
- [x] add a simple desing doc | the PR description should be more clear, tell us what is the expected behaviors before and after this PR. (it's better if we make it a small design doc)
- [x] it fixed | check will this PR fix  #876 (found a bug #919)
- [x] make sure the diff code coverage  larger than 70% 
- [WIP]  send it back to oam-k8s-runtime.

wonderflow (approved):
Excellent Job! Please also put what describes in PR body into the design doc.

PR #858, opened 2021-01-02
Add provider parameter to route trait

Incubator4th:
#849 add provider to route,and default with nginx

### DIFF
--- a/hack/vela-templates/cue/route.cue
+++ b/hack/vela-templates/cue/route.cue
@@ -26,5 +26,5 @@ parameter: {
 		path:          string
 		rewriteTarget: *"" | string
 	}]
-	provider: *"" | string
+	provider?: string
 }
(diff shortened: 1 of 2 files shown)


wonderflow:
>  @wonderflow I wonder why git diff workflow failed by only change cue template

Because the file will automatically formatted. You can try:

```
make reviewable
```

The command will format automatically.

wonderflow:
You should change the file here: https://github.com/oam-dev/kubevela/blob/master/hack/vela-templates/cue/route.cue

hack/vela-templates/cue/route.cue
```diff
@@ -24,4 +26,5 @@ parameter: {
 		path:          string
 		rewriteTarget: *"" | string
 	}]
+	provider: *"" | string
```
wonderflow, line 29:
```suggestion
	provider?: string
```

I think you can write like this for optional field.

wonderflow (approved):
LGTM

PR #861, opened 2021-01-04
feature: add healthcheck policy for workload and trait by cue template

rain18:
- [x] Support Custom Health Check Logic for Appliction by Cue Template

Vela is used for application release, and how to evaluate the success of an application release? Not only Vela creates workloads and traits, but more importantly, Vela should check the health of workloads and traits. HealthScope is hardcoded for the health check logic of various resources. Here providers a custom health check plan.

### Health Check Policy

After the K8s resource is created, the relevant data will be written back to the status of CR. Whether the release is successful is actually a series of judgments and filter conditions combination, called health policy. Vela should combine with the data in CR and the policy to determine whether a resource object meets the conditions. For example, when the values of replicas and readyReplicas in the status of Deployment CR are equal, the release is successful.

Developers could customize the health check policy in the definition, which has high scalability.
``` yaml
kind: WorkloadDefinition
metadata:
  name: worker
  annotations:
    definition.oam.dev/description: "Long-running scalable backend worker without network endpoint"
spec:
  definitionRef:
    name: deployments.apps
  extension:
    healthPolicy: |
      isHealth: output.status.replicas == output.status.readyReplicas
    template
      ... 
```
* **healthPolicy** is the customized policy by developer 
* **output** represents the CR
### Example
* Workload(deployment) health policy
```yaml
kind: WorkloadDefinition
metadata:
  name: worker
  annotations:
    definition.oam.dev/description: "Long-running scalable backend worker without network endpoint"
spec:
  definitionRef:
    name: deployments.apps
  extension:
    healthPolicy: |
      isHealth: output.status.replicas == output.status.readyReplicas
    template: |
      output: {
      	apiVersion: "apps/v1"
      	kind:       "Deployment"
      	spec: {
      		selector: matchLabels: {
      		}
      		template: {
      			metadata: labels: {
      			}
      			spec: {
      				containers: [{
      					name:  context.name
      					image: parameter.image

      					if parameter["cmd"] != _|_ {
      						command: parameter.cmd
      					}
      				}]
      			}
      		}

      		selector:
      			matchLabels:
      	}
      }

      parameter: {
      	// +short=i
      	image: string

      	cmd?: [...string]
      }`
```
* Trait(ManualScalerTrait) health policy
It is assumed that the trait will return whether the release is successful through the **phase** of status
```yaml
kind: TraitDefinition
metadata:
  annotations:
    definition.oam.dev/description: "Manually scale the app"
  name: scaler
spec:
  appliesToWorkloads:
    - webservice
    - worker
  definitionRef:
    name: manualscalertraits.core.oam.dev
  workloadRefPath: spec.workloadRef
  extension:
    healthPolicy: |
      isHealth: output.status.phase == "ready"
    template: |-
      output: {
      	apiVersion: "core.oam.dev/v1alpha2"
      	kind:       "ManualScalerTrait"
      	spec: {
      		replicaCount: parameter.replicas
      	}
      }
      parameter: {
      	//+short=r
      	replicas: *1 | int
      }
```

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/application_controller_test.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/application_controller_test.go
@@ -25,9 +25,8 @@ import (
 	"net/http/httptest"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
-
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	"github.com/google/go-cmp/cmp"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/apps/v1"
@@ -41,6 +40,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
+	"github.com/oam-dev/kubevela/pkg/oam"
 	"github.com/oam-dev/kubevela/pkg/oam/util"
 )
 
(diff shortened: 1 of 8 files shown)


ryanzhang-oss:
Good idea overall. I wonder what type of health check does this support? There must be a limit.

pkg/dsl/definition/template_test.go
```diff
@@ -133,3 +133,51 @@ parameter: {
 		}}
 	assert.Equal(t, expect, obj)
 }
+
+//
```
ryanzhang-oss, line 137: why is this commented out?

pkg/dsl/definition/template.go
```diff
@@ -160,3 +229,82 @@ func (td *traitDef) Complete(ctx process.Context) error {
...
+		}
+	} else {
+		return errors.WithMessagef(fmt.Errorf("there is no trait output cr for health check"), "trait %s health check", td.name)
+	}
+	insts := cue.Build([]*build.Instance{bi})
+	for _, inst := range insts {
+		if err := inst.Value().Err(); err != nil {
+			return errors.WithMessagef(err, "trait %s check", td.name)
+		}
+		isHealthVal := inst.Lookup("isHealth")
```
ryanzhang-oss, line 276: IIRC, this means the "health" template is a cue expression, right?

ryanzhang-oss (changes requested):

ryanzhang-oss:
> Good idea overall. I wonder what type of health check does this support? There must be a limit.

I wonder how can the "health" expression work when the fields are in a YAML/Json format instead of cue?

pkg/dsl/definition/template.go
rain18: Yes.  And the "health" check just is used for the output of CR, which could be any kind of CR because of powerful cue expression.

rain18:
> > Good idea overall. I wonder what type of health check does this support? There must be a limit.
> 
> I wonder how can the "health" expression work when the K8s CR output are in a YAML/JSON format instead of cue?
>
> The "health" expression is cue template, but data(k8s CR output) is JSON used to combine with "health" expression.

ryanzhang-oss:
> > > Good idea overall. I wonder what type of health check does this support? There must be a limit.
> 
> Thanks, the "health" check just is used for the output of CR, which could be any kind of CR as long as you can write relevant expressions correctly with powerful cue expression.
> 
> > > I wonder how can the "health" expression work when the K8s CR output are in a YAML/JSON format instead of cue?
> 
> The "health" expression is cue expression, but data(k8s CR output) is JSON used to combine with "health" expression.

I like the idea that we can add "health check"  CUE formula to the template. This is useful for the health scope. 

However, the application controller shouldn't care about each workload/trait's health. We have a healthScope that handles that.

I wonder if you can make the UT test work and then integrate this formula with the healthScope implementation instead (in a separate PR)?

pkg/controller/core.oam.dev/v1alpha2/application/application_controller_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/oam-dev/kubevela/pkg/oam"
```
wonderflow, line 25: nit

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -4,6 +4,9 @@ import (
 	"context"
 	"time"
 
+	"github.com/oam-dev/kubevela/pkg/controller/core.oam.dev/v1alpha2/application/parser"
+	"github.com/oam-dev/kubevela/pkg/dsl/process"
+
```
wonderflow, line 9: nit

wonderflow (approved):
Generally LGTM, thanks for your effort @rain18 !  ping @ryanzhang-oss  I think we can merge it and refactor in the following PRs.

pkg/dsl/definition/template_test.go
rain18: I have remove the unwanted code

pkg/controller/core.oam.dev/v1alpha2/application/application_controller_test.go
rain18: done

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
rain18: done

ryanzhang-oss (approved):

PR #882, opened 2021-01-13
fix CI script

ryanzhang-oss:
- Fix CI script
- Better organize tests in CI
- Move tests to better places


wonderflow:
fixed by #887

PR #886, opened 2021-01-14
Generate the reference doc for a workload type or trait in the local host website and in a table style on Cli console

zzxwill:
By Cli `vela reference` to generate reference docs for all workload
types and traits, and host them in a local website.

To fix #880


zzxwill:
@wonderflow I've made a quick e2e solution PoC to feature #880. Please take a look at it. I will store capability reference to ConfigMap later. Thanks.

```
$ vela reference
```

Open a browser to visit the reference docs site which is hosted locally.

image

resouer:
My 2c is to show the doc of independent workload type or trait instead of full list.

```console
$ vela show webservice
```

Then it just pop up this page: https://kubevela.io/#/en/developers/references/workload-types/webservice

zzxwill:
> My 2c is to show the doc of independent workload type or trait instead of full list.

Sure, I can show the exact doc for a workload type or trait.

> 
> ```
> $ vela show webservice
> ```
> 
> Then it just pop up this page: https://kubevela.io/#/en/developers/references/workload-types/webservice

Am I understanding correctly?
- for built-in workload type or trait, we should redirect to `kubevela.io`.
- for extended workload type or trait, we can load the responding doc in our local hosted doc site `127.0.0.1:18080`.

hongchaodeng:
@zzxwill

> for built-in workload type or trait, we should redirect to kubevela.io
 
AFAICT, it is good enough to show the reference docs as the website does, not about showing a link or redirecting via browser.

Currently, the reference docs in the website is auto generated and put into the website files. We can do the same for the CLI/UI as well by saving those data on server side. For the long term, we should let the /definitions API endpoint return the documentation too. For now, the CLI doesn't talk to the API Server yet, so we could probably save it as a ConfigMap on k8s.

wonderflow:
The general idea to show docs in a generated website is good. But most cli user maybe don't have a GUI browser, what if the user only have a terminal console and hope to see the docs?

wonderflow:
By the way, we're improving the code quality for the whole project, we need unit-tests before merge.

zzxwill:
> The general idea to show docs in a generated website is good. But most cli user maybe don't have a GUI browser, what if the user only have a terminal console and hope to see the docs?
> 
> So I think the cli still need to show a simple workload/trait reference in terminal level.

Let me implement this feature and try to make it as simple and clear as the website does, but it will take more time for the table-style formatting.

zzxwill:
> By the way, we're improving the code quality for the whole project, we need unit-tests before merge.

Of course, working on it. Thanks for the kind reminder.

zzxwill:
Seems `CodeCov` is broken as the data  of the latest commit is not refreshed.
image

And the detailed coverage statistics could not be visited. 
image

zzxwill:
> The general idea to show docs in a generated website is good. But most cli user maybe don't have a GUI browser, what if the user only have a terminal console and hope to see the docs?
> 
> So I think the cli still need to show a simple workload/trait reference in terminal level.

Here is a snapshot to display `webservice` on console.

image

wonderflow:
> Seems CodeCov is broken as the data of the latest commit is not refreshed.

If the unit-test broken, the coverage data doesn't make sense so it won't refresh.

wonderflow:
> > The general idea to show docs in a generated website is good. But most cli user maybe don't have a GUI browser, what if the user only have a terminal console and hope to see the docs?
> > So I think the cli still need to show a simple workload/trait reference in terminal level.
> 
> Here is a snapshot to display `webservice` on console.
> 
> image

It will also close #896

pkg/plugins/cluster.go
```diff
@@ -198,3 +198,72 @@ func HandleTemplate(in *runtime.RawExtension, name, syncDir string) (types.Capab
 	}
 	return tmp, nil
 }
+
+// SyncDefinitionsToLocal sync definitions to local
+func SyncDefinitionsToLocal(ctx context.Context, c types.Args, localDefinitionDir string) ([]types.Capability, []string, error) {
```
wonderflow, line 203: please write a test for this func

pkg/plugins/cluster.go
```diff
@@ -198,3 +198,72 @@ func HandleTemplate(in *runtime.RawExtension, name, syncDir string) (types.Capab
...
+			warnings = append(warnings, fmt.Sprintf("WARN: %v, you will unable to use this trait capability\n", e))
+		}
+	}
+	syncedTemplates = append(syncedTemplates, templates...)
+	SinkTemp2Local(templates, localDefinitionDir)
+	return syncedTemplates, warnings, nil
+}
+
+// SyncDefinitionToLocal sync definitions to local
+func SyncDefinitionToLocal(ctx context.Context, c types.Args, localDefinitionDir string, capabilityName string) (*types.Capability, error) {
```
wonderflow, line 234: please write a test for this func

wonderflow (approved):
LGTM

PR #891, opened 2021-01-16
roadmap: add template, rewrite 2020 winter, add 2021 spring

hongchaodeng:


docs/en/roadmap/template.md
```diff
@@ -0,0 +1,39 @@
...
+- CI/CD, GitOps
+- Application development framework
+- Third party workloads/traits
+
+## Testing
+
+- Test infrastructure
+  - CI (Github Actions) and hosts
+  - codecov
+- Unit/e2e test scripts and go code
```
zzxwill, line 34: How about we simply use `Unit/e2e test cases`? I got confused the first time I read `go code`. Now I realized `scripts` and `go code` are used to write unit/e2e test cases.

docs/en/roadmap/2020-12-roadmap.md
```diff
@@ -0,0 +1,27 @@
+# Roadmap
+
+Date: 2020-10-01 to 2021-12-30
```
zzxwill, line 3:
```suggestion
Date: 2020-10-01 to 2021-12-31
```

zzxwill (approved):
LGTM

docs/en/roadmap/2020-12-roadmap.md
wonderflow, line 3:
```suggestion
Date: 2020-10-01 to 2020-12-31
```

?

wonderflow (changes requested):
There are some confusions I have commented.

docs/en/roadmap/template.md
hongchaodeng: Good idea!
Updated.

docs/en/roadmap/2020-12-roadmap.md
hongchaodeng: > Do you mean 2020 or the whole year of 2021?

2020

> If you means 2020, it has been passed, why do we need to write a separate part of roadmap for it?

There is a roadmap for it already. I only changed the format to make all roadmaps consistent.

> I guess you want to make it as a record, but some of the features in roadmap haven't been finished, I think we need to move them out.

That's fine. It is just for history records. For example, in the future we might write more roadmaps, and we probably won't go back to delete those items that we haven't finished..

wonderflow (approved):
LGTM

PR #892, opened 2021-01-16
Rollout Design Doc

ryanzhang-oss:
Rollout Design Doc

### DIFF
--- a/design/vela-core/rollout-design.md
+++ b/design/vela-core/rollout-design.md
@@ -1,0 +1,269 @@
+# OAM Rollout Controller Design
+
+- Owner: Ryan Zhang (@ryangzhang-oss)
+- Date: 01/14/2021
+- Status: Draft
+
+## Table of Contents
+
+- [Introduction](#introduction)
+   - [Two flavors of Rollout](#two-flavors-of-rollout)
+- [Goals](#goals)
+- [Proposal](#proposal)
+  - [Registration via Definition/Capability](#registration-via-definitioncapability)
+  - [Templating](#templating)
+  - [CLI/UI interoperability](#cliui-interoperability)
+  - [vela up](#vela-up)
+- [Examples](#examples)
+
+## Introduction
+
+`Rollout` or `Upgrade` is one of the most essential "day 2" operation on any application
+. KubeVela, as an application centric platform, definitely needs to provide a customized solution
+to alleviate the  burden on the application operators. There are several popular rollout solutions
+, i.e. [flagger](https://flagger.app/),  in the open source community. However, none of them
+work directly with our OAM framework. Therefore, we propose to create an OAM native rollout
+ framework that can address all the application rollout/upgrade needs in Kubevela.
+    
+### Two flavors of Rollout
+After hearing from the OAM community, it is clear to us that there are two flavors of rollout
+ that we need to support. 
+ - One way is through an OAM trait. This flavor works for existing OAM applications that don't
+  have many specialized requirements such as interaction with other traits. For
+  example, most rollout operations don't work well with scaling operations. Thus, the
+  application operator needs to remove any scalar traits from the component before applying the
+  rollout trait. Another example is rollout operations usually also involve traffic spliting
+  of sorts. Thus, the application operators might also need to manually adjust the related
+  traffic trait before and after applying the rollout trait. 
+ - The other ways is through a new applicationDeployment CR which directly reference different
+  versions of applications instead of workloads. This opens up the possibility for the controller
+  to solve conflicts between traits automatically. This resource, however, not a trait and requires
+  the application to be immutable. 
+
+### Design Principles and Goals
+We design our controllers with the following principles in mind 
+  - First, we want all flavors of rollout controllers share the same core rollout
+   related logic. The trait and application related logic can be easily encapsulated into its own
+   package.
+  - Second, the core rollout related logic is easily extensible to support different type of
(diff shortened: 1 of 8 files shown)


design/vela-core/rollout-design.md
```diff
@@ -0,0 +1,269 @@
...
+	// All pods in the batches up to the batchPartition (included) will have
+	// the target resource specification while the rest still have the source resource
+	// This is designed for the operators to manually rollout
+	// Default is the the number of batches which will rollout all the batches
+	// +optional
+	BatchPartition *int32 `json:"lastBatchToRollout,omitempty"`
+
+	// RevertOnDelete revert the rollout when the rollout CR is deleted, default is false 
+	//+optional 
+	RevertOnDelete bool `json:"revertOnDelete,omitempty"`
```
wonderflow, line 121: Do we have real scenario on this field?

design/vela-core/rollout-design.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+## User Experience Workflow
+OAM rollout experience is different from flagger in a few different ways and here are the
+ implications on its impact on the user experience.
+- We assume that the resources it refers to are **_immutable_**. In contrast, flagger watches over a
+  target resource and reacts whenever the target's specification changes.
+    - The trait version of the controller refers to componentRevision.
+    - The application version of the controller refers to immutable application. 
+- The rollout logic **_works only once_** and stops after it reaches a terminal state. One can
```
wonderflow, line 145: Does it mean if a rollout CR works completely, then it won't work even the CR spec is changed?

design/vela-core/rollout-design.md
```diff
@@ -0,0 +1,269 @@
...
+  appDeployment controll has full control of the new application from the beginning.
+ However, there is no kubernetes systematic way of acquiring a distributed lock for a controller.  
+- The appDeployment controller will make sure it marks itself as the owner of the
+ application and changes the appDeployment CR state to be initialized. The application controller
+  can be aware of appDeployment and have build-in logics in case.
+- The appDeployment controller can change the application fields. For example, it can remove all
+ the conflict traits, such as HPA, during the upgrade period. 
+- Upon a successful rollout, the appDeployment controller can remove all the traits in the old
+  application and leave no pods in the workload.
+- Upon a failed rollout, the condition is not determined, it could result in an unstable state.
```
wonderflow, line 180: Why it's unstable? Even a failed rollout should make sure the replicas alive larger or equal then it should be or maxUnavaliable defined, right?

design/vela-core/rollout-design.md
```diff
@@ -0,0 +1,269 @@
...
+
+	// RolloutWebhooks provides a way for the rollout to interact with an external process
+	// +optional
+	RolloutWebhooks []RolloutWebhook `json:"rolloutWebhooks,omitempty"`
+
+	// CanaryMetric provides a way for the rollout process to automatically check certain metrics
+	// before complete the process
+	// +optional
+	CanaryMetric []CanaryMetric `json:"canaryMetric,omitempty"`
+}
```
wonderflow, line 135: I don't see some fields like `maxUnavailable` here? Maybe we need to consider rollout strategies contained in statefulset or cloneset?

design/vela-core/rollout-design.md
```diff
@@ -0,0 +1,269 @@
...
+  - [Templating](#templating)
+  - [CLI/UI interoperability](#cliui-interoperability)
+  - [vela up](#vela-up)
+- [Examples](#examples)
+
+## Introduction
+
+`Rollout` or `Upgrade` is one of the most essential "day 2" operation on any application
+. KubeVela, as an application centric platform, definitely needs to provide a customized solution
+to alleviate the  burden on the application operators. There are several popular rollout solutions
```
zzxwill, line 23:
```suggestion
to alleviate the burden on the application operators. There are several popular rollout solutions
```

design/vela-core/rollout-design.md
```diff
@@ -0,0 +1,269 @@
...
+ 
+## Notable implementation level details
+Here are some high level implementation decisions
+
+### Rollout on different resources
+As we mentioned in the introduction section, we will implement two rollout controllers that work
+on different levels. At the end, they both emit an in-memory rollout plan object which includes
+references to the target and source kubernetes resources that the rollout planner will execute
+upon. For example, the applicationDeployment controller will get the component from the
+ application and extract the real workload from it before passing it to the rollout plan object.
```
zzxwill, line 163: So the trait controller will only refer to the upper workload and vela-core/OAM runtime will do the extraction of the real underlying workload?

design/vela-core/rollout-design.md
ryanzhang-oss: the only way to trigger it to restart is when the target or source resource is changed

design/vela-core/rollout-design.md
ryanzhang-oss: the controller will do the extraction, not the OAM runtime

design/vela-core/rollout-design.md
ryanzhang-oss: I didn't paste all the types, the maxUnavailable is in the RolloutBatch structure. However, since cloneset already has its own strategy, I don't want to repeat it here as it may not apply to other resources.

design/vela-core/rollout-design.md
ryanzhang-oss: since the appDeployment creates/remove traits, I am not sure if there is a way to guarantee the intermediate state with two application running is stable. It's not just the replica number.

PR #901, opened 2021-01-20
Add `Writing Appfile` based on capability references docs

zzxwill:
Introduced `vela show` with and without flag `--no-website`, and
official site to help setting properties for workload types and
traits.

PR #916, opened 2021-01-21
Fix ref doc guide

resouer:
Fixes #901.

The original PR is lack of design and has bugs, please review carefully next time.

Also, in general we should not add more end user docs before 1.0 unless significant feature are released (Helm or Terraform support etc), the platform builder docs need more efforts.


wonderflow (approved):
LGTM

docs/en/_sidebar.md
```diff
@@ -36,7 +35,7 @@
 
 - References
   - [Appfile](/en/developers/references/devex/appfile.md)
-  - [Capabilities](./developers/references/README.md)
+  - [Capabilities](/en/developers/references/README.md)
```
zzxwill, line 38: This relative path is working and I have verified. @sunny0826 could you please double check it? Absolute path is not friendly for docs writer and to verify the link before deploying it to Github pages or running `docsify server`.

sunny0826: Both ways of writing are valid

Issue #913, opened 2021-01-21
[Feature] Application creation OpenAPI based on new Application object


zzxwill:
Based on [Application](https://kubevela.io/#/en/concepts?id=application) object, implement OpenAPI to create an application.

Follow KubeVela Restful API specification https://kubevela.io/en/developers/references/restful-api/index.html

zzxwill:
@leejanee is interested in it.

PR #924, opened 2021-01-22
OpenAPI: Application creation based on Application object

zzxwill:
Implemented Application creation OpenAPI based on new Application object
To #913


hongchaodeng (approved):
lgtm

PR #926, opened 2021-01-22
update APIServer + Catalog design

hongchaodeng:
- add application creation workflow
- make application adaptable to multiple env and clusters
- add module into catalog package format and adapt to helm/terraform

The design will fix https://github.com/oam-dev/kubevela/issues/864 https://github.com/oam-dev/kubevela/issues/895

### DIFF
--- a/design/vela-core/APIServer-Catalog.md
+++ b/design/vela-core/APIServer-Catalog.md
@@ -291,8 +291,10 @@ The structure of one package version contains:
         repo: https://kubernetes.github.io/ingress-nginx
         name: ingress-nginx
   - terraform:
-      source: terraform-aws-modules/rds/aws
-      version: "~> 2.0"
+      path: ./tf_modules/rds/ # local path in this package
+      remote:
+        source: terraform-aws-modules/rds/aws
+        version: "~> 2.0"
   ```
 
 #### Register a catalog in APIServer
--- a/design/vela-core/APIServer-Catalog.md
+++ b/design/vela-core/APIServer-Catalog.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-In KubeVela, APIServer provides the RESTful API for external systems (e.g. UI) to manage Vela abstractions like Applications, Definitions; Catalog stores templates to install common-off-the-shell (COTS) applications on Kubernetes.
+In KubeVela, APIServer provides the RESTful API for external systems (e.g. UI) to manage Vela abstractions like Applications, Definitions; Catalog stores templates to install common-off-the-shell (COTS) capabilities on Kubernetes.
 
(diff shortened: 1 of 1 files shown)


design/vela-core/APIServer-Catalog.md
```diff
@@ -233,33 +246,11 @@ A catalog contains multiple packages, and a package contains multiple versions.
...
-      name: redis
   ```
 
-  Note that we choose to integrate with existing community solutions instead of inventing another packaging solution (e.g. Helm, Terraform). In this way we can adopt the reservoir of community resources and also be extensible to more packaing formats.
-
-- Each package could expose multiple workload and trait definitions under `definitions/`. These definitions will be parsed on APIServer side to check if a cluster needs install or upgrade this package by comparing definitions.
+- `definitions`: definition files that describe the capabilities from this package to enable on a cluster. Note that these definitions will compared against a cluster on APIServer side to see if a cluster can install or upgrade this package.
 
-- The parameters of this package are defined in `parameters.yaml` . It will be used as the values to render the manifest source (e.g. Helm Chart).
-
```
resouer, line 250: why parameters are deleted?

hongchaodeng: This is not useful, and we shouldn't require users to write this file.
Because we can just parse the schema of parameters from Helm Chart or Terraform. For example, Helm supports [value schema file](https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/) for input validation and there is an [automation tool](https://github.com/karuppiah7890/helm-schema-gen] to generate the schema.

wonderflow (approved):
LGTM

design/vela-core/APIServer-Catalog.md
```diff
@@ -311,8 +317,14 @@ Vela APIServer aggregates package information from multiple catalog servers. To
 
 In our future roadmap, we will build a catalog controller for each k8s cluster. Then we will add API endpoint to install the package in APIServer which basically creates a CR to trigger the controller to reconcile package installation into the cluster. We choose this instead of APIServer installing the package because in this way we can bypass the APIServer in the package data transfer path and avoid APIServer becoming single point of failure.
 
+## Examples
+
 ## Considerations
 
+### Package parameters
+
+We can parse the schema of parameters from Helm Chart or Terraform. For example, Helm supports [value schema file](https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/) for input validation and there is an [automation tool](https://github.com/karuppiah7890/helm-schema-gen] to generate the schema.
```
vnzongzna, line 326: automation tool](https://github.com/karuppiah7890/helm-schema-gen] -> [automation tool](https://github.com/karuppiah7890/helm-schema-gen)

PR #927, opened 2021-01-22
some initial code for app deployment controller/webhook

ryanzhang-oss:
added code dealing with fetching component from application

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationdeployment/applicationdeployment_controller.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationdeployment/applicationdeployment_controller.go
@@ -8,17 +8,20 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	ktypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
+	"k8s.io/kubectl/pkg/util/slice"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
 	corev1alpha2 "github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
 	controller "github.com/oam-dev/kubevela/pkg/controller/core.oam.dev"
 	"github.com/oam-dev/kubevela/pkg/oam/discoverymapper"
 )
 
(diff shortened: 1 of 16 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationdeployment/applicationdeployment_controller.go
```diff
@@ -32,26 +35,68 @@ type Reconciler struct {
...
+		klog.InfoS("get the source workload we need to work on", "sourceWorkload", klog.KObj(sourceWorkload))
+	}
+	// TODO: pass these two object to the rollout plan
 	return ctrl.Result{}, nil
 }
 
+func (r *Reconciler) handleFinalizer(appDeploy *corev1alpha2.ApplicationDeployment) {
+	if appDeploy.DeletionTimestamp.IsZero() {
+		if !slice.ContainsString(appDeploy.Finalizers, appDeployfinalizer, nil) {
+			// TODO: add finalizer
```
wonderflow, line 85: do you mean remove finalizer

pkg/controller/core.oam.dev/v1alpha2/applicationdeployment/applicationdeployment_controller.go
```diff
@@ -32,26 +35,68 @@ type Reconciler struct {
...
+	}
+	klog.InfoS("get the target workload we need to work on", "targetWorkload", klog.KObj(targetWorkload))
+	if sourceWorkload != nil {
+		klog.InfoS("get the source workload we need to work on", "sourceWorkload", klog.KObj(sourceWorkload))
+	}
+	// TODO: pass these two object to the rollout plan
 	return ctrl.Result{}, nil
 }
 
+func (r *Reconciler) handleFinalizer(appDeploy *corev1alpha2.ApplicationDeployment) {
```
wonderflow, line 82: what do you prepare to use finalizer for?

wonderflow (approved):

pkg/controller/core.oam.dev/v1alpha2/applicationdeployment/applicationdeployment_controller.go
ryanzhang-oss: this is the case that the object is not deleted (the delete timestamp is zero) and the finalizer is not there yet, so we need to add the finalizer there.

pkg/controller/core.oam.dev/v1alpha2/applicationdeployment/applicationdeployment_controller.go
ryanzhang-oss: I plan to make sure that any changes we make to the application is reverted

wonderflow:
please resolve the conflict @ryanzhang-oss and we can merge this PR

Issue #928, opened 2021-01-22
[Feature] cert-manager install optional


hprotzek:
**Is your feature request related to a problem? Please describe.**
Trying to install kubevela into our existing gke dev cluster, which already has the cert-manager installed, results in an error.

```
Failed to install the chart with error: ClusterRole "cert-manager-cainjector" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "kubevela": current value is "cert-manager"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "vela-system": current value is "ee-system"
rendered manifests contain a resource that already exists. Unable to continue with install
```

**Describe the solution you'd like**
Make the cert-manager optional in the helm chart.

resouer:
I am fully supporting this. 

As a plus, I think we should make default installation of KubeVela does not include any 3rd party dependency, and only support basic traits (ingress, hpa, simple rollout etc.) since now we are pretty good on modeling any built-in capabilities.

And as the next step, user can choose to install different capabilities step by step.

/cc @hongchaodeng @wonderflow

wonderflow:
kubevela webhook depends on cert-manager, do you mean we should disable webhook by default? @resouer

resouer:
@wonderflow webhook should be there, but we should allow user install cert-manager by himself.

LeoLiuYan:
> kubevela webhook depends on cert-manager, do you mean we should disable webhook by default? @resouer

Maybe `KubeVela` can generate certs by itself, just like `Kruise` do?

wonderflow:
I have supported cert-manager as optional, by default we will install it.  I suggest we raise a new issue for this.

> I would support this.
> 
> As a plus, I think we should make default installation of KubeVela does not include any 3rd party dependency, and only support basic traits (ingress, hpa, simple rollout etc.) since now we are pretty good on modeling any built-in capabilities.
> 
> And as the next step, user can choose to install different capabilities step by step.
> 
> /cc @hongchaodeng @wonderflow

hprotzek:
Can all dependencies, be optional? Because while testing kubevela in our existing gke cluster, also the already existing `kube-prometheus-stack` got broken due duplicate installed webhooks from the prometheus operator. 

So maybe https://github.com/oam-dev/kubevela/pull/934 can be extended to make the other helm chart dependencies from  `velaConfig` also optional?

resouer:
[...]

PR #934, opened 2021-01-25
support vela install with helm args and make cert-manager optional

wonderflow:
1. support `vela install ` with helm args.

For example:

```
vela install --set useWebhook=false --set image.pullPolicy=always 
```

Use `--set` to sepcify "key=value" pairs just like what `helm install` supports.

2. support cert manager to be optional, specify it as false will not install it. part of #928 

```
vela install --set installCertManager=false
```

3. remove webhook secret volume when webhook not enabled.

### DIFF
--- a/pkg/commands/system.go
+++ b/pkg/commands/system.go
@@ -48,6 +48,7 @@ type chartArgs struct {
 	imageRepo       string
 	imageTag        string
 	imagePullPolicy string
+	more            []string
 }
 
 type infoCmd struct {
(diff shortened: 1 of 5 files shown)


pkg/commands/system.go
```diff
@@ -225,7 +227,7 @@ func (i *initCmd) resolveValues() (map[string]interface{}, error) {
 	if i.chartArgs.imagePullPolicy != "" {
 		valuesConfig = append(valuesConfig, fmt.Sprintf("image.pullPolicy=%s", i.chartArgs.imagePullPolicy))
 	}
-	// TODO(wonderflow) values here could give more arguments in command line
+	valuesConfig = append(valuesConfig, i.chartArgs.more...)
```
ryanzhang-oss, line 230: Do you want to put any checks on the format "xxx=xx"?

wonderflow: will not, I'm preparing to move it out, and deprecate `vela install`

ryanzhang-oss (approved):

PR #937, opened 2021-01-26
allow multiple outputs for workloaddefintion

### LINKED ISSUE #903: [Feature] Support multiple resources outputs in cue template for workloadDefinition
Now our TraitDefinition can support output multiple K8s resources by define `outputs` in TraitDefinition. But we don't support this way in workloadDefinition. The main reason is we convert the resource in workloaddefinition into Component which can only hold only one K8s object.

We can support this feature after  #867

wonderflow:
1. trait name should not contain dummy when traitdefinition not found 
Before change, it was `xxx-dummy-xxx`, after change, it was `xxx-trait-xxx`, and the existing instance won't be affected.

image

2. allow multiple outputs for workloaddefintion, which means a workload can contain deployment+service

PLEASE DON'T SQUASH THE COMMITS.

fix #903


pkg/dsl/definition/template.go
```diff
@@ -29,6 +29,12 @@ const (
 	PatchFieldName = "patch"
 )
 
+const (
+	// ExtraWorkloadObj defines the extra workload obj from a workloadDefinition,
+	// e.g. a workload composed by deployment and service, the service will be marked as ExtraWorkloadObj
+	ExtraWorkloadObj = "ExtraWorkloadObj"
```
zzxwill, line 35: How about we using `CompoundWorkloadObj`?

wonderflow: No， ExtraWorkloadObj is better

ryanzhang-oss: How about "AuxilaryWorkload", extra makes it sound like unnecessary

wonderflow: AuxiliaryWorkload is good!

pkg/dsl/definition/template.go
```diff
@@ -99,6 +105,24 @@ func (wd *workloadDef) Complete(ctx process.Context) error {
 			return errors.WithMessagef(err, "workloadDef %s new base", wd.name)
 		}
 		ctx.SetBase(base)
+
+		// we will support outputs for workload composition, and it will become trait in AppConfig.
+		outputs := inst.Lookup(OutputsFieldName)
```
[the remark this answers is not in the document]
wonderflow: outputs defines multiple object

pkg/dsl/definition/template.go
```diff
@@ -99,6 +105,24 @@ func (wd *workloadDef) Complete(ctx process.Context) error {
...
+			for i := 0; i < st.Len(); i++ {
+				fieldInfo := st.Field(i)
+				if fieldInfo.IsDefinition || fieldInfo.IsHidden || fieldInfo.IsOptional {
+					continue
+				}
+				other, err := model.NewOther(fieldInfo.Value)
+				if err != nil {
+					return errors.WithMessagef(err, "parse WorkloadDefinition %s outputs(%s)", wd.name, fieldInfo.Name)
+				}
+				// extra workload CR object will not have type
```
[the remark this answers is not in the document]
wonderflow: It's type is AuxiliaryWorkload

pkg/dsl/definition/template.go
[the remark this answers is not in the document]
ryanzhang-oss: maybe consider using some keywords like "companions" or even "traits"

ryanzhang-oss (approved):

PR #943, opened 2021-01-26
validating webhook for TraitDefinition

### LINKED ISSUE #907: traitdefinition without definitionRef field should work
Currently, we have already supported patch trait which will not really create resource.

It means the patch trait don't rely on definitionRef to refer to a CRD name or anything.

Currently, we just make the field spec.definitionRef of TraitDefinition optional, but it's really a required field if the trait is not a patch trait. We don't have any test or logic that can make sure this won't broke if we define in a wrong way.  It's dangerous.

Actions to fix this issue should contain:

1. Add TraitDefinition webhook logic, any traitdefinition without defintionRef field must meet one of the requirements:
    - it's a patch trait without output
    - it's output already have group/version/kind(GVK) that don't rely on definitionRef.
2. make sure our controller don't rely on definitionRef, make sure it will not broke if the trait doesn't have GVK defined in cue template.
3. Add unit-test or e2e test that can cover these cases.
4. Add/Fix related docs, some examples are wrong.

captainroy-hy:
related to #907 
Add a validating webhook for TraitDefintion, and a validator:
TraitDefinition without `.spec.definitionRef` should satisfy one of below:
- it has at least one output, and all outputs must have GVK
- it has no output but has a patch
- it has a patch and outputs, and all outputs must have GVK

### DIFF
--- a/pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
@@ -1,0 +1,140 @@
+package traitdefinition
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/pkg/errors"
+
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
+	"github.com/oam-dev/kubevela/pkg/oam/discoverymapper"
+)
+
+const (
+	errValidateDefRef = "error occurs when validating definition reference"
+
+	failInfoDefRefOmitted = "if definition reference is omitted, patch or output with GVK is required"
+)
+
+var traitDefGVR = v1alpha2.SchemeGroupVersion.WithResource("traitdefinitions")
+
+// ValidatingHandler handles validation of trait definition
+type ValidatingHandler struct {
+	Client client.Client
+	Mapper discoverymapper.DiscoveryMapper
+
+	// Decoder decodes object
+	Decoder *admission.Decoder
+	// Validators validate objects
+	Validators []TraitDefValidator
+}
+
+// TraitDefValidator validate trait definition
+type TraitDefValidator interface {
+	Validate(context.Context, v1alpha2.TraitDefinition) error
+}
+
+// TraitDefValidatorFn implements TraitDefValidator
+type TraitDefValidatorFn func(context.Context, v1alpha2.TraitDefinition) error
+
+// Validate implements TraitDefValidator method
+func (fn TraitDefValidatorFn) Validate(ctx context.Context, td v1alpha2.TraitDefinition) error {
+	return fn(ctx, td)
+}
+
+var _ admission.Handler = &ValidatingHandler{}
+
+// Handle validate trait definition
+func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	obj := &v1alpha2.TraitDefinition{}
+	if req.Resource.String() != traitDefGVR.String() {
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("expect resource to be %s", traitDefGVR))
+	}
+
+	if req.Operation == admissionv1beta1.Create || req.Operation == admissionv1beta1.Update {
+		err := h.Decoder.Decode(req, obj)
+		if err != nil {
+			return admission.Errored(http.StatusBadRequest, err)
+		}
+		klog.Info("validating ", " name: ", obj.Name, " operation: ", string(req.Operation))
+		for _, validator := range h.Validators {
+			if err := validator.Validate(ctx, *obj); err != nil {
+				klog.Info("validation failed ", " name: ", obj.Name, " errMsgi: ", err.Error())
+				return admission.Denied(err.Error())
+			}
+		}
+		klog.Info("validation passed ", " name: ", obj.Name, " operation: ", string(req.Operation))
+	}
+	return admission.ValidationResponse(true, "")
+}
+
+var _ inject.Client = &ValidatingHandler{}
+
+// InjectClient injects the client into the ValidatingHandler
+func (h *ValidatingHandler) InjectClient(c client.Client) error {
+	h.Client = c
+	return nil
+}
+
+var _ admission.DecoderInjector = &ValidatingHandler{}
+
+// InjectDecoder injects the decoder into the ValidatingHandler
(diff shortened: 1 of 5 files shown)


pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	switch req.Operation { //nolint:exhaustive
+	case admissionv1beta1.Delete:
+		if len(req.OldObject.Raw) != 0 {
+			if err := h.Decoder.DecodeRaw(req.OldObject, obj); err != nil {
+				return admission.Errored(http.StatusBadRequest, err)
+			}
+		} else {
+			klog.Info("deleting Trait Definition", req.Name)
+		}
```
wonderflow, line 78: why do we need to check delete behaviar? This won't happen as you only add CREATE and  UPDATE into webhook .

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
```diff
@@ -0,0 +1,219 @@
...
+	default:
```
wonderflow, line 79:
```suggestion
	case admissionv1beta1.Create, admissionv1beta1.Update:
```

Please make it clean, you should not include `Option` event here.

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// ValidateDefinitionReference validates whether the trait definition is valid if
+// its `.spec.reference` field is unset.
+// It's valid if
+// it has at least one output, and all outputs must have GVK
+// or it has no output but has a patch
+// or it has a patch and outputs, and all outputs must have GVK
+func ValidateDefinitionReference(_ context.Context, td v1alpha2.TraitDefinition) error {
+	if td.Spec.Reference.Name == "" {
```
wonderflow, line 134:
```suggestion
	if td.Spec.Reference.Name != "" {
	  return nil
	}
```

We prefer to write clean code with less `{}`

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
```diff
@@ -0,0 +1,219 @@
...
+						return errors.Wrap(err, errDefinitionReferenceRequired)
+					}
+					// TODO(roywang) how to validate structs in loop?
+					// if definitionRef is missing, each trait in outputs must have GVK
+					if !fieldExistInStruct(s, apiVersionFieldName) ||
+						!fieldExistInStruct(s, kindFieldName) {
+						return errors.New(errGVKRequired)
+					}
+				}
+			}
```
wonderflow, line 189: Please write a common function for it, nameing `validateOutputsWithGVK`

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
```diff
@@ -0,0 +1,219 @@
...
+				oStruct, err := output.Struct()
+				if err != nil {
+					return errors.Wrap(err, errDefinitionReferenceRequired)
+				}
+				// if definitionRef is missing, output must have GVK
+				if !fieldExistInStruct(oStruct, apiVersionFieldName) ||
+					!fieldExistInStruct(oStruct, kindFieldName) {
+					return errors.New(errGVKRequired)
+				}
+			}
```
wonderflow, line 167: It should be sub-function of `validateOutputsWithGVK`

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
```diff
@@ -0,0 +1,219 @@
...
+			if !hasOutput {
+				// if definitionRef is missing and no output, a patch is required
+				patch := inst.Lookup("patch")
+				if patch.Exists() {
+					// TODO(roywang) is it necessary to validate patch furthermore?
+					continue
+				}
+				return errors.New(errDefinitionReferenceRequired)
+			}
```
wonderflow, line 198: don't need to check here.

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validator_test.go
```diff
@@ -0,0 +1,201 @@
...
+			want: errors.New(errGVKRequired),
+		},
+		"NoPatch_NoOutput": {
+			reason: "An error should be returned if definitionRef is missing",
+			template: `
+      fakefield: fakefieldvalue`,
+			want: errors.New(errDefinitionReferenceRequired),
+		},
+	}
+
```
wonderflow, line 167: please add two cases with multiple `outputs`:

1. all `outputs` are all valid
2. one is invalid and the whole to be invalid

wonderflow:
This PR can only fix partial of  #943

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validator_test.go
captainroy-hy: Fixed.

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
```diff
@@ -0,0 +1,237 @@
...
+			if f.IsDefinition || f.IsHidden || f.IsOptional {
+				continue
+			}
+			hasOutput = true
+			s, innerErr := f.Value.Struct()
+			if innerErr != nil {
+				return hasOutput, false, innerErr
+			}
+			// TODO(roywang) how to validate structs in for/if comprehension?
+			// if definitionRef is missing, each trait in outputs must have GVK
```
wonderflow, line 198: can we detect there's a for loop or if?

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validating_handler.go
```diff
@@ -0,0 +1,141 @@
...
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	"k8s.io/klog"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/pkg/errors"
```
wonderflow, line 18: merge these imports

pkg/webhook/core.oam.dev/v1alpha2/traitdefinition/validator_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"testing"
+
+	"github.com/crossplane/crossplane-runtime/pkg/test"
+	"github.com/google/go-cmp/cmp"
+	"github.com/pkg/errors"
+
+	"github.com/oam-dev/kubevela/pkg/oam/util"
+)
+
+func TestValidateListComprehension(t *testing.T) {
```
wonderflow, line 15: please remove if not used

wonderflow (approved):
Generally LGTM

PR #952, opened 2021-01-28
remove unused dependency installation 

wonderflow:
we have reduce 3rd party dependency , they will be installed as capability from registry


pkg/serverlib/capability.go
```diff
@@ -158,6 +161,25 @@ func InstallCapability(client client.Client, mapper discoverymapper.DiscoveryMap
 	return nil
 }
 
+// HackForStandardTrait will do some hack install for standard capability
+func HackForStandardTrait(tp types.Capability, client client.Client) error {
```
ryanzhang-oss, line 165: what about the other traits (autoscale that relies on Keda)?

wonderflow: They don't rely on it. Promethues is special

zzxwill: Autoscale replies on KEDA.

wonderflow: let me explain more clear:

autoscale relies on KEDA, it's installed by helm in the definition file https://github.com/oam-dev/catalog/blob/master/registry/autoscale.yaml#L17 The helm chart is running the keda instance.

Promethues is special, when we install the helm chart, we just installed an operator which will be able to create promethus instance, it needs an extra CR to create an instance for runing promethus instance. 

They are in different level.

zzxwill (approved):
LGTM

PR #954, opened 2021-01-28
fix quick start guide as we don't have route trait installed as default

wonderflow:


charts/vela-core/templates/defwithtemplate/ingress.yaml
```diff
@@ -0,0 +1,57 @@
+# Code generated by KubeVela templates. DO NOT EDIT.
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  name: ingress
```
zzxwill, line 5: It would be better if an annotation `definition.oam.dev/description` is set.

docs/examples/vela.yaml
```diff
@@ -4,5 +4,7 @@ services:
     type: webservice
     image: crccheck/hello-world
     port: 8000
-    route:
-      domain: testsvc.example.com
\ No newline at end of file
+    ingress:
+      domain: testsvc.example.com
+      http:
+        "/": 8000
```
zzxwill, line 10: I ran the AppFile, and hit the following issue.

```
➜  /Users/zhouzhengxi/Programming/golang/src/github.com/wonderflow/kubevela/docs/examples git:(fixexample) vela status first-vela-app
About:

  Name:      	first-vela-app
  Namespace: 	env-e2e-hello
  Created at:	2021-01-28 19:47:04.552585 +0800 CST
  Updated at:	2021-01-28 19:47:04.552585 +0800 CST

Services:

Health checking failed!
Error: service: cannot apply components: cannot apply trait "v1" "Service" "testsvc": cannot update object: Service "testsvc" is invalid: spec.clusterIP: Invalid value: "": field is immutable not ready
```

zzxwill: 另外，我没有看到签证书的逻辑，是在 controller 里吗？我拿的是一个老的 vela binary 运行的。

wonderflow: 那个trait错误你本地run的server端没更新吧？

wonderflow: 这个没证书，http访问

docs/en/design.md
```diff
@@ -103,13 +103,13 @@ services:
     type: webservice
     image: oamdev/testapp:v1
 
-    route:
+    ingress:
```
hongchaodeng, line 106: Why change route to ingress? Ingress is not a developer-facing abstraction.

wonderflow: we don't have route now, and it's a conflict name if we install it from cap-center, so I use ingress as a new name

hongchaodeng: Sounds good

charts/vela-core/templates/defwithtemplate/ingress.yaml
wonderflow: added

hongchaodeng (approved):
lgtm

PR #958, opened 2021-01-28
add staticcheck CI action

captainroy-hy:
- add staticcheck CI action
- add staticcheck in Makefile
- fix staticcheck issue

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -2135,8 +2135,6 @@ golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208/go.mod h1:RxMgew5VJxzue5/jJ
 golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9 h1:SQFwaSi55rU7vdNs9Yr0Z324VNlrF+0wMqRXT4St8ck=
 golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sys v0.0.0-20200826173525-f9321e4c35a6 h1:DvY3Zkh7KabQE/kfzMvYvKirSiguP9Q/veMtkYyf0o8=
-golang.org/x/sys v0.0.0-20200826173525-f9321e4c35a6 h1:DvY3Zkh7KabQE/kfzMvYvKirSiguP9Q/veMtkYyf0o8=
-golang.org/x/sys v0.0.0-20200826173525-f9321e4c35a6/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20200826173525-f9321e4c35a6/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/term v0.0.0-20201117132131-f5c789dd3221/go.mod h1:Nr5EML6q2oocZ2LXRh80K7BxOlk5/8JxuGnuhpl+muw=
 golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1 h1:v+OssWQX+hTHEmOBgwxdZxK4zHq3yOs8F9J7mk0PY8E=
--- a/pkg/appfile/addon_suit_test.go
+++ b/pkg/appfile/addon_suit_test.go
@@ -45,7 +45,7 @@ func TestAppFile(t *testing.T) {
 }
 
 var _ = BeforeSuite(func(done Done) {
-	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+	logf.SetLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(GinkgoWriter)))
 	ctx := context.Background()
 	By("bootstrapping test environment")
 	useExistCluster := false
--- a/pkg/builtin/registry/registry.go
+++ b/pkg/builtin/registry/registry.go
@@ -84,6 +84,7 @@ func newCallCtx(io util.IOStreams, data map[string]interface{}) CallCtx {
 	}
 }
 
+// Run executes tasks
 // Deprecated: Run is deprecated, you should use DoTasks is builtin package, it will automatically register all internal functions
 func Run(spec map[string]interface{}, io util.IOStreams) (map[string]interface{}, error) {
 	var (
--- a/staticcheck.conf
+++ b/staticcheck.conf
@@ -1,0 +1,8 @@
+# This is config file for staticcheck.
+# If you need to add ignored checks, pls also add explaination in comments. 
+
+checks = ["all", "-ST1000", "-ST1003", "-ST1016"]
+
+# ST1000 - Incorrect or missing package comment (non-default)
+# ST1003 – Poorly chosen identifier (non-default)
+# ST1016 – Use consistent method receiver names (non-default)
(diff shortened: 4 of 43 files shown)


go.mod
```diff
@@ -84,8 +85,6 @@ replace (
 	github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
 	github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
 	github.com/wercker/stern => github.com/oam-dev/stern v1.13.0-alpha
-	// fix build issue https://github.com/ory/dockertest/issues/208
-	golang.org/x/sys => golang.org/x/sys v0.0.0-20200826173525-f9321e4c35a6
```
captainroy-hy, line 88: I remove this replacement because the [issue ](https://github.com/ory/dockertest/issues/208) seems fixed already. And honnef.co/go/tools depends on newer version of golang.org/x/sys. 
@zzxwill  Could you please help to verify this fix would not bring breaking issues?

staticcheck.conf
```diff
@@ -0,0 +1,9 @@
+# This is config file for staticcheck.
+# If you need to add ignored checks, pls also add explaination in comments. 
+
+checks = ["all", "-ST1000", "-ST1003", "-ST1016", "-SA1019"]
+
+# ST1000 - Incorrect or missing package comment (non-default)
+# ST1003 – Poorly chosen identifier (non-default)
+# ST1016 – Use consistent method receiver names (non-default)
+# SA1019 - Using a deprecated function, variable, constant or field
```
captainroy-hy, line 9: Ignore this check because we are using several deprecated functions currently and I'm not sure how to handle them.

go.sum
```diff
@@ -2134,10 +2136,104 @@ golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208 h1:qwRHBd0NqMbJxfbotnDhm2By
...
+golang.org/x/sys v0.0.0-20200802091954-4b90ce9b60b3/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20200803210538-64077c9b5642/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20200828194041-157a740278f4/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20200905004654-be1d3432aa8f/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20200930185726-fdedc70b468f/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20201119102817-f84b799fce68/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20210119212857-b64e53b001e4 h1:myAQVi0cGEoqQVR5POX+8RR2mrocKqNN1hmeMqhX27k=
+golang.org/x/sys v0.0.0-20210119212857-b64e53b001e4/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c h1:VwygUrnw9jn88c4u8GD3rZQbqrP/tgas88tPUbBxQrk=
+golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
```
wonderflow, line 2236: why do we need to add so many libs?

staticcheck.conf
wonderflow: let‘s remove the deprecate mark and enbale the check

pkg/builtin/registry/registry.go
```diff
@@ -84,6 +84,7 @@ func newCallCtx(io util.IOStreams, data map[string]interface{}) CallCtx {
 	}
 }
 
+// Run executes tasks
 // Deprecated: Run is deprecated, you should use DoTasks is builtin package, it will automatically register all internal functions
```
wonderflow, line 88: remove the Deprecated mark is OK, but add an annotation that tell users they should import the builtin package if directly use it

go.sum
```diff
@@ -2555,6 +2651,8 @@ honnef.co/go/tools v0.0.1-2020.1.4 h1:UoveltGrhghAA7ePc+e+QYDHXrBps2PqFZiHkGR/xK
 honnef.co/go/tools v0.0.1-2020.1.4/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9vFzvIQ3k=
 honnef.co/go/tools v0.0.1-2020.1.5 h1:nI5egYTGJakVyOryqLs1cQO5dO0ksin5XXs2pspk75k=
 honnef.co/go/tools v0.0.1-2020.1.5/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9vFzvIQ3k=
+honnef.co/go/tools v0.1.1 h1:EVDuO03OCZwpV2t/tLLxPmPiomagMoBOgfPt0FM+4IY=
+honnef.co/go/tools v0.1.1/go.mod h1:NgwopIslSNH47DimFoV78dnkksY2EFtX0ajyb3K/las=
```
wonderflow, line 2655: do we really need to add it into modules?

go.mod
zzxwill: Go head:)

go.sum
captainroy-hy: that's the problem of golang.org/x/sys , because it doesn't use tag as release version, go mod use commit as its release version. Each indirect dependency on it will bring such a item in go.sum...
A workaround is to pin golang.org/x.sys to a fixed version, e.g., latest version is v0.0.0-20210124154548-22da62e12c0c. Then all indirect dependency will redirect to this version and tidy go.sum up.

pkg/builtin/registry/registry.go
captainroy-hy: got it.

staticcheck.conf
captainroy-hy: sure.

go.sum
wonderflow: > golang.org/x.sys to a fixed version

Sure, please do it

go.sum
captainroy-hy: I see. It's not necessary to add it. Let me remove it. Then I can revert everything in go.mod & go.sum to keep the minimal change. Thx!

wonderflow:
yes, please also fix the conflicts

pkg/appfile/addon_suit_test.go
```diff
@@ -45,7 +45,7 @@ func TestAppFile(t *testing.T) {
 }
 
 var _ = BeforeSuite(func(done Done) {
-	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+	logf.SetLogger(zap.New(zap.UseDevMode(true), zap.WriteTo(GinkgoWriter)))
```
wonderflow, line 48: why?

captainroy-hy: zap.LoggerTo is deprecated.

wonderflow (approved):
LGTM

Issue #922, opened 2021-01-22
[Feature] Detailed Documentation Structure for KubeVela 1.0


resouer:
For 1.0 release, I'd propose a new sidebar (doc) structure like this:

- Overview
  - [Introduction](/en/introduction.md)
  - [Getting Started](/en/quick-start.md)
  - [Concepts and Glossaries](/en/concepts.md)
- Platform Team Guide
  - [Overview](/en/platform-engineers/overview.md) *# overview*
  - Progressive Rollout
    - [ ] [AppDeployment CRD]() *# AppDeployment (with Traffic?)*
  - Application Encapsulation
    - [x] [Application CRD]() *# Application CRD and its relationship with Templates and Definitions, revisioning mechanism, labels and conventions (standard contract)**
    - [x] Managing Template Modules
      - [x] [Definition and Templates]() *# definition objects metadata (including appliestTo, etc.)*
      - [x] CUE
        - [x] [CUE Basic]() *# why CUE and a basic CUE guide*
        - [x] Workload Type
          - [x] [Basic](/en/platform-engineers/workload-type.md) *# Worker = StatefulSet/Deployment*
          - [x] [Composition]() *# Web Service = Deployment + Service*
        - [x] Trait
          - [x] [Basic](/en/platform-engineers/trait.md) *# consider replace with HPA*
          - [x] [Composition]() *# think sth else, ingress + svc is not good*
          - [x] [Patch]() *# node-affinity and sidecar, replicas is not good*
          - [x] [Advance Features]() *# health check, http task, etc.*
      - [x] Cloud Resources
        - [ ] [Terraform]()
        - [x] [Crossplane](/en/platform-engineers/cloud-services.md)
  - Defining Deployment Environment
    - [ ] [Environment CRD]() *# how to register environment (k8s cluster, SA etc)*
  - Generating API and Doc
    - [x] [API generation]()
    - [x] [Doc generation]()

Open for discussion and revise of course.

I also suggest we leave the end user doc untouched until the platform team doc is finished. Also, the platform team doc should be just doc, not tutorial. It's boring, but accurate and detailed.

Fixes: https://github.com/oam-dev/kubevela/issues/905

/cc @oam-dev/vela-maintainers

PR #960, opened 2021-01-28
Explain KubeVela

resouer:
Explain KubeVela for platform builders.

ref: #922


docs/en/platform-engineers/overview.md
```diff
@@ -1,3 +1,101 @@
...
+
+Hence, the encapsulation engine of KubeVela is designed to help to make building abstractions easy, in a highly extensible approach. 
+
+#### Build Extensible Abstraction
+
+First of all, with KubeVela, you will never create monolithic abstraction which is restricted and can't be extended. In detail, the encapsulation engine introduced a extensible app-centric model behind the abstraction, this makes the abstraction is essentially assembled by components (workload modules) and traits (operational modules), a example is like below:
```
zzxwill, line 25:
```suggestion
First of all, with KubeVela, you will never create monolithic abstraction which is restricted and can't be extended. In detail, the encapsulation engine introduced an extensible app-centric model behind the abstraction, this makes the abstraction is essentially assembled by components (workload modules) and traits (operational modules), an example is like below:
```

zzxwill (approved):
LGTM

docs/en/platform-engineers/overview.md
```diff
@@ -1,3 +1,101 @@
...
+        - name: scaler
+          properties:
+            replicas: 10
+        - name: sidecar
+          properties:
+            name: "sidecar-test"
+            image: "nginx"
+    - name: bar
+      type: aliyun-oss # component type
+      bucket: "xxxxx"
```
wonderflow, line 51:
```suggestion
      bucket: "my-bucket"
```

docs/en/platform-engineers/overview.md
```diff
@@ -1,3 +1,101 @@
...
+```
+
+Every `component` and `trait` in above abstraction is defined by platform team via `Definition` objects. For example, [`WorkloadDefinition`](https://github.com/oam-dev/kubevela/tree/master/config/samples/application#workload-definition) and [`TraitDefinition`](https://github.com/oam-dev/kubevela/tree/master/config/samples/application#scaler-trait-definition). As the end user, they only need to assemble these modules into an application. Also, if end user has any requirement, the platform team could customize module template in definitions by any time.
```
wonderflow, line 54:
```suggestion
Every `component` and `trait` in above abstraction is defined by platform team via `Definition` objects. For example, [`WorkloadDefinition`](https://github.com/oam-dev/kubevela/tree/master/config/samples/application#workload-definition) and [`TraitDefinition`](https://github.com/oam-dev/kubevela/tree/master/config/samples/application#scaler-trait-definition). As the end user, they only need to assemble these modules into an application. Also, if end user has any new requirements, the platform team could customize the module template in definitions by any time.
```

docs/en/platform-engineers/overview.md
```diff
@@ -1,3 +1,101 @@
...
+
+#### A Unified Abstraction For All
+
+KubeVela intends to support any possible module type by natural, for example `CUE`, `Terraform`, `Helm`, etc and of course by a plain `Kubernetes CRD`. In order to define modules and parameters orgnized, we also introduced a [`catalog` structure](https://github.com/oam-dev/kubevela/blob/master/design/vela-core/APIServer-Catalog.md#catalog-structure). KubeVela will load such catalog via Git repo URL.
```
wonderflow, line 58:
```suggestion
KubeVela intends to support popular module types by natural, for example `CUE`, `Terraform`, `Helm`, etc and of course with a plain `Kubernetes CRD`. In order to define modules and parameters orgnized, we also introduced a [`catalog` structure](https://github.com/oam-dev/kubevela/blob/master/design/vela-core/APIServer-Catalog.md#catalog-structure). KubeVela will load such catalog via Git repo URL.
```

docs/en/platform-engineers/overview.md
```diff
@@ -1,3 +1,101 @@
...
+|-- <package>
+```
+
+Hence, it's straightforward that you could use KubeVela to create unified abstraction that can deploy any kind of resource, including cloud services, as long as they could be encapsulated by a module and placed in the catalog above. Actually, in the `application-sample` above it defined a OSS bucket on Alibaba Cloud for the other component to consume, this is powered by Terraform module.
+
+#### No Configuration Drift
+
+Many of the existing modules today are defined by client side Infrastructure-as-Code (IaC) tools and even Kubernetes tool like Helm sits at client side as well. So in the nutshell, KubeVela encapsulation engine can just be implemented at client side which would be easier to be adopted.
+
+But client side abstractions, though light-weighted, always lead to a issue called infrastructure/configuration drift, i.e. the generated component instances are not in line with the expected configuration. This could be caused by incomplete coverage, less-than-perfect processes or emergency changes.
```
wonderflow, line 83:
```suggestion
But client side abstractions, though light-weighted, always lead to an issue called infrastructure/configuration drift, i.e. the generated component instances are not in line with the expected configuration. This could be caused by incomplete coverage, less-than-perfect processes or emergency changes.
```

docs/en/platform-engineers/overview.md
```diff
@@ -1,3 +1,101 @@
...
+
+In KubeVela, the encapsulation engine is intended to be implemented in a [Kubernetes Control Loop](https://kubernetes.io/docs/concepts/architecture/controller/). This is the key for KubeVela to eliminate the issue of configuration drifting but still keeps the simplicity and software delivery velocity enabled by IaC (and Helm) modules.
```
wonderflow, line 85: Maybe we are not following the control loop, once deployed, the underlying resouces can be manuelly changed and will not be recovered by the kubevela controller

resouer: We always go the control loop, in detail:
1. C/AC are **immutable** objects, i.e. they cannot be modified  by human (should ensure this in admission). So vela should at least reconcile delete events of these immutable objects (in case they were deleted accidently).
2. C/AC controller will track and reconcile underlying resources are up-to-date by control loop, unless the admin chose to disable this. 

It also worth to take a look at the behavior of helm operator, since we plan to support similar thing in the future https://toolkit.fluxcd.io/guides/helmreleases/

PR #963, opened 2021-01-29
further develop appDeployment controller

ryanzhang-oss:


pkg/controller/common/rollout/rollout_plan_init.go
```diff
@@ -0,0 +1,19 @@
+package rollout
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
```
zzxwill, line 7: I am curious why we choose this lib as it seems not yet widely adopted:)

wonderflow: we need to clean up the log system in the whole project.

wonderflow (approved):

PR #972, opened 2021-01-29
This PR spells out the rollout states

ryanzhang-oss:


pkg/controller/common/rollout/rollout_state.go
```diff
@@ -0,0 +1,197 @@
...
+)
+
+const invalidRollingStateTransition = "the rollout state transition from `%s` state  with `%s` is invalid"
+
+const invalidBatchRollingStateTransition = "the batch rolling state transition from `%s` state  with `%s` is invalid"
+
+// StateTransition is the center place to do rollout state transition
+// it returns an error if the transition is invalid
+// it changes the coming rollout state if it's valid
+// nolint: gocyclo
```
wonderflow, line 66: it's better to split into small functions

pkg/controller/common/rollout/rollout_state.go
```diff
@@ -0,0 +1,197 @@
...
+func StateTransition(rolloutStatus *v1alpha1.RolloutStatus, event rolloutEvent) error {
```
wonderflow, line 67:
```suggestion
func StateMachine(rolloutStatus *v1alpha1.RolloutStatus, event rolloutEvent) error {
```

What about this name?

wonderflow (approved):

pkg/controller/common/rollout/rollout_state.go
ryanzhang-oss: either way is fine.

pkg/controller/common/rollout/workloads/factory.go
```diff
@@ -0,0 +1,47 @@
...
+		rolloutSpec:    f.rolloutSpec,
+		targetWorkload: f.targetWorkload,
+	}
+
+	switch kind.Kind {
+	case "CloneSet":
+		return cloneSetCtrl
+
+	default:
+		return cloneSetCtrl
```
zzxwill, line 45: These two cases return a same controller. Is it by design?

ryanzhang-oss: yes and no, this is just a framework.

pkg/controller/common/rollout/rollout_state.go
ryanzhang-oss: done

PR #1000, opened 2021-02-03
Cloneset rollout

ryanzhang-oss:
Fleshed out more details on how to handle workload rollout.  We use Cloneset as an example.


pkg/controller/common/rollout/workloads/cloneset_controller.go
```diff
@@ -1,32 +1,199 @@
...
+	// get the cloneSet
+	workload := kruise.CloneSet{}
+	err := c.client.Get(ctx, c.workloadNamespacedName, &workload)
+	if err != nil {
+		klog.CalculateMaxSize()
+		if !apierrors.IsNotFound(err) {
+			c.rolloutStatus.RolloutRetry(err.Error())
+			c.recorder.Event(c.parentController, event.Warning("Failed to get the Cloneset", err))
+		}
+		c.rolloutStatus.RolloutFailed(err.Error())
```
wonderflow, line 194: if it's not found, why don't we create it? There can be a case that we use rollout to create or scale up.

1. create: there will no source workload, and the target workload is also not exist.
2. scale up: target workload will equal with source worklaod and only replicas diff

pkg/controller/common/rollout/workloads/cloneset_controller.go
```diff
@@ -1,32 +1,199 @@
...
+	}
+	return *c.rolloutStatus
 }
 
-// RolloutPods calculates the number of pods we can upgrade once according to the rollout spec
+// RolloutOneBatchPods calculates the number of pods we can upgrade once according to the rollout spec
 // and then set the partition accordingly
-func (c *CloneSetController) RolloutPods() (int32, error) {
-	return 0, nil
+func (c *CloneSetController) RolloutOneBatchPods(ctx context.Context) v1alpha1.RolloutStatus {
```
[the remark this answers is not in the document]
ryanzhang-oss: yes, I will definitely add unit-tests for all the cases. I haven't fleshed out all the cases yet so I will add tests after that.

pkg/controller/common/rollout/workloads/cloneset_controller.go
ryanzhang-oss: The contract between the rollout controller and the outside (either application/appDepolyment/rolloutTrait) is that the outside controller creates the resources. 

I think we can retry on the NotFound case though.

wonderflow (approved):

PR #1001, opened 2021-02-03
fix cannot sync trait without definitionRef

### LINKED ISSUE #998: trait capability without definitionRef cannot be synced locally
**Describe the bug**

`vela cap cent sync <center>` cannot sync trait capabilities that have no `definitionRef` into local.

**To Reproduce**

- remove local cap center(`https://github.com/oam-dev/catalog/tree/master/registry`) if added before 
- add cap center again `vela cap center config catalog https://github.com/oam-dev/catalog/tree/master/registry`
- sync cap `vela cap center sync catalog`
- list caps `vela cap ls`

Then we will noticed that several trait capabilities are missed, e.g., ingress

**Reason**

`vela cap center sync` will use the CRD name of cap as file name saved locally. But currently, it's allowed for trait to omit definitionRef field, namely, it will omit CRD name because CRD name is `definitionRef.name` So vela cannot save it into local correctly.

captainroy-hy:
to fix #998 
solution: use trait cap name instead of trait's CRD name as file name saved locally

because unit tests in pkg `capability` are skipped by CI, I just run those tests locally.

### DIFF
--- a/pkg/serverlib/capability.go
+++ b/pkg/serverlib/capability.go
@@ -91,9 +91,9 @@ func InstallCapability(client client.Client, mapper discoverymapper.DiscoveryMap
 	switch tp.Type {
 	case types.TypeWorkload:
 		var wd v1alpha2.WorkloadDefinition
-		workloadData, err := ioutil.ReadFile(filepath.Clean(filepath.Join(repoDir, tp.CrdName+".yaml")))
+		workloadData, err := ioutil.ReadFile(filepath.Clean(filepath.Join(repoDir, tp.Name+".yaml")))
 		if err != nil {
-			return nil
+			return err
 		}
 		if err = yaml.Unmarshal(workloadData, &wd); err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


pkg/serverlib/capability.go
```diff
@@ -119,7 +119,7 @@ func InstallCapability(client client.Client, mapper discoverymapper.DiscoveryMap
 		}
 	case types.TypeTrait:
 		var td v1alpha2.TraitDefinition
-		traitdata, err := ioutil.ReadFile(filepath.Clean(filepath.Join(repoDir, tp.CrdName+".yaml")))
+		traitdata, err := ioutil.ReadFile(filepath.Clean(filepath.Join(repoDir, tp.Name+".yaml")))
```
wonderflow, line 122: what about the workload in L94

captainroy-hy: I keep all code related to workload unchanged. So workloadDef file is still stored as CRD name as well as read.

wonderflow (approved):
LGTM

PR #1005, opened 2021-02-03
Removed prometheus and certmanager crd's

hprotzek:
Removed the obsolete Prometheus crd and replaced cert-manager crd copy with optional helm chart dependency.


wonderflow:
This PR is great!

Please add `helm dependency update` into the installation guide and our e2e CI system.

hprotzek:
@wonderflow Unfortunately this was not as straightforward as I thought. When using the `cert-manager` helm chart as dependency, the crd's are still missing, when `kubevela` tries to create the `Issuer` and it fails.
It seems, that's because the crd's from `cert-manager` are provided by normal helm templates instead in the crds folder.

We could just addthe manual install of the certmanager crds to the documentation? Or automate that with a Job and a helm `pre-install` hook, to install the certmanager crds?

wonderflow:
> Or automate that with a Job and a helm pre-install hook, to install the certmanager crds?

Is that possible? this would be the best solution. If one user try install kubevela helm chart with webhook enabled, we should check there MUST have a cert-manager installed. If not, we should fail the installation.

hprotzek:
Unfortunately the idea with the helm pre-install hook didn't work neither. It feels all my changes made it worse, so I will close this PR, maybe there are better solutions in the future.

wonderflow:
@hprotzek  It seems this PR is very important as we still conflict with the cluster which has the existing cert-manager CRD installed? 

> that's because the crd's from cert-manager are provided by normal helm templates instead in the crds folder.

Can I ask why ？ If the cert-manager is installed, the crd in its template will still be there and the vela-core shouldn't fail?

hprotzek:
@wonderflow The problem is, when cert-manager is not installed and helm tries to install it as a dependency, the cert-manager crd's are not available when kubevela tries to render the Issuer and Certificate resources. 

I had a look at the prometheus operator helm chart, as they also use cert-manager for the webhook. There they don't install cert-manager as a helm dependency, you need to install it before manually. But they are using [kube-webhook-certgen](https://hub.docker.com/r/jettech/kube-webhook-certgen) as a default (if cert-manager is not installed) to generate the self signed certs. We could use the same mechanism also for kubevela?

wonderflow:
@hprotzek  Sure I think that works for KubeVela.

wonderflow:
issue refer to #1035

hprotzek:
sorry it took so long, I will look into this today :-)

hprotzek:
@wonderflow I have now a working version without the helm crd's from Prometheus and Certmanager. 

If no Certmanager is installed, a helm hook will generate and setup a self signed cert (https://github.com/jet/kube-webhook-certgen). 

If the Certmanager is installed, then the old setup is used, but the Certmanager must be installed manually before Kubevela. 

But during the e2e tests I saw, that the cli is using the certmanager to create certs. Not sure about this functionality and if then the version without Certmanager makes sense at all?

Makefile
```diff
@@ -122,7 +122,10 @@ docker-push:
 	docker push ${IMG}
 
 e2e-setup:
-	bin/vela install --set installCertManager=true --image-pull-policy IfNotPresent --image-repo vela-core-test --image-tag $(GIT_COMMIT)
+	helm repo add jetstack https://charts.jetstack.io
+	helm repo update
+	helm upgrade --install --create-namespace --namespace cert-manager cert-manager jetstack/cert-manager --version v1.2.0  --set installCRDs=true --wait
+	helm upgrade --install --create-namespace --namespace vela-system --set image.pullPolicy=IfNotPresent --set admissionWebhooks.certManager.enabled=true --set image.repository=vela-core-test --set image.tag=$(GIT_COMMIT) --wait kubevela ./charts/vela-core
```
wonderflow, line 128: why use `helm upgrade` instead of `helm install`

Makefile
```diff
@@ -122,7 +122,10 @@ docker-push:
 	docker push ${IMG}
 
 e2e-setup:
-	bin/vela install --set installCertManager=true --image-pull-policy IfNotPresent --image-repo vela-core-test --image-tag $(GIT_COMMIT)
+	helm repo add jetstack https://charts.jetstack.io
+	helm repo update
+	helm upgrade --install --create-namespace --namespace cert-manager cert-manager jetstack/cert-manager --version v1.2.0  --set installCRDs=true --wait
```
wonderflow, line 127: I think we can use the `self signed cert` version to do e2e-test, will that work?

charts/vela-core/templates/kubevela-controller.yaml
```diff
@@ -142,16 +142,16 @@ spec:
             initialDelaySeconds: 90
             periodSeconds: 5
           volumeMounts:
-            - mountPath: {{ .Values.certificate.mountPath }}
```
wonderflow, line 145: why remove the mountPath helm value and use a hardcode one

docs/en/install.md
```diff
@@ -96,9 +96,9 @@ These steps will install KubeVela controller and its dependency.
     helm install -n vela-system kubevela kubevela/vela-core --set useWebhook=false
```
wonderflow, line 96: The above description about the default case also need to update , please describe the new mechanism a bit

wonderflow (commented):
This is a great PR, some minor comments left before merge

wonderflow:
> @wonderflow I have now a working version without the helm crd's from Prometheus and Certmanager.
> 
> If no Certmanager is installed, a helm hook will generate and setup a self signed cert (https://github.com/jet/kube-webhook-certgen).
> 
> If the Certmanager is installed, then the old setup is used, but the Certmanager must be installed manually before Kubevela.
> 
> But during the e2e tests I saw, that the cli is using the certmanager to create certs. Not sure about this functionality and if then the version without Certmanager makes sense at all?

You can change the e2e test to use the `self signed certificate` instead of cert-manager

Makefile
hprotzek: Sorry for the late reply, was on vacation :-) Running helm upgrade with the option --install will do a normal install if the release is missing, if it's already existing, it will upgrade it. This comes handy when you test and run in locally and also works fine in a fresh ci env.

Makefile
hprotzek: I tried, but one of the e2e tests failed, when creating an environment and it tries to generate a certificate, expecting an existing `Issuer`.

charts/vela-core/templates/kubevela-controller.yaml
hprotzek: I didn't see any case, to set the mount path manually, but I can re-add it again.

Makefile
wonderflow: cool, that's good

Makefile
wonderflow: OK, maybe we can refactor that in the future

charts/vela-core/templates/kubevela-controller.yaml
wonderflow: thanks, please add it back,

wonderflow (approved):
Thanks for your great work,  LGTM after the doc updated

charts/vela-core/templates/kubevela-controller.yaml
```diff
@@ -107,10 +107,10 @@ spec:
           args:
             - "--metrics-addr=:8080"
             - "--enable-leader-election"
-            {{ if .Values.useWebhook }}
+            {{ if .Values.admissionWebhooks.enabled }}
             - "--use-webhook=true"
             - "--webhook-port={{ .Values.webhookService.port }}"
-            - "--webhook-cert-dir={{ .Values.certificate.mountPath }}"
+            - "--webhook-cert-dir=/etc/k8s-webhook-certs"
```
ryanzhang-oss, line 113: why hard code this?

wonderflow: yes, please add back, thanks, we need it in inner use case

wonderflow:
@hprotzek  Hi, can you help fix the final comment so we can merge this PR, thanks

wonderflow:
Great Job

PR #1006, opened 2021-02-03
Add latest release tag to be the prefix of latest version of Chart

zzxwill:
Currently the latest chart is vela-core:latest, which could
not be supported by Artifacthub, add latest tag and make the
lastest chart version as `vx.y.z-latest`

### DIFF
--- a/.github/workflows/registry.yml
+++ b/.github/workflows/registry.yml
@@ -71,6 +71,11 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@master
+      - uses: oprypin/find-latest-tag@v1
+          with:
+            repository: oam-dev/kubevela
+            releases-only: true
+          id: latest_tag
       - name: Get the version
         id: get_version
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/registry.yml
```diff
@@ -71,14 +71,19 @@ jobs:
...
+          id: latest_tag
       - name: Get the version
         id: get_version
         run: |
           VERSION=${GITHUB_REF#refs/tags/}
           if [[ ${GITHUB_REF} == "refs/heads/master" ]]; then
             VERSION=latest
           fi
-          echo ::set-output name=VERSION::${VERSION}
+          echo ::set-output name=VERSION::${{ steps.latest_tag.outputs.tag }}-${VERSION}
```
wonderflow, line 86: revert this line

wonderflow: we should add a new Variable for this case. The old version(latest) we still need for image.

zzxwill: Fixed.

PR #1012, opened 2021-02-04
remove redundant image push operation

dylandee:
since image push operation is done here(https://github.com/oam-dev/kubevela/blob/a761e75c01b9305fe616f32d2646131ab6986fe4/pkg/builtin/build/build.go#L39)

it seems redudant to push image in image build function


wonderflow (approved):
Cool， good catch

PR #1014, opened 2021-02-04
move template out from extension to spec.template

wonderflow:
fix https://github.com/oam-dev/kubevela/issues/976

I still keep the compatibility for extension.template, so the existing definition file won't fail.


config/samples/app-with-status/template.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: core.oam.dev/v1alpha2
 kind: WorkloadDefinition
 metadata:
-  name: worker
+  name: nworker
```
zzxwill, line 4: typo?

zzxwill (approved):
LGTM

PR #1015, opened 2021-02-04
fix: update CR's status with retry.RetryOnConflict

dylandee:
It's would be better to update CR's status with retry.RetryOnConflict to avoid the conflicts caused by other code making unrelated updates to the resource at the same.

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/component.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/component.go
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -191,7 +192,7 @@ func (c *ComponentHandler) createControllerRevision(mt metav1.Object, obj runtim
 		return nil, false
 	}
 
-	err = c.Client.Status().Update(context.Background(), comp)
+	err = c.UpdateStatus(context.Background(), comp)
 	if err != nil {
 		c.Logger.Info(fmt.Sprintf("update component status latestRevision %s err %v", revisionName, err), "componentName", mt.GetName())
 		return nil, false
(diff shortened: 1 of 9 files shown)


wonderflow:
Can you describe more detail about:

> the conflicts caused by other code making unrelated updates to the resource at the same.

dylandee:
> Can you describe more detail about:
> 
> > the conflicts caused by other code making unrelated updates to the resource at the same.
> 
> Update with retry blindly will gloss over problems some of which we should notice and fix.

As what is used in client-go examples(https://github.com/kubernetes/client-go/blob/93ce9718ffcde5a4465be99b1e0238a3a042ad22/examples/create-update-delete-deployment/main.go#L126)

this function is  used to retry status update only if CR's version is not the latest and other errors wont be missed if raised

Waht's more, this kind of function is widely used like in OpenKruise
https://github.com/openkruise/kruise/blob/faa8045ae74b1f2d57a173ea22fa1439580580e7/pkg/controller/advancedcronjob/advancedcronjob_controller.go#L165

wonderflow:
Yes, the example in openkruise makes sense, can you help make the function to be a method of those reconciler and name it like update<XXX>Status?

I don't want the function to be abused

dylandee:
> Yes, the example in openkruise makes sense, can you help make the function to be a method of those reconciler and name it like updateStatus?
> 
> I don't want the function to be abused

I have tried to update status with one function at first time, but i find the function would be complex since i have to use 'switch...case...' to distinguish different CR's struct. 
Is it better to define one status update function in every  reconciler file?

wonderflow:
> Is it better to define one status update function in every reconciler file?

Yes, of course. You could write one for each reconciler

dylandee:
> Yes, the example in openkruise makes sense, can you help make the function to be a method of those reconciler and name it like updateStatus?
> 
> I don't want the function to be abused

Done.

wonderflow (approved):
Great Job!  LGTM

ping @ryanzhang-oss @hongchaodeng  @captainroy-hy  would you like to have a second review?
The status update function widely changed our old way, I'll leave it until next Monday before merge.

captainroy-hy (approved):
LGTM.

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/component.go
```diff
@@ -277,6 +278,18 @@ func (c *ComponentHandler) cleanupControllerRevision(curComp *v1alpha2.Component
 	return nil
 }
 
+// UpdateStatus updates v1alpha2.Component's Status with retry.RetryOnConflict
+func (c *ComponentHandler) UpdateStatus(ctx context.Context, comp *v1alpha2.Component, opts ...client.UpdateOption) error {
+	status := comp.DeepCopy().Status
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() (err error) {
```
ryanzhang-oss, line 284: what exactly does this function do?

pkg/controller/core.oam.dev/v1alpha2/core/scopes/healthscope/healthscope_controller.go
```diff
@@ -257,3 +259,15 @@ func (r *Reconciler) GetScopeHealthStatus(ctx context.Context, healthScope *v1al
 
 	return scopeCondition, workloadHealthConditions
 }
+
+// UpdateStatus updates v1alpha2.HealthScope's Status with retry.RetryOnConflict
+func (r *Reconciler) UpdateStatus(ctx context.Context, hs *v1alpha2.HealthScope, opts ...client.UpdateOption) error {
```
ryanzhang-oss, line 264: is there a way to not repeat this over and over?

wonderflow: That's because Go don't have the generic.

ryanzhang-oss: we could use a common function

PR #1016, opened 2021-02-04
improvement: make 'RolloutReconcileWaitTime' configurable

dylandee:
would its be better to make 'RolloutReconcileWaitTime' configurable?


wonderflow:
thanks for your contributions! seems it's not used any where, closing, feel free to reopen if you want to improve it.

PR #1027, opened 2021-02-05
add more docs about apllication and CUE

wonderflow:


zzxwill (approved):
LGTM

hongchaodeng (commented):
The doc generally looks good.
One nit is that the doc reads like technical reasoning. I think this doc is for Vela users, right? If so I would suggest weakening the reasoning part and enhance the hands-on experience.

PR #1031, opened 2021-02-07
add webhooks to the rollout plan

ryanzhang-oss:
1. Refactor and enhance the webhook lib
2. Add the webhook logic to the rollout plan controller


pkg/controller/common/rollout/workloads/cloneset_controller.go
```diff
@@ -58,8 +58,10 @@ func (c *CloneSetController) Size(ctx context.Context) (int32, error) {
 }
 
 // Verify verifies that the target rollout resource is consistent with the rollout spec
-func (c *CloneSetController) Verify(ctx context.Context) *v1alpha1.RolloutStatus {
+func (c *CloneSetController) Verify(ctx context.Context) (status *v1alpha1.RolloutStatus) {
```
wonderflow, line 61: for my experiense, use a named return argument is anti-partten.

1. reviewer can hardly understand what value is really returned, they have to check code all the time.
2. it's easier to write bugs because you will ignore the return without variable.

ryanzhang-oss: https://golang.org/doc/effective_go.html#named-results

I actually use the named results to make the code clean and easier to check. I feel that there are too many places that return value. Instead of checking the return value at every place, we can check the returned value at only one place.   

I wonder if that makes sense?

cmd/core/main.go
```diff
@@ -235,7 +235,7 @@ func registerHealthChecks(mgr ctrl.Manager) error {
 	if err := mgr.AddReadyzCheck("ping", healthz.Ping); err != nil {
 		return err
 	}
-
+	// TODO: change the health check to be different from readiness check
```
wonderflow, line 238: I think it could be a good first issue if you have more detailed design about how do we define the controller is readiness.

pkg/controller/common/rollout/rollout_webhook.go
```diff
@@ -4,94 +4,107 @@ import (
...
-	ctx, cancel := context.WithTimeout(req.Context(), t)
-	defer cancel()
-
-	r, err := http.DefaultClient.Do(req.WithContext(ctx))
+	// issue request with retry
+	var r *http.Response
+	var body []byte
+	err = retry.OnError(retry.DefaultBackoff,
+		func(error) bool {
+			// not sure what not to retry on
```
wonderflow, line 42: maybe the error code 500 that we should not retry on.

pkg/controller/common/rollout/workloads/cloneset_controller.go
wonderflow: This works when error return is always the same. When the named return values differ on different errors, it can be very hard to maintain.

Anyway, I'm fine with this, but the named-results always need to be reviewed cautiously. I'm afrait this is not a good start.

ryanzhang-oss: The error only depends on the returned status, thus I put that logic at the only exit point instead of every return point.

cmd/core/main.go
ryanzhang-oss: good idea

pkg/controller/common/rollout/rollout_webhook.go
ryanzhang-oss: I think we need to retry 5xx but not 4xx.  That's not determined here though.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller_test.go
```diff
@@ -36,18 +36,18 @@ import (
...
 
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
+	"github.com/oam-dev/kubevela/pkg/controller/common"
 	"github.com/oam-dev/kubevela/pkg/oam"
 	"github.com/oam-dev/kubevela/pkg/oam/util"
 )
 
 var _ = Describe("Test Application Controller", func() {
-	ctx := context.Background()
+	ctx := context.TODO()
```
wonderflow, line 50: what's the difference？

pkg/webhook/common/rollout/rollout_plan.go
```diff
@@ -12,7 +12,7 @@ import (
 func DefaultRolloutPlan(rollout *v1alpha1.RolloutPlan) {
 	if rollout.TargetSize != nil && rollout.NumBatches != nil && rollout.RolloutBatches == nil {
 		// create the rollout  batch based on the total size and num batches if it's not set
-		// leave it for the validator to reject if they are both set
+		// leave it for the validator to valiate if they are both set
```
wonderflow, line 15:
```suggestion
		// leave it for the validator to validate if they are both set
```

wonderflow (approved):
Great Job

pkg/controller/core.oam.dev/v1alpha2/application/application_controller_test.go
ryanzhang-oss: TODO is the recommended way to add context back (more detail in the TODO()'s source code comment)

pkg/controller/common/util.go
```diff
@@ -0,0 +1,60 @@
...
+func ExtractComponentName(revisionName string) string {
+	splits := strings.Split(revisionName, "-")
+	return strings.Join(splits[0:len(splits)-1], "-")
+}
+
+// CompareWithRevision compares a component's spec with the component's latest revision content
+func CompareWithRevision(ctx context.Context, c client.Client, logger logging.Logger, componentName, nameSpace,
+	latestRevision string, curCompSpec *v1alpha2.ComponentSpec) (bool, error) {
+	oldRev := &v1.ControllerRevision{}
+	// retry on NotFound since we update the component last revision first
```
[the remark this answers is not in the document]
ryanzhang-oss: this is related to the order of the "update component status" before "create a new revision" change which is the right way to update.  In that case, Not found is expected if the component's status revision is ahead of the real revision creation. 

 I reverted that change since it broke too many tests.   The old order can lead to permanent new revision creation failure unless we handle the "already created" error with special logic.

PR #1036, opened 2021-02-07
quick fix unstable unit test

captainroy-hy:

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_trait_test.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_trait_test.go
@@ -355,13 +355,19 @@ spec:
 				}
 				return appConfig.GetGeneration()
 			}, time.Second, 300*time.Millisecond).Should(Equal(int64(2)))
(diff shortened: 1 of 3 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_trait_test.go
```diff
@@ -355,13 +355,19 @@ spec:
...
-			By("Reconcile")
-			reconcileRetry(reconciler, req)
 
-			changedTrait = unstructured.Unstructured{}
-			changedTrait.SetAPIVersion("example.com/v1")
-			changedTrait.SetKind("Bar")
-			Expect(k8sClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: traitName}, &changedTrait)).Should(Succeed())
+			Eventually(func() int64 {
+				By("Reconcile")
+				reconcileRetry(reconciler, req)
```
wonderflow, line 361: why do we need to reconcile every time? what will be wrong if we only reconcile once? The above has the same question

wonderflow (approved):
LGTM

PR #1041, opened 2021-02-08
fix unstable unit test

captainroy-hy:
Fix :
- combine reconcile trigger and the following validation into one gomega.Eventually 
- extend retry interval of gomega.Eventually 

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/dependency_test.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/dependency_test.go
@@ -98,8 +98,23 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
 	AfterEach(func() {
 		logf.Log.Info("Clean up resources")
 		// delete the namespace with all its resources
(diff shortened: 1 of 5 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/dependency_test.go
```diff
@@ -98,8 +98,23 @@ var _ = Describe("Resource Dependency in an ApplicationConfiguration", func() {
...
+		Expect(k8sClient.DeleteAllOf(ctx, foo, client.InNamespace(namespace))).Should(Succeed())
+		Eventually(func() bool {
+			l := &unstructured.UnstructuredList{}
+			l.SetAPIVersion("example.com/v1")
+			l.SetKind("Foo")
+			if err := k8sClient.List(ctx, l, client.InNamespace(namespace)); err != nil {
+				return false
+			}
+			return len(l.Items) == 0
+		}, 3*time.Second, time.Second).Should(BeTrue())
```
captainroy-hy, line 117: Why make this change? Move "cleaning up resource" operation into AfterEach to ensure the resources created by each ginkgo.It spec can be deleted even the spec failed. Otherwise, residual resources may block other cases.

wonderflow (approved):

PR #1046, opened 2021-02-09
Revise the doc site and app crd doc

resouer:
Revise the whole doc site and updated App CRD doc.


resouer:
I left several TBD in the app crd doc which needs e2e sample instead of long narrative. /cc @wonderflow

wonderflow (approved):
great job

PR #1076, opened 2021-02-19
Use context.output instead of context.input to refer to the output object defined in the WorkloadDefinition

leejanee:
use context.output instead of context.input to refer to the output object defined in the WorkloadDefinition

use case
```
kind: WorkloadDefinition
metadata:
  name: worker
spec:
  definitionRef:
    name: deployments.apps
  extension:
    template: |
      output: {
      	apiVersion: "apps/v1"
      	kind:       "Deployment"
      	spec: {
      		selector: matchLabels: {
      		}

      		template: {
      			spec: {
      				containers: [{
      					name:  context.name
                                         ports:[{containerPort: 443}]
      				}]
      			}
      		}
      	}
      }
---
kind: TraitDefinition
metadata:
  name: service
spec:
  extension:
    template: |
      parameter: {
        exposePort: int
      }
      outputs: service: {
        apiVersion: "v1"
        kind: "Service"
        spec: {
          selector:
            app: context.name
          ports: [
            {
              port: paramter.exposePort
              // Previously used context.input
              targetPort: context.output.spec.template.spec.containers[0].ports[0].containerPort
            }
          ]
        }
      }
```

### DIFF
--- a/pkg/dsl/process/handle_test.go
+++ b/pkg/dsl/process/handle_test.go
@@ -41,7 +41,7 @@ image: "myserver"
 	myAppName, err := ctxInst.Lookup("context", "appName").String()
 	assert.Equal(t, nil, err)
 	assert.Equal(t, "myapp", myAppName)
-	inputJs, err := ctxInst.Lookup("context", "input").MarshalJSON()
+	inputJs, err := ctxInst.Lookup("context", "output").MarshalJSON()
 	assert.Equal(t, nil, err)
 	assert.Equal(t, `{"image":"myserver"}`, string(inputJs))
 }
--- a/pkg/dsl/process/handle_test.go
+++ b/pkg/dsl/process/handle_test.go
@@ -34,14 +34,14 @@ image: "myserver"
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/dsl/process/handle_test.go
```diff
@@ -41,7 +41,7 @@ image: "myserver"
 	myAppName, err := ctxInst.Lookup("context", "appName").String()
 	assert.Equal(t, nil, err)
 	assert.Equal(t, "myapp", myAppName)
-	inputJs, err := ctxInst.Lookup("context", "input").MarshalJSON()
+	inputJs, err := ctxInst.Lookup("context", "output").MarshalJSON()
```
wonderflow, line 44: please use a const variable

wonderflow: by the way, please add a test for it. It seems we change from `input` to `output` but no test broken. That could be hard to maintain

wonderflow: OutputFieldName

wonderflow (approved):
LGTM

PR #1077, opened 2021-02-19
Use "context.outputs" to reference auxiliary objects defined in the definition template

leejanee:
use "context.outputs" to reference auxiliary objects defined in the definition template
reference issue: [context.output refer to workload output  object](https://github.com/oam-dev/kubevela/pull/1076)

Implement  refer to objects defined in the definition template also achieves the purpose of
 [support simple data passing from workload to trait](https://github.com/oam-dev/kubevela/issues/1026)

fix https://github.com/oam-dev/kubevela/issues/1026


wonderflow (commented):
Good job。

1. Please add doc for this great feature, maybe add an `advance` section in  https://github.com/oam-dev/kubevela/tree/master/docs/en/cue
2. Please add the use case in issue as a more detailed test case

PR #1078, opened 2021-02-20
improvement: reduce io operation when load appfile

dylandee:


pkg/serverlib/application.go
```diff
@@ -3,6 +3,7 @@ package serverlib
 import (
 	"bytes"
 	"context"
+	j "encoding/json"
```
wonderflow, line 6: I think `json` is better than `j` here

pkg/serverlib/application.go
```diff
@@ -327,20 +329,34 @@ func GetServicesWhenDescribingApplication(cmd *cobra.Command, app *api.Applicati
...
+	switch ext {
+	case ".json":
 		dest = "vela.json"
-	} else if ext == ".yaml" || ext == ".yml" {
+		af, err = api.JSONToYaml(body, af)
+	case ".yaml", ".yml":
+		dest = "vela.yaml"
+		err = yaml.Unmarshal(body, af)
+	default:
 		dest = "vela.yaml"
```
wonderflow, line 348: Does it should be `Appfile` ?

pkg/serverlib/application.go
```diff
@@ -327,20 +329,33 @@ func GetServicesWhenDescribingApplication(cmd *cobra.Command, app *api.Applicati
 	return targetServices, nil
 }
 
-func saveRemoteAppfile(url string) (string, error) {
+func saveAndLoadRemoteAppfile(url string) (string, *api.AppFile, error) {
```
wonderflow, line 332: if we already return the AppFile, we don't need the filePath as return value

wonderflow (approved):
LGTM after the import header changed

dylandee (commented):
> LGTM after the import header changed

'json' is used below for package '"k8s.io/apimachinery/pkg/runtime/serializer/json"'

what about 'json2' or 'buildinjson' ?

wonderflow:
> > LGTM after the import header changed
> 
> 'json' is used below for package '"k8s.io/apimachinery/pkg/runtime/serializer/json"'
> 
> what about 'json2' or 'buildinjson' ?

I didn't see it was occupied in this file.

dylandee:
> > > LGTM after the import header changed
> > 
> > 
> > 'json' is used below for package '"k8s.io/apimachinery/pkg/runtime/serializer/json"'
> > what about 'json2' or 'buildinjson' ?
> 
> I didn't see it was occupied in this file.

Im confused..Are we discussing the same thing? 😵
image

wonderflow:
@dylandee Oh, sorry, I made the mistake

PR #1081, opened 2021-02-20
remove go-header package dependency

wonderflow:


ryanzhang-oss (approved):

go.mod
```diff
@@ -39,7 +40,6 @@ require (
 	github.com/onsi/ginkgo v1.13.0
 	github.com/onsi/gomega v1.10.3
 	github.com/openkruise/kruise-api v0.7.0
-	github.com/openservicemesh/osm v0.3.0
```
ryanzhang-oss, line 42: what's the security issue in the OSM?

ryanzhang-oss (approved):

go.mod
wonderflow: osm rely on go-header

PR #1087, opened 2021-02-22
Use "context.outputs" to reference auxiliary objects defined in the definition template

leejanee:
use "context.outputs" to reference auxiliary objects defined in the definition template
reference issue: [context.output refer to workload output  object](https://github.com/oam-dev/kubevela/pull/1076)

Implement  refer to objects defined in the definition template also achieves the purpose of
 [support simple data passing from workload to trait](https://github.com/oam-dev/kubevela/issues/1026)

fix https://github.com/oam-dev/kubevela/issues/1026

### DIFF
--- a/docs/en/cue/trait.md
+++ b/docs/en/cue/trait.md
@@ -349,6 +349,9 @@ spec:
 
 The trait can use the data of workload output and outputs to fill itself.
 
+There are two keywords `output` and `outputs` in the rendering context.
+You can use `context.output` refer to the output-object, and use `context.outputs.<xx>` refer to the outputs-object.
+
 Below is an example
 1. the main workload object(Deployment) in this example will render into the context.output before rendering traits.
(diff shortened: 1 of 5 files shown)


wonderflow (approved):
LGTM ping @ryanzhang-oss please  review

docs/en/cue/trait.md
```diff
@@ -345,6 +345,126 @@ spec:
     }
 ```
 
+## Simple data passing
+
+The trait can use the data of workload output and outputs to fill itself.
+
+There are two keywords `output` and `outputs` in the rendering context.
+You can use `context.output` refer to the output-object, and use `context.outputs.<xx>` refer to the outputs-object.
```
wonderflow, line 353:
```suggestion
You can use `context.output` refer to the workload object, and use `context.outputs.<xx>` refer to the trait object.
please make sure the trait resource name is unique, or the former data will be covered by the latter one.
```

PR #1088, opened 2021-02-22
improvement: specify 'vela exec' error hint

dylandee:
I am used to use command ’kubectl exec ${pod} -c ${container}  ${command}‘
But i found error hint in 'vela exec' is not specified.

Shall we make it more clear?


wonderflow (approved):
Good catch

PR #1091, opened 2021-02-23
render the workload and trait name differently

ryanzhang-oss:
Special treatment for names in appConfig generated by an Application
1. in-place upgradable workload name will be component 
2. otherwise, it will be componentRevision Name
3. We ban name in the metadata field

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/render.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/render.go
@@ -27,6 +27,7 @@ import (
 	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	kruise "github.com/openkruise/kruise-api/apps/v1alpha1"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/render.go
```diff
@@ -342,6 +322,30 @@ func isRevisionEnabled(traitDefs []v1alpha2.TraitDefinition) bool {
...
+// setAppWorkloadInstanceName sets the name of the workload instance depends on the component revision
+// and the workload kind
+func setAppWorkloadInstanceName(componentName string, w *unstructured.Unstructured, revision int) {
+	// TODO: we can get the workloadDefinition name from w.GetLabels()["oam.WorkloadTypeLabel"]
+	// and use a special field like "use-inplace-upgrade" in the definition to allow configurable behavior
+
+	// we hard code the behavior depends on the workload group/kind for now. The only in-place upgradable resources
+	// we support is cloneset for now. We can easily add more later.
+	if w.GroupVersionKind().Group == kruise.GroupVersion.Group {
+		if w.GetKind() == reflect.TypeOf(kruise.CloneSet{}).Name() {
```
zzxwill, line 334: Cool.

wonderflow, line 334: also consider statefulset

wonderflow (approved):
good job

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/render.go
ryanzhang-oss: I was going to but not 100% sure if statefulset should be upgraded in-place

PR #1092, opened 2021-02-23
output not allowed in traitdefinition, should use outputs

### LINKED ISSUE #1023: only outputs allowed in trait template

Now we have supported output in CUE template for both workload and trait definition. While the output and outputs in trait definition can be the same, it's confusing.

For example, in this case, the `output` will create a `Service` object, while the `outputs` will create a `Ingress`. They are almost the same, but with different `keyword`.

Another reason is the `output` can be a keyword of workload output if no traits can use it. So I propose we deprecate the output field in trait side, and only allow using `outputs` in the trait template.

Before:

```
kind: TraitDefinition
metadata:
  name: ingress
spec:
  appliesToWorkloads:
    - webservice
    - worker
  extension:
    template: |
      parameter: {
      	domain: string
      	http: [string]: int
      }
      
      // trait template can have multiple outputs in one trait
      output: {
      	apiVersion: "v1"
      	kind:       "Service"
      	metadata:
      		name: context.name
      	spec: {
      		selector:
      		ports: [
      			for k, v in parameter.http {
      				port:       v
      				targetPort: v
      			},
      		]
      	}
      }
      
      outputs: ingress: {
      	apiVersion: "networking.k8s.io/v1beta1"
      	kind:       "Ingress"
      	metadata:
      		name: context.name
      	spec: {
      		rules: [{
      			host: parameter.domain
      			http: {
      				paths: [
      					for k, v in parameter.http {
      						path: k
      						backend: {
      							serviceName: context.name
      							servicePort: v
      						}
      					},
      				]
      			}
      		}]
      	}
      }
```

```
# Code generated by KubeVela templates. DO NOT EDIT.
kind: TraitDefinition
metadata:
  annotations:
    definition.oam.dev/description: "Configures K8s ingress and service to enable web traffic for your service.
    Please use route trait in cap center for advanced usage."
  name: ingress
spec:
  status:
    customStatus: |-
      if len(context.outputs.ingress.status.loadBalancer.ingress) > 0 {
      	message: "Visiting URL: " + context.outputs.ingress.spec.rules[0].host + ", IP: " + context.outputs.ingress.status.loadBalancer.ingress[0].ip
      }
      if len(context.outputs.ingress.status.loadBalancer.ingress) == 0 {
      	message: "No loadBalancer found, visiting by using 'vela port-forward " + context.appName + " --route'\n"
      }
    healthPolicy: |
      isHealth: len(context.outputs.service.spec.clusterIP) > 0
  appliesToWorkloads:
    - webservice
    - worker
  extension:
    template: |
      parameter: {
      	domain: string
      	http: [string]: int
      }
      
      // trait template can have multiple outputs in one trait
      outputs: service: {
      	apiVersion: "v1"
      	kind:       "Service"
      	metadata:
      		name: context.name
      	spec: {
      		selector:
      		ports: [
      			for k, v in parameter.http {
      				port:       v
      				targetPort: v
      			},
      		]
      	}
      }
      
      outputs: ingress: {
      	apiVersion: "networking.k8s.io/v1beta1"
      	kind:       "Ingress"
      	metadata:
      		name: context.name
      	spec: {
      		rules: [{
      			host: parameter.domain
      			http: {
      				paths: [
      					for k, v in parameter.http {
      						path: k
      						backend: {
      							serviceName: context.name
      							servicePort: v
      						}
      					},
      				]
      			}
      		}]
      	}
      }
      

```

**Additional context**

wonderflow:
fix #1023

### DIFF
--- a/pkg/dsl/definition/template.go
+++ b/pkg/dsl/definition/template.go
@@ -119,7 +119,7 @@ func (wd *workloadDef) Complete(ctx process.Context, abstractTemplate string) er
 			if err != nil {
 				return errors.WithMessagef(err, "invalid outputs(%s) of workload %s", fieldInfo.Name, wd.name)
 			}
-			ctx.AppendAuxiliaries(process.Auxiliary{Ins: other, Type: AuxiliaryWorkload, Name: fieldInfo.Name, IsOutputs: true})
+			ctx.AppendAuxiliaries(process.Auxiliary{Ins: other, Type: AuxiliaryWorkload, Name: fieldInfo.Name})
 		}
 	}
 	return nil
@@ -291,14 +291,6 @@ func (td *traitDef) Complete(ctx process.Context, abstractTemplate string) error
 			}
 		}
 
-		output := inst.Lookup(OutputFieldName)
-		if output.Exists() {
-			other, err := model.NewOther(output)
-			if err != nil {
-				return errors.WithMessagef(err, "invalid output of trait %s", td.name)
-			}
-			ctx.AppendAuxiliaries(process.Auxiliary{Ins: other, Type: td.name, IsOutputs: false})
-		}
 		outputs := inst.Lookup(OutputsFieldName)
 		if outputs.Exists() {
 			st, err := outputs.Struct()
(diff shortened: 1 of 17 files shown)


pkg/dsl/definition/template.go
```diff
@@ -361,11 +353,7 @@ func (td *traitDef) getTemplateContext(ctx process.Context, cli client.Reader, n
 		if err != nil {
 			return nil, err
 		}
-		if assist.IsOutputs {
-			outputs[assist.Name] = object
-		} else {
-			root[OutputFieldName] = object
-		}
+		outputs[assist.Name] = object
```
zzxwill, line 356: I wonder whether we should check there are same assist.Name. If yes, errors should be returned.

wonderflow: Yes, that should be checked in webhook, let me add a TODO here.

PR #1098, opened 2021-02-24
cherry-pick(#1094) fix apply only once observedGeneration should mark after meet all dependency requirements

wonderflow:
add log for apply only once

cherry-pick(#1094)


zzxwill (approved):
LGTM

PR #1099, opened 2021-02-24
add unit test

captainroy-hy:
add a unit test to cover: if ApplyOnceOnlyForce is enabled, for the AC with unsatisfied dependencies, it should create the workloads correctly once the dependencies are all satisfied.

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_once_only_test.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_once_only_test.go
@@ -287,6 +287,131 @@ var _ = Describe("Test apply (workloads/traits) once only", func() {
 	})
 
 	When("ApplyOnceOnlyForce is enabled", func() {
+		It("tests the situation where workload is not applied at the first because of unsatisfied dependency",
+			func() {
+				By("Enable ApplyOnceOnlyForce")
+				reconciler.applyOnceOnlyMode = core.ApplyOnceOnlyForce
+
+				tempFoo := &unstructured.Unstructured{}
+				tempFoo.SetAPIVersion("example.com/v1")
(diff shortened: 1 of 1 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/apply_once_only_test.go
```diff
@@ -287,6 +287,131 @@ var _ = Describe("Test apply (workloads/traits) once only", func() {
...
+					reconcileRetry(reconciler, reqDep)
+					acWithDep = v1alpha2.ApplicationConfiguration{}
+					if err := k8sClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: acWithDepName}, &acWithDep); err != nil {
+						return false
+					}
+					return len(acWithDep.Status.Workloads) == 1
+				}, time.Second, 300*time.Millisecond).Should(BeTrue())
+
+				// because dependency is not satisfied so the workload should not be created
+				By("Check the workload is NOT created")
```
ryanzhang-oss, line 381: We should already have tests cover this. AFAIK is the bug is related to "ApplyOnceOnlyForce" when there is an unsatisfied component. I am not sure where that is tested in this Unit test

captainroy-hy: In line 391, the component became satisfied, so it should be created. Before the bug fix, once the component is transited from unsatisfied into satisfied, it's supposed to be created but `ApplyOnceOnlyForce` still blocks creating it. After the bug fix, it can create the component correctly when it's satisfied.

wonderflow (approved):
LGTM please fix the CI

wonderflow:
please cherry-pick to master

PR #1112, opened 2021-02-27
Rollout e2e demo

ryanzhang-oss:
1. change the behavior of the target appConfig to apply once with disabled workloads on templated resource
2. skip applying until the annotation is removed to avoid conflicts with rollout controller and the source controller


pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -252,8 +248,8 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
 	// make sure this is the last functional defer function to be called
 	defer func() {
 		// Make sure if error occurs, reconcile will not happen too frequency
-		if returnErr != nil && result.RequeueAfter < shortWait {
-			result.RequeueAfter = shortWait
+		if returnErr != nil {
+			result.RequeueAfter = 0
```
wonderflow, line 252: why don't need requeue here?

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -112,32 +105,14 @@ func (h *appHandler) apply(ctx context.Context, ac *v1alpha2.ApplicationConfigur
...
-		}
-		if exist {
-			continue
-		}
-		// Component not exits in current Application, should be deleted
-		var oldC = &v1alpha2.Component{ObjectMeta: metav1.ObjectMeta{Name: comp.Name, Namespace: ac.Namespace}}
-		if err := h.r.Delete(ctx, oldC); err != nil {
-			return err
-		}
-	}
```
wonderflow, line 140: why delete?

wonderflow (approved):
Generally LGTM, with two questions left

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
ryanzhang-oss: because controller runtime has auto requeue for any error with ratelimit

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
ryanzhang-oss: no need to garbage collect any components as they all have owner reference to the application. They will be gced automatically.  GC here will make the old AC not working (comp1-> comp2 change then comp1 is deleted which leaves ac-v1 hanging)

PR #1114, opened 2021-03-01
update CRD and fix workload not create first time

wonderflow:
fix test case and fix a bug that workload was not created the first time AC created

### DIFF
--- a/apis/core.oam.dev/v1alpha2/core_types.go
+++ b/apis/core.oam.dev/v1alpha2/core_types.go
@@ -412,6 +412,16 @@ type WorkloadTrait struct {
 
 	// Message will allow controller to leave some additional information for this trait
 	Message string `json:"message,omitempty"`
+
+	// AppliedGeneration indicates the generation observed by the appConfig controller.
+	// The same field is also recorded in the annotations of traits.
+	// A trait is possible to be deleted from cluster after created.
+	// This field is useful to track the observed generation of traits after they are
+	// deleted.
+	AppliedGeneration int64 `json:"appliedGeneration,omitempty"`
+
+	// DependencyUnsatisfied notify does the trait has dependency unsatisfied
+	DependencyUnsatisfied bool `json:"dependencyUnsatisfied,omitempty"`
 }
 
 // A ScopeStatus represents the state of a scope.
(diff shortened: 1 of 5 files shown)


apis/core.oam.dev/v1alpha2/core_types.go
```diff
@@ -412,6 +412,16 @@ type WorkloadTrait struct {
...
+
+	// AppliedGeneration indicates the generation observed by the appConfig controller.
+	// The same field is also recorded in the annotations of traits.
+	// A trait is possible to be deleted from cluster after created.
+	// This field is useful to track the observed generation of traits after they are
+	// deleted.
+	AppliedGeneration int64 `json:"appliedGeneration,omitempty"`
+
+	// DependencyUnsatisfied notify does the trait has dependency unsatisfied
+	DependencyUnsatisfied bool `json:"dependencyUnsatisfied,omitempty"`
```
ryanzhang-oss, line 424: normally it's better for a bool name to be positive (DependencySatisfied) so we don't need to use double negative

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -392,8 +392,13 @@ func updateObservedGeneration(ac *v1alpha2.ApplicationConfiguration) {
 	}
 	for i, w := range ac.Status.Workloads {
 		// only the workload meet all dependency can mean the generation applied successfully
-		if w.AppliedGeneration != ac.Generation && !w.DependencyUnsatisfied {
-			ac.Status.Workloads[i].AppliedGeneration = ac.GetGeneration()
+		if w.AppliedComponentRevision != w.ComponentRevisionName && !w.DependencyUnsatisfied {
```
ryanzhang-oss, line 395: like here, it can be w.DependencySatisfied

ryanzhang-oss (approved):

apis/core.oam.dev/v1alpha2/core_types.go
wonderflow: it will be uncompatiable, the default case should be false which is satisfied

PR #1109, opened 2021-02-26
parent override child when annotation/labels conflicts && one revision will apply once only in force mode && AC.status CRD updated

wonderflow:
1) AC will override its component/workload; 
2) AC will override its trait; 
3) workload will override child resources

image

---

4) apply once only will check component revision
5) refactor CRD status of AppConfig to make workload has Dependency and appliedGeneration

PR #1115, opened 2021-03-01
cherry-pick(1109 and 1114)

wonderflow:
parent override child when annotation/labels conflicts && one revision will apply once only in force mode && AC.status CRD updated (#1109)

* when annotation/labels passthrough from parent to child conflicts, the parent will override the child.
e.g. 1) AC will override its component/workload; 2) workload will override child-resource; 3) AC will override its trait

* apply once only force will block workload apply when revision not changed even ac generation updated

cherry-pick(#1109 and #1114)

please don't merge until #1114 merged


zzxwill (approved):
LGTM

PR #1119, opened 2021-03-02
fine tune rollout experience with demo

ryanzhang-oss:
Refine rollout experience by adding status and make the rollout annotation semantic more intuitive


pkg/controller/common/rollout/rollout_webhook_test.go
```diff
@@ -0,0 +1,42 @@
...
+		wantErr    bool
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// generate a test server so we can capture and inspect the request
+			testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+				res.WriteHeader(tt.statusCode)
+				res.Write([]byte(tt.body))
+			}))
+			defer func() { testServer.Close() }()
```
wonderflow, line 35:
```suggestion
			defer testServer.Close()
```

wonderflow (approved):

PR #1120, opened 2021-03-02
Applying klog in core workloads/traits/scopes

### LINKED ISSUE #378: [Feature] Improve the logging system
The current OAM runtime logging system doesn't print out the file and line numbers, it's very cumbersome to debug and can potentially lead to wrong conclusions.

Use Klog (https://github.com/kubernetes/klog) instead. It seems that it's a very simple one to use that can printout the line numbers by default (l saw it in Kruise). It's also super straightforward to use.
```
import "k8s.io/klog"

klog.InitFlags(nil)
```

Use zap logger like flagger or EDAS.	
```
import go.uber.org/zap/zapcore
logger, err := logger.NewLoggerWithEncoding(logLevel, zapEncoding)
```
I didn't look deep into either solution so I am not sure who performs better and who has a better feature set but either is way better than the one we have now.

**Additional context**
I can take a pass to convert all our logs to the new logging system when we agree on the approach.

zzxwill:
Used more friendly klog as logging

Partly fix #378

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/core/traits/manualscalertrait/manualscalertrait_controller.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/core/traits/manualscalertrait/manualscalertrait_controller.go
@@ -24,13 +24,13 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	cpmeta "github.com/crossplane/crossplane-runtime/pkg/meta"
-	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
+	"k8s.io/klog"
 	"k8s.io/kube-openapi/pkg/util/proto"
 	"k8s.io/kubectl/pkg/explain"
 	"k8s.io/kubectl/pkg/util/openapi"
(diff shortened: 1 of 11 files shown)


pkg/controller/standard.oam.dev/v1alpha1/autoscaler/autoscaler_controller.go
```diff
@@ -87,7 +87,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 
 	// Fetch the instance to which the trait refers to
-	workload, err := util.FetchWorkload(ctx, r, log, &scaler)
+	workload, err := util.FetchWorkload(ctx, r, &scaler)
```
zzxwill, line 90: @ryanzhang-oss I need more test to verify this change.

pkg/controller/core.oam.dev/v1alpha2/core/traits/manualscalertrait/manualscalertrait_controller.go
```diff
@@ -24,13 +24,13 @@ import (
...
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	cpmeta "github.com/crossplane/crossplane-runtime/pkg/meta"
-	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
+	"k8s.io/klog"
```
ryanzhang-oss, line 33: please use k8s.io/klog/v2 which has a structured (key=value) way to do log

pkg/controller/core.oam.dev/v1alpha2/core/traits/manualscalertrait/manualscalertrait_controller.go
```diff
@@ -86,35 +84,33 @@ type Reconciler struct {
...
-	mLog.Info("Reconcile manualscalar trait")
+	klog.Info("Reconcile manualscalar trait")
 
 	var manualScalar oamv1alpha2.ManualScalerTrait
 	if err := r.Get(ctx, req.NamespacedName, &manualScalar); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
-	r.log.Info("Get the manualscalar trait", "ReplicaCount", manualScalar.Spec.ReplicaCount,
-		"Annotations", manualScalar.GetAnnotations())
+	klog.Infof("Get the manualscalar trait, ReplicaCount: %d, Annotations: %v", manualScalar.Spec.ReplicaCount,
```
ryanzhang-oss, line 93: please use structured printout  InfoS(xx)

ryanzhang-oss (changes requested):

PR #1121, opened 2021-03-02
API: make Component type|settings omitempty

hongchaodeng:
I'm importing the oamcore ApplicationComponent type for overlay patch.

But the field is required even though there is no patch for it:
```
      components:
        - name: backend
          type: "" 
          settings:
            image: busybox
        - name: frontend
          type: ""
          settings: {} 
          traits:
            - name: sidecar
              properties:
                image: "fluentd"
```

With this change it would be simplified:

```
      components:
        - name: backend
          settings:
            image: busybox
        - name: frontend
          traits:
            - name: sidecar
              properties:
                image: "fluentd"
```

### DIFF
--- a/apis/core.oam.dev/v1alpha2/application_types.go
+++ b/apis/core.oam.dev/v1alpha2/application_types.go
@@ -86,9 +86,9 @@ type ApplicationTrait struct {
 // ApplicationComponent describe the component of application
 type ApplicationComponent struct {
 	Name         string `json:"name"`
-	WorkloadType string `json:"type"`
+	WorkloadType string `json:"type,omitempty"`
 	// +kubebuilder:pruning:PreserveUnknownFields
-	Settings runtime.RawExtension `json:"settings"`
+	Settings runtime.RawExtension `json:"settings,omitempty"`
 
 	// Traits define the trait of one component, the type must be array to keep the order.
 	Traits []ApplicationTrait `json:"traits,omitempty"`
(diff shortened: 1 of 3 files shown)


ryanzhang-oss:
The WorkloadType field is not an optional field per se. So we need at least a webhook to check that if you remove the check at the CRD level.

hongchaodeng:
@ryanzhang-oss 

Fine. If they are for different purposes, I think it is better to create two types even if they have the same fields with different json tags.

> So we need at least a webhook to check that if you remove the check at the CRD level.

This is not different without this PR. Adding this PR is to omit the check.

wonderflow:
I can't very understand why the patch is related with the CRD validation? Do you mean the patch will be a normal Application CR?

hongchaodeng:
@wonderflow 

See this example: https://github.com/oam-dev/velacp/blob/main/docs/examples/app-env/setup/app_template.yaml
```yaml
apiVersion: velacp.oam.dev/v1alpha1
kind: AppTemplate
metadata:
  name: app-template-1
spec:
  template:
    components:
      - name: backend
        type: worker
        settings:
          cmd:
            - sleep
            - '1000'
      - name: frontend
        type: webservice
        settings:
          image: nginx
        traits:
          - name: autoscaler
            properties:
              min: 1
              max: 10
  patch:
    - envs:
        - env-1
      components:
        - name: backend
          type: "" # TODO: oamcore types require this. Make this optional
          settings:
            image: busybox
        - name: frontend
          type: "" # TODO: oamcore types require this. Make this optional
          settings: {} # TODO: oamcore types require this. Make this optional
          traits:
            - name: sidecar
              properties:
                image: "fluentd"
```

I'm directly importing and using ApplicationComponent for the `components` field

wonderflow:
It's `AppTemplate` not `Application`, does it because the `AppTemplate` reference the structure?

PR #1132, opened 2021-03-03
fix(api-server): application meta list reads from application crd

### LINKED ISSUE #1129: [Dashboard] Application can't be deleted
**Describe the bug**
Cannot delete the application listed on the page. Seems like there is a bug in api-server, mismatching the `app name` and `app with version name`

**To Reproduce**
image

**Expected behavior**
Delete the application correctly.

kinsolee:
Fix #1129

Since we use Application CRD to generate `ApplicationConfiguration`, the `ListApplications` method should get `ApplicationMeta` from `Application` rather than `ApplicationConfiguration`.

The dashboard will also be able to delete `Application` as expected.

### DIFF
--- a/references/common/application.go
+++ b/references/common/application.go
@@ -23,7 +23,6 @@ import (
 
 	corev1alpha2 "github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
 	"github.com/oam-dev/kubevela/apis/types"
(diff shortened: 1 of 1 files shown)


wonderflow:
@kinsolee  Hi， would you like to rebase this PR?

kinsolee:
@wonderflow  Done.

wonderflow (approved):
LGTM

PR #1135, opened 2021-03-04
Add webhook test

ryanzhang-oss:
1. Add webhook funcitonal UT

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -105,7 +105,7 @@ jobs:
         run: |
           make e2e-cleanup
           make e2e-setup
-
+          
       - name: Wait for e2e preparation ready
         run: |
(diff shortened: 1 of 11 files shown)


.github/workflows/go.yml
```diff
@@ -106,6 +106,11 @@ jobs:
           make e2e-cleanup
           make e2e-setup
 
+      - name: Helm 3
+        uses: WyriHaximus/github-action-helm3@v2.0
+        with:
+          exec: helm install kruise https://github.com/openkruise/kruise/releases/download/v0.7.0/kruise-chart.tgz
```
wonderflow, line 112: where does it be used?

wonderflow (approved):

.github/workflows/go.yml
ryanzhang-oss: I am writing an e2e test that needs to use cloneset

PR #1136, opened 2021-03-04
fix garbage collection for apply once force model

wonderflow:
- [x] 1. reduce reconcile too frequency, error will requeue automatically
- [x] 2. check annotation for revision workload
- [x] 3. only delete workload with zero replicas in apply-only-once force model here.

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
@@ -36,6 +36,7 @@ import (
 
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/event"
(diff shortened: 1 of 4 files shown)


ryanzhang-oss (approved):

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -340,6 +348,41 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
 	return reconcile.Result{RequeueAfter: waitTime}, nil
 }
 
+// confirmDeleteOnApplyOnceMode will confirm whether the workload can be delete or not in apply once only enabled mode
+// currently only workload replicas with 0 can be delete
+func (r *OAMApplicationReconciler) confirmDeleteOnApplyOnceMode(ctx context.Context, namespace string, u *unstructured.Unstructured) error {
```
Diddaa, line 353: use error control logic procedure is a bad practice IMO.

PR #1141, opened 2021-03-04
cherry-pick(#1136) fix revision enable workload check and prevent GC if replica>0

wonderflow:


pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -316,6 +317,13 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (result reco
 		log := log.WithValues("kind", e.GetKind(), "name", e.GetName())
 		record := r.record.WithAnnotations("kind", e.GetKind(), "name", e.GetName())
 
+		err := r.confirmDeleteOnApplyOnceMode(ctx, ac.GetNamespace(), &e)
+		if err != nil {
+			log.Debug("confirm component can't be garbage collected", "error", err)
+			record.Event(ac, event.Warning(reasonCannotGGComponents, err))
+			ac.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errGCComponent)))
```
ryanzhang-oss, line 324: Do we need to fail the reconcile logic just because we fail to gc? They will be gced next time.

ryanzhang-oss (approved):

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
wonderflow: it looks like the GC is blocking the reconcile because we reconstruct the status.workloads from scratch .

> 如果GC 不返回error，是不是再也不能GC了？

it shouldn't be but the way we re-create status.workload makes it this way

PR #1143, opened 2021-03-05
fix #1140 delete image after e2e test

### LINKED ISSUE #1140: delete image after e2e test
we are using a custom runner for e2e test, the disk space is easily to be full, please delete the image after the e2e test finished. It's better if we can also delete the image even the CI failed

yangsoon:
add step **cleanup** to delete image after step **e2e-test** , the **cleanup** step will always executes(even if the previous step fails).

fix #1140


wonderflow (approved):
good job

PR #1162, opened 2021-03-09
fix vela system dryrun command cannot get namespaced definition

wangyikewxgm:
quick fix vela system dryrun command cannot get namespaced definition, use env namespace for application namespace


wonderflow (approved):
LGTM

PR #1174, opened 2021-03-11
align cli to the application object

### LINKED ISSUE #877: [Feature] Refactor vela cli to interact with Application instead of AppConfig and Component
Now we have successfully built `Application` Object which has done most appfile logic in controller. Our CLI should work with the Application object instead of AppConfig and Component.

## Commands should be refactored

- [x]     init   	Create scaffold for an application
- [x]     delete APP_NAME                           	Delete an application  @zzxwill 
- [x]     exec [flags] APP_NAME -- COMMAND [args...]	Execute command in a container
- [x]     logs                                      	Tail logs for application
- [x]     ls                                        	List services of all applications
- [x]     port-forward APP_NAME                     	Forward local ports to services in an application
- [x]     status APP_NAME                           	Show status of an application, including workloads and

## Commands should be deprecated

we should use appfile instead of these commands.

- [x] vela svc deploy
- [x] vela route APP_NAME
- [x] vela scale APP_NAME
- [x] vela metrics APP_NAME
- [x] vela rollout APP_NAME

### LINKED ISSUE #964: delete health scope when app was deleted
Currently, we will create a default health scope for every app created by vela cli. But we don't delete it in `vela delete` command, so there will be lots of no used health scope CR instance, we need to clean them.

1. check the health scope after `vela delete ` command.
2. when there's no working app in the health scope, delete it.

### LINKED ISSUE #1164: Kubevela client CI/CD integration and distributed developers workflow
**Describe the bug**

We are a bit confused about what the intended workflow is for developers/applications teams working with Kubevela.

We expect `vela ls` to show all the applications of an environment (kubernetes namespace) that a developer is part of. This way developers can get the same view of their applications and their states.  Currently `vela ls` only shows apps which have been deployed by local user.

**To Reproduce**

1. UserA deploys *app1* with `vela up` at time T1.
2. UserB wants to check the status and logs of the app *app1*, performs `vela ls` and *app1* is not there.
3. UserB pushes  *app1* again and sees `App exists, updating existing deployment...` at  time T2.
4. UserB sees *app1* created at time T1 with `vela ls`
5. UserB  runs `vela status app1` and sees T2 timestamps

**Expected behavior**

According to the previous steps to reproduce the issue, *User B* would see the same timestamps and be able to run actions like `vela logs` or `vela exec` without deploying the app again.

**Additional context**

Pushing apps via an automated CI/CD system, where pipelines run tests on git push/commits, why developers need to push the same app again manually, with the same code and parameters (environment variables and/or secrets) as the CI/CD pipeline in order to be able just to check the status of the app?

Can you point us to a suitable workflow for this scenario?

wonderflow:
fix #877

1. delete health scope when delete app,  fix #964
2. cherry-pick(#1162) fix vela system dryrun command cannot get namespaced definition 
3. support `vela ls --namespace=xxx` to list applications from cluster, fix #1164


references/apiserver/capabilityHandlers.go
```diff
@@ -63,7 +63,8 @@ func (s *APIServer) DeleteCapabilityCenter(c *gin.Context) {
 // RemoveCapabilityFromCluster remove a specific capability from cluster
 func (s *APIServer) RemoveCapabilityFromCluster(c *gin.Context) {
 	capabilityCenterName := c.Param("capabilityName")
-	msg, err := common.RemoveCapabilityFromCluster(s.KubeClient, capabilityCenterName)
+	// TODO get namespace from env
+	msg, err := common.RemoveCapabilityFromCluster("default", s.c, s.KubeClient, capabilityCenterName)
```
zzxwill, line 67: Let's define `default` as a constant string as it's used in many lines.

hongchaodeng (approved):
lgtm

references/cli/exec.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
```
zzxwill, line 9: Move it to L24:)

references/cli/ls.go
```diff
@@ -33,98 +32,64 @@ func NewListCommand(c types.Args, ioStreams cmdutil.IOStreams) *cobra.Command {
...
-		return app, err
-	}
-	all := mergeStagingComponents(deployedComponentList, env, ioStreams, fetcher)
+func printApplicationList(ctx context.Context, c client.Reader, namespace string, ioStreams cmdutil.IOStreams) error {
 	table := newUITable()
-	table.AddRow("SERVICE", "APP", "TYPE", "TRAITS", "STATUS", "CREATED-TIME")
-	for _, a := range all {
-		traitAlias := strings.Join(a.TraitNames, ",")
-		table.AddRow(a.Name, a.App, a.WorkloadName, traitAlias, a.Status, a.CreatedTime)
+	table.AddRow("APP", "COMPONENT", "TYPE", "TRAITS", "PHASE", "HEALTHY", "STATUS", "CREATED-TIME")
```
zzxwill, line 58:
```suggestion
	table.AddRow("APP", "COMPONENT", "TYPE", "TRAITS", "PHASE", "HEALTHY", "ERROR-MESSAGE", "CREATED-TIME")
```

I learnt from your implementation, `HEALTHY` stands whether the component is healthy, and `STATUS` is its error message. But if there are data listed, I think you can ignore this suggestion.

PR #1179, opened 2021-03-11
fix dockerfile of vela-core  to use 'CMD' instead of 'ENTRYPOINT'

shikanon:
refer to https://github.com/oam-dev/kubevela/issues/1177
change the dockerfile to use 'CMD' instead of 'ENTRYPOINT' and vela-core deployment to use `command` instead of `args`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,4 +35,4 @@ ARG TARGETARCH
 COPY --from=builder /workspace/manager-${TARGETARCH} /manager
 USER nonroot:nonroot
 
-ENTRYPOINT ["/manager"]
+CMD ["/manager","--metrics-addr=:8080","--enable-leader-election"]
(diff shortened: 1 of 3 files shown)


shikanon:
@wonderflow hi,  Do you konw why the Test of HealthScope  is Failed when this code change?

wonderflow:
I can't understand the relationship between this PR and Issue. Does it really fix the issue, have you tried?

shikanon:
> I can't understand the relationship between this PR and Issue. Does it really fix the issue, have you tried?

ye, this PR and Issue is discuss how to provide a convenient way to enter the container for troubleshooting.  But I think this change can not let me entry the container because of using the image `Distroless` which have not shell and any other programs.

I will close this PR and discuss this solution in the issue continue.

Issue #1157, opened 2021-03-08
[Feature] `vela env init` should only create namespace if not existing


jriguera:
**Is your feature request related to a problem? Please describe.**

We want to manage user using RBAC, each user would have access to a pre-created namespace (but they do not have access to create new ones), when we use `vela env init --namespace TEAM  TEAM`, vela cli fails with:

```
Error: namespaces is forbidden: User "user@account.com" cannot create resource "namespaces" in API group "" at the cluster scope: requires one of ["container.namespaces.create"] permission(s).
```

**Describe the solution you'd like**
Kubevela should check if the namespace is already existing and if so try to get it and not created (the user has `container.namespaces.get` access).

**Describe alternatives you've considered**
We have created this workaround instead of using `vela env init`:
```
TEAM="my-team"
mkdir -p ~/.vela/envs/$TEAM
echo {"name":"$TEAM","namespace":"$TEAM","domain":"","issuer":""}  > ~/.vela/envs/$TEAM/config.json
```
But we would like a more user friendly solution with `vela env init`.

**Additional context**
This is on GKE with RBAC and role bindings.

wonderflow:
Yes, thanks for reporting

wonderflow:
@mosesyou Hi, are you working on that? I can fix along with #1174

mosesyou:
> @mosesyou Hi, are you working on that? I can fix along with #1174

Sry, because something delayed doing this. Have you completed it? If not I will fix it in a while.

wonderflow:
@mosesyou not yet, please go ahead

PR #1187, opened 2021-03-12
fix: check if namespace exists

mosesyou:
fix: #1157 
Check if namespace exists before create it.

### DIFF
--- a/pkg/utils/env/env.go
+++ b/pkg/utils/env/env.go
@@ -78,12 +78,14 @@ func CreateOrUpdateEnv(ctx context.Context, c client.Client, envName string, env
 	var message = ""
 	// Check If Namespace Exists
 	if err := c.Get(ctx, k8stypes.NamespacedName{Name: envArgs.Namespace}, &corev1.Namespace{}); err != nil {
-		if apierrors.IsNotFound(err) {
-			// Create Namespace
-			if err := c.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: envArgs.Namespace}}); err != nil {
-				return message, err
-			}
+		if !apierrors.IsNotFound(err) {
+			return message, err
 		}
+		// Create Namespace if not found
+		if err := c.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: envArgs.Namespace}}); err != nil {
+			return message, err
+		}
+	}
 		return message, err
 	}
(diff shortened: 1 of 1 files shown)


wonderflow (approved):
LGTM, thanks

pkg/utils/env/env.go
```diff
@@ -75,8 +76,14 @@ func CreateOrUpdateEnv(ctx context.Context, c client.Client, envName string, env
...
-	// Create Namespace
-	if err := c.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: envArgs.Namespace}}); err != nil && !apierrors.IsAlreadyExists(err) {
+	// Check If Namespace Exists
+	if err := c.Get(ctx, k8stypes.NamespacedName{Name: envArgs.Namespace}, &corev1.Namespace{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			// Create Namespace
+			if err := c.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: envArgs.Namespace}}); err != nil {
+				return message, err
+			}
+		}
```
wonderflow, line 86:
```suggestion
	// Check If Namespace Exists
	if err := c.Get(ctx, k8stypes.NamespacedName{Name: envArgs.Namespace}, &corev1.Namespace{}); err != nil {
		if !apierrors.IsNotFound(err) {
			return message, err
		}
		// Create Namespace if not found
		if err := c.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: envArgs.Namespace}}); err != nil {
			return message, err
		}
	}
```

pkg/utils/env/env.go
```diff
@@ -75,8 +76,16 @@ func CreateOrUpdateEnv(ctx context.Context, c client.Client, envName string, env
...
+	if err := c.Get(ctx, k8stypes.NamespacedName{Name: envArgs.Namespace}, &corev1.Namespace{}); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return message, err
+		}
+		// Create Namespace if not found
+		if err := c.Create(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: envArgs.Namespace}}); err != nil {
+			return message, err
+		}
+	}
 		return message, err
```
wonderflow, line 89:
```suggestion
```

PR #1192, opened 2021-03-14
add podDisruptive to traitdefinition

### LINKED ISSUE #1123: [Feature] TraitDefinition should tell if it will restart the workload pod

Now when we deploy/upgrade an Application, only change some fields of trait , we don't know whether the underlying pod will restart or not, we should let the TraitDefinition tell.

**Additional context**

yangsoon:
fix #1123

Now when we deploy/upgrade an Application, only change some fields of trait , we don't know whether the underlying pod will restart or not, we should let the TraitDefinition tell. So we add a podDisruptive field  to TraitDefinition.

- [X] add podDisruptive field into traitdefinition.
- [X] display in cli
- [x] add docs
- [x] check all traits in kubevela and registry

### DIFF
--- a/apis/core.oam.dev/common/types.go
+++ b/apis/core.oam.dev/common/types.go
@@ -112,22 +112,6 @@ type WorkloadGVK struct {
 	Kind       string `json:"kind"`
 }
 
-// DisruptiveProperty describes whether when a trait is applied to an Application,
-// it will cause the pod to restart due to changes in some fields.
-// +kubebuilder:validation:Enum=unknown;yes;no;
-type DisruptiveProperty string
-
-const (
-	// UnknownDisruptive means disruptive unknown.
-	UnknownDisruptive DisruptiveProperty = "unknown"
-
-	// Disruptive means will cause the pod to restart.
-	Disruptive DisruptiveProperty = "yes"
-
-	// NonDisruptive means will not cause the pod restart
-	NonDisruptive DisruptiveProperty = "no"
-)
-
 // A DefinitionReference refers to a CustomResourceDefinition by name.
 type DefinitionReference struct {
 	// Name of the referenced CustomResourceDefinition.
--- a/apis/core.oam.dev/common/types.go
+++ b/apis/core.oam.dev/common/types.go
@@ -112,6 +112,22 @@ type WorkloadGVK struct {
 	Kind       string `json:"kind"`
 }
 
+// DisruptiveProperty describes whether when a trait is applied to an Application,
+// it will cause the pod to restart due to changes in some fields
+// +kubebuilder:validation:Enum=unknown;yes;no;
+type DisruptiveProperty string
+
+const (
+	// UnknownDisruptive means disruptive unknown.
+	UnknownDisruptive DisruptiveProperty = "unknown"
+
+	// Disruptive means will cause the pod to restart.
+	Disruptive DisruptiveProperty = "yes"
+
+	// NonDisruptive means will not cause the pod restart
+	NonDisruptive DisruptiveProperty = "no"
+)
+
 // A DefinitionReference refers to a CustomResourceDefinition by name.
 type DefinitionReference struct {
 	// Name of the referenced CustomResourceDefinition.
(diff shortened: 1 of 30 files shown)


references/cli/traits.go
```diff
@@ -46,9 +46,9 @@ func printTraitList(userNamespace string, c types.Args, ioStreams cmdutil.IOStre
 	if err != nil {
 		return err
 	}
-	table.AddRow("NAME", "NAMESPACE", "APPLIES-TO", "CONFLICTS-WITH", "DESCRIPTION")
+	table.AddRow("NAME", "NAMESPACE", "APPLIES-TO", "CONFLICTS-WITH", "POD-DISRUPTIVE", "DESCRIPTION")
 	for _, t := range traitDefinitionList {
-		table.AddRow(t.Name, t.Namespace, strings.Join(t.Spec.AppliesToWorkloads, ","), strings.Join(t.Spec.ConflictsWith, ","), plugins.GetDescription(t.Annotations))
+		table.AddRow(t.Name, t.Namespace, strings.Join(t.Spec.AppliesToWorkloads, ","), strings.Join(t.Spec.ConflictsWith, ","), t.Spec.PodDisruptive, plugins.GetDescription(t.Annotations))
```
wonderflow, line 51: if `t.Spec.PodDisruptive` is empty string, use `Unknown` instead

apis/core.oam.dev/v1alpha2/core_types.go
```diff
@@ -155,6 +171,14 @@ type TraitDefinitionSpec struct {
 	// +optional
 	WorkloadRefPath string `json:"workloadRefPath,omitempty"`
 
+	// PodDisruptive specifies whether using the trait will cause the pod to restart
+	// Valid values are:
+	// - "Unknown" (default): disruptive unknown
+	// - "Yes": cause the pod to restart
+	// - "No": will not cause the pod restart
```
wonderflow, line 178: please use lowercase

wonderflow (approved):
LGTM, please add this feature into docs.

apis/core.oam.dev/common/types.go
```diff
@@ -112,6 +112,22 @@ type WorkloadGVK struct {
 	Kind       string `json:"kind"`
 }
 
+// DisruptiveProperty describes whether when a trait is applied to an Application,
```
zzxwill, line 115: // DisruptiveProperty describes whether a trait will cause the pod to restart due to changes in some fields 
// when it is applied to an Application. 

Changing `when` order looks more clear.

apis/core.oam.dev/common/types.go
```diff
@@ -112,6 +112,22 @@ type WorkloadGVK struct {
...
+
+const (
+	// UnknownDisruptive means disruptive unknown.
+	UnknownDisruptive DisruptiveProperty = "unknown"
+
+	// Disruptive means will cause the pod to restart.
+	Disruptive DisruptiveProperty = "yes"
+
+	// NonDisruptive means will not cause the pod restart
+	NonDisruptive DisruptiveProperty = "no"
```
zzxwill, line 128: Looks like `true` and `false` are much more popular in CRD. I can get that you wanna set three values.
How about `unknown` meaning `false`? This is also the default behavior for those traits without setting it.

apis/core.oam.dev/common/types.go
wonderflow: good idea

docs/en/cue/trait.md
```diff
@@ -141,4 +141,7 @@ spec:
               "/api": 8080
 ```
 
+When we deploy/upgrade an Application, some field of trait changed will cause the pod to restart, We should add `podDisruptive`
```
zzxwill, line 144:
```suggestion
When we deploy/upgrade an Application, if changes on some field of a trait will cause the pod to restart, We should add `podDisruptive`
```

apis/core.oam.dev/common/types.go
zzxwill: It seems `unknown` is necessary as we could not tell whether a trait will restart its pods or not. But whether a trait will cause its Pods to restart is so essential to business, we should emphasizing platform team to clearly set the field, so leave it to default `false`. `unknown` should not be an option.

wonderflow: after a deep thought, I think yes, we should use bool for this field.

wonderflow:
@zzxwill thanks for your advise, all good ideas and adopted

PR #1205, opened 2021-03-16
move out rollout e2e

ryanzhang-oss:
move out rollout e2e

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -125,9 +125,6 @@ jobs:
 
   e2e-rollout-tests:
     runs-on: aliyun
-      needs: detect-noop
-      if: needs.detect-noop.outputs.noop != 'true'
-
     steps:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -123,6 +123,62 @@ jobs:
       - name: Clean up testdata
         run: make compatibility-testdata-cleanup
 
+  e2e-rollout-tests:
+    runs-on: aliyun
```
wonderflow, line 127: separate out doesn't help, the CI runs in the same queue  on `aliyun`

ryanzhang-oss: won't they run in parallel? Also, it's at least good to see that the original e2e failed in this PR.

wonderflow: They are not parallel in `aliyun` runner

ryanzhang-oss: now there are more than one machine

hongchaodeng (approved):
lgtm

PR #1208, opened 2021-03-17
Refactor appConfig Controller to make it reusable and rename appDeployment to appRollout

ryanzhang-oss:
1. Refactor appConfig Controller to make it reusable
2. Rename appDeployment to appRollout


wonderflow (approved):
LGTM

PR #1210, opened 2021-03-17
init applicationRevision Object as revision of Application

wonderflow:
add applicationRevision CRD to record the snapshot of the application. That is a revision of Application.


hongchaodeng (approved):
lgtm

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -0,0 +1,156 @@
...
+		if err := r.Get(ctx, client.ObjectKey{Name: application.Status.LatestRevision.Name, Namespace: application.Namespace}, lastAppRevision); err != nil {
+			return false, nil, errors.Wrapf(err, "fail to get applicationRevision %s", application.Status.LatestRevision.Name)
+		}
+		if DeepEqualRevision(lastAppRevision, appRev) {
+			// No difference on spec, will not create a new revision
+			return false, nil, nil
+		}
+	}
+
+	return true, newRevision, r.Create(ctx, appRev)
```
hongchaodeng, line 105: > r.Create(ctx, appRev)

This might return error if it is created once in second reconciles.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -96,12 +103,19 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 	applog.Info("build template")
 	// build template to applicationconfig & component
-	ac, comps, err := appParser.GenerateApplicationConfiguration(appfile, app.Namespace)
+	ac, comps, err := appParser.GenerateApplicationConfiguration(generatedAppfile, app.Namespace)
 	if err != nil {
 		applog.Error(err, "[Handle GenerateApplicationConfiguration]")
 		app.Status.SetConditions(errorCondition("Built", err))
 		return handler.handleErr(err)
 	}
+	handler.specDiff, handler.newRevision, err = r.GenerateRevision(ctx, generatedAppfile, app, ac, comps)
```
hongchaodeng, line 112: Should add a if-else here:
- If it is rollout-template mode, just create the revision, no appConfig
- otherwise, create appConfig

Right?

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
+	"github.com/oam-dev/kubevela/pkg/appfile"
+	"github.com/oam-dev/kubevela/pkg/controller/utils"
+)
+
+// GenerateRevision will generate revision for an Application when created/updated
+func (r *Reconciler) GenerateRevision(ctx context.Context, appfile *appfile.Appfile, application *v1alpha2.Application, ac *v1alpha2.ApplicationConfiguration, comps []*v1alpha2.Component) (bool, *v1alpha2.AppRevisionHash, error) {
```
hongchaodeng, line 33: The names of comps being passed in are not versioned. Should they be?

hongchaodeng: Also note that the ac passed in here will use the ComponentName instead of the revisionName.
The revisionName is only modified after apply():

```go
func (h *appHandler) apply(ctx context.Context, ac *v1alpha2.ApplicationConfiguration, comps []*v1alpha2.Component) error {

	ac.SetOwnerReferences(owners)
	for _, comp := range comps {
		...
		revisionName, err := h.createOrUpdateComponent(ctx, newComp)
		if err != nil {
			return err
		}
		// find the ACC that contains this component
		for i := 0; i < len(ac.Spec.Components); i++ {
			// update the AC using the component revision instead of component name
			// we have to make AC immutable including the component it's pointing to
			if ac.Spec.Components[i].ComponentName == newComp.Name {
				ac.Spec.Components[i].RevisionName = revisionName
				ac.Spec.Components[i].ComponentName = ""
			}
		}
	}
```

apis/core.oam.dev/v1alpha2/core_types.go
```diff
@@ -349,9 +349,29 @@ type ComponentStatus struct {
...
+	Name     string `json:"name"`
+	Revision int64  `json:"revision"`
+
+	// RevisionHash record the hash value of the spec of ApplicationConfiguration object.
+	// We're going to deprecate that by using ApplicationRevision
 	RevisionHash string `json:"revisionHash,omitempty"`
+
+	// AppRevisionHash record the detailed hash value of the spec of ApplicationRevision Object
+	// We should replace it with 'revisionHash' after the refactor finished
+	AppRevisionHash *AppRevisionHash `json:"appRevisionHash,omitempty"`
```
ryanzhang-oss, line 361: Not sure why we need to have a complex hash structure that we can't compare.  We end up not even using it to compare.

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+		}
+	}
+	newRevision := &v1alpha2.AppRevisionHash{
+		ApplicationSpec:      appHash,
+		ComponentDefinitions: componentDefinitionHash,
+		TraitDefinitions:     traitDefinitionHash,
+	}
+	// check API spec difference
+	if application.Status.LatestRevision != nil && application.Status.LatestRevision.AppRevisionHash != nil {
```
ryanzhang-oss, line 93: we should compare the newRevision and the latest AppRevision before doing any deepEqual comparison. The fact that the Revision Hash is a structure makes it not very intuitive (and might even be incorrect)

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -0,0 +1,156 @@
...
+			// No difference on spec, will not create a new revision
+			return false, nil, nil
+		}
+	}
+
+	return true, newRevision, r.Create(ctx, appRev)
+}
+
+// DeepEqualRevision will check the Application and Definition to see if the Application is the same revision
+// We don't check the rendered AC and Component spec here, because their spec type is runtime.RawExtension which is hard to check.
```
ryanzhang-oss, line 109: All the definitions contain runtime.RawExtension, why can we compare there but not in the AC? I think it's actually enough to just compare AC if it refers to component revision

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -0,0 +1,156 @@
...
+	for idx, td := range new.Spec.TraitDefinitions {
+		if !cmp.Equal(&old.Spec.TraitDefinitions[idx].Spec, &td.Spec) {
+			return false
+		}
+	}
+	return cmp.Equal(&old.Spec.Application.Spec, &new.Spec.Application.Spec)
+}
+
+// DupWorkloadDef will check if the workloadDefinition is duplicated
+func DupWorkloadDef(appRev *v1alpha2.ApplicationRevision, wd *v1alpha2.WorkloadDefinition) bool {
```
ryanzhang-oss, line 139: isWorkloadDefDup sounds like a better name

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+	}
+	appRev.Name = appfile.RevisionName
+	appRev.Namespace = application.Namespace
+
+	for _, w := range appfile.Workloads {
+		if w == nil {
+			continue
+		}
+		if !DupWorkloadDef(appRev, w.FullTemplate.WorkloadDefinition) {
```
ryanzhang-oss, line 58: I don't think we need this but if we have to, why not make them a map[definiiton]string instead of an array? We are clearly using it as a map.

pkg/appfile/parser.go
```diff
@@ -44,6 +44,8 @@ type Workload struct {
 
 	Helm                *v1alpha2.Helm
 	DefinitionReference v1alpha2.DefinitionReference
+
+	FullTemplate *util.Template
```
ryanzhang-oss, line 48: why do we need any of the above when we have the fullTemplate?

pkg/appfile/parser.go
```diff
@@ -93,6 +95,8 @@ type Trait struct {
 	Template           string
 	HealthCheckPolicy  string
 	CustomStatusFormat string
+
+	FullTemplate *util.Template
```
ryanzhang-oss, line 99: why do we need any of the above when we have the fullTemplate when all of them are copied from the fullTemplate.

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
ryanzhang-oss: yes, this will cause problems. We compute the revision with AC pointing to component (mutable) when the real AC created pointing to componentRevision (imutable)

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
ryanzhang-oss: yes but that logic is in apply.go

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
ryanzhang-oss, line 93: The application's lastestRevision is only bumped after a new AC is created, this will cause inconsistency.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller_test.go
```diff
@@ -47,7 +48,7 @@ import (
 	"github.com/oam-dev/kubevela/pkg/oam/util"
 )
 
-var _ = Describe("Test Application Controller", func() {
+var _ = FDescribe("Test Application Controller", func() {
```
wonderflow, line 51:
```suggestion
var _ = Describe("Test Application Controller", func() {
```

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -265,85 +278,28 @@ func (h *appHandler) createOrUpdateAppConfig(ctx context.Context, appConfig *v1a
...
-	if _, exist := h.app.GetAnnotations()[oam.AnnotationAppRollout]; exist || h.app.Spec.RolloutPlan != nil {
-		h.logger.Info(fmt.Sprintf("The application %s rolling out is controlled by a rollout plan", h.app.Name))
-		appConfig.SetAnnotations(oamutil.MergeMapOverrideWithDst(appConfig.GetAnnotations(), map[string]string{
-			oam.AnnotationAppRollout: strconv.FormatBool(true),
-		}))
-	}
-
-	// record that last appConfig we created first in the app's status
-	// make sure that we persist the latest revision first
-	if err := h.r.UpdateStatus(ctx, h.app); err != nil {
```
wonderflow, line 340: do we need to keep this logic? If not, the status won't recorded

wonderflow: I got it

PR #1214, opened 2021-03-18
init ApplicationRevision CRD Object as revision of Application

wonderflow:


hongchaodeng (approved):
lgtm

apis/core.oam.dev/v1alpha2/core_types.go
```diff
@@ -349,9 +349,29 @@ type ComponentStatus struct {
 
 // Revision has name and revision number
 type Revision struct {
```
ryanzhang-oss, line 351: I don't think we really need this

PR #1218, opened 2021-03-18
Revise the doc site with recent features

resouer:
Re-write some doc for v1.0 release.


docs/en/helm/component.md
```diff
@@ -11,7 +11,7 @@ so we provide a chart which only includes minimal dependencies KubeVela relies o
 
 Install the minimal flux2 chart provided by KubeVela:
 ```shell
-helm install --create-namespace -n flux-system helm-flux http://oam.dev/catalog/helm-flux2-0.1.0.tgz
+$ helm install --create-namespace -n flux-system helm-flux http://oam.dev/catalog/helm-flux2-0.1.0.tgz
```
wonderflow, line 14: without the `$` , the user can copy-paste  directly

wonderflow (approved):
Cool

docs/en/helm/component.md
resouer: Then we should have consistency, for now all other docs have $

wonderflow: Some of the commands along with output should keep `$`

PR #1236, opened 2021-03-19
refactor ac reconcile so that we can reuse it in new AC

wangyikewxgm:
fix applicationContext pr e2e-test

reference https://github.com/oam-dev/kubevela/pull/1215/


ryanzhang-oss (approved):
good job

PR #1246, opened 2021-03-20
more useful vela system dry-run

wonderflow:
1. Add a demo for `vela system dry-run` which is handy for `debug`
2. fix a bug of `CreateOrUpdateObjects` that only update the first object.

### DIFF
--- a/pkg/utils/common/common.go
+++ b/pkg/utils/common/common.go
@@ -16,6 +16,7 @@ import (
 	"cuelang.org/go/cue"
 	"cuelang.org/go/encoding/openapi"
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/ghodss/yaml"
 	certmanager "github.com/wonderflow/cert-manager-api/pkg/apis/certmanager/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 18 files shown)


pkg/utils/common/common.go
```diff
@@ -16,6 +16,7 @@ import (
 	"cuelang.org/go/cue"
 	"cuelang.org/go/encoding/openapi"
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/ghodss/yaml"
```
ryanzhang-oss, line 19: is this the right yaml package?

references/cli/dryrun.go
```diff
@@ -1,26 +1,37 @@
...
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 
+	"k8s.io/apimachinery/pkg/runtime"
```
ryanzhang-oss, line 12: please reorg goimports

ryanzhang-oss (approved):

pkg/utils/common/common.go
wonderflow: yes

ryanzhang-oss:
btw,  do we have any examples using the "kube" package in cue?

wonderflow:
@ryanzhang-oss  Will improve it in another PR. Working for it now. For quick glance:

```
kind: WorkloadDefinition
metadata:
  name: myworker
spec:
  definitionRef:
    name: deployments.apps
  schematic:
    cue:
      template: |
        import ("kube")
        output: kube.#Deployment

        output: {
        	spec: {
        		selector: matchLabels: {
        		}

        		template: {
        			metadata: labels: {
        			}

        			spec: {
        				containers: [{
        					name:  context.name
        					image: parameter.image

        					if parameter["cmd"] != _|_ {
        						command: parameter.cmd
        					}
        				}]
        			}
        		}
        	}
        }

        parameter: {
        	// +short=i
        	image: string
        	// +usage=Commands to run in the container
        	cmd?: [...string]
        }
```

PR #1256, opened 2021-03-21
docs: minor fixes on introduction and concepts

unknwon:
I think there is a typo in "introduction" and made other minor wording fixes in "concepts".


wonderflow (approved):
good catch! Thanks

PR #1261, opened 2021-03-22
Compatible with Application containing WorkloadDefinition type helm moudle

yangsoon:
When Application containing helm moudle which defined by WorkloadDefinition, we set workload.DefinitionReference refer to
the DefinitionReference in workloadDefinition.


wonderflow (approved):
LGTM

PR #1267, opened 2021-03-23
fix kube-webhook-certgen to patch CRD conversation && remove cert-manager in CI e2e test && remove issuer create from CLI env

wonderflow:
remove cert-manager in CI, by default, the webhook is enabled  in chart, it will use `wonderflow/kube-webhook-certgen:v2.1` if cert-manager is not enabled.

Repo: https://github.com/wonderflow/kube-webhook-certgen

fix https://github.com/oam-dev/kubevela/issues/1361


zzxwill (approved):
/lgtm

PR #1271, opened 2021-03-24
Restore rollout after the change to point to appRevision

ryanzhang-oss:


charts/vela-core/values.yaml
```diff
@@ -16,6 +16,7 @@ image:
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
+useAppConfig: true
```
wonderflow, line 19: cool

charts/vela-core/templates/kubevela-controller.yaml
```diff
@@ -112,6 +112,9 @@ spec:
             - "--webhook-port={{ .Values.webhookService.port }}"
             - "--webhook-cert-dir={{ .Values.admissionWebhooks.certificate.mountPath }}"
             {{ end }}
+            {{ if not .Values.useAppConfig }}
+            - "--app-config-installed=false"
+            {{ end }}
```
wonderflow, line 117: can we add a conditional statement for the crd ?

wonderflow (approved):
LGTM, the CI seems still failing

charts/vela-core/templates/kubevela-controller.yaml
ryanzhang-oss: is there a way to not install the appConfig CRD?

wonderflow: I'm not sure, maybe we need create two charts

wonderflow: One is application only(name is `kubevela`), one is application with AppConfig(name is `kubevela-compatible`).

PR #1272, opened 2021-03-24
AppDeployment: Add definitions and utilities code

hongchaodeng:
This takes out all the definitions and utilities code from https://github.com/oam-dev/kubevela/pull/1193 .
The controller code needs to be rewritten so let's get this part in first.


apis/core.oam.dev/v1alpha2/appdeployment_types.go
```diff
@@ -0,0 +1,197 @@
+/*
+Copyright 2020 The KubeVela Authors.
```
wonderflow, line 2:
```suggestion
Copyright 2021 The KubeVela Authors.
```

wonderflow (changes requested):
I didn't see the added imports in `go.mod` used any where?

hongchaodeng:
> I didn't see the added imports in go.mod used any where?

@wonderflow 

in common.Scheme

hongchaodeng:
@wonderflow 

No everything in go.mod is used because some of them are indirect imports.

hongchaodeng:
@wonderflow 

They are automatically imported via `go get` command. Let me give you one example, if we removed `github.com/go-logr/zapr`, then we will have the following error:

```
# github.com/go-logr/zapr
../../../../pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:136:10: cannot use &infoLogger{...} (type *infoLogger) as type logr.Logger in return argument:
	*infoLogger does not implement logr.Logger (missing Error method)
../../../../pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:141:2: cannot use disabledInfoLogger (type *noopInfoLogger) as type logr.Logger in return argument:
	*noopInfoLogger does not implement logr.Logger (missing Error method)
```

wonderflow (approved):
LGTM after CI passed

PR #1275, opened 2021-03-24
bump go version to 1.16

hongchaodeng:
go 1.16 brings important features like `embed`, and improves module system and `go get` cmd (which honors the retract directives).

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -13,7 +13,7 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.14'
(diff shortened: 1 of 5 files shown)


hongchaodeng:
Hold this PR until 1.16 becomes more stable

wonderflow:
I think we can continue this PR, as it seems more stable now and 1.14 will not be maintained https://golang.org/doc/devel/release.html#policy

hongchaodeng:
Quite busy this following week. Looking for someone to take it over. Thanks!

PR #1277, opened 2021-03-24
re-open e2e-test for helm schematic

captainroy-hy:
as title.

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationcontext/applicationcontext_controller.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationcontext/applicationcontext_controller.go
@@ -11,7 +11,6 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationcontext/applicationcontext_controller.go
```diff
@@ -83,7 +82,9 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 	appConfig.Status = appContext.Status
 	// the name of the appConfig has to be the same as the appContext
-	appConfig.ObjectMeta = metav1.ObjectMeta{Namespace: appContext.Namespace, Name: appContext.Name, UID: appContext.UID}
```
captainroy-hy, line 86: AC controller needs ownerRef to determine isControlledByApp. So we should not override ownerRef here.

wonderflow:
Does it the same with #1261 ? Can you include changes in that one

captainroy-hy:
> Does it the same with #1261 ? Can you include changes in that one

Not exactly same. After offline discussion, @yangsoon will include his change after the helm e2e issue fixed.

wonderflow:
Cool， please fix the conflicts @

> > Does it the same with #1261 ? Can you include changes in that one
> 
> Not exactly same. After offline discussion, @yangsoon will include his change after the helm e2e issue fixed.

wonderflow (approved):
Cool job!

PR #1279, opened 2021-03-24
add hpa trait and fix render template with nil params

wonderflow:
1. add a hpa trait in readme
2. fix render template with nil params.

### DIFF
--- a/docs/examples/registry/hpa.yaml
+++ b/docs/examples/registry/hpa.yaml
@@ -1,0 +1,43 @@
+apiVersion: core.oam.dev/v1alpha2
+kind: TraitDefinition
+metadata:
+  annotations:
+    definition.oam.dev/description: "configure k8s HPA for Deployment"
+  name: hpa
+spec:
+  appliesToWorkloads:
+    - webservice
+    - worker
+  schematic:
+    cue:
+      template: |
+        outputs: hpa: {
+        	apiVersion: "autoscaling/v2beta2"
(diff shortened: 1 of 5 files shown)


docs/examples/registry/hpa.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - worker
+  schematic:
+    cue:
+      template: |
+        outputs: hpa: {
+        	apiVersion: "autoscaling/v2beta2"
+        	kind:       "HorizontalPodAutoscaler"
+        	metadata: name: context.name
+        	spec: {
+        		scaleTargetRef: {
```
ryanzhang-oss, line 19: hard code this in the defnition?

pkg/dsl/definition/template_test.go
```diff
@@ -496,6 +513,57 @@ parameter: {
...
+		maxReplicas: parameter.max
+	}
+}
+parameter: {
+	min:     *1 | int
+	max:     *10 | int
+}`,
+			params: nil,
+			expWorkload: &unstructured.Unstructured{
+				Object: map[string]interface{}{
```
ryanzhang-oss, line 533: why not build a real deployment and then marshal it to map

ryanzhang-oss (approved):

docs/examples/registry/hpa.yaml
wonderflow: yes， it's apply to Deployment kind workload.

PR #1280, opened 2021-03-24
feat: Support CloneSet scale with AppRollout

jasenzhang:
1. support CloneSet scale with AppRollout


pkg/controller/common/rollout/rollout_plan_controller.go
```diff
@@ -380,16 +380,27 @@ func (r *Controller) GetWorkloadController() (workloads.WorkloadController, erro
...
+			// TODO: create deployment scale controller when current rollout plan is for scale
 			return workloads.NewDeploymentController(r.client, r.recorder, r.parentController,
 				r.rolloutSpec, r.rolloutStatus, source, target), nil
 		}
 	}
 	return nil, fmt.Errorf("the workload kind `%s` is not supported", kind)
 }
+
+func (r *Controller) isRolling() bool {
+	return r.sourceWorkload != nil && r.targetWorkload != nil
```
ryanzhang-oss, line 405: only sourceWorkload can be nil

pkg/controller/common/rollout/workloads/cloneset_scale_controller.go
```diff
@@ -0,0 +1,49 @@
...
+		recorder:               recorder,
+		parentController:       parentController,
+		rolloutSpec:            rolloutSpec,
+		rolloutStatus:          rolloutStatus,
+		workloadNamespacedName: workloadName,
+		handler:                csScaleHandler,
+	}
+}
+
+func (c *cloneSetScaleHandler) verifySpec(rolloutSpec *v1alpha1.RolloutPlan, cloneSet *kruise.CloneSet) error {
```
ryanzhang-oss, line 36: why return nil here?

pkg/controller/common/rollout/workloads/cloneset_scale_controller.go
```diff
@@ -0,0 +1,49 @@
...
+	return nil
+}
+
+func (c *cloneSetScaleHandler) initialize(ctx context.Context, cloneSet *kruise.CloneSet) error {
```
ryanzhang-oss, line 40: the scale controller still needs to initialize the clonset. It is very similar to the rollout controller ( or identical but I haven't gone through the details)

pkg/controller/common/rollout/workloads/cloneset_scale_controller.go
```diff
@@ -0,0 +1,49 @@
...
+	return nil
+}
+
+// rolloutOneBatchPods update CloneSet spec replicas directly
+func (c *cloneSetScaleHandler) rolloutOneBatchPods(ctx context.Context, cloneSet *kruise.CloneSet, newPodTargets int) error {
+	targetReplicas := int32(newPodTargets)
```
ryanzhang-oss, line 46: we need more logs and events

pkg/controller/common/rollout/workloads/cloneset_controller.go
```diff
@@ -138,6 +131,10 @@ func (c *CloneSetController) RolloutOneBatchPods(ctx context.Context) (bool, err
 	newPodTarget := c.calculateNewPodTarget(int(cloneSetSize))
 	// set the Partition as the desired number of pods in old revisions.
 	clonePatch := client.MergeFrom(c.cloneSet.DeepCopyObject())
+
+	if err := c.handler.rolloutOneBatchPods(ctx, c.cloneSet, newPodTarget); err != nil {
+		return false, err
+	}
 	c.cloneSet.Spec.UpdateStrategy.Partition = &intstr.IntOrString{Type: intstr.Int,
```
ryanzhang-oss, line 138: why is this still here? It seems that the scale part shouldn't touch the partition.

pkg/controller/common/rollout/rollout_plan_controller.go
jasenzhang: Ok. I'll update it.
```suggestion
	return r.sourceWorkload != nil 
```

pkg/controller/common/rollout/workloads/cloneset_scale_controller.go
jasenzhang: It seems that it doesn't need verification.

pkg/controller/common/rollout/workloads/cloneset_controller.go
jasenzhang: OK. I'll delete it.

pkg/controller/common/rollout/workloads/cloneset_scale_controller.go
jasenzhang: OK

PR #1281, opened 2021-03-24
[DO NOT MERGE] Add webhook version convert methods for Application

yangsoon:
add convert methods to convert different version


wonderflow:
prefer  #1282

PR #1287, opened 2021-03-24
feat: Support CloneSet scale with AppRollout

jasenzhang:
Support CloneSet scale with AppRollout


ryanzhang-oss:
I am closing  this PR as we will have to rewrite the scale controller as an independent one

PR #1290, opened 2021-03-25
add abandoning state to the rollout

ryanzhang-oss:
we need an abandoning state in the rollout to handle user abandon the rollout in the middle of it

### DIFF
--- a/test/e2e-test/rollout_plan_test.go
+++ b/test/e2e-test/rollout_plan_test.go
@@ -351,7 +351,7 @@ var _ = Describe("Cloneset based rollout tests", func() {
 				k8sClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: newAppRollout.Name}, &appRollout)
 				return appRollout.Status.RollingState
 			},
-			time.Second*10, time.Millisecond*50).Should(BeEquivalentTo(oamstd.RollingInBatchesState))
+			time.Second*10, time.Millisecond).Should(BeEquivalentTo(oamstd.RollingInBatchesState))
 
 		By("Pause the rollout")
 		Eventually(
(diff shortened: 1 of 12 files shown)


wonderflow (approved):
LGTM

test/e2e-test/rollout_plan_test.go
```diff
@@ -426,7 +426,7 @@ var _ = Describe("Cloneset based rollout tests", func() {
 				appRollout.Spec.TargetAppRevisionName = utils.ConstructRevisionName(app.GetName(), 2)
 				appRollout.Spec.RolloutPlan.BatchPartition = nil
 				return k8sClient.Update(ctx, &appRollout)
-			}, time.Second*5, time.Millisecond*500).Should(Succeed())
+			}, time.Second*5, time.Millisecond).Should(Succeed())
```
wonderflow, line 429: two frequency? `time.Second`?

test/e2e-test/rollout_plan_test.go
```diff
@@ -351,7 +351,7 @@ var _ = Describe("Cloneset based rollout tests", func() {
 				k8sClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: newAppRollout.Name}, &appRollout)
 				return appRollout.Status.RollingState
 			},
-			time.Second*10, time.Millisecond*50).Should(BeEquivalentTo(oamstd.RollingInBatchesState))
+			time.Second*10, time.Millisecond).Should(BeEquivalentTo(oamstd.RollingInBatchesState))
```
wonderflow, line 354: same do you mean second?

test/e2e-test/rollout_plan_test.go
ryanzhang-oss: one of the e2e failed because the state has passed the "rolling" before the check below...

test/e2e-test/rollout_plan_test.go
ryanzhang-oss: no, I meant to check faster but the same e2e failed anyway, need to see why

PR #1294, opened 2021-03-25
fix isControlledByApp function in AC controller

captainroy-hy:


ryanzhang-oss (approved):
Good catch

ryanzhang-oss (commented):
Please also fix the unit tests

yangsoon:
Good job!

Issue #1278, opened 2021-03-24
[Feature] Add extensions guides back when vela is ready


resouer:
**Is your feature request related to a problem? Please describe.**

In PR #1270 , I temporarily deleted these docs because they are not built-in capabilities anymore: https://github.com/oam-dev/kubevela/pull/1270/files#diff-30a5a76180d46829fe7d37ea60f40850d75858723ed7914ae3ea476a65d9c145L37

But we should add them back as tutorials of add existing community capabilities to KubeVela, especially for Flagger and KEDA.

The user story is like:
1. I found a useful capability in community,
2. I want to use it with my application abstraction,
3. I just need to define a template and install this definition in vela,
4. Then I can use it as trait in application.

Similar story also applies to components but I guess it's already supported.

/cc @wonderflow 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

wonderflow:
Yes, we've already supported , just add a tutorial document will fix this issue.

PR #1297, opened 2021-03-25
add extend component definition in KubeVela

wonderflow:
add extend component definition in KubeVela, fix part of #1278


zzxwill (approved):
LGTM

PR #1299, opened 2021-03-25
fix rollout e2e

captainroy-hy:
Maybe we need other fixes on webhooks. This PR just try to get rollout e2e pass. 


charts/vela-core/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
```diff
@@ -52,6 +52,7 @@ webhooks:
           - core.oam.dev
         apiVersions:
           - v1alpha2
```
ryanzhang-oss, line 54: we can actually move appRollout to v1beta1, no need to support v1alpha2, no one is using it.

captainroy-hy:
Close for all changes are included in #1303 .

wonderflow:
Thanks for the effort, really great @captainroy-hy

PR #1307, opened 2021-03-26
fix vela components/workloads and vela cap

yangsoon:
1. fix the problem which vela components/workloads repeat get components/workloads list.
2. `vela cap center` only deal with ComponentDefinition and TraitDefinition


yangsoon:
fix https://github.com/oam-dev/kubevela/issues/1262

wonderflow (approved):
LGTM

PR #1310, opened 2021-03-26
Add license header check script 

wangyikewxgm:
1、add license header check script  and add a CI check rule
2、add all missed header


wonderflow (approved):
LGTM

PR #1315, opened 2021-03-26
fix app upgrade without rollout and rollout plan will create multiple version resources

wonderflow:
fix https://github.com/oam-dev/kubevela/issues/1196


zzxwill (approved):
/lgtm

PR #1317, opened 2021-03-26
fix CI check-diff

### LINKED ISSUE #1311: Why application CRD description field have line break when it's too long
As title said,  but other CRDs don't. This broke my CI pipeline

https://github.com/oam-dev/kubevela/blob/20a31d4554466206fe939dae141eee379fe45a7f/charts/vela-core/crds/core.oam.dev_applications.yaml#L60

captainroy-hy:
In CI image "ubuntu-20.04", kustomize v4.0.5 is pre-installed. So current makefile skips installing v3.8.2
> {Version:kustomize/v4.0.5 GitCommit:9e8e7a7fe99ec9fbf801463e8607928322fc5245 BuildDate:2021-03-08T20:53:03Z GoOs:linux GoArch:amd64}

And diff from v3.8.2, customize v4 has the problem that break too long sentences. 
That will fail any PR containing changes on crds/application.yml while author's local kustomize is truly v3.8.2.

to fix #1311 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -277,7 +277,7 @@ KUSTOMIZE_VERSION ?= 3.8.2
 
 .PHONY: kustomize
 kustomize:
-ifeq (, $(shell which kustomize))
+ifeq (, $(shell kustomize version | grep $(KUSTOMIZE_VERSION)))
 	@{ \
 	set -e ;\
 	echo 'installing kustomize-v$(KUSTOMIZE_VERSION)' ;\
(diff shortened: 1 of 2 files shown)


wonderflow:
So do you mean all the contributors need to update their local Kustomize?

Makefile
```diff
@@ -277,7 +277,7 @@ KUSTOMIZE_VERSION ?= 3.8.2
 
 .PHONY: kustomize
 kustomize:
-ifeq (, $(shell which kustomize))
+ifeq (, $(shell kustomize version | grep $(KUSTOMIZE_VERSION)))
```
wonderflow, line 280: It's really smart to do so!

wonderflow (approved):
Good job!

PR #1320, opened 2021-03-26
add appdeployment control logic

hongchaodeng:
This is part 2 of splitting https://github.com/oam-dev/kubevela/pull/1193 . part 1: #1272 . Ref issue: https://github.com/oam-dev/kubevela/issues/1144

- This PR adds the appDeployment control logic, so it implements multi-revision and multi-cluster features.
- Instead of relying on AppConfig and Component, the controller now renders k8s resources directly (see workloadRenderer).
- It also adds an integration test to cover the main logic.
- It relies on appRevision label which has a bug (https://github.com/oam-dev/kubevela/issues/1253). But this won't affect the main logic testing since we are generating AppRevisions there.


hongchaodeng:
@wonderflow All tests passed

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
```diff
@@ -0,0 +1,561 @@
...
+			d.Del = append(d.Del, newRevision(p.RevisionName, c.ClusterName, c.Replicas))
+		}
+	}
+	return d
+}
+
+// UpdateStatus updates AppDeployment's Status with retry.RetryOnConflict
+func (r *Reconciler) updateStatus(ctx context.Context, appd *oamcore.AppDeployment, opts ...client.UpdateOption) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() (err error) {
+		return r.Client.Status().Update(ctx, appd, opts...)
```
wonderflow, line 445: do we need to get the newer and update the resourceVersion before retry?

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -152,14 +156,18 @@ func (h *appHandler) apply(ctx context.Context, appRev *v1beta1.ApplicationRevis
...
+		if _, exist := h.app.GetAnnotations()[oam.AnnotationAppRollout]; !exist && h.app.Spec.RolloutPlan == nil {
+			h.setInplace(true)
+			return h.createOrUpdateAppContext(ctx, owners)
+		}
+		h.setInplace(false)
 	}
-	h.setInplace(false)
+
 	return nil
 }
```
wonderflow, line 173: ping @ryanzhang-oss  please  also review here

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -107,27 +107,31 @@ func (h *appHandler) apply(ctx context.Context, appRev *v1beta1.ApplicationRevis
...
-		// find the ACC that contains this component
-		for i := 0; i < len(ac.Spec.Components); i++ {
-			// update the AC using the component revision instead of component name
-			// we have to make AC immutable including the component it's pointing to
-			if ac.Spec.Components[i].ComponentName == newComp.Name {
-				ac.Spec.Components[i].RevisionName = revisionName
-				ac.Spec.Components[i].ComponentName = ""
+	isRevisionOnly := ac.Annotations[oam.AnnotationAppRevisionOnly] == "true"
+	// don't create components if revision-only annotation is set
+	if !isRevisionOnly {
```
wonderflow, line 112: I think we can make a function for the revisionOnly case, so the code could more clear like:

```
if isRevisionOnly{
    revisionOnlyApply(..)
}

the reset code...
```

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
```diff
@@ -0,0 +1,561 @@
...
+
+	key := client.ObjectKey{
+		Name:      c.Spec.KubeconfigSecretRef.Name,
+		Namespace: ns,
+	}
+	secret := &corev1.Secret{}
+	if err := r.Client.Get(ctx, key, secret); err != nil {
+		return nil, err
+	}
+	return clustermanager.GetClient(secret.Data["config"])
```
wonderflow, line 201: please make a const var for it. What about naming it `kubeconfig` or `token`?

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
```diff
@@ -0,0 +1,561 @@
...
+		Name:      name,
+		Namespace: ns,
+	}
+	if err := r.Client.Get(ctx, key, &obj); err != nil {
+		return nil, err
+	}
+	return &obj, nil
+}
+
+func (r *Reconciler) getWorkloadsForRevision(ctx context.Context, revName, ns string) ([]*workload, error) {
```
wonderflow, line 304:
```suggestion
func (r *Reconciler) getWorkloadsFromRevision(ctx context.Context, revName, ns string) ([]*workload, error) {
```

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
```diff
@@ -0,0 +1,561 @@
...
+		if isHostCluster(rev.ClusterName) {
+			kubecli = r.Client
+		} else {
+			kubecli, err = r.getClientForCluster(ctx, rev.ClusterName, appd.Namespace)
+			if err != nil {
+				return err
+			}
+		}
+
+		for _, wl := range workloads {
```
wonderflow, line 223: There's a case, that the workload instance is resued in the new revision.  If so , it will be deleted here while it should still be in use.

 I think we need to check for it.

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
```diff
@@ -0,0 +1,561 @@
...
+		For(&oamcore.AppDeployment{}).
+		Complete(r)
+}
+
+func (r *Reconciler) applyTraffic(ctx context.Context, appd *oamcore.AppDeployment) (err error) {
+
+	vsvc := &istioclientv1beta1.VirtualService{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: istioclientv1beta1.SchemeGroupVersion.String(),
+			Kind:       "VirtualService",
```
wonderflow, line 461: Do we have any extensibility for other service mesh implementation here? I means, if we wan to support e.g. OSM, how to extend the code.

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
```diff
@@ -0,0 +1,561 @@
...
+			continue
+		}
+		result = append(result, item)
+	}
+	return
+}
+
+// Setup adds a controller that reconciles AppDeployment.
+func Setup(mgr ctrl.Manager, _ controller.Args, _ logging.Logger) error {
+	dm, err := discoverymapper.New(mgr.GetConfig())
```
wonderflow, line 555: the discoverymapper is in `controller.Args`

wonderflow (commented):
Really Great Job, I think it's generally LGTM, ping @ryanzhang-oss  for a second round review

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
hongchaodeng: Removing retry here. The controller is able to reconcile.

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
hongchaodeng: Sure

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
hongchaodeng: I can't think of any case like that. Can you give a more concrete example?

pkg/controller/core.oam.dev/v1alpha2/appdeployment/appdeployment_controller.go
hongchaodeng: Yes. I think there should be a service mesh interface here to manage traffic. This will need more user input to design a generic interface which we should not over-optimize for now.

ryanzhang-oss (approved):

PR #1321, opened 2021-03-26
fix e2e test of kube schematic

captainroy-hy:
fix to catch up with the change related to workload name mechanism.


wonderflow (approved):
LGTM after CI

PR #1322, opened 2021-03-27
Rewrite the doc for 1.0 release

resouer:
Rewrite the doc for 1.0 release.

/cc @wonderflow


docs/en/_sidebar.md
```diff
@@ -11,26 +11,36 @@
...
+  - Defining Traits
+  - [How-to](/en/cue/trait.md)
+  - [Patch Traits](/en/cue/patch-trait.md)
+  - [Status Write Back](/en/cue/status.md)
+  - [Advanced Features](/en/cue/advanced.md)
+
+  - Hands-on Lab
+  - [Debug, Test and Dry-run](/en/platform-engineers/debug-test-cue.md)
+  - [Defining KEDA as Autoscaling Trait](/en/platform-engineers/keda.md)  
+<!-- - [Defining Cloud Service](/en/platform-engineers/cloud-services.md) -->
```
wonderflow, line 42: we have finished the cloud service docs now, please unquote it

resouer: I have read the doc and I don't think it's ready, especially this one:

> ** Note: We highly recommend that you should split the cloud resource provision and consuming in different applications.** ** Because the cloud resources can have standalone Lifecycle Management.** 

The recommended way (at least for RDS use case) should be defining the workload and cloud service in same YAML and let the platform to handle everything.

resouer: And the data binding mechanism demonstrated in that doc is too restricted and barely usable (i.e. you have to create a new `ComponentDefinition` for different Secret key). The definition as a template should be very generic.

The old approach (a template for `valueFrom`) makes more sense (many many charts have similar approach) if we do not have a generic service binding capability.

xref: https://github.com/vmware-labs/service-bindings

resouer:
The conflict has been fixed.

wonderflow (approved):
LGTM

PR #1323, opened 2021-03-27
propagate AppRevision Into context.AppRevision 

### LINKED ISSUE #1253: context.AppRevision is not propagated correctnly
The AppRevision should be propagated to the `context.AppRevision`: https://github.com/oam-dev/kubevela/blob/379e1432221b00baa433ddf595cdcbc369f1d393/pkg/dsl/process/handle.go#L24

Nonetheless, currently when app controller reconciles, it calls GenerateAppFile() where the RevisionName is not set: https://github.com/oam-dev/kubevela/blob/379e1432221b00baa433ddf595cdcbc369f1d393/pkg/appfile/parser.go#L143-L156

Then the generateComponentFromCUEModule() function will pass in empty revision string: https://github.com/oam-dev/kubevela/blob/379e1432221b00baa433ddf595cdcbc369f1d393/pkg/appfile/parser.go#L252

wonderflow:
fix #1253

- [x] refactor: calculate revision first before render AC and Components
- [x] Add AppRevision Into conext

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
@@ -102,6 +102,16 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	app.Status.SetConditions(readyCondition("Parsed"))
 	handler.appfile = generatedAppfile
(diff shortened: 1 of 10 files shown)


hongchaodeng (approved):
lgtm

PR #1325, opened 2021-03-27
requeue soon after error occurs

### LINKED ISSUE #1188: requeue mechanism when error occurs in reconciliation
AppConfig controller uses its own mechanism to re-queue obj during reconciliation. Once an error occurs, an error condition will be added into `appconfig.status`. And status changing will trigger requeue then reconcile again.
But in the situation where reconcile proceed very fast, the same error occurs, and the new error condition is created.
Since the duration time is very small, the timestamp(lastTransactionTime) in the error condition is the same as last time. That means, no change happens to the appconfig status, so it will not trigger requeue.
image

Then the appconfig controller has to wait a bit long time until an informer re-sync happens. Currently, the default informer resync interval is 5min. 
`	flag.DurationVar(&syncPeriod, "informer-re-sync-interval", 5*time.Minute,
		"controller shared informer lister full re-sync period")
`

#### solution
If error occurs, track it into resource's status (condition) to trigger requeue.
To avoid the issue aforementioned, check the previous error msg before setting status.

captainroy-hy:
to fix #1188 
For helm based app, it really replies on reconcile in a short time until the workload is found.
So I add this util that allows an error to request a reconciliation in a short time


wonderflow:
ping @ryanzhang-oss  please review

pkg/controller/core.oam.dev/v1alpha2/applicationconfiguration/applicationconfiguration.go
```diff
@@ -250,7 +252,14 @@ func (r *OAMApplicationReconciler) Reconcile(req reconcile.Request) (reconcile.R
 // ACReconcile contains all the reconcile logic of an AC, it can be used by other controller
 func (r *OAMApplicationReconciler) ACReconcile(ctx context.Context, ac *v1alpha2.ApplicationConfiguration,
 	log logging.Logger) (result reconcile.Result) {
```
wonderflow, line 254: the key point is we need to refactor the return value here. We should return the reconcileErr in this function, and let the outer logic to handle.

captainroy-hy:
Closed. And will figure out a better solution in the future.

captainroy-hy:
Close for working on a better solution.

PR #1327, opened 2021-03-27
fix vela show align with ComponentDefinition

yangsoon:
as title


.github/workflows/go.yml
```diff
@@ -1,4 +1,4 @@
-name: Go
+  name: Go
```
wonderflow, line 1: please fix the format to let the workflow run

wonderflow, line 1:
```suggestion
name: Go
```

wonderflow (approved):
LGTM

PR #1328, opened 2021-03-28
action for auto docs

sunny0826:
Required document structure
- docs/
  - en/
  - zh-CN/
  - sidebar.js

### DIFF
--- a/.github/workflows/docs-versioning.yml
+++ b/.github/workflows/docs-versioning.yml
@@ -1,0 +1,17 @@
+name: docs-versioning
+on:
+  push:
+    branches:
+      - 'release-**'
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-versioning.yml
```diff
@@ -0,0 +1,17 @@
+name: docs-versioning
+on:
+  push:
+    branches:
+      - 'release-**'
```
wonderflow, line 5:
```suggestion
      - 'release-**'
      - master
    paths:
      - 'docs/**'      
```

wonderflow: let's merge them in one file

wonderflow:
prefer #1331

PR #1338, opened 2021-03-29
fix install-kind cmd in docs and add vela components related doc

yangsoon:
1. add vela_components.md
2. fix install-kind cmd


wonderflow (approved):
LGTM

PR #1353, opened 2021-03-29
add AppRevision label to cue templates

hongchaodeng:


hongchaodeng:
I'm closing this one. I figured out deployment.spec.selector field is immutable. We might need to add a new workload for AppDeployment case.

PR #1354, opened 2021-03-29
use ApplicationContext as the owner of all the resources if it controlled by Application

wonderflow:
as the title said

fix https://github.com/oam-dev/kubevela/issues/1314


hongchaodeng (approved):
lgtm

PR #1362, opened 2021-03-30
add initial finalizer and abandon support

ryanzhang-oss:
add initial finalizer and abandon support

need to completely separate out the initial target/source app logic for the deleting and abandoning state later


pkg/controller/core.oam.dev/v1alpha2/applicationrollout/applicationrollout_controller.go
```diff
@@ -123,19 +128,26 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (res reconcile.Result, retErr e
...
+		} else {
+			// so that we don't think we are modified again
+			appRollout.Status.LastUpgradedTargetAppRevision = targetAppRevisionName
+			appRollout.Status.LastSourceAppRevision = sourceAppRevisionName
 		}
 		appRollout.Status.StateTransition(v1alpha1.RollingModifiedEvent)
 	}
 
-	// Get the source application
+	//TODO: handle deleting/abandoning state differently
```
wonderflow, line 139: what does it mean? In which case we need this?

wonderflow (approved):

pkg/controller/core.oam.dev/v1alpha2/applicationrollout/applicationrollout_controller.go
ryanzhang-oss: the finalizer will stuck if the underlying resources (appRevision, etc) are deleted

PR #1371, opened 2021-03-31
fix: add "system-definition-namespace" flag to container args

kinsolee:
The `system-definition-namespace` flag of `vela-core` should be added into `kubevela-controller.yaml` of the helm chart, and specified with chart value `systemDefinitionNamespace`, so that the parser can find the definitions from the same namespace which defined in the `ComponentDefinition/TraitDefinition`.

Without this, Application cannot be processed properly if the definition crds are not installed in `vela-system` namespace:
```
kubectl get application first-vela-app -o yaml
kind: Application
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"core.oam.dev/v1beta1","kind":"Application","metadata":{"annotations":{},"name":"first-vela-app","namespace":"default"},"spec":{"components":[{"name":"express-server","properties":{"image":"crccheck/hello-world","port":8000},"traits":[{"properties":{"domain":"testsvc.example.com","http":{"/":8000}},"type":"ingress"}],"type":"webservice"}]}}
  creationTimestamp: "2021-03-31T03:02:04Z"
  generation: 1
spec:
  components:
  - name: express-server
    properties:
      image: crccheck/hello-world
      port: 8000
    traits:
    - properties:
        domain: testsvc.example.com
        http:
          /: 8000
      type: ingress
    type: webservice
status:
  batchRollingState: ""
  conditions:
  - lastTransitionTime: "2021-03-31T03:02:04Z"
    message: 'fetch type of express-server: LoadTemplate from workloadDefinition [webservice] : WorkloadDefinition.core.oam.dev "webservice" not found'
    reason: ReconcileError
    status: "False"
    type: Parsed
  currentBatch: 0
  rollingState: ""
  status: rendering
  upgradedReadyReplicas: 0
  upgradedReplicas: 0
```

### DIFF
--- a/charts/vela-core/templates/kubevela-controller.yaml
+++ b/charts/vela-core/templates/kubevela-controller.yaml
@@ -120,6 +120,7 @@ spec:
             {{ if ne .Values.disableCaps "" }}
             - "--disable-caps={{ .Values.disableCaps }}"
             {{ end }}
+            - "--system-definition-namespace={{ .Values.systemDefinitionNamespace }}"
           image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
           imagePullPolicy: {{ quote .Values.image.pullPolicy }}
           resources:


wonderflow (approved):
good catch! thanks

PR #1374, opened 2021-03-31
fix cannot update cross namespace workload application issue

### LINKED ISSUE #1372: Cross namespace workload cannot update
Workload  in a namespace that is different with application  cannot be updated. An reconcile error will set in appContext status.
image

wangyikewxgm:
fix #1372


wonderflow (approved):
LGTM

PR #1393, opened 2021-04-01
Remove outputSecretName in cloud service provisoning and consuming

zzxwill:
Refer to the reason https://github.com/oam-dev/kubevela/issues/1128#issuecomment-811590631


wonderflow (approved):
thanks

PR #1396, opened 2021-04-01
add missing newline for configs in BaseContextFile

plpan:


pkg/dsl/process/handle.go
```diff
@@ -146,7 +146,7 @@ func (ctx *templateContext) BaseContextFile() string {
 
 	if len(ctx.configs) > 0 {
 		bt, _ := json.Marshal(ctx.configs)
-		buff += ConfigFieldName + ": " + string(bt)
+		buff += ConfigFieldName + ": " + string(bt) + "\n"
```
wonderflow, line 149: good catch, please also add a new line for outputSecretName in  L160

wonderflow (approved):

pkg/dsl/process/handle.go
plpan: ok

PR #1413, opened 2021-04-02
Add unit-tests for ApplicationRollout related packages

zzxwill:
Part I: pkg/controller/common/rollout

### DIFF
--- a/pkg/controller/common/rollout/workloads/cloneset_rollout_controller.go
+++ b/pkg/controller/common/rollout/workloads/cloneset_rollout_controller.go
@@ -188,6 +188,11 @@ func (c *CloneSetRolloutController) CheckOneBatchPods(ctx context.Context) (bool
 	newPodTarget := calculateNewBatchTarget(c.rolloutSpec, 0, int(cloneSetSize), int(c.rolloutStatus.CurrentBatch))
 	// get the number of ready pod from cloneset
 	readyPodCount := int(c.cloneSet.Status.UpdatedReadyReplicas)
+	if len(c.rolloutSpec.RolloutBatches) <= int(c.rolloutStatus.CurrentBatch) {
+		klog.Info("somehow, currentBatch number exceeded the rolloutBatches spec.", "current batch",
+			c.rolloutStatus.CurrentBatch)
+		return false, nil
+	}
 	currentBatch := c.rolloutSpec.RolloutBatches[c.rolloutStatus.CurrentBatch]
 	unavail := 0
 	if currentBatch.MaxUnavailable != nil {
--- a/pkg/controller/common/rollout/workloads/deployment_controller.go
+++ b/pkg/controller/common/rollout/workloads/deployment_controller.go
@@ -73,9 +73,7 @@ func (c *DeploymentController) VerifySpec(ctx context.Context) (bool, error) {
 	defer func() {
 		if verifyErr != nil {
 			klog.Error(verifyErr)
-			if c.recorder != nil && c.parentController != nil {
-				c.recorder.Event(c.parentController, event.Warning("VerifyFailed", verifyErr))
-			}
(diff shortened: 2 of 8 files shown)


zzxwill:
@ryanzhang-oss I slightly modified some code, please help review it:)

pkg/controller/common/rollout/workloads/deployment_controller.go
```diff
@@ -73,15 +73,19 @@ func (c *DeploymentController) VerifySpec(ctx context.Context) (bool, error) {
 	defer func() {
 		if verifyErr != nil {
 			klog.Error(verifyErr)
-			c.recorder.Event(c.parentController, event.Warning("VerifyFailed", verifyErr))
+			if c.recorder != nil && c.parentController != nil {
```
ryanzhang-oss, line 76: when will the recorder be nil?

pkg/controller/common/rollout/workloads/deployment_controller.go
```diff
@@ -73,15 +73,19 @@ func (c *DeploymentController) VerifySpec(ctx context.Context) (bool, error) {
...
+			}
 		}
 	}()
 
 	// check if the rollout spec is compatible with the current state
 	targetTotalReplicas, verifyErr := c.calculateTargetTotalSize(ctx)
 	if verifyErr != nil {
 		// do not fail the rollout just because we can't get the resource
-		c.rolloutStatus.RolloutRetry(verifyErr.Error())
+		if c.rolloutSpec != nil {
```
ryanzhang-oss, line 86: when will the spec be nil?

pkg/controller/common/rollout/workloads/deployment_controller.go
ryanzhang-oss: btw, the deployment rollout is not really finished, it doesn't cover many corner cases (i.e. revert in the middle of rollout)

zzxwill: @ryanzhang-oss From the normal logic, application rollout plan controller --> rollout plan controller --> deployment controller, the recorder won't be nil.
But from the filed type, it's possible.

>btw, the deployment rollout is not really finished, it doesn't cover many corner cases (i.e. revert in the middle of rollout)

Got it, I will focus on adding unit-tests on Cloneset.

pkg/controller/common/rollout/workloads/deployment_controller.go
zzxwill: Same reason as above. Should I remove the check nil logic?

ryanzhang-oss: yeah, we should not check (and allow) conditions that should not happen.

pkg/controller/common/rollout/workloads/cloneset_rollout_controller.go
```diff
@@ -188,6 +188,11 @@ func (c *CloneSetRolloutController) CheckOneBatchPods(ctx context.Context) (bool
 	newPodTarget := calculateNewBatchTarget(c.rolloutSpec, 0, int(cloneSetSize), int(c.rolloutStatus.CurrentBatch))
 	// get the number of ready pod from cloneset
 	readyPodCount := int(c.cloneSet.Status.UpdatedReadyReplicas)
+	if len(c.rolloutSpec.RolloutBatches) <= int(c.rolloutStatus.CurrentBatch) {
+		klog.Info("somehow, currentBatch number exceeded the rolloutBatches spec.", "current batch",
+			c.rolloutStatus.CurrentBatch)
+		return false, nil
+	}
```
zzxwill, line 195: @ryanzhang-oss I kept this checking and removed all other nil checking.

pkg/controller/common/rollout/workloads/deployment_controller.go
zzxwill: Okay, removed.

PR #1414, opened 2021-04-02
Fix the bug that the registered k8s built-in gvk does not exist

leejanee:
fix the bug that the registered k8s built-in gvk does not exist

### DIFF
--- a/pkg/dsl/definition/package.go
+++ b/pkg/dsl/definition/package.go
@@ -105,7 +105,10 @@ func (pd *PackageDiscover) Exist(gvk metav1.GroupVersionKind) bool {
 	importPath := genStandardPkgName(dgvk)
 	pd.mutex.RLock()
 	defer pd.mutex.RUnlock()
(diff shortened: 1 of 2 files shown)


pkg/dsl/definition/package.go
```diff
@@ -105,7 +105,10 @@ func (pd *PackageDiscover) Exist(gvk metav1.GroupVersionKind) bool {
 	importPath := genStandardPkgName(dgvk)
 	pd.mutex.RLock()
 	defer pd.mutex.RUnlock()
-	pkgKinds := pd.pkgKinds[importPath]
+	pkgKinds, ok := pd.pkgKinds[importPath]
+	if !ok {
+		pkgKinds = pd.pkgKinds[genOpenPkgName(dgvk)]
```
wonderflow, line 110: please write a test for it

wonderflow (approved):
LGTM

PR #1415, opened 2021-04-02
update: add helm test step

charles-woshicai:
Create an action to test helm charts, as required in https://github.com/oam-dev/kubevela/issues/1406

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -167,6 +167,8 @@ jobs:
         run: |
           make e2e-cleanup
           make e2e-setup
+          helm lint ./charts/vela-core
+          helm test -n vela-system kubevela --timeout 5m
 
       - name: Wait for e2e preparation ready
         run: |
--- a/charts/vela-core/templates/test/test-application.yaml
+++ b/charts/vela-core/templates/test/test-application.yaml
@@ -1,0 +1,60 @@
+apiVersion: core.oam.dev/v1beta1
+kind: Application
+metadata:
+  annotations:
+    helm.sh/hook: test-success
+  name: first-vela-app
+spec:
+  components:
+    - name: express-server
+      type: webservice
+      properties:
+        image: crccheck/hello-world
+        port: 8000
+      traits:
+        - type: ingress
+          properties:
+            domain: testsvc.example.com
+            http:
+              "/": 8000
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ .Release.Name }}-application-test"
+  annotations:
+    "helm.sh/hook": test
+spec:
+  serviceAccountName: kubevela-vela-core
+  containers:
+    - name: {{ .Release.Name }}-application-test
+      image: alpine/k8s:1.18.2
+      imagePullPolicy: IfNotPresent
+      command:
+        - /bin/bash
+        - -ec
+        - |
+
+          set -e
+
+          echo "Waiting application is ready..."
+
+          echo "waiting for application being Applied"
+          kubectl -n vela-system wait --for=condition=Applied application first-vela-app --timeout=3m
+          echo "application being Applied"
+
+          # wait for deploy being created
+          echo "waiting for deployment being available"
+          kubectl -n vela-system wait --for=condition=available deploy express-server --timeout 3m
+          echo "deployment being available"
+
+          # wait for ingress being created
+          while ! [ `kubectl -n vela-system get ing express-server | grep -v NAME | wc -l` = 1 ]; do
+            echo "waiting for ingress being created"
+            sleep 1
+          done
+
+
+
+          echo "Application and its components are created"
+  restartPolicy: Never


charles-woshicai:
Add an action by following the doc of [this page](https://github.com/marketplace/actions/helm-chart-testing-helm-3)

charles-woshicai:
Use chart-test-action to test charts, [here](https://github.com/helm/chart-testing-action) is the docs.

charles-woshicai:
Review needed @wonderflow .

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          python-version: 3.7
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.0.1
+
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(ct list-changed --chart-dirs charts)
```
wonderflow, line 36: what is the comand `ct` ?

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,55 @@
...
+        uses: helm/kind-action@v1.0.0
+        if: steps.list-changed.outputs.changed == 'true'
+        with:
+          node_image: kindest/node:v1.18.15
+
+      # don't run `ct install` if charts hasn't been changed
+      - name: Run chart-testing (install)
+        run: |
+          kubectl create ns vela-system
+          ct install --namespace vela-system
```
wonderflow, line 54: It should have a `helm test` runing here to validate, I didn't see them?

https://helm.sh/docs/topics/chart_tests/

charts/vela-core/templates/test/test-application.yaml
```diff
@@ -0,0 +1,19 @@
...
+      type: webservice
+      properties:
+        image: crccheck/hello-world
+        port: 8000
+      traits:
+        - type: ingress
+          properties:
+            domain: testsvc.example.com
+            http:
+              "/": 8000
```
wonderflow, line 19: As the docs said, the test should be a K8s job with command running.  https://helm.sh/docs/topics/chart_tests/

wonderflow (changes requested):
We could add the chart test in `e2e-test` instead of  creating new one

.github/workflows/lint-test.yaml
charles-woshicai: `ct` is equal to `chart testing`, a command-line tool which is maintained by `helm`.

[Here](https://github.com/helm/chart-testing) is the repository of `ct`.

.github/workflows/lint-test.yaml
charles-woshicai: `ct install` would run both helm `install` and `test`.

[Here](https://github.com/helm/chart-testing/blob/master/doc/ct_install.md#ct-install) is the doc ref.

charts/vela-core/templates/test/test-application.yaml
charles-woshicai: > We could add the chart test in `e2e-test` instead of creating new one

Now we detect whtehre there is a change in charts via `ct list-change`, which comares the charts between `master` branch and `PR` branch. If the charts is not changed,  there is no necessary to run `helm install & helm test`. 

Is it necessary to run `helm install & helm test` in `e2e-test`?

> As the docs said, the test should be a K8s job with command running. https://helm.sh/docs/topics/chart_tests/

I will fix it when I am free.

wonderflow: > Is it necessary to run helm install & helm test in e2e-test?

I think we can always run the check. Does it cost too much time in CI?

.github/workflows/lint-test.yaml
wonderflow: I see, thanks, but use a yaml about `Application` seems not check the check well.

wonderflow: We should at least check the Application is running and has a deployment and ingress created

charles-woshicai: I agree, an application just test whether crd is installed.

Will fix it.

charts/vela-core/templates/test/test-application.yaml
charles-woshicai: > > Is it necessary to run helm install & helm test in e2e-test?
> 
> I think we can always run the check. Does it cost too much time in CI?

I review the charts and it makes sense to run check every time for the image tag of core controller is `latest` in the master branch.

Will adopt to move charts check to `e2e-test`.

.github/workflows/go.yml
```diff
@@ -132,6 +132,30 @@ jobs:
...
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.0.1
+
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.0.0
```
wonderflow: the kind already setup in L174

.github/workflows/go.yml
```diff
@@ -132,6 +132,30 @@ jobs:
...
+        with:
+          node_image: kindest/node:v1.18.15
+
+      # don't run `ct install` if charts hasn't been changed
+      - name: Run chart-testing (install)
+        run: |
+          kubectl create ns vela-system
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/kind/deploy.yaml
```
wonderflow: why need to apply this file?

charles-woshicai: FYI the code is not the final state now.

I want to check the `PHASE` and `HEALTH` of `applicatoin` to be `running` and `healthy` in test pod.
In my local environment the `HEALTH` of application stuck in `checkHealth`, this line is just having a try. If it's no necessary, I will  remove it.

charles-woshicai:
Review needed. @wonderflow 
I just chcek whether vela-controller created `depoloy` and `ing` without waiting for `application` ready.

Could I rebase total commits in this PR to one?  There  are too many debug commits, which is useless.

wonderflow:
> BTW, could I rebase total commits in this PR to one? There are too many debug commits, which is useless.

Sure, please go ahead. I can also help squash before merge.

.github/workflows/go.yml
```diff
@@ -147,6 +147,31 @@ jobs:
...
+          helm lint ./charts/vela-core
+
+      - name: helm install
+        run: |
+          helm install --create-namespace -n vela-system kubevela ./charts/vela-core/ --timeout 15m
+
+      - name: helm test
+        run: |
+          helm test -n vela-system kubevela --timeout 5m
+          kubectl -n vela-system logs kubevela-application-test
```
wonderflow, line 173: please check the `make e2e-setup` command, in the makefile, it has already install the helm chart. We can just merge `helm lint` and `helm test` into the `e2e-setup`.

wonderflow, line 173: BTW, why check the logs?

charts/vela-core/templates/test/test-application.yaml
```diff
@@ -0,0 +1,62 @@
...
+          while ! [ `kubectl -n vela-system get deploy express-server | grep -v NAME | wc -l` = 1 ]; do
+            echo "waiting for deployment being created"
+            sleep 1
+          done
+
+          # wait for ingress being created
+          while ! [ `kubectl -n vela-system get ing express-server | grep -v NAME | wc -l` = 1 ]; do
+            echo "waiting for ingress being created"
+            sleep 1
+          done
```
wonderflow, line 57: will the `helm test` fail if the job is hanging on sleep? Do we have any timeout settings?

.github/workflows/go.yml
```diff
@@ -147,6 +147,31 @@ jobs:
           version: ${{ env.KIND_VERSION }}
           skipClusterCreation: true
 
+#      - name: Set up Helm
+#        uses: azure/setup-helm@v1
+#        with:
+#          version: v3.4.0
```
wonderflow, line 153: please clean these unused code

.github/workflows/go.yml
charles-woshicai: It's not necessary, I will remove it.

charles-woshicai: OK, I will take a look.

charts/vela-core/templates/test/test-application.yaml
charles-woshicai: Yeah the timeout is `5m` for now.

wonderflow: That's good

charles-woshicai:
I have rebased all commits into one.

wonderflow (approved):
LGTM after CI passed

wonderflow:
@charles-woshicai great job, thanks for your contributions!

PR #1431, opened 2021-04-08
stop zh-CN docs sync

sunny0826:
Stop Chinese document synchronization for translation activities.


wonderflow (approved):
LGTM, it's better to delete these lines

PR #1436, opened 2021-04-08
fix CI: Add clean-image action to clean up dangling docker image

yangsoon:
fix https://github.com/oam-dev/kubevela/issues/1429 

add timed task, clean up dangling docker image

### DIFF
--- a/.github/workflows/timed-task.yml
+++ b/.github/workflows/timed-task.yml
@@ -1,0 +1,10 @@
+name: Timed Task
+on:
+  schedule:
+    - cron: '0 16 * * *'
+jobs:
+  clean-image:
+    runs-on: aliyun
+    steps:
+      - name: Cleanup image
+        run: docker image prune -f


.github/workflows/timed-task.yml
```diff
@@ -0,0 +1,10 @@
+name: Timed Task
+on:
+  schedule:
+    - cron: '0 16 * * *'
+jobs:
+  clean-image:
+    runs-on: aliyun
+    steps:
+      - name: Cleanup image
+        run: docker rmi $(docker images --filter "dangling=true" -q --no-trunc)
```
chengshiwen, line 10: If `docker images --filter "dangling=true" -q --no-trunc` return `empty`, error info `"docker rmi" requires at least 1 argument` will occur. Maybe `docker image prune -f` is enough

chengshiwen (changes requested):

.github/workflows/timed-task.yml
wonderflow, line 10:
```suggestion
        run: docker image prune -f
```

yangsoon: good suggestion

wonderflow (approved):

PR #1441, opened 2021-04-09
fix webhooks and try to fix e2e

ryanzhang-oss:
1. Add default for rolloutStrategy
2. Attempt to fix flaky e2e validation

### DIFF
--- a/apis/standard.oam.dev/v1alpha1/rollout_state.go
+++ b/apis/standard.oam.dev/v1alpha1/rollout_state.go
@@ -265,9 +265,12 @@ func (r *RolloutStatus) StateTransition(event RolloutEvent) {
 	}
 	// special handle modified event here
 	if event == RollingModifiedEvent {
+		if r.RollingState == RolloutDeletingState {
+			panic(fmt.Errorf(invalidRollingStateTransition, rollingState, event))
+		}
 		if r.RollingState == RolloutFailedState || r.RollingState == RolloutSucceedState {
 			r.ResetStatus()
-		} else if r.RollingState != RolloutDeletingState {
+		} else {
 			r.SetRolloutCondition(NewNegativeCondition(r.getRolloutConditionType(), "Rollout Spec is modified"))
 			r.RollingState = RolloutAbandoningState
 			r.BatchRollingState = BatchInitializingState
--- a/apis/standard.oam.dev/v1alpha1/rollout_state.go
+++ b/apis/standard.oam.dev/v1alpha1/rollout_state.go
@@ -242,7 +242,11 @@ func (r *RolloutStatus) SetRolloutCondition(new runtimev1alpha1.Condition) {
 	if !exists {
 		r.Conditions = append(r.Conditions, new)
 	}
+}
 
+// we can't panic since it will crash the other controllers
+func (r *RolloutStatus) illegalStateTransition(err error) {
+	r.RolloutFailed(err.Error())
 }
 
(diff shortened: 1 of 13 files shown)


apis/standard.oam.dev/v1alpha1/rollout_state.go
```diff
@@ -265,9 +265,12 @@ func (r *RolloutStatus) StateTransition(event RolloutEvent) {
 	}
 	// special handle modified event here
 	if event == RollingModifiedEvent {
+		if r.RollingState == RolloutDeletingState {
+			panic(fmt.Errorf(invalidRollingStateTransition, rollingState, event))
```
wonderflow, line 269: Does it necessary to panic? Can we set some state that represent this `panic`? My concern is the approllout panic will affect the whole kubevela controller

pkg/controller/common/rollout/workloads/cloneset_scale_controller.go
```diff
@@ -272,12 +272,20 @@ func (s *CloneSetScaleController) Finalize(ctx context.Context, succeed bool) bo
...
 	for _, owner := range s.cloneSet.GetOwnerReferences() {
 		if owner.Kind == v1beta1.AppRolloutKind && owner.APIVersion == v1beta1.SchemeGroupVersion.String() {
+			isOwner = true
 			continue
 		}
 		newOwnerList = append(newOwnerList, owner)
 	}
+	if !isOwner {
+		// nothing to do if we are already not the owner
+		klog.InfoS("the cloneset is already released", "cloneSet", s.cloneSet.Name)
```
wonderflow, line 285:
```suggestion
		klog.InfoS("the cloneset is already released and not controlled by rollout", "cloneSet", s.cloneSet.Name)
```

pkg/controller/core.oam.dev/v1alpha2/applicationrollout/applicationrollout_controller.go
```diff
@@ -111,10 +112,12 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (res reconcile.Result, retErr e
...
-	// handle rollout target/source change,
-	if appRollout.Status.LastUpgradedTargetAppRevision != "" &&
-		appRollout.Status.LastUpgradedTargetAppRevision != targetAppRevisionName ||
-		(appRollout.Status.LastSourceAppRevision != "" && appRollout.Status.LastSourceAppRevision != sourceAppRevisionName) {
+	// handle rollout target/source change (only if it's not deleting already)
+	if appRollout.Status.RollingState != v1alpha1.RolloutDeletingState &&
+		((appRollout.Status.LastUpgradedTargetAppRevision != "" &&
+			appRollout.Status.LastUpgradedTargetAppRevision != targetAppRevisionName) ||
+			(appRollout.Status.LastSourceAppRevision != "" && appRollout.Status.
+				LastSourceAppRevision != sourceAppRevisionName)) {
```
wonderflow, line 120: this is too complex to read, please write a function

wonderflow (commented):
please consider the panic statement

apis/standard.oam.dev/v1alpha1/rollout_state.go
ryanzhang-oss: panic makes sense in terms of state transition but it is a problem when all controllers run in the same executable

wonderflow (approved):

PR #1442, opened 2021-04-09
fix controller: update crd from v1alpha2 to v1beta1 in ComponentDefinition/TraitDefinition Controller

yangsoon:
update crd(ComponentDefinition, TraitDefinition) from v1alpha2 to v1beta1


wonderflow (approved):
LGTM

Issue #1177, opened 2021-03-11
[Feature] request to change base image of  vela-core for debug convenience


shikanon:
**Is your feature request related to a problem? Please describe.**

The vela-core images to use `ENTRYPOINT `, we can not access to the container. When the vela-core have some problem, I cannot troubleshoot problems quickly.

For exmaple, when I install kubevela and meet the problem and I cannot not find problem easy:
```bash
$ kubectl --namespace vela-system port-forward $POD_NAME 8080:80
Forwarding from 127.0.0.1:8080 -> 80
Forwarding from [::1]:8080 -> 80
Handling connection for 8080
Handling connection for 8080
E0311 18:08:22.897968   20928 portforward.go:400] an error occurred forwarding 8080 -> 80: error forwarding port 80 to pod d07025c7648f5e2b6ece1fe8f3caf1210b462ce5b4df19826b0253a1176dfa89, uid : exit status 1: 2021/03/11 18:08:24 socat[31899] E connect(5, AF=2 127.0.0.1:80, 16): Connection refused
E0311 18:08:22.898883   20928 portforward.go:400] an error occurred forwarding 8080 -> 80: error forwarding port 80 to pod d07025c7648f5e2b6ece1fe8f3caf1210b462ce5b4df19826b0253a1176dfa89, uid : exit status 1: 2021/03/11 18:08:24 socat[31900] E connect(5, AF=2 127.0.0.1:80, 16): Connection refused
```

**Describe the solution you'd like**

Change  `CMD` of Dockerfile instead of `ENTRYPOINT`. And the yaml file of vela-core deployment use `command` instead of `args`.

shikanon:
@wonderflow can we provide other base images like alpine instead of "Distroless". when occur problem we cannot entry the container to network and service.

wonderflow:
I think this need to be discussed. WDYT?  ping @hongchaodeng @ryanzhang-oss

wonderflow:
It seems we can use `ubuntu:latest` as our base image, its size is very small now. 

https://hub.docker.com/_/ubuntu?tab=tags&page=1&ordering=last_updated

shikanon:
> It seems we can use `ubuntu:latest` as our base image, its size is very small now.
> 
> https://hub.docker.com/_/ubuntu?tab=tags&page=1&ordering=last_updated

yes, the command of ubuntu is more popular than alpine which is only 2MB. But if consider the balance of  image size and propularity, is alpine better choice?

[...]

Issue #1392, opened 2021-04-01
[Feature]Using the Kubevela CLI as docker image


sunny0826:
**Describe the solution you'd like**

Run kubevela cil with docker.

**Additional context**

Reference : [Using the official AWS CLI version 2 Docker image](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-docker.html)

PR #1445, opened 2021-04-09
Improve makefile and dockerfile

### LINKED ISSUE #1448: [Docker] Docker base image discussion
@wonderflow @shikanon 
After comparison and discussion as follows, `alpine` base image will be a better choice. And the `BASE_IMAGE` argument can be used for users to override the base image. So the current known needs have been met.

|                                            | Ubuntu   | Debian  | Alpine  | Distroless |
| ------------------------------------------ | -------- | ------- | ------- | ---------- |
| Popularity                                 | Low      | High    | High    | Medium     |
| Image Size (latest, uncompressed)          | 72.9MB   | 114MB   | 5.61MB  | 1.82MB     |
| Image Size (latest, compressed)            | 27.25 MB | 48.1 MB | 2.68 MB | 625KB      |
| Security                                   | Medium   | Medium  | High    | High       |
| Install Package (like ca-certificates)     | Yes      | Yes     | Yes     | No         |
| Debug (like attaching container by `bash`) | Easy     | Easy    | Easy    | Hard       |
| Extensibility (like 2nd development)               | Easy     | Easy    | Easy    | Hard       |

chengshiwen:
Brief changes:

1. Improve makefile
  - Add `-s -w` in `LDFLAGS` to strip debug information and reduce release binary size
    - In Mac, the binary size of the `vela` is reduced from `79M` to `67M`, and the binary size of the `manager` is reduced from `53M` to `43M`
  - Define docker image name such as `VELA_IMAGE`, `VELA_CORE_IMAGE` and `VELA_CORE_TEST_IMAGE`
  - Define go build env `GOBUILD_ENV`
image

2.  Improve dockerfile
  - Replace base image `ubuntu:latest` with `alpine:latest`, since in many docker libraries, debian and alpine are more popular
  - The docker image size of `vela-core` is reduced from `140M` to `51.1M`
image

3. Add entrypoint.sh to support entrypoint and debugging
  - Solve #1177 better
  - The following commands can be supported:

```
docker run -it --rm vela-core:latest manager -use-webhook
docker run -it --rm vela-core:latest -use-webhook
docker run -it --rm vela-core:latest sh
docker run -it --rm vela-core:latest bash
docker run -it --rm vela-core:latest -h
```

image

This closes #1448

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,6 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
-# Set GOPROXY if downloading slowly
-RUN go env -w GOPROXY=https://goproxy.cn,https://goproxy.io,direct
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
@@ -32,11 +30,15 @@ RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE:-alpine:latest}
 # This is required by daemon connnecting with cri
-RUN apk add --no-cache ca-certificates
+RUN apk add --no-cache ca-certificates bash
 
 WORKDIR /
 
 ARG TARGETARCH
-COPY --from=builder /workspace/manager-${TARGETARCH} /manager
+COPY --from=builder /workspace/manager-${TARGETARCH} /usr/local/bin/manager
 
-ENTRYPOINT ["/manager"]
+COPY entrypoint.sh /usr/local/bin/
+
+ENTRYPOINT ["entrypoint.sh"]
+
+CMD ["manager"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,12 @@
 # Build the manager binary
-FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.14 as builder
+FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.14-alpine as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
+# Set GOPROXY if downloading slowly
+RUN go env -w GOPROXY=https://goproxy.cn,https://goproxy.io,direct
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
@@ -19,20 +21,18 @@ COPY version/ version/
 ARG TARGETARCH
 ARG VERSION
 ARG GITVERSION
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} GO111MODULE=on \
-    go build -a -ldflags "-X github.com/oam-dev/kubevela/version.VelaVersion=${VERSION:-undefined} -X github.com/oam-dev/kubevela/version.GitRevision=${GITVERSION:-undefined}" \
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
+    go build -a -ldflags "-s -w -X github.com/oam-dev/kubevela/version.VelaVersion=${VERSION:-undefined} -X github.com/oam-dev/kubevela/version.GitRevision=${GITVERSION:-undefined}" \
     -o manager-${TARGETARCH} main.go
 
-# Use ubuntu as base image for convenience.
+# Use alpine as base image to reduce image size
 # You can replace distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 # Could use `--build-arg=BASE_IMAGE=gcr.io/distroless/static:nonroot` to overwrite
 ARG BASE_IMAGE
-FROM ${BASE_IMAGE:-ubuntu:latest}
+FROM ${BASE_IMAGE:-alpine:latest}
 # This is required by daemon connnecting with cri
-RUN ln -s /usr/bin/* /usr/sbin/ && apt-get update -y \
-  && apt-get install --no-install-recommends -y ca-certificates \
-  && apt-get clean && rm -rf /var/log/*log /var/lib/apt/lists/* /var/log/apt/* /var/lib/dpkg/*-old /var/cache/debconf/*-old
+RUN apk add --no-cache ca-certificates
 
 WORKDIR /
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,10 +23,10 @@ RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
     go build -a -ldflags "-s -w -X github.com/oam-dev/kubevela/version.VelaVersion=${VERSION:-undefined} -X github.com/oam-dev/kubevela/version.GitRevision=${GITVERSION:-undefined}" \
     -o manager-${TARGETARCH} main.go
 
-# Use alpine as base image to reduce image size
+# Use alpine as base image due to the discussion in issue #1448
 # You can replace distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-# Could use `--build-arg=BASE_IMAGE=gcr.io/distroless/static:nonroot` to overwrite
+# Overwrite `BASE_IMAGE` by passing `--build-arg=BASE_IMAGE=gcr.io/distroless/static:nonroot`
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE:-alpine:latest}
 # This is required by daemon connnecting with cri
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,11 @@ VELA_VERSION ?= master
 # Repo info
 GIT_COMMIT          ?= git-$(shell git rev-parse --short HEAD)
 GIT_COMMIT_LONG     ?= $(shell git rev-parse HEAD)
-VELA_VERSION_VAR    := github.com/oam-dev/kubevela/version.VelaVersion
-VELA_GITVERSION_VAR := github.com/oam-dev/kubevela/version.GitRevision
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -19,20 +21,18 @@ COPY version/ version/
...
     -o manager-${TARGETARCH} main.go
 
-# Use ubuntu as base image for convenience.
+# Use alpine as base image to reduce image size
 # You can replace distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 # Could use `--build-arg=BASE_IMAGE=gcr.io/distroless/static:nonroot` to overwrite
 ARG BASE_IMAGE
-FROM ${BASE_IMAGE:-ubuntu:latest}
+FROM ${BASE_IMAGE:-alpine:latest}
```
wonderflow, line 33: Here is the reason why we change the base image, https://github.com/oam-dev/kubevela/issues/1177

Does this change fit for this issue?

Dockerfile
```diff
@@ -1,10 +1,12 @@
 # Build the manager binary
-FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.14 as builder
+FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.14-alpine as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
+# Set GOPROXY if downloading slowly
+RUN go env -w GOPROXY=https://goproxy.cn,https://goproxy.io,direct
```
wonderflow, line 9: will it slow down the speed outside of China?

Makefile
```diff
@@ -3,10 +3,11 @@ VELA_VERSION ?= master
 # Repo info
 GIT_COMMIT          ?= git-$(shell git rev-parse --short HEAD)
 GIT_COMMIT_LONG     ?= $(shell git rev-parse HEAD)
-VELA_VERSION_VAR    := github.com/oam-dev/kubevela/version.VelaVersion
-VELA_GITVERSION_VAR := github.com/oam-dev/kubevela/version.GitRevision
-LDFLAGS             ?= "-X $(VELA_VERSION_VAR)=$(VELA_VERSION) -X $(VELA_GITVERSION_VAR)=$(GIT_COMMIT)"
+VELA_VERSION_KEY    := github.com/oam-dev/kubevela/version.VelaVersion
+VELA_GITVERSION_KEY := github.com/oam-dev/kubevela/version.GitRevision
+LDFLAGS             ?= "-s -w -X $(VELA_VERSION_KEY)=$(VELA_VERSION) -X $(VELA_GITVERSION_KEY)=$(GIT_COMMIT)"
```
wonderflow, line 8: can you explain a bit more about `-s -w` flags? Or give some docs.

Makefile
```diff
@@ -34,6 +35,11 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Image URL to use all building/pushing image targets
+VELA_IMAGE           ?= vela:latest
```
wonderflow, line 39: where is it used?

Makefile
```diff
@@ -42,16 +48,13 @@ test: vet lint staticcheck
 	go test -race -covermode=atomic ./references/apiserver/... ./references/appfile/... ./references/cli/... ./references/common/... ./references/plugins/...
 	@$(OK) unit-tests pass
 
-# Build manager binary
-build: fmt vet lint staticcheck
-	go run hack/chart/generate.go
-	go build -o bin/vela -ldflags ${LDFLAGS} references/cmd/cli/main.go
-	git checkout references/cmd/cli/fake/chart_source.go
+# Build vela cli binary
+build: fmt vet lint staticcheck vela-cli
```
wonderflow, line 52: I like this refactor, cool!

Makefile
```diff
@@ -179,25 +184,22 @@ e2e-cleanup:
...
 
 # Run tests
 core-test: fmt vet manifests
 	go test ./pkg/... -coverprofile cover.out
 
-# Build manager binary
+# Build vela core manager binary
 manager: fmt vet lint manifests
-	go build -o bin/manager ./cmd/core/main.go
+	$(GOBUILD_ENV) go build -o bin/manager -a -ldflags $(LDFLAGS) ./cmd/core/main.go
```
wonderflow, line 202: does the ldflags really be used?

wonderflow (commented):
Generally I love this refactor very much, leave few questions in comments

Makefile
chengshiwen: The original plan was not to add `ldflags`, but vela-core has no version information. I plan to add the `manager version` subcommand, so I added it in advance

chengshiwen: Otherwise I have to distinguish between `-s -w` and `-X`

wonderflow: Sure, go ahead, sounds very good

wonderflow: maybe print the version information at startup time is enough.

Makefile
chengshiwen: Considering #1392, I added it in advance. `VELA_IMAGE` can also be removed first

wonderflow: Let's remove it cause it's confusion , we can add it at that time.

Dockerfile
chengshiwen: If we need to solve #1177 better, support entrypoint and debugging like `docker run -it --rm vela-core:latest sh` at the same time, we need to add an `entrypoint.sh` file. So both `docker run -it --rm vela-core:latest -h` and `docker run -it --rm vela-core:latest sh` can run well

chengshiwen: I will make this change

Dockerfile
chengshiwen: `GOPROXY` is really frustrating, so I will remove it first

Makefile
chengshiwen: Please refer to [go cmd link](https://golang.org/cmd/link)
```
-s
	Omit the symbol table and debug information.
-w
	Omit the DWARF symbol table.
```

chengshiwen: [Kubernetes Makefile](https://github.com/kubernetes/kubernetes/blob/master/build/root/Makefile)

```
#   make all GOLDFLAGS=""
#     Note: Specify GOLDFLAGS as an empty string for building unstripped binaries, which allows
#           you to use code debugging tools like delve. When GOLDFLAGS is unspecified, it defaults
#           to "-s -w" which strips debug information. Other flags that can be used for GOLDFLAGS 
#           are documented at https://golang.org/cmd/link/
```

Makefile
chengshiwen: Done

Dockerfile
chengshiwen: Done

Dockerfile
chengshiwen: Done

chengshiwen:
@wonderflow 
The following commands can be supported:
```
docker run -it --rm vela-core:latest manager -use-webhook
docker run -it --rm vela-core:latest -use-webhook
docker run -it --rm vela-core:latest sh
docker run -it --rm vela-core:latest bash
docker run -it --rm vela-core:latest -h
```

image

wonderflow:
ping @shikanon would you like to review the dockfile changes?

Dockerfile
```diff
@@ -19,24 +19,26 @@ COPY version/ version/
...
 ARG VERSION
 ARG GITVERSION
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} GO111MODULE=on \
-    go build -a -ldflags "-X github.com/oam-dev/kubevela/version.VelaVersion=${VERSION:-undefined} -X github.com/oam-dev/kubevela/version.GitRevision=${GITVERSION:-undefined}" \
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
+    go build -a -ldflags "-s -w -X github.com/oam-dev/kubevela/version.VelaVersion=${VERSION:-undefined} -X github.com/oam-dev/kubevela/version.GitRevision=${GITVERSION:-undefined}" \
     -o manager-${TARGETARCH} main.go
 
-# Use ubuntu as base image for convenience.
+# Use alpine as base image to reduce image size
```
shikanon, line 26: @chengshiwen I think not only consider the image size, we can open the issue for discuss the choice of base image base on size, security and propularity.

shikanon (commented):
I think we can discuss the choice of base image between ubuntu, alpine and distroless etc. from image size, security and propularity.

*the size*: distroless < alpine < ubuntu
Distroless image build size is 642.67 KB
alpine image size is ~2.5MB
ubuntu image size is ~25MB

*the security*: distroless > alpine > ubuntu

*the propularity*: distroless < alpine < ubuntu

wonderflow:
Hi,  @chengshiwen , you're more than welcome to join our developer community, would you like to add my Dingtalk (wonderflow) for further discussion?

Dockerfile
chengshiwen: I just optimize this work with best practices, since in many docker libraries, debian and alpine are more popular, while images based on ubuntu are rare. As for security, no images will be safe and reliable. It doesn’t seem to make sense to only consider whether the base image is security.

Dockerfile
```diff
@@ -19,24 +19,26 @@ COPY version/ version/
...
 # You can replace distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
```
wonderflow, line 28: we could add the discussion into comments here. For this PR, LGTM.

wonderflow (approved):

Dockerfile
shikanon: > I just optimize this work with best practices, since in many docker libraries, debian and alpine are more popular, while images based on ubuntu are rare. As for security, no images will be safe and reliable. It doesn’t seem to make sense to only consider whether the base image is security.

would you add the more issue annotation for the discussion not only for the reason for the `to reduce image size`, I think this discussion will help the other.

chengshiwen: OK

Dockerfile
chengshiwen: OK

shikanon (approved):

Dockerfile
chengshiwen: See #1448, done

Dockerfile
chengshiwen: See #1448, done

PR #1446, opened 2021-04-09
Fix bug: fix data-race of package discover

yangsoon:
when `cue` build a `*build.Instance` which contains build-in Kube package, if other thread refresh the package discover,
It will lead to data race, Because the different thread will read or write the `Imports` field of build.Instance.

This PR is to fix this problem.

### DIFF
--- a/pkg/dsl/definition/package.go
+++ b/pkg/dsl/definition/package.go
@@ -82,6 +82,20 @@ func (pd *PackageDiscover) ImportBuiltinPackagesFor(bi *build.Instance) {
 	bi.Imports = append(bi.Imports, pd.velaBuiltinPackages...)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/dsl/definition/package.go
```diff
@@ -82,6 +82,20 @@ func (pd *PackageDiscover) ImportBuiltinPackagesFor(bi *build.Instance) {
 	bi.Imports = append(bi.Imports, pd.velaBuiltinPackages...)
 }
 
+// ImportPackagesAndBuildInstance will add KubeVela built-in packages and build your CUE instance
```
wonderflow, line 85: comment the data race thing

wonderflow (approved):

Issue #1093, opened 2021-02-23
check outputs:resource should be unique in application webhook


wonderflow:
The trait can use `context.outputs.<xx>`  to do simple data passing.

This rely on the trait resource name is unique, or the former outputs resource data will be covered by the latter one.

We should add [a webhook validate rule](https://github.com/oam-dev/kubevela/tree/master/pkg/webhook/core.oam.dev/v1alpha2/application) to check for it.

https://github.com/oam-dev/kubevela/pull/1087/files#diff-05b89d808deaa1f5a7e06ea0b6de559283ab1e2ea28e006299af38a200fa9bcaR354

_Originally posted by @zzxwill in https://github.com/oam-dev/kubevela/pull/1092#r580846972_

captainroy-hy:
May I help with this issue? If @Cweiping is busy with other tasks.

wonderflow:
Sure, please carry on @captainroy-hy Thanks!

Issue #733, opened 2020-12-07
[Feature] Vela-Cli should support live-diff（dry-run） for users to see the difference of underlying resource after upgrade


wonderflow:
This could be very helpful for production serious use.

For example, if a user has already running a deployment like below:

```
kind: Deployment
metadata:
  name: myweb
spec:
  template:
     containers:
     - image: nginx:v1
```

When running a new application called `myapp.yaml`

```
kind: Application
metadata:
  name: myapp
spec:
  components:
  - name: myweb:
    properties:
      image: "nginx:v2"
```

We could support command like this:

```
$ vela system live-diff -f myapp.yaml
DIFF
---
kind: Deployment
metadata:
  name: myweb
spec:
  template:
     containers:
-    - image: nginx:v1
+    - image: nginx:v2
```

This command will not actually update, but will execute like `dry-run` and do live diff for underlying resource.

What's more, we can specify the revision using `appRevision`

```
$ vela system live-diff -f myapp.yaml --revision myapp-v1
```

wonderflow:
Now we have  [`vela system dry-run`](https://github.com/oam-dev/kubevela/tree/master/docs/examples/dry-run) which will render K8s resource.  And we have [AppRevision](https://github.com/oam-dev/kubevela/blob/master/apis/core.oam.dev/v1alpha2/applicationrevision_types.go) records all the revision information(including Application and Definition snapshot).

The live diff could be implemented now.

```
vela system live-diff -f <myapp.yaml> --revision <app-revision-name>
```

Then we can render K8s resource out by using `vela system dry-run` for the specified revision and the local application. Then, we will get the diff.

wonderflow:
live diff for helm https://github.com/databus23/helm-diff

PR #1447, opened 2021-04-09
refactor: Split appfile.Parser into appfile.Appfile and Parser

captainroy-hy:
Major changes:

- Split appfile.Parser into appfile.Appfile and Parser. Parser is responsible for generating Appfile, especially getting all data required by Appfile to generate AppConfig. And Appfile becomes a self-contained object that can generate AppConfig and do other works without interacting with clusters by client.
- Move pkg/oam/util/template.go to pkg/appfile. It seems appfile pkg is the only consumer of this util, and all logic in template.go is relevant to getting data for Appfile/Workloads. And this util references another util in pkg/controllers/util, that makes the source code hard to read.
- Other fixes around the above two.

Why make these changes?
In pkg/appfile,  parser handles two tasks, generating Appfile and generating AC from Appfile. A problem is that, we get data (from cluster) of Workloads in generating Appfile phase, but still use client to get other data again in generating AppConfig phase. It passes a client interface deep into the logic of generating AppConfig. But indeed, we can prepare all data ready in generating Appfile phase and get rid of client interface in generating AppConfig phase.  
Once Appfile becomes a self-contained object, we can do mocking data and validating easier.

That can help in #733 and #1093. 
  


pkg/appfile/parser.go
```diff
@@ -579,43 +183,54 @@ func GetOutputSecretNames(workloads *Workload) (string, error) {
 
 func parseWorkloadInsertSecretTo(ctx context.Context, c client.Client, namespace string, wl *Workload) ([]process.RequiredSecrets, error) {
```
captainroy-hy, line 184: Hi @zzxwill 
This function is re-wrote. Instead of generating schema, we can get the secret name from CUE template comments through CUE SDK dircectly. Please help to review in case of my misunderstanding on this feature. Thx!

wonderflow (approved):
LGTM, ping @zzxwill @ryanzhang-oss , please double check this refactor

ryanzhang-oss:
The overall idea is great, the parser part definitely needs a lot of refactoring.

pkg/appfile/parser.go
zzxwill: That's better. Thanks.

zzxwill (approved):
/lgtm

PR #1450, opened 2021-04-09
Fix bug: fix bug which Component/TraitDefinition won't work when contains "_|_" in value

yangsoon:
fix https://github.com/oam-dev/kubevela/issues/1439

when CUE Template in ComponentDefinition/TraitDefinition contains "\_\|\_" in value, KubeVela will mistakenly think that the Template is compiled incorrectly.

This PR fix the bug, use `_\|_[\s]//.*` to match compile error like:

```
_|_ // undefined field "#Deployment"
```

### DIFF
--- a/pkg/dsl/model/instance.go
+++ b/pkg/dsl/model/instance.go
@@ -17,15 +17,14 @@ limitations under the License.
 package model
 
 import (
+	"regexp"
 	"strings"
-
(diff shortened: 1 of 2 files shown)


wonderflow (approved):
GOOD

wonderflow (approved):

PR #1452, opened 2021-04-10
make sure we don't create more pods

ryanzhang-oss:


wonderflow (approved):
will it fix the flaky test?

ryanzhang-oss:
> will it fix the flaky test?

that's the theory. The failed test error message seems to indicate that more pods are created than expceted.

PR #1458, opened 2021-04-11
[Improvement-1457] Use kind cmd code to load docker image instead of local kind binary

### LINKED ISSUE #413: [Feature] Command 'vela up' needs 'docker', 'kubectl','kind' as prerequisite
1. For 'docker', I can say 90% developer will have docker installed in their environment, make it as prerequisite in readme is OK.

2. For 'kubectl' and 'kind', we must remove them.

### LINKED ISSUE #1457: [Improvement] Use kind cmd code to load docker image instead of local kind binary
The dependency on the kind binary needs to be removed as follows:
```
cmd := exec.Command("kind", "load", "docker-image", image)
```

The related issue: #413

chengshiwen:
Just use
```
err := kind.LoadDockerImage(image)
```
instead of
```
cmd := exec.Command("kind", "load", "docker-image", image)
```

Refer to https://github.com/kubernetes-sigs/kind/blob/main/pkg/cmd/kind/load/docker-image/docker-image.go

This closes #413 and closes #1457

The `docker` command will be resolved after `kind`

### DIFF
--- a/pkg/builtin/kind/client.go
+++ b/pkg/builtin/kind/client.go
@@ -1,5 +1,7 @@
 /*
-Copyright 2020 The KubeVela Authors.
+Copyright 2021 The KubeVela Authors.
+
+The code is inspired from the Kind repo at https://github.com/kubernetes-sigs/kind/blob/main/pkg/cmd/kind/load/docker-image/docker-image.go
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


pkg/builtin/kind/client.go
```diff
@@ -0,0 +1,163 @@
+/*
+Copyright 2020 The KubeVela Authors.
```
wonderflow, line 2:
```suggestion
Copyright 2021 The KubeVela Authors.
```

BTW, do you copy-paste this code from `kind` repo? If so, we should declare it here.

wonderflow (approved):
great job!

pkg/builtin/kind/client.go
chengshiwen: Yes, I copy-paste the code from `kind` repo and make a few changes. Is it ok to modify it to `Copyright 2021 The KubeVela and Kind Authors`?

wonderflow: Not like that , the Copyright line should keep `Copyright 2021 The KubeVela Authors.` While you can add a line such as `The code is inspired from The kind repo at <the original source code linke>` below it.

chengshiwen: Done

PR #1461, opened 2021-04-11
skip applying helm-release if app spec not change

### LINKED ISSUE #1389: do we still need to apply helm resource if spec no difference
[pkg/controller/core.oam.dev/v1alpha2/application/apply.go:L135](https://github.com/oam-dev/kubevela/blob/0629049e1f9fd1529a2e285f569258896296a566/pkg/controller/core.oam.dev/v1alpha2/application/apply.go#L135)

captainroy-hy:
to fix #1389 


wonderflow (approved):
I suggest add some explanation here. For example, the `isNewRevision` always promise there's any changes or not on the spec.

captainroy-hy:
> I suggest add some explanation here. For example, the `isNewRevision` always promise there's any changes or not on the spec.

Sure, let me add comments.

Issue #1351, opened 2021-03-29
[Feature] add workload type for components in Application .status


zzxwill:
**Is your feature request related to a problem? Please describe.**

When I'd like to check whether all components of an application are provisioned, I normally use `kubectl get components`.

```
✗ k get components
NAME             WORKLOAD-KIND   AGE
express-server   Deployment      54s
sample-db        RDSInstance     53s
sample-oss       Bucket          52s
```
Then I can quickly check Crossplane managed resource `RDSInstance` and `Bucket` in a convenient way.

```
✗ k get rdsinstance
NAME        READY   SYNCED   STATE     ENGINE   VERSION   AGE
sample-db   True    True     Running   mysql    8.0       116m
```

As application is the main entry object, so its status should also contain the workload kind of a component.

```
status:
  batchRollingState: ""
  components:
  - apiVersion: core.oam.dev/v1alpha2
    kind: Component
    name: express-server
    uid: e9b9cf7a-0ff5-4aac-8fc8-0a030bb09387
  - apiVersion: core.oam.dev/v1alpha2
    kind: Component
    name: sample-db
    uid: e9b9cf7a-0ff5-4aac-8fc8-0a030bb09387
  - apiVersion: core.oam.dev/v1alpha2
    kind: Component
    name: sample-oss
    uid: e9b9cf7a-0ff5-4aac-8fc8-0a030bb09387
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

wonderflow:
@charles-woshicai  We can list the componentName only like below:

```
status:
  batchRollingState: ""
  components:
  - name: express-server
  - name: sample-db
  - name: sample-oss
```

charles-woshicai:
> @charles-woshicai We can list the componentName only like below:
> 
> ```
> status:
>   batchRollingState: ""
>   components:
>   - name: express-server
>   - name: sample-db
>   - name: sample-oss
> ```

Is this sample the expectation?

wonderflow:
@charles-woshicai  Not very sure, you can also check the current status and find a better place to record.

PR #1471, opened 2021-04-12
update: try to add workloadDefinitionRef to application status

charles-woshicai:
ref: https://github.com/oam-dev/kubevela/issues/1351


pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -188,14 +188,24 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-		})
+			UID:        comp.UID,
+		}
+		// Add workload kind to component status. Related issue:
+		// https://github.com/oam-dev/kubevela/issues/1351
+		var workloadMap map[string]interface{}
+		workloadMap, err = oamutil.RawExtension2Map(&comp.Spec.Workload)
+		if err == nil {
+			if wlRef, ok := workloadMap["definitionRef"]; ok {
+				refComp.WorkloadDefinition = oamutil.Object2RawExtension(wlRef)
```
ryanzhang-oss, line 205: why is it a raw Extention? Shouldn't it just be a string?

charles-woshicai: It depends on what information is needed here.

I thought before that provider an object of `workloadDefinition` here could be better than just provide a workload kind, but now I find that  if we provide object of workload, it seems workload name repeates twice.

Maybe you are right, I will change to provide just a name of workload name rather than total object.

apis/core.oam.dev/common/types.go
```diff
@@ -198,7 +208,7 @@ type AppStatus struct {
 	Phase ApplicationPhase `json:"status,omitempty"`
 
 	// Components record the related Components created by Application Controller
-	Components []v1alpha12.TypedReference `json:"components,omitempty"`
+	Components []ComponentRef `json:"components,omitempty"`
 
 	// Services record the status of the application services
 	Services []ApplicationComponentStatus `json:"services,omitempty"`
```
wonderflow, line 214: does the information already contained in `Services`?  Is it the name field in it?

https://github.com/oam-dev/kubevela/issues/1351

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -188,14 +188,24 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	var refComps []common.ComponentRef
 	for _, comp := range comps {
-		refComps = append(refComps, v1alpha1.TypedReference{
+		refComp := common.ComponentRef{
 			APIVersion: comp.APIVersion,
 			Kind:       comp.Kind,
 			Name:       comp.Name,
-			UID:        app.UID,
-		})
+			UID:        comp.UID,
```
wonderflow, line 197: BTW, the UID is wrong

charles-woshicai: Sry, will fix it.

apis/core.oam.dev/common/types.go
charles-woshicai: Let me check.

charles-woshicai: I think `Services` just record the name of workload such as `webservice`, what we want to add here is the type of `workloadDefinition` such as `deployments.app`.

Ref:
https://github.com/oam-dev/kubevela/blob/63b348f4d887bf522d476f0614de112293e7ead3/pkg/controller/core.oam.dev/v1alpha2/application/apply.go#L197-L204

wonderflow: In the issue #1351  we just want to know the instance name, I think the name in service is enough

charles-woshicai: I don't think so, according to the isse #1351 , `WORKLOAD-KIND` is actually wanted in my opinion.
image

The name is already exist, if you think `WORKLOAD-KIND` is not necessary, I will close this PR and I think the issue #1351 could be closed.

wonderflow: If so, I think we could add the workloadDef info into `ApplicationComponentStatus` struct

charles-woshicai: ok, I will try.

charles-woshicai:
I have a question, do I need to commit  files which are below `legacy/charts`?

charles-woshicai:
> I have a question, do I need to commit files which are below `legacy/charts`?

I see, just run `make reviewable` before reviewing.

wonderflow:
@charles-woshicai of course, they're automatically modified with CRD

wonderflow:
please fix tests

charles-woshicai:
> please fix tests

I will fix it in the evening.

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -199,8 +199,11 @@ func (h *appHandler) statusAggregate(appFile *appfile.Appfile) ([]common.Applica
 	var healthy = true
 	for _, wl := range appFile.Workloads {
 		var status = common.ApplicationComponentStatus{
-			Name:    wl.Name,
-			Healthy: true,
+			Name: wl.Name,
+			// add workload definition to status, related issue:
+			// https://github.com/oam-dev/kubevela/issues/1351
```
wonderflow, line 204: this is not needed

wonderflow (commented):
Generally LGTM. ping @zzxwill  please check if it resolves the issue.

zzxwill:
@charles-woshicai Thanks for taking up this feature.

I've executed your PR and get the status as below.

```
status:
  batchRollingState: ""
  components:
  - apiVersion: core.oam.dev/v1alpha2
    kind: Component
    name: express-server
    uid: 2d0f1a2c-b51c-4f1c-b7ed-cf6268808e6a
  - apiVersion: core.oam.dev/v1alpha2
    kind: Component
    name: sample-oss
    uid: 2d0f1a2c-b51c-4f1c-b7ed-cf6268808e6a
```

I expect to have clear information of all components of an application as below.

```
➜  /Users/zhouzhengxi k get component
NAME             WORKLOAD-KIND   AGE
express-server   Deployment      29m
sample-oss       Configuration   29m
```

`Name` and `Workload-kind` are essential. So changing the status to something like this might be better.

```
status:
  batchRollingState: ""
  components:
  - apiVersion: v1/apps
    kind: Deployment
    name: express-server
    uid: 2d0f1a2c-b51c-4f1c-b7ed-cf6268808e6a
  - apiVersion: terraform.core.oam.dev/v1beta1
    kind: Configuration
    name: sample-oss
    uid: 2d0f1a2c-b51c-4f1c-b7ed-cf6268808e6a
```

charles-woshicai:
> @charles-woshicai Thanks for taking up this feature.
> 
> I've executed your PR and get the status as below.
> 
> ```
> status:
>   batchRollingState: ""
>   components:
>   - apiVersion: core.oam.dev/v1alpha2
>     kind: Component
>     name: express-server
>     uid: 2d0f1a2c-b51c-4f1c-b7ed-cf6268808e6a
>   - apiVersion: core.oam.dev/v1alpha2
>     kind: Component
>     name: sample-oss
>     uid: 2d0f1a2c-b51c-4f1c-b7ed-cf6268808e6a
> ```
> 
> I expect to have clear information of all components of an application as below.
> 
> ```
> ➜  /Users/zhouzhengxi k get component
> NAME             WORKLOAD-KIND   AGE
> express-server   Deployment      29m
> sample-oss       Configuration   29m
> ```
> 
> `Name` and `Workload-kind` are essential. Adding `apiVersion` of the workload is better. So changing the status to something like this might be better.
> 
> ```
> status:
>   batchRollingState: ""
>   components:
>   - apiVersion: v1/apps
>     kind: Deployment
>     name: express-server
>     uid: 2d0f1a2c-b51c-4f1c-b7ed-cf6268808e6a
>   - apiVersion: terraform.core.oam.dev/v1beta1
>     kind: Configuration
>     name: sample-oss
>     uid: 2d0f1a2c-b51c-4f1c-b7ed-cf6268808e6a
> ```

Seems more friendly for user, let me try.

wonderflow:
@charles-woshicai Not like that, the comments point to the wrong place, we should check `status.services`

charles-woshicai:
> @charles-woshicai Not like that, the comments point to the wrong place, we should check `status.services`

OK, won't change code until final expected status to be decided.

zzxwill:
@charles-woshicai Doubled checked with @wonderflow. Putting the definition information in `status.service` is the most right place.

```
  services:
  - healthy: true
    name: express-server
    workloadDefinition:
      apiVersion: apps/v1
      kind: Deployment
```

I like it very much. Thanks for fulfilling my requirement #1351. Great job.

zzxwill (approved):
/lgtm

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
wonderflow, line 204:
```suggestion
```

wonderflow (approved):
Great job, thanks

PR #1481, opened 2021-04-14
refactor appRevison extract func to utils

wangyikewxgm:
move ExtractRevisionNum func to utils/helper


pkg/oam/util/helper_test.go
```diff
@@ -2054,3 +2054,29 @@ spec:
...
+	}, {
+		appRevision:     "v1-v10",
+		wantRevisionNum: "10",
+	}, {
+		appRevision:     "v10-v1-v1",
+		wantRevisionNum: "1",
+	}, {
+		appRevision:     "myapp-v1-v2",
+		wantRevisionNum: "2",
+	}}
```
wonderflow, line 2077: please add a case

```
		appRevision:     "v1",
		wantRevisionNum: "0",
```

pkg/oam/util/helper.go
```diff
@@ -721,6 +721,13 @@ func ConvertComponentDef2WorkloadDef(dm discoverymapper.DiscoveryMapper, compone
 	return nil
 }
 
+// ExtractRevisionNum  extract revision number from appRevision name
+func ExtractRevisionNum(appRevision string) string {
```
wonderflow, line 725: the return value should be a int

pkg/oam/util/helper_test.go
wonderflow:
```
appRevision:     "vv",
wantRevisionNum: "0",
```

pkg/oam/util/helper.go
wonderflow: we can return error if parse fail

wangyikewxgm: Got it

wonderflow (approved):
good job

yangsoon:
LGTM

PR #1483, opened 2021-04-14
Fix helm install error when disable webhook patch

YaoC:
[failurePolicy](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy) should be `Fail` instead of `Fails`.


wonderflow (approved):
good catch, thanks

PR #1489, opened 2021-04-14
merge commits from master into release-1.0

wonderflow:
Most of them are bugfixes and docs improvement.

The only big feature is `live-diff` which is very useful client side feature, so I decide to merge it in v1.x release.


zzxwill (approved):
/lgtm

Issue #1451, opened 2021-04-09
update application will see appContext update failed error-log


wangyikewxgm:
**Describe the bug**
update a application, vela-core will print a applicationContext update status error

**To Reproduce**
1. create application
2. update application
3. check vela-core log See error

**Screenshots**
![image](https://user-images.githubusercontent.com/77846369/114163436-c4a32800-995c-11eb-808b-49c791665f77.png)

**Cluster information**
Describe your kubernetes cluster information.
- Kubernetes Version [e.g. 1.18.9]

ryanzhang-oss:
this error should be transiant

PR #1500, opened 2021-04-15
fix appContext update failed error

xiqiu618:
Fixes: #1451

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationcontext/applicationcontext_controller.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationcontext/applicationcontext_controller.go
@@ -109,8 +109,8 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	reconResult := acReconciler.ACReconcile(ctx, appConfig, r.log)
 	appContextPatch := client.MergeFrom(appContext.DeepCopy())
 	appContext.Status = appConfig.Status
-         // always update ac status and set the error
-         // this should be the only place for status of AppContext to update, so we can patch to avoid update conflicts caused by `resourceVersion` changed by spec.
+	// always update ac status and set the error
+	// this should be the only place for status of AppContext to update, so we can patch to avoid update conflicts caused by `resourceVersion` changed by spec.
 	err = errors.Wrap(r.client.Status().Patch(ctx, appContext, appContextPatch), errUpdateAppContextStatus)
 	// use the controller build-in backoff mechanism if an error occurs
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationcontext/applicationcontext_controller.go
```diff
@@ -107,9 +107,10 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	// call into the old ac Reconciler and copy the status back
 	acReconciler := ac.NewReconciler(r.mgr, dm, r.log, ac.WithRecorder(r.record), ac.WithApplyOnceOnlyMode(r.applyMode))
 	reconResult := acReconciler.ACReconcile(ctx, appConfig, r.log)
+
+	appContextPatch := client.MergeFrom(appContext.DeepCopy())
 	appContext.Status = appConfig.Status
-	// always update ac status and set the error
-	err = errors.Wrap(r.client.Status().Update(ctx, appContext), errUpdateAppContextStatus)
+	err = errors.Wrap(r.client.Patch(ctx, appContext, appContextPatch), errUpdateAppContextStatus)
```
wonderflow, line 113: why it can fix the issue? the status is a sub-resource

xiqiu618: sorry , i has been modified to be a sub resource  patch，the cause of this error Update() may fail a few times before passing due to the cache being stale,Patch() can avoid

pkg/controller/core.oam.dev/v1alpha2/applicationcontext/applicationcontext_controller.go
```diff
@@ -107,9 +107,9 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	// call into the old ac Reconciler and copy the status back
 	acReconciler := ac.NewReconciler(r.mgr, dm, r.log, ac.WithRecorder(r.record), ac.WithApplyOnceOnlyMode(r.applyMode))
 	reconResult := acReconciler.ACReconcile(ctx, appConfig, r.log)
+	appContextPatch := client.MergeFrom(appContext.DeepCopy())
 	appContext.Status = appConfig.Status
-	// always update ac status and set the error
-	err = errors.Wrap(r.client.Status().Update(ctx, appContext), errUpdateAppContextStatus)
+	err = errors.Wrap(r.client.Status().Patch(ctx, appContext, appContextPatch), errUpdateAppContextStatus)
```
wonderflow, line 112:
```suggestion
         // always update ac status and set the error
         // this should be the only place for status of AppContext to update, so we can patch to avoid update conflicts caused by `resourceVersion` changed by spec.
	err = errors.Wrap(r.client.Status().Patch(ctx, appContext, appContextPatch), errUpdateAppContextStatus)
```

wonderflow: Let's keep the old commnets and add new to explain this behavior

wonderflow (approved):
LGTM

PR #1507, opened 2021-04-16
Refactor template_test.go to make it readable

zzxwill:
Some tests are not named which make it difficult to
read test cases and add more


wonderflow (approved):
good job

PR #1510, opened 2021-04-16
Add cloud service docs for end-users

zzxwill:
Split cloud service into two section, one for platform engineers,
and the other for end-users

### DIFF
--- a/docs/en/platform-engineers/cloud-services.md
+++ b/docs/en/platform-engineers/cloud-services.md
@@ -111,80 +111,6 @@ spec:
 
 ```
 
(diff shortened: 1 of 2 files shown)


docs/en/platform-engineers/cloud-services.md
```diff
@@ -1,5 +1,5 @@
 ---
-title:  Crossplane
+title: Provision and Consume Cloud Resources
```
wonderflow, line 2: 就用Crossplane吧，等terraform做好再改标题

wonderflow (approved):

PR #1514, opened 2021-04-16
add video from DevOps Toolkit

wonderflow:
this is really good video talking about kubevela and oam. add it to readme. \cc @resouer


hongchaodeng (approved):
lgtm

PR #1523, opened 2021-04-18
add action: sync vela api code to the repo(kubevela-core-api)

yangsoon:
fix https://github.com/oam-dev/kubevela/issues/1428

EveryTime a new version is released and apis is modified, 
it will be automatically sycn to https://github.com/oam-dev/kubevela-core-api

### DIFF
--- a/hack/apis/sync.sh
+++ b/hack/apis/sync.sh
@@ -1,0 +1,60 @@
+#!/bin/bash -l
+#
+# Copyright 2021. The KubeVela Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


hack/apis/sync.sh
```diff
@@ -0,0 +1,60 @@
...
+
+echo "push to kubevela-core-api"
+if git diff --quiet
+then
+  echo "nothing need to push, finished!"
+else
+  git add .
+  git commit -m "align with kubevela-$VERSION from commit $COMMIT_ID"
+  git push origin main
+fi
```
wonderflow, line 60: we should  add a tag for it.

yangsoon: got it

wonderflow (approved):

PR #1539, opened 2021-04-20
Finalize deployment rollout

ryanzhang-oss:
1. Complete the deployment rollout test
2. Refactor the controller into a common base paves the way for a deployment scalar controller


wonderflow (approved):
cool

Issue #1454, opened 2021-04-10
Healthy status is healthy, but vela status appname show error msg.


rapkey:
```
$ vela ls
APP     	COMPONENT            	TYPE     	TRAITS	PHASE  	HEALTHY	STATUS	CREATED-TIME
demo-app	demo-app-debug-ubuntu	stateless	      	running	healthy	      	2021-04-09 14:41:02 +0800 CST
├─      	demo-app-mysql       	stateless	      	running	healthy	      	2021-04-09 14:41:02 +0800 CST
└─      	demo-app-webapp      	stateless	      	running	healthy	      	2021-04-09 14:41:02 +0800 CST
```
```
$ vela status demo-app
About:

  Name:      	demo-app
  Namespace: 	default
  Created at:	2021-04-09 14:41:02 +0800 CST

Services:

Health checking failed!
Error: cannot get health condition from the health scope: demo-app-default-health
```
```
$ kubectl get healthscopes demo-app-default-health -o json
{
    "apiVersion": "core.oam.dev/v1alpha2",
    "kind": "HealthScope",
    "metadata": {
        "creationTimestamp": "2021-04-09T06:41:02Z",
        "generation": 1,
        "managedFields": [
            {
                "apiVersion": "core.oam.dev/v1alpha2",
                "fieldsType": "FieldsV1",
                "fieldsV1": {
                    "f:status": {
                        ".": {},
                        "f:scopeHealthCondition": {
                            ".": {},
                            "f:healthStatus": {}
                        }
                    }
                },
                "manager": "manager",
                "operation": "Update",
                "time": "2021-04-09T06:41:02Z"
            },
            {
                "apiVersion": "core.oam.dev/v1alpha2",
                "fieldsType": "FieldsV1",
                "fieldsV1": {
                    "f:spec": {
                        ".": {},
                        "f:workloadRefs": {}
                    }
                },
                "manager": "vela",
                "operation": "Update",
                "time": "2021-04-09T06:41:02Z"
            }
        ],
        "name": "demo-app-default-health",
        "namespace": "default",
        "resourceVersion": "1649693",
        "uid": "1bccf2f4-292f-4811-8d36-8cff738cb827"
    },
    "spec": {
        "workloadRefs": []
    },
    "status": {
        "scopeHealthCondition": {
            "healthStatus": "HEALTHY"
        }
    }
}
```

The healthy status in applications are also ture.
```
k get applications demo-app -o json
{
[...]

PR #1540, opened 2021-04-21
Set scpoe reference in application status.Service

wangyikewxgm:
Set components scope reference into application `status.service`. ref: #1454 

- [x] set scope in app status and write test
- [x] add define healthy docs 

```shell
%  kubectl get app website -o yaml
kind: Application
...
status:
 services:
   - healthy: true
     name: frontend
     scopes:
     - apiVersion: core.oam.dev/v1alpha2
       kind: HealthScope
       name: health-check
```


wonderflow (approved):
LGTM

PR #1553, opened 2021-04-22
calculate e2e-test into coverage

### LINKED ISSUE #1443: collect test coverage for e2e-test
# Background

We have many test cases in e2e but they are not counted in coverage. So I think it will be very cool if we can add them.

# Solution Reference

After some research, I found there's a solution: https://www.cyphar.com/blog/post/20170412-golang-integration-coverage

And the pounch team have followed the blog and tried successfully https://github.com/alibaba/pouch/pull/1338

# What Should We do?

1. create a new main function and build the binary (`test-main`) with the `go test` command.
2. run ginkgo test and send a stop signal at the end.
3. the coverage report is generated by the `test-main` in docker , so we need to export it from docker and upload coverage report, refer to https://github.com/oam-dev/kubevela/pull/1423

chivalryq:
Hey, I wrote a process to obtain e2e test coverage profile.

By this pr, total coverage will raise from **58.0%** to **61.5%**

fix #1443

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,7 @@ docker-push:
 e2e-setup:
 	helm install --create-namespace -n flux-system helm-flux http://oam.dev/catalog/helm-flux2-0.1.0.tgz
 	helm install kruise https://github.com/openkruise/kruise/releases/download/v0.7.0/kruise-chart.tgz
+	sh ./hack/e2e/modify_charts.sh
 	helm upgrade --install --create-namespace --namespace vela-system --set image.pullPolicy=IfNotPresent --set image.repository=vela-core-test --set applicationRevisionLimit=5 --set image.tag=$(GIT_COMMIT) --wait kubevela ./charts/vela-core
 	ginkgo version
 	ginkgo -v -r e2e/setup
@@ -197,9 +198,12 @@ ifneq ($(shell docker images -q $(VELA_CORE_TEST_IMAGE)),)
 	docker rmi -f $(VELA_CORE_TEST_IMAGE)
 endif
 
+end-e2e:
+	sh ./hack/e2e/end_e2e.sh
+
 # load docker image to the kind cluster
 kind-load:
-	docker build -t $(VELA_CORE_TEST_IMAGE) .
+	docker build -t $(VELA_CORE_TEST_IMAGE) -f Dockerfile.e2e .
 	kind load docker-image $(VELA_CORE_TEST_IMAGE) || { echo >&2 "kind not installed or error loading image: $(VELA_CORE_TEST_IMAGE)"; exit 1; }
 
 # Run tests
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -149,7 +149,7 @@ jobs:
       - name: Setup Kind Cluster
         run: |
           kind delete cluster
-          kind create cluster
+          kind create cluster --image kindest/node:v1.18.15@sha256:5c1b980c4d0e0e8e7eb9f36f7df525d079a96169c8a8f20d8bd108c0d0889cc4
           kubectl version
           kubectl cluster-info
 
@@ -178,6 +178,20 @@ jobs:
 
       - name: Run e2e tests
         run: make e2e-test
+
+      - name: Stop kubevela, get profile
+        run: make end-e2e
+
+      - name: Upload coverage report
+        uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          file: /tmp/e2e-profile.out
+          flags: e2etests
+          name: codecov-umbrella
+
+      - name: Clean e2e profile
+        run: rm /tmp/e2e-profile.out
 
       - name: Cleanup image
         if: ${{ always() }}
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -19,9 +20,16 @@ COPY version/ version/
...
+ARG is_e2e=false
+
+RUN if [[ ${is_e2e} == "true" ]]; then\
+     apk add gcc musl-dev libc-dev;\
+     go test -c -o manager-${TARGETARCH}  -cover -covermode=atomic ./cmd/core/; \
+ else\
+     GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
+     go build -a -ldflags "-s -w -X github.com/oam-dev/kubevela/version.VelaVersion=${VERSION:-undefined} -X github.com/oam-dev/kubevela/version.GitRevision=${GITVERSION:-undefined}" \
+     -o manager-${TARGETARCH} main.go;\
+  fi
```
wonderflow, line 32: Keep the original line in elseif ?

Dockerfile
```diff
@@ -39,6 +47,8 @@ COPY --from=builder /workspace/manager-${TARGETARCH} /usr/local/bin/manager
 
 COPY entrypoint.sh /usr/local/bin/
 
+VOLUME ["/workspace/data"]
```
wonderflow, line 50: This is only for e2e test, we should not keep it in normal dockerfile. I suggest we create a new dockerfile for e2e test

Makefile
```diff
@@ -197,17 +199,17 @@ ifneq ($(shell docker images -q $(VELA_CORE_TEST_IMAGE)),)
...
+
+
 # load docker image to the kind cluster
 kind-load:
-	docker build -t $(VELA_CORE_TEST_IMAGE) .
+	docker build -t $(VELA_CORE_TEST_IMAGE) --build-arg=is_e2e=true .
 	kind load docker-image $(VELA_CORE_TEST_IMAGE) || { echo >&2 "kind not installed or error loading image: $(VELA_CORE_TEST_IMAGE)"; exit 1; }
 
-# Run tests
-core-test: fmt vet manifests
```
wonderflow, line 206: why deleted?

Makefile
```diff
@@ -197,17 +199,17 @@ ifneq ($(shell docker images -q $(VELA_CORE_TEST_IMAGE)),)
...
-	go test ./pkg/... -coverprofile cover.out
-
 # Build vela core manager binary
-manager: fmt vet lint manifests
```
wonderflow, line 210: why deleted

charts/vela-core/values.yaml
```diff
@@ -81,4 +81,7 @@ admissionWebhooks:
 
 systemDefinitionNamespace: vela-system
 
+
+e2eTestEnable: false
```
wonderflow, line 85: It's not good to add test related thing in user fields.

charts/vela-core/templates/kubevela-controller.yaml
```diff
@@ -105,6 +105,11 @@ spec:
           securityContext:
           {{- toYaml .Values.securityContext | nindent 12 }}
           args:
+            {{ if .Values.e2eTestEnable }}
+            - "-test.coverprofile=/workspace/data/e2e-profile.out"
+            - "__DEVEL__E2E"
+            - "-test.run=E2EMain"
+            {{ end }}
```
wonderflow, line 112: we can write a script and use string replace for this work instead of using chart values.

Makefile
chivalryq: I thought this test isn't called by anywhere? maybe called manully？

cmd/core/main.go
```diff
@@ -127,6 +128,8 @@ func main() {
 
 	restConfig := ctrl.GetConfigOrDie()
 	restConfig.UserAgent = kubevelaName + "/" + version.GitRevision
+	fmt.Println("$$$$$")
+	fmt.Println(restConfig.Host)
```
wonderflow, line 132: remove this debug info after you finished

wonderflow (changes requested):
Great job

Makefile
wonderflow: oh, yes, it could be deleted

Makefile
chivalryq: All "lint" like pre action deleted juest for accelerate test process，i forgot to recover, sorry

charts/vela-core/values.yaml
chivalryq: right, I'll try to turn the e2e start option into a environment variable.

ryanzhang-oss (commented):
Please highlight the before and after coverage percentage in the description.

ryanzhang-oss:
The build seems to be broke

chivalryq:
> The build seems to be broke

I resume kubevela-controller.yaml. Now instead of hard code in kubevela-controller.yaml, I run a script to substitude args in that when e2e test.

charts/vela-core/templates/kubevela-controller.yaml
chivalryq: done!

wonderflow (approved):
great job

yangsoon:
good job, it's very meaningful

Issue #1100, opened 2021-02-25
[Feature] Consider built-in components w/o template module - so we can reference a workload as template


resouer:
**Is your feature request related to a problem? Please describe.**

This is a way that `Rollout` can work with raw workload instance, it will be helpful to demos and onboarding users. 

**Describe the solution you'd like**

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
    rollout.kubevela.io: true # the flag
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```

Then it will begin to rollout with immutable instances!

**What happened**

Because we have several built-in `ComponentDefinition`, let's say:

```yaml
apiVersion: core.oam.dev/v1alpha3
kind: ComponentDefinition
metadata:
  name: built-in-deployment # reserved name
spec:
  workload:
    definition:
      apiVersion: apps/v1
      kind: Deployment
    characteristics:
      replicable: true
      podspecable: true
      revisionable: true # means it will be revisioned
  schematic: {} # empty means it will reference a raw resource as template, no component package needed
```

And, we also have a adaptor (webhook?). It will convert (create, update, delete) any Deployment with `rollout.kubevela.io: true` to an `Application` with built-in `ComponentDefinition`, e.g.

```yaml
apiVersion: core.oam.dev/v1alpha2
kind: Application
metadata:
  name: nginx-deployment
spec:
  components:
    - name: nginx-deployment
      type: built-in-deployment
      settings:
        metadata: # name and `rollout` label should be removed
          labels:
            app: nginx
        spec:
          replicas: 3
          selector:
            matchLabels:
              app: nginx
          template:
            metadata:
              labels:
                app: nginx
            spec:
              containers:
              - name: nginx
                image: nginx:1.14.2
                ports:
                - containerPort: 80
```

Then `Application` controller should be able to render such `module: {}` app. The reason I want to reuse `ComponentDefinition` is it can still be used to control the behavior of the workload.

And when user attach a `Rollout` to it:

```yaml
apiVersion: core.oam.dev/v1alpha1
kind: RolloutTrait # we really need to rename it to Rollout ...
metadata:
  name: canary
spec:
  application: nginx-deployment # name of the application
  ... # cool rollout spec
```
[...]

PR #1555, opened 2021-04-22
implement nomad component

captainroy-hy:
related to #1100 
to do 
- [ ] unit test
- [ ] doc


ryanzhang-oss:
I am not sure if this is something of high priority and we may need a bit more design on exactly what does this special case componentDef enabled.

docs/examples/app-with-nomad/mycloneset-adopted.yaml
```diff
@@ -0,0 +1,45 @@
...
+spec:
+  components:
+  - name: mycloneset
+    properties:
+      apiVersion: apps.kruise.io/v1alpha1
+      kind: CloneSet
+      metadata:
+        name: mycloneset
+        namespace: default
+      spec:
```
resouer, line 15: These boilerplate lines should be removed and they should be stored in component definition.

resouer:
@ryanzhang-oss Agree, the direction dose not fully align with the original issue #1100

@wonderflow @captainroy-hy The intention of #1100 is to enable user to use KubeVela as an independent rollout controller in his existing system. This assumption is only valid when `Rollout` becomes a highly demanded feature but user already has a legacy system and can't migrate.

wonderflow:
closing as this is not the intended design.

PR #1557, opened 2021-04-23
add deployment scale controller

captainroy-hy:
add deployment scale controller
add ut latter

### DIFF
--- a/pkg/controller/common/rollout/workloads/deployment_scale_controller.go
+++ b/pkg/controller/common/rollout/workloads/deployment_scale_controller.go
@@ -1,0 +1,334 @@
+/*
+Copyright 2021 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workloads
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/crossplane/crossplane-runtime/pkg/event"
+	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
+	"github.com/oam-dev/kubevela/apis/standard.oam.dev/v1alpha1"
+	"github.com/oam-dev/kubevela/pkg/oam"
+	"github.com/oam-dev/kubevela/pkg/oam/util"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/controller/common/rollout/workloads/deployment_scale_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	err := s.fetchDeployment(ctx)
+	if err != nil {
+		s.rolloutStatus.RolloutRetry(err.Error())
+		// nolint:nilerr
+		return false, nil
+	}
+	newPodTarget := calculateNewBatchTarget(s.rolloutSpec, int(s.rolloutStatus.RolloutOriginalSize),
+		int(s.rolloutStatus.RolloutTargetSize), int(s.rolloutStatus.CurrentBatch))
+	// get the number of ready pod from deployment
+	// TODO: should we use the replica number when we shrink?
```
wonderflow, line 195: what does it mean?

pkg/controller/common/rollout/workloads/deployment_scale_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	if currentBatch.MaxUnavailable != nil {
+		unavail, _ = intstr.GetValueFromIntOrPercent(currentBatch.MaxUnavailable,
+			util.Abs(int(s.rolloutStatus.RolloutTargetSize-s.rolloutStatus.RolloutOriginalSize)), true)
+	}
+	klog.InfoS("checking the scaling progress", "current batch", s.rolloutStatus.CurrentBatch,
+		"new pod count target", newPodTarget, "new ready pod count", readyPodCount,
+		"max unavailable pod allowed", unavail)
+	s.rolloutStatus.UpgradedReadyReplicas = int32(readyPodCount)
+	targetReached := false
+	// nolint
```
wonderflow, line 208: nolint for what?

wonderflow (commented):
ping @ryanzhang-oss please review

pkg/controller/common/rollout/workloads/controller.go
```diff
@@ -104,3 +105,39 @@ func (c *cloneSetController) fetchCloneSet(ctx context.Context) error {
...
+	err := c.client.Get(ctx, c.targetNamespacedName, &workload)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			c.recorder.Event(c.parentController, event.Warning("Failed to get the Deployment", err))
+		}
+		return err
+	}
+	c.deploy = &workload
+	return nil
+}
```
ryanzhang-oss, line 143: Why are those here? Looks like they are only used by the deploymentScaler

pkg/controller/common/rollout/workloads/deployment_scale_controller.go
ryanzhang-oss: This is copied from the clonesetController. We use the "readyReplicas" to check if the scaleup succeeded or not. it seems that we should use "replicas" to check if we scale down

pkg/controller/common/rollout/workloads/deployment_scale_controller.go
ryanzhang-oss: if/else

ryanzhang-oss (commented):
LGTM. Please see if you can run the deployment examples with this check-in. 
The next step is to make this more generic.

pkg/controller/common/rollout/workloads/controller.go
captainroy-hy: Fixed. I merged these helper functions into DeploymentScaleController.

pkg/controller/common/rollout/workloads/deployment_scale_controller.go
wonderflow: yes, check `replicas` in scale down case is better

wonderflow (approved):
lgtm @ryanzhang-oss  please merge after check

ryanzhang-oss (approved):

PR #1558, opened 2021-04-23
docs: remove the command prompt to support fast copy/paste

SataQiu:
The copied command that starts with `$` will not execute correctly.
This PR removes the command prompt to support fast copy/paste.


wonderflow (approved):
LGTM

Issue #1552, opened 2021-04-22
[Feature] Embed rollout in an application


wangyikewxgm:
**Is your feature request related to a problem? Please describe.**
Support embedded rollout settings in an application. In this way, any changes to the
application will naturally roll out in a controlled manner instead of a sudden replace.

**Describe the solution you'd like**
IMO we can split the solution to 2 pr
1.refactor rollout Reconcile func to let it could be called by application. Guarantee  logic work well  and pass all tests
2.write application side logic and add tests.

**Describe alternatives you've considered**
If the solution of refactor rollout reconcile cannot work well. Consider create a AppRollout object in application control loop.

ryanzhang-oss:
> If the solution of refactor rollout reconcile cannot work well. Consider create a AppRollout object in application control loop.

There might be a few problems with this approach

- appRollout has finalizer logics that won't apply when it's not the entire object
- appRollout reconciler has defer functions that might not apply when it's not the entire object

wonderflow:
Let me clarify the design:

```
apiVersion: core.oam.dev/v1beta1
kind: ScopeDefinition
metadata:
  name: rollout
spec:
  definitionRef:
    name: approllouts.core.oam.dev
    version: v1beta1
```

Then:

```
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: example-app
spec:
  components:
    - name: testsvc
      type: webservice
      properties:
        addRevisionLabel: true
        image: crccheck/hello-world
        port: 8000
  scopes:
    - type: rollout
      properties:
        sourceAppRevisionName: test-rolling-v1
        targetAppRevisionName: test-rolling-v2
        componentList:
          - metrics-provider
        rolloutPlan:
          targetSize: 5
          rolloutBatches:
            - replicas: 1
            - replicas: 3
```

Then we will create the rollout object, the name of the rollout is automatically generated:

```
apiVersion: core.oam.dev/v1beta1
kind: AppRollout
metadata:
  name: example-app-rollout-1
spec:
    sourceAppRevisionName: test-rolling-v1
    targetAppRevisionName: test-rolling-v2
    componentList:
      - metrics-provider
    rolloutPlan:
      targetSize: 5
      rolloutBatches:
        - replicas: 1
        - replicas: 3
```

resouer:
[...]

PR #1559, opened 2021-04-23
 Refactor rollout reconcile 

wangyikewxgm:
ref #1552 
this is step1 to embed rollout in an application


ryanzhang-oss:
should this be closed?

wangyikewxgm:
> should this be closed?
 Yes, contained in #1568

PR #1575, opened 2021-04-28
Check whether parameter.cmd is nill for sidecar trait

cnbailian:
Deploying error with https://kubevela.io/zh/docs/application#step-2-design-and-deploy-application , error message:
```
evaluate template trait=sidecar app=frontend: invalid patch trait sidecar into workload: compile patch file: spec.template.spec.containers.0.command: reference \"parameter\" not found
```
This PR is referenced from https://github.com/oam-dev/kubevela/pull/1067/files

### DIFF
--- a/charts/vela-core/templates/defwithtemplate/sidecar.yaml
+++ b/charts/vela-core/templates/defwithtemplate/sidecar.yaml
@@ -16,9 +16,11 @@ spec:
         patch: {
         	// +patchKey=name
         	spec: template: spec: containers: [{
-        		name:    parameter.name
-        		image:   parameter.image
-        		command: parameter.cmd
+        		name:  parameter.name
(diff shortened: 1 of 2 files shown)


captainroy-hy:
Good catch! @cnbailian  Pls fix it [here ](https://github.com/oam-dev/kubevela/blob/master/hack/vela-templates/cue/sidecar.cue) And run `make manifests` to generate automatically, instead of modifying template YAML in charts directly.

wonderflow:
please format the cue file by `cue fmt`

wonderflow (approved):
good catch, lgtm after CI

cnbailian:
OK, I'm sorry I didn't learn the working mechanism here in advance. Thanks @captainroy-hy @wonderflow

PR #1576, opened 2021-04-28
Change metric-server port in unit test

zzxwill:
When running `make test` locally, this change can avoid the conflicts with local 8080 port by running kubevela
core and by other test cases .


wonderflow (approved):

go.sum
```diff
@@ -1,4 +1,3 @@
-bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898 h1:SC+c6A1qTFstO9qmB86mPV2IpYme/2ZoEQ0hrP+wo+Q=
```
ryanzhang-oss, line 1: why does the port change PR introduce that many go.sum changes?

zzxwill: I did `go clean -modcache` before `go mod tidy`, and I did it twice to make sure `go.sum` needs to be updated. So I was guessing master branch might be not clean.

wonderflow: feel free to update the makefile if our `make reviewable` not accurate

PR #1577, opened 2021-04-28
Go1.16latest

wangyuan249:


Makefile
```diff
@@ -93,6 +93,7 @@ generate-source:
 
 cross-build:
 	rm -rf _bin
+	go get github.com/mitchellh/gox@v0.4.0
```
yangsoon, line 96: Why not directly modify the go.mod and go.sum?

wonderflow (approved):

PR #1580, opened 2021-04-29
Lowercase "Properties" in code and docs for definition reference

zzxwill:


wonderflow:
Why we need to change?

docs/en/developers/references/README.md
```diff
@@ -0,0 +1,113 @@
...
++------+------------------------------------------------------------------+--------+----------+---------+
+```
+
+## For Built-in Capabilities
+
+Note that for all the built-in capabilities, we already published their reference docs below based on the same doc generation mechanism.
+
+
+- Workload Types
+	- [webservice](component-types/webservice)
```
yangsoon, line 105:
```suggestion
	- [webservice](../../end-user/components/webservice)
```
the link should be set to a relative path

PR #1582, opened 2021-04-29
Add krew auto release and docs

sunny0826:
[kubectl vela krew plugin manifests](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/vela.yaml) has been merged into krew-index, and it is now possible to install kubectl vela via krew, now with the addition of an automatic release action that will automatically push each subsequent release to krew.


wonderflow (approved):

yangsoon:
Good Job, we can not only use `krew` to download vela plugin but also automatically release new version plugin. \lgtm

PR #1584, opened 2021-04-29
fix controller: fix the bug which the controller cannot start due to the format error of the third-party CRD 

yangsoon:
Refer to https://github.com/oam-dev/kubevela/issues/1313 , the format error of third-party CRD will lead to vela-controller exited. In this case, we need to keep the vela-controller run normally.

### DIFF
--- a/pkg/dsl/definition/package.go
+++ b/pkg/dsl/definition/package.go
@@ -70,7 +70,7 @@ func NewPackageDiscover(config *rest.Config) (*PackageDiscover, error) {
 		pkgKinds: make(map[string][]VersionKind),
 	}
 	if err = pd.RefreshKubePackagesFromCluster(); err != nil {
-		return nil, err
+		return pd, err
 	}
 	return pd, nil
 }
--- a/pkg/dsl/definition/package.go
+++ b/pkg/dsl/definition/package.go
@@ -42,6 +42,9 @@ const (
 	BuiltinPackageDomain = "kube"
 	// K8sResourcePrefix Indicates that the definition comes from kubernetes
 	K8sResourcePrefix = "io_k8s_api_"
+
+	// ParseJSONSchemaErr describes the error that occurs when cue parses json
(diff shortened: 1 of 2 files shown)


pkg/dsl/definition/package.go
```diff
@@ -70,7 +70,7 @@ func NewPackageDiscover(config *rest.Config) (*PackageDiscover, error) {
 		pkgKinds: make(map[string][]VersionKind),
 	}
 	if err = pd.RefreshKubePackagesFromCluster(); err != nil {
-		return nil, err
+		return pd, err
```
wonderflow, line 73: will the pd be nil?

yangsoon: `pd` will be nil, i will add some check and prevent it

yangsoon:
I add an new error type `CUEParseErr` to represent the error when cue parse k8s openapiv2, When we start vela-controller,
If the packageDiscover fails to initialize due to an error in parsing the json schema, it will not affect the startup of the controller.

wonderflow (approved):
LGTM

pkg/dsl/definition/package.go
```diff
@@ -59,6 +62,25 @@ type VersionKind struct {
...
+	return fmt.Sprintf("%s: %s", cueErr.errType, cueErr.err.Error())
+}
+
+// IsCUEParseErr returns true if the specified error is CUEParseError type.
+func IsCUEParseErr(err error) bool {
+	return errors.As(err, &CUEParseError{})
+}
+
 // NewPackageDiscover will create a PackageDiscover client with the K8s config file.
 func NewPackageDiscover(config *rest.Config) (*PackageDiscover, error) {
```
ryanzhang-oss, line 85: better to have a UT to make sure it works

wonderflow: yes, ping @yangsoon

yangsoon: got it i will add UT

PR #1596, opened 2021-05-01
[Design] Environment-Based Application Configuration Management

hongchaodeng:
This is a continuation based on the discussion from https://github.com/oam-dev/kubevela/issues/1048 and https://github.com/oam-dev/velacp/issues/85.

### DIFF
--- a/design/vela-core/environment_template_appgroup.md
+++ b/design/vela-core/environment_template_appgroup.md
@@ -1,0 +1,132 @@
+# Environment-Based Application Configuration Management
+
+## Background
+
+KubeVela users want an "Environment" concept to define shared-base for application deployments. For example, different applications might share the same cluster, DB connection secrets, capability definitions, network boundaries (VPC), etc. An environment encapsulates these shared resources into one concept and enables better manageability.
+
+Here are some user scenarios from our offline survey:
+
+- Users have a template base, and want to make modifications based on selected environment. For example, set different image names for dev/prod environments.
+- Users want the definitions in dev/prod environments to be different. For example, the logging trait might be uploaded to OSS/S3 buckets in dev or sinked to ELK in prod.
+- Users want the secrets in dev/prod environments to be different. For example, the DB connection secrets are different in dev/prod environments.
+
+
+## Proposal
+
+This design proposes to add the following CRDs to satisfy user requirements.
+
+### 1 Environment CRD
+
+Environment CRD defines the shared-base environment to deploy applications to.
+
+```yaml
+kind: Environment
+spec:
+  clusters: # Apply to the following clusters.
+    - prod-cluster
+
+  secrets: # The secrets that will be created/updated.
+    - name: redis
+      data:
+        url: redis-url
+        password: redis-password
+
+  definitions: # The definitions that will be created/updated.
+    - type: Component
+      name: function
+      source:
+        kube: # using kubectl apply 
+          git: git-repo-url
+          path: path/to/definition # path to a file or a directory
+    - type: Trait
+      name: logging
+      source:
+        helm: # using helm install/upgrade
+          git: git-repo-url
+          path: path/to/chart
+          valuePath: value.yaml
+```
+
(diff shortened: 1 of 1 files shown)


hongchaodeng:
cc @wonderflow @yangsoon @resouer @LeoLiuYan @Ghostbaby

hongchaodeng:
Ping @wonderflow for review

design/vela-core/environment_template_appgroup.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+Note that this is a data object to reference to. It wouldn't have any effects if applied by itself.
+
+### 3. ApplicationSetting CRD
+
+With the above ApplicationTemplate, how could users pick a template and define environment-specific parameters? This is what ApplicationSetting CRD defines.
+
+```yaml
+kind: ApplicationSetting
```
wonderflow, line 94: do we have a better name?

wonderflow: Name candidates from my mind:

- ApplicationSet
- ApplicationBundle
- ApplicationBinding

design/vela-core/environment_template_appgroup.md
```diff
@@ -0,0 +1,132 @@
...
+        project: dev
+        databaseType: mysql
+  - targetEnvs:
+      - prod
+    template: my-app-tpl # name of the ApplicationTemplate
+    config:
+      parameters:
+        project: prod
+        databaseType: rds
+      patch: # kustomize-style overlay patch to base template
```
wonderflow, line 111: we may only implement the ApplicationTemplate first, and leave the patch feature in the future when users really need it.

wonderflow (commented):
Generally LGTM, please create a better name for the `ApplicationSetting` in design.

design/vela-core/environment_template_appgroup.md
```diff
@@ -0,0 +1,132 @@
...
+Note that this is a data object to reference to. It wouldn't have any effects if applied by itself.
+
+### 3. ApplicationSetting CRD
+
+With the above ApplicationTemplate, how could users pick a template and define environment-specific parameters? This is what ApplicationSetting CRD defines.
+
+```yaml
+kind: ApplicationSetting
+spec:
+  applications:
```
resouer, line 96: Are we able to make `Application` the entry point? We should avoid having too many entries in general.

hongchaodeng:
A note from community chat:

To support flexibility in templating options, we should not have a specific method like `ApplicationSetting`. Instead, we probably should base on the policy design in https://github.com/oam-dev/kubevela/issues/1613, and have specify policy to render it:

```yaml
kind: Application
spec:
  policy:
  - name: cueTemplate
    properties:
      template: my-app-tpl # name of the ApplicationTemplate
      parameters: # cue parameters
        project: dev
        databaseType: mysql
  - name: kustomizePatch
    properties:
      template: my-app-tpl # name of the ApplicationTemplate
      patch: # kustomize-style overlay patch to base template
        spec:
          components:
            - name: instance
              properties:
                image: prod-image
```

design/vela-core/environment_template_appgroup.md
LeoLiuYan: In my opinion, `ApplicationTemplate`  is a top-level object. Users could directly create `Application` object, or use `ApplicationTemplate`+`ApplicationSettings` to generate `Application` in various environments.  

It may be strange to somebody if  `ApplicationTemplate` as a policy of `Application`, at least for me.

@hongchaodeng @resouer

LeoLiuYan: > It is not about making ApplicationTemplate as a policy. It is about making the templating method as a policy, e.g. cue/helm/kustomize. We can still have the upper layer like ApplicationSet to manage config.

@hongchaodeng  Got it.

wonderflow (changes requested):
Let's refine the doc, keep the environment CRD part and refactor the left with the workflow design

design/vela-core/environment_template_appgroup.md
hongchaodeng: Adding a note here. As I'm implementing the workflow, I realize that this is not intuitive in workflow. The config templating or overlay could be better done at admission layer.

hongchaodeng:
Closing this. I will come up with a new design which is much better.

PR #1619, opened 2021-05-08
Enable Dynamic Admission Control for Application

yangsoon:
enable dynamic admission control for Application,  the admissionWebhooks will check the correctness of the fields in Application.

### DIFF
--- a/pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler.go
@@ -98,5 +98,5 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
 // RegisterValidatingHandler will register application validate handler to the webhook
 func RegisterValidatingHandler(mgr manager.Manager, args controller.Args) {
 	server := mgr.GetWebhookServer()
-	server.Register("/validating-core-oam-dev-v1alpha2-applications", &webhook.Admission{Handler: &ValidatingHandler{dm: args.DiscoveryMapper, pd: args.PackageDiscover}})
+	server.Register("/validating-core-oam-dev-v1beta1-applications", &webhook.Admission{Handler: &ValidatingHandler{dm: args.DiscoveryMapper, pd: args.PackageDiscover}})
 }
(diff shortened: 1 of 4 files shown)


pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler.go
```diff
@@ -85,9 +85,10 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
 		if err := h.Decoder.DecodeRaw(req.AdmissionRequest.OldObject, oldApp); err != nil {
 			return admission.Errored(http.StatusBadRequest, err)
 		}
-
-		if allErrs := h.ValidateUpdate(ctx, app, oldApp); len(allErrs) > 0 {
-			return admission.Errored(http.StatusUnprocessableEntity, allErrs.ToAggregate())
+		if app.ObjectMeta.DeletionTimestamp.IsZero() {
+			if allErrs := h.ValidateUpdate(ctx, app, oldApp); len(allErrs) > 0 {
```
zzxwill, line 89:
```suggestion
			if len(h.ValidateUpdate(ctx, app, oldApp)) > 0 {
```
Better? :)

pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler.go
```diff
@@ -98,5 +99,5 @@ func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) a
 // RegisterValidatingHandler will register application validate handler to the webhook
 func RegisterValidatingHandler(mgr manager.Manager, args controller.Args) {
 	server := mgr.GetWebhookServer()
-	server.Register("/validating-core-oam-dev-v1alpha2-applications", &webhook.Admission{Handler: &ValidatingHandler{dm: args.DiscoveryMapper, pd: args.PackageDiscover}})
+	server.Register("/validating-core-oam-dev-v1beta1-applications", &webhook.Admission{Handler: &ValidatingHandler{dm: args.DiscoveryMapper, pd: args.PackageDiscover}})
```
zzxwill, line 102: What will happed if there the old webhook is already registered and running in a cluster?

yangsoon: In fact, vela has never enable the Admission Control for Application

pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler.go
yangsoon: This cannot be modified here, because the following logic needs to return the error in `allErrs`

wonderflow (approved):
good job

PR #1633, opened 2021-05-10
fix application rollout annotation false issue

### LINKED ISSUE #1632: [Bug] application  "app.oam.dev/rollout-template": "false" annotation don't works
**Describe the bug**

**To Reproduce**
apply the application yaml
```yaml
kind: Application
metadata:
  name: test-rolling
  namespace: vela-test
  annotations:
    "app.oam.dev/rolling-components": "metrics-provider"
    "app.oam.dev/rollout-template": "false"
spec:
  components:
    - name: metrics-provider
      type: worker
      properties:
        cmd:
          - ./podinfo
          - stress-cpu=1
        image: stefanprodan/podinfo:4.0.6
        port: 8080
        updateStrategyType: InPlaceIfPossible
        replicas: 3
```
application controller still assume the app is not  inplace one.  And don't create appContext for it

wangyikewxgm:
fix #1632

### DIFF
--- a/pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
@@ -42,7 +42,7 @@ func (h *ValidatingHandler) ValidateCreate(ctx context.Context, app *v1beta1.App
 		componentErrs = append(componentErrs, field.Invalid(field.NewPath("schematic"), app, err.Error()))
 	}
 	if v := app.GetAnnotations()[oam.AnnotationAppRollout]; len(v) != 0 && v != "true" {
-		componentErrs = append(componentErrs, field.Invalid(field.NewPath("annotation:app.oam.dev/rollout-template"), app, "true is the only valid value"))
+		componentErrs = append(componentErrs, field.Invalid(field.NewPath("annotation:app.oam.dev/rollout-template"), app, "the annotation value of rollout-template must be true"))
 	}
 	return componentErrs
 }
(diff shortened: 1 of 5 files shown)


wonderflow (commented):
ping @ryanzhang-oss

test/e2e-test/testdata/app/app5.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: core.oam.dev/v1beta1
+kind: Application
+metadata:
+  annotations:
+    "app.oam.dev/rollout-template": "false"
```
ryanzhang-oss, line 5: What does "false" mean? I think it's better to add a webhook to check instead of allowing this.

wangyikewxgm: Because some users assume that the false  annotations means inplace upgrade

wangyikewxgm:
This annotation value truly make user confused.We can tell user `true` is the only valid value in docs. I  close this pr.

pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
```diff
@@ -40,7 +41,9 @@ func (h *ValidatingHandler) ValidateCreate(ctx context.Context, app *v1beta1.App
 	if err := appParser.ValidateCUESchematicAppfile(af); err != nil {
 		componentErrs = append(componentErrs, field.Invalid(field.NewPath("schematic"), app, err.Error()))
 	}
-
+	if v := app.GetAnnotations()[oam.AnnotationAppRollout]; len(v) != 0 && v != "true" {
+		componentErrs = append(componentErrs, field.Invalid(field.NewPath("annotation:app.oam.dev/rollout-template"), app, "true is the only valid value"))
```
wonderflow, line 45:
```suggestion
		componentErrs = append(componentErrs, field.Invalid(field.NewPath("annotation:app.oam.dev/rollout-template"), app, "the annotation value of rollout-template must be true"))
```

wonderflow (approved):

PR #1636, opened 2021-05-11
fix docs:  add how to specify definition revision in Application

### LINKED ISSUE #1637: [Bug] Worker component should not expose port
It's wrong to add a `port` to worker to demo definition revisions: https://kubevela.io/docs/next/end-user/definition-revision#usage-of-definition

For example, copied from https://docs.digitalocean.com/products/app-platform/concepts/worker/:

- (Web) Service: a type of component that is accessible via public or internal ports, and running server-side code written in a supported programming language, such as Python, Ruby, Go, Node.js, or PHP. Because services are accessible, they need to be configured with HTTP port and HTTP route values, so that the load-balancer can distribute traffic properly or define internal ports which are only accessible by other components such as workers or other services within the app.
- Worker:  a worker is a type of component that is running server-side code written in a supported programming language, such as Python, Ruby, Go, Node.js, or PHP, and is not internet-accessible. Because workers can can not accept HTTP requests, they are only suitable for doing background tasks, such as queue processing or video encoding.

Please fix this issue as well as the description of definitions.

yangsoon:
add docs about how to specify definition revision in Application.

fix #1637,  fix ComponentDefinition `worker` that `worker` should not expose port


docs/en/end-user/application.md
```diff
@@ -68,6 +68,87 @@ spec:
           - '1000'
 ```
 
+### Specify Definition Revision in Application
```
resouer, line 71: This doc should be moved to platform team doc and simplified, an example could be: https://github.com/resouer/kubevela/blob/97037b2efe10bee3791fbe3b985be8fabc90049d/docs/en/platform-engineers/definition-revision.md

Also, please fix this issue: https://github.com/oam-dev/kubevela/issues/1637

yangsoon: got it

docs/en/platform-engineers/definition-and-templates.md
```diff
@@ -234,3 +234,41 @@ spec:
...
+## Definition Revisions
+
+In KubeVela, definition entities are mutable. Each time a `ComponentDefinition` or `TraitDefinition` is updated, a corresponding `DefinitionRevision` will be generated to snapshot this change. Hence, KubeVela allows user to reference a specific revision of definition to declare an application.
+
+### Specify Definition Revision in Application
+
+For example, we can update the worker component definition by applying a new definition with same name as below.
+
+```shell
+kubectl apply -f https://raw.githubusercontent.com/oam-dev/kubevela/master/docs/examples/definition-revision/worker-v2.yaml
```
wonderflow, line 247: please use a webservice as example because the above is webservice

yangsoon: got it

wonderflow (approved):
good job

PR #1656, opened 2021-05-13
fix bug: remove unneeded workload definitions

yangsoon:
fix https://github.com/oam-dev/kubevela/issues/1563.

The previous logic of controller will create a workloadDefinition with the same name as ComponentDefinition.
But this will cause confusion about the concepts of ComponentDefinition and workloadDefinition.

now this pr remove the unneeded workload definitions and align with https://github.com/oam-dev/kubevela/issues/1563.

```shell
$ kubectl get workload -n vela-system
NAME                                  DEFINITION-NAME                       DESCRIPTION
containerizedworkloads.core.oam.dev   containerizedworkloads.core.oam.dev
deployments.apps                      deployments.apps
jobs.batch                            jobs.batch
```

 for details, please see:

https://github.com/oam-dev/spec/blob/master/3.component_model.md
https://github.com/oam-dev/spec/blob/master/4.workload_types.md


pkg/controller/utils/capability.go
```diff
@@ -67,6 +66,26 @@ type CapabilityComponentDefinition struct {
...
+	}
+	if componentDefinition.Spec.Schematic != nil && componentDefinition.Spec.Schematic.HELM != nil {
+		def.WorkloadType = util.HELMDef
+		def.Helm = componentDefinition.Spec.Schematic.HELM
+	}
+	if componentDefinition.Spec.Schematic != nil && componentDefinition.Spec.Schematic.KUBE != nil {
+		def.WorkloadType = util.KubeDef
+		def.Kube = componentDefinition.Spec.Schematic.KUBE
+	}
+	def.ComponentDefinition = *componentDefinition
```
captainroy-hy, line 85: `def.ComponentDefinition = *componentDefinition.DeepCopy()` is safer since comdDef is passed from external as a pointer.

yangsoon: good suggestion, I will fix it

wonderflow (approved):

PR #1657, opened 2021-05-14
fix staticcheck lint

captainroy-hy:
fix staticcheck lint ST1023.
make CI work consistenly with Makefile/staticcheck

Signed-off-by: roy wang <seiwy2010@gmail.com>

PR #1658, opened 2021-05-14
fix golint error

wangyikewxgm:
try to fix golint error ref to https://github.com/oam-dev/kubevela/runs/2581810914


captainroy-hy:
dup of #1657

wangyikewxgm:
> dup of #1657
cool, closed

PR #1677, opened 2021-05-18
fix makefile: update docker image in make docs-start/docs-build

yangsoon:
we provide a new docker image `oamdev/vela-webtool:v1`, you can find `Dockerfile` in `/hack/website`.


wonderflow (approved):
LGTM

PR #1678, opened 2021-05-18
Add expose

chivalryq:
Add one trait: "expose" as KubeVela's inner trait


hack/vela-templates/definitions/expose.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: core.oam.dev/v1beta1
+kind: TraitDefinition
+metadata:
+  annotations:
+    definition.oam.dev/description: "expose port to enable web traffic for your service."
```
wonderflow, line 5:
```suggestion
    definition.oam.dev/description: "Expose port to enable web traffic for your component."
```

hack/vela-templates/definitions/expose.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: expose
+  namespace: {{.Values.systemDefinitionNamespace}}
+spec:
+  appliesToWorkloads:
+    - webservice
```
yangsoon, line 10:
```suggestion
    - deployments.apps
```
please change to workloadDefinition name, please refer to https://github.com/oam-dev/kubevela/issues/1635

chivalryq: yes, i'll fix

hack/vela-templates/definitions/expose.yaml
chivalryq: 👌

wonderflow (approved):
lgtm

charts/vela-core/templates/defwithtemplate/expose.yaml
```diff
@@ -0,0 +1,36 @@
+# Code generated by KubeVela templates. DO NOT EDIT.
+apiVersion: core.oam.dev/v1beta1
+kind: TraitDefinition
+metadata:
+  annotations:
+    definition.oam.dev/description: "Expose port to enable web traffic for your component."
```
resouer, line 6: This will be really confusing with web service component, not sure how we will expect user use this. To expose port in a chart? I believe most chart will package service as part of the component.

PR #1686, opened 2021-05-19
fix webhook: set a default workload.type for componentdefinition which doesn't refer to a workload

yangsoon:
refer to https://github.com/oam-dev/kubevela/pull/1680#discussion_r634808401

The validating logic of the componentdeifnition webhook will reject if the componentDefinition doesn't refer to a workload.
Now we will set a default workloaddefinition `autodetect.core.oam.dev` for the componentDefinition.

### DIFF
--- a/charts/vela-core/templates/definitions/autodetect.yaml
+++ b/charts/vela-core/templates/definitions/autodetect.yaml
@@ -1,0 +1,8 @@
+apiVersion: core.oam.dev/v1beta1
+kind: WorkloadDefinition
+metadata:
+  name: autodetect.core.oam.dev
+  namespace: {{.Values.systemDefinitionNamespace}}
+spec:
+  definitionRef:
+    name: autodetect.core.oam.dev
--- a/charts/vela-core/templates/definitions/autodetect.yaml
+++ b/charts/vela-core/templates/definitions/autodetect.yaml
@@ -1,8 +1,10 @@
 apiVersion: core.oam.dev/v1beta1
 kind: WorkloadDefinition
 metadata:
-  name: autodetect.core.oam.dev
+  annotations:
+    definition.oam.dev/description: "autodetects.core.oam.dev is the default workload type of ComponentDefinition"
+  name: autodetects.core.oam.dev
   namespace: {{.Values.systemDefinitionNamespace}}
 spec:
   definitionRef:
-    name: autodetect.core.oam.dev
+    name: autodetects.core.oam.dev
--- a/pkg/controller/core.oam.dev/v1alpha2/application/apply.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/apply.go
@@ -86,6 +86,7 @@ type appHandler struct {
 	revisionHash             string
 	acrossNamespaceResources []v1beta1.TypedReference
 	resourceTracker          *v1beta1.ResourceTracker
+	autodetect               bool
 }
 
 // setInplace will mark if the application should upgrade the workload within the same instance(name never changed)
@@ -132,12 +133,21 @@ func (h *appHandler) apply(ctx context.Context, appRev *v1beta1.ApplicationRevis
 		return h.createOrUpdateAppRevision(ctx, appRev)
 	}
 
+	var needTracker bool
+	var err error
 	for _, comp := range comps {
 		comp.SetOwnerReferences(owners)
-		needTracker, err := h.checkAndSetResourceTracker(&comp.Spec.Workload)
+		if h.checkAutoDetect(comp) && h.isNewRevision {
+			if err = h.applyHelmModuleResources(ctx, comp, owners); err != nil {
+				return errors.Wrap(err, "cannot apply Helm module resources")
+			}
+			continue
+		}
+		needTracker, err = h.checkAndSetResourceTracker(&comp.Spec.Workload)
 		if err != nil {
 			return err
 		}
+
 		newComp := comp.DeepCopy()
 		// newComp will be updated and return the revision name instead of the component name
 		revisionName, err := h.createOrUpdateComponent(ctx, newComp)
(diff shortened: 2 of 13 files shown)


wonderflow (changes requested):
please make sure the helm component still works if users don't fill any workload definition info \cc @captainroy-hy

captainroy-hy:
> please make sure the helm component still works if users don't fill any workload definition info \cc @captainroy-hy

That may cause a problem if the workload type ref is not a valid resource registered in API-server. In the stage where App controller checks whether a workload raw is cluster-scope or namespace-scope, discovermapper will return an error if it's an unknown resource to API-server.

wonderflow:
@captainroy-hy  do we have any solution? That we just need to deploy successfully without any traits

charts/vela-core/templates/definitions/autodetect.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: core.oam.dev/v1beta1
+kind: WorkloadDefinition
+metadata:
+  name: autodetect.core.oam.dev
```
zzxwill, line 4: Add an annotation description might help to explain the definition to those who are interested in it.

yangsoon: got it

captainroy-hy:
> @captainroy-hy do we have any solution? That we just need to deploy successfully without any traits

@wonderflow Yes, minor changes can help. During app controller apply, we can skip several steps and just keep applying helm resources. I would like to work with @yangsoon to fix that.

wonderflow:
> > @captainroy-hy do we have any solution? That we just need to deploy successfully without any traits
> 
> @wonderflow Yes, minor changes can help. During app controller apply, we can skip several steps and just keep applying helm resources. I would like to work with @yangsoon to fix that.

very appreciate for that

yangsoon:
With the help of @captainroy-hy , I fix part of the controller logic, please \cc @captainroy-hy

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -494,6 +508,9 @@ func (h *appHandler) generateResourceTrackerName() string {
 // with application
 func (h *appHandler) checkCrossNamespace(u *unstructured.Unstructured) (bool, error) {
 	gk := u.GetObjectKind().GroupVersionKind().GroupKind()
+	if gk == types.AutoDetectWorkloadDefinitionGVK.GroupKind() {
```
captainroy-hy, line 511: This is workable. But it's better to avoid invading `checkCXNamespace`'s logic with "check auto-detect" logic, these two logic are not closely relevant, even though `checkCXNamespace` is the place where an error occurs.  How about moving this check into an independent function?

yangsoon: good suggestion, it's better to implement an independent function

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -176,13 +184,19 @@ func (h *appHandler) apply(ctx context.Context, appRev *v1beta1.ApplicationRevis
...
-	if h.inplace {
+	// `!h.autodetect`: If the workload type of the helm mode component is not clear, an autodetect type workload will be specified by default
+	// In this case, the traits attached to the helm mode component will fail to generate,
+	// so we only call applyHelmModuleResources to create the helm resource, don't generate ApplicationContext.
+	if h.inplace && !h.autodetect {
 		return h.createOrUpdateAppContext(ctx, owners)
 	}
 
+	if h.autodetect {
+		return ErrUseAutoDetect
```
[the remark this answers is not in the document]
yangsoon: got it, the name `ErrUseAutoDetect`  seems to be confused. I will fix it

captainroy-hy (commented):
LGTM.

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -172,17 +182,24 @@ func (h *appHandler) apply(ctx context.Context, appRev *v1beta1.ApplicationRevis
 	ac.SetOwnerReferences(owners)
 	h.FinalizeAppRevision(appRev, ac, comps)
 
+	if h.autodetect {
+		// TODO(yangsoon) autodetect is temporarily not implemented
+		return fmt.Errorf("helm mode component doesn't specify workload")
```
wonderflow, line 187: So the helm based component still can't work in this PR?

wonderflow (commented):
Please clarify which feature do we gain in this PR?

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
yangsoon: The helm-based component still works. This error will be recorded in application CR, to tell the reason why the traits attached to the helm mode component were failed to work.

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -132,12 +133,21 @@ func (h *appHandler) apply(ctx context.Context, appRev *v1beta1.ApplicationRevis
 		return h.createOrUpdateAppRevision(ctx, appRev)
 	}
 
+	var needTracker bool
+	var err error
 	for _, comp := range comps {
 		comp.SetOwnerReferences(owners)
-		needTracker, err := h.checkAndSetResourceTracker(&comp.Spec.Workload)
+		if h.checkAutoDetect(comp) && h.isNewRevision {
```
wonderflow, line 140: why it's related with `h.isNewRevision` ?

test/e2e-test/helm_app_test.go
```diff
@@ -337,6 +337,58 @@ var _ = Describe("Test application containing helm module", func() {
 		}, 240*time.Second, 5*time.Second).Should(Succeed())
 	})
 
+	It("Test deploy an application containing helm module and the componet refer to autodetect type worklaod", func() {
+		cd := v1alpha2.ComponentDefinition{}
+		cd.SetName("wrong-cd")
```
wonderflow, line 342: wrong cd?

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
wonderflow: please clarify in the code

yangsoon: sorry, i will fix it

test/e2e-test/helm_app_test.go
yangsoon: sorry, i will fix the name

wonderflow (approved):

PR #1687, opened 2021-05-19
Minor fix on wording

resouer:
As the title said.

### DIFF
--- a/docs/en/end-user/traits/service-binding.md
+++ b/docs/en/end-user/traits/service-binding.md
@@ -2,57 +2,7 @@
 title:  Service Binding
 ---
 
-Service binding trait will bind data from Kubernetes `Secret` to the application container's ENV.
-
-```yaml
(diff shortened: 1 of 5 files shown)


docs/en/end-user/traits/service-binding.md
```diff
@@ -2,57 +2,7 @@
...
-
-        parameter: {
-        	// +usage=The mapping of environment variables to secret
-        	envMappings: [string]: [string]: string
-        }
-
-```
-
-With the help of this `service-binding` trait, you can explicitly set parameter `envMappings` to mapping all
-environment names with secret key. Here is an example.
```
wonderflow, line 55: please add service-binding as inner trait, put it in kubevela's chart

wonderflow: @zzxwill

wonderflow (approved):

PR #1701, opened 2021-05-24
use custom domain for vela charts

wonderflow:
```
helm repo add https://charts.kubevela.net/core
```


zzxwill (approved):
It's better. /lgtm

Issue #1705, opened 2021-05-24
[Umbrella Issue] Refactor Application to deploy resources directly without AppContext layer


wonderflow:
Currently the Application object(OAM v0.3) still relies on AppContext which is actually AppConfig(OAM v0.2).
As we want to make KubeVela as control plane, we should refactor the Application controller and remove the dependency of AppContext.

## Here's the working items that we have to finish:

- [ ] 1. AppRollout should not rely on AppContext and should be able to work in runtime cluster. #1605
- [x] 2. generate/set name&namespace of workload/trait instances in Application
- [x] 3. set workload reference to trait (TraitDefinition WorkloadRefPath) in Application
- [x] 4. add workload reference to scope (scope feature) in Application
- [x] 5. enable trait system still works on helm-based workload when workload type is given and not rely on AppContext
- [x] 6. controller-owner references of workload/trait instances should be Application instead of AppContext
- [x] 7. add a series of Label/Annotation to workload/trait instances in Application
- [x] 8. convert useful webhook logic in AppConfig to Application
- [x] 9. After all the above was done, let's remove the AppContext dependency in Application Controller.
- [x] 10. package the AppConfig based controller in a separated helm chart #1305
- [x] 11. backward compatibility test (upgrade clusters in each version v1.0.0, ..., v1.0.6)

## Features that contained in AppConfig but different in Application:

1. [DAG-based dependency(data input and output)](https://github.com/crossplane/oam-kubernetes-runtime/blob/master/design/resource-dependency.md): this should be replaced by [workflow](https://github.com/oam-dev/kubevela/blob/master/design/vela-core/workflow_policy.md).

## Features that should be evaluated

The following features work in AppConfig but not have clear use cases, we should evaluate and hold until we have real use cases:

1. [apply-once-only mechanism](https://github.com/oam-dev/kubevela/blob/master/design/vela-core/apply-once-only.md): it works for use cases which the OAM controller should cooperate with others and do not control after deployed. 
2. [component-level revision mechanism](https://github.com/crossplane/oam-kubernetes-runtime/blob/master/design/one-pager-component-mutable-and-versioning.md): in Application of KubeVela, we have AppRevision which records the whole Application, so we don't seem to see there's cases that still need the component level revision.

captainroy-hy:
Here is a basic proposal for the logic workflow of app controller after refactor. The core opinion is aggregating closely-relevant logic into independent and reusable modules.
[...]

PR #1708, opened 2021-05-25
add workflow and policy

hongchaodeng:
ref: https://github.com/oam-dev/kubevela/issues/1613


hongchaodeng:
The code is ready for review.

hongchaodeng:
There is an example tutorial: https://github.com/oam-dev/kubevela/blob/2e2bac2cca9da1a167b8f596e07f99ea07c3862b/docs/examples/workflow/README.md

apis/core.oam.dev/v1beta1/application_types.go
```diff
@@ -52,18 +52,20 @@ type ApplicationComponent struct {
...
 	Type string `json:"type"`
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Properties runtime.RawExtension `json:"properties,omitempty"`
 }
 
 // WorkflowStep defines how to execute a workflow step.
 type WorkflowStep struct {
-	Type string `json:"type"`
+	// Name is the unique name of the workflow step.
+	Name string `json:"name"`
```
wonderflow, line 66: do we need to provide some automatic way to generate the name? e.g the index number

design/vela-core/workflow_policy.md
```diff
@@ -147,13 +147,13 @@ Each workflow step has the following interactions with the app controller:
     - type: workflow-finish
       status: 'True'
       reason: 'Succeeded'
-    - ...
+      message: '{"generation":1}'
```
wonderflow, line 150:
```suggestion
      message: '{"observedGeneration":1}'
```

pkg/appfile/parser.go
```diff
@@ -98,33 +99,67 @@ func (p *Parser) GenerateAppFile(ctx context.Context, app *v1beta1.Application)
 		wds = append(wds, wd)
 	}
 	appfile.Workloads = wds
+
+	var err error
+
+	appfile.Policies, err = p.parsePolicies(ctx, appName, ns, app.Spec.Policies)
+	if err != nil {
+		return nil, err
```
wonderflow, line 107: I  think we need some logs and error wrapper here

pkg/appfile/parser.go
```diff
@@ -98,33 +99,67 @@ func (p *Parser) GenerateAppFile(ctx context.Context, app *v1beta1.Application)
...
+	}
+
+	appfile.WorkflowSteps, err = p.parseWorkflow(ctx, appName, ns, app.Spec.Workflow)
```
wonderflow, line 110: I  think we need some logs and error wrapper here

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -794,6 +795,53 @@ outLoop:
 	return nil
 }
 
+func (h *appHandler) execWorkflows(ctx context.Context, rev string, objects []*unstructured.Unstructured) (bool, error) {
```
wonderflow, line 798: why not create a new file for the workflow logics

wonderflow: we can even make the workflow as a standalone lib

wonderflow (commented):
Great Job! The workflow logic generally Good and Clear for me, we can start to write tests.

ping @captainroy-hy @wangyikewxgm  please also review as it contains Application render refactor

design/vela-core/workflow_policy.md
hongchaodeng: > why don't we use yaml format? 

This is json marshalled data. Yaml is not easy to put in string.

>  json is hard to read and decode mixed with yaml

This is not for human to read and write. This part should be automatically generated by controller.

apis/core.oam.dev/v1beta1/application_types.go
hongchaodeng: One more reason: the name is used to pass down cue context.

apis/core.oam.dev/v1beta1/applicationrevision_types.go
```diff
@@ -50,6 +50,9 @@ type ApplicationRevisionSpec struct {
 	// +kubebuilder:validation:EmbeddedResource
 	// +kubebuilder:pruning:PreserveUnknownFields
 	ApplicationConfiguration runtime.RawExtension `json:"applicationConfiguration"`
+
+	// Resources records the finalized resources from Application.
+	Resources []common.RawComponent `json:"resources,omitempty"`
```
captainroy-hy, line 55: The resources recorded here are duplicated with those recorded in other fields of `appRevision.Spec`. This new field makes it convenient to consume these raw resources. But may I ask if there's other usage or purpose of it?

IMO, these "finalized" resources are not truly finalized, they are not ready to emit to k8s cluster until they are processed (set metadata, set references, etc) by AppContext(AppConfig) currently. In this `Resources` field, different kinds (workloads,traits,policies) of resources are mixed together, which makes it hard to be processed by AppConfig logic.

captainroy-hy: If the goal is to provide a convenient and flexible way to get finalized resources (ready to emit), maybe assemble pkg as we introduced in #1705  can help. It will play as a standalone lib, the only input is an AppRevision, the outcome is a bundle of ready-to-go resources, and they are categorized.

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -80,9 +82,34 @@ func (h *appHandler) setRevisionMetadata(appRev *v1beta1.ApplicationRevision) {
...
+	u, err := util.Object2Unstructured(c.Spec.Workload)
+	if err != nil {
+		panic(err)
+	}
+	u.SetName(c.Name)
+	u.SetNamespace(c.Namespace)
+	return util.Object2RawExtension(u)
 }
 
 // gatherRevisionSpec will gather all revision spec withouth metadata and rendered result.
```
captainroy-hy, line 115: We will take workload/policy into account for revision spec and computing hash, right?

apis/core.oam.dev/v1beta1/applicationrevision_types.go
hongchaodeng: Hi @captainroy-hy . Thanks for your review. Those are very good questions. Let me share my reasoning and thoughts below:

 > The resources recorded here are duplicated with those recorded in other fields of appRevision.Spec.  This new field makes it convenient to consume these raw resources. But may I ask if there's other usage or purpose of it?

Yes. The goal is to use rendered resources to replace those fields like components or appConfig. From controller's perspective, it would be **EASIER** to just emit or use those resources without the structure. This is very different from user/developers' perspective.

Thus, I envision that we would eventually deprecate the Component or AppConfig field in AppRev, and replace them with ready-to-use resources.

> If the goal is to provide a convenient and flexible way to get finalized resources (ready to emit), maybe assemble pkg as we introduced in #1705 can help. It will play as a standalone lib, the only input is an AppRevision, the outcome is a bundle of ready-to-go resources, and they are categorized.

That's fine. I agree with this for the current situation. But again, this seems like workaround to me. Think about how much work this has made users to do: 1. learn about Component and AppConfig in AppRev; 2. learn about the assemble pkg in order to emit them.

Thus, I don't think that makes any sense. The Component/Trait/Workflow/etc. concepts are UI objects. For infra level control logic, we probably don't need them

pkg/controller/core.oam.dev/v1alpha2/application/revision.go
hongchaodeng: Yes. Everything except Workflow will be counted in revision hash.

apis/core.oam.dev/v1beta1/applicationrevision_types.go
captainroy-hy: Sure, I got it! Thx for the details.

And this is a crucial signal ... I should take care.
> Thus, I envision that we would eventually deprecate the Component or AppConfig field in AppRev, and replace them with ready-to-use resources.

leejanee (commented):
will it produce unexpected results? when the workflow executor modifies the derived object of the application, and application reconciler apply rendering results every time

For example, the workflow executor wants to implement Gray Release by modifying the partition of deployment's updatePolicy

hongchaodeng:
@leejanee Not catching your question well. Can you give an example?

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,166 @@
...
+
+	cond, found, err := utils.GetUnstructuredObjectStatusCondition(obj, "workflow-finish")
+	if err != nil {
+		return nil, err
+	}
+
+	if !found || cond.Status != "True" {
+		status.Phase = common.WorkflowStepPhaseRunning
+		return status, nil
+	}
```
cnbailian, line 138: Hi, is there a way to skip this validation if the resource created in step has no status?

hongchaodeng: No. We need to enforce this because the the workflow step needs some feedback on whether it finishes or not.

leejanee:
> @leejanee Not catching your question well. Can you give an example?

Can the workload executor modify the output object of the application? 

> @leejanee Not catching your question well. Can you give an example?

_针对下面场景可能是我有一些点没想明白，不过先抛出来_

一般通过deployment发布应用的场景，会通过动态修改如下的字段实现滚动发布
```
spec: updatePolicy: {
rollingupdate: partition: $(stepNumber)
}
```

在这个场景下 理想的流程是application在集群中创建或者修改deployment，然后workflow再去按节奏修改这个deployment的updatePolicy

按我对这个pr的理解再结合上面的例子
1 不是要等到workflow步骤全部执行完再apply deployment
2 当workflow步骤都完成后如何再次触发，如果诉求是镜像变更希望能够重新触发workflow

pkg/workflow/workflow.go
cnbailian: Thanks for your reply, but then resources such as ConfigMap, TektonTask etc. will not be available in the workflow, and they will always running.

Is it possible to add a field to the WorkflowStepDefinition for skipping the validation of these resources?

hongchaodeng:
@leejanee 非常感谢你的例子！这些问题都问的非常好。我来分享一下我的思路：

> 1 workflow步骤全部执行完才会apply deployment, 如何实现上面滚动升级的例子

workflow 并没有强制最后才 apply deployment。相反，有了 workflow 将不会主动 apply 任何 resource，apply resource 的步骤将交给用户自己去编排 workflow steps。比如说，第一步先 apply deployment，第二步去修改 deployment，第三步 approval。详情请看设计文档：https://github.com/oam-dev/kubevela/blob/master/design/vela-core/workflow_policy.md

> 2 当workflow步骤都完成后如何再次触发，如果诉求是镜像变更希望能够重新触发workflow

我理解当镜像变更后 -> update application image field -> 触发 application event -> trigger workflow，因为版本不一样了。

pkg/workflow/workflow.go
hongchaodeng: For such scenarios they are not put into workflow. They should be put into Component or Policy.

cnbailian: I see, I will rethink the usage scenario of workflow, thanks!

hongchaodeng:
Closing this PR. It is already separated into smaller PRs and merged.

PR #1711, opened 2021-05-25
fix show annotations

### LINKED ISSUE #1618: The properties table for "vela show annotations" is empty
**Describe the bug**

Applying [annotation trait](https://github.com/oam-dev/kubevela/blob/3be63bba91117272b2704953d2255d73ddbd304f/charts/vela-core/templates/defwithtemplate/annotations.yaml), "vela show annotations" display none of properties.

```
vela show annotations
# Properties
+------+-------------+------+----------+---------+
| NAME | DESCRIPTION | TYPE | REQUIRED | DEFAULT |
+------+-------------+------+----------+---------+
+------+-------------+------+----------+---------+
```

wangyuan249:
###New
Only fix map type parameter (the resolution for cue parameter need to be struct kind ), for  single string , int and bool type  we need to fix in  pkg/cue/convert.go  and add  non-struct kind logic.

We should declare the parameter name and ”+usage"  in cue template
fix #1618 
image
the resolution for cue template in function GetCUEParameterValue has no problem

example for two parameter declare and resolution：
image

### DIFF
--- a/e2e/plugin/plugin_test.go
+++ b/e2e/plugin/plugin_test.go
@@ -140,6 +140,12 @@ var _ = Describe("Test Kubectl Plugin", func() {
 			Expect(output).Should(ContainSubstring("targetPort"))
 			Expect(output).Should(ContainSubstring("target port num for service provider."))
 		})
+		It("Test show traitDefinition def with cue single map parameter", func() {
+			tdName := "annotations"
+			output, err := e2e.Exec(fmt.Sprintf("kubectl-vela show %s", tdName))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(output).Should(ContainSubstring("map[string]string"))
+		})
 	})
 
 	Context("Test kubectl vela comp discover", func() {
--- a/references/plugins/references.go
+++ b/references/plugins/references.go
@@ -541,31 +541,8 @@ func (ref *ParseReference) parseParameters(paraValue cue.Value, paramKey string,
 			param.Name = "undefined"
 			param.Required = true
 			tl := paraValue.Template()
-			if tl == nil {
-				sourceStr := paraValue.Source() // original node string for this value
-				typeContain := fmt.Sprintf("%s", sourceStr)
-				switch {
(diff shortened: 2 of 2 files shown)


wangyuan249:
Actually the change can not pass "make reviewable",  after make reviewable,  it will degenerate back to the original cue template..  I am looking for a better solution.

charts/vela-core/templates/defwithtemplate/annotations.yaml
```diff
@@ -8,16 +8,19 @@ metadata:
...
-        		for k, v in parameter {
+        		for k, v in parameter.annotations {
         			"\(k)": v
         		}
         	}
         }
-        parameter: [string]: string
+        parameter:
+            // +usage= Specify annotations for your Workload
+            annotations: [string]: string
```
zzxwill, line 25: Why does parameter  `annotations` have to be added?

e2e/plugin/plugin_suit_test.go
```diff
@@ -96,6 +98,16 @@ var _ = BeforeSuite(func(done Done) {
...
+	traitDefCueMap, err := ioutil.ReadFile("../../charts/vela-core/templates/defwithtemplate/annotations.yaml")
+	Expect(err).Should(BeNil())
+	traitDefCueMapStr := string(traitDefCueMap)
+	traitDefCueMapStr = strings.Replace(traitDefCueMapStr, "{{.Values.systemDefinitionNamespace}}", "default", -1)
+
+	Expect(yaml.Unmarshal([]byte(traitDefCueMapStr), &testTdDefCueMap)).Should(BeNil())
+	testTdDefCueMap.Namespace = "default"
+	err = k8sClient.Create(ctx, &testTdDefCueMap)
+	Expect(err).Should(SatisfyAny(BeNil(), &util.AlreadyExistMatcher{}))
+
```
yangsoon, line 110: it seems no need, e2e environment already has `annotations` trait

leejanee:
LGTM

PR #1712, opened 2021-05-26
duplicate AppConfig logic into app/assemble pkg

captainroy-hy:
related to #1705 

init [assemble phase](https://github.com/oam-dev/kubevela/issues/1705#issuecomment-847574326)

- [x] dup logic of AppConfig render
- [x] handle scope
- [x] unit test

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
@@ -1,0 +1,400 @@
+/*
+Copyright 2021 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package assemble
+
+import (
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
+	ctrlutil "github.com/oam-dev/kubevela/pkg/controller/utils"
+	"github.com/oam-dev/kubevela/pkg/oam"
+	"github.com/oam-dev/kubevela/pkg/oam/util"
+)
+
+// NewAppManifests create a AppManifests
+func NewAppManifests(appRevision *v1beta1.ApplicationRevision) *AppManifests {
+	return &AppManifests{AppRevision: appRevision}
+}
+
+// AppManifests contains configuration to assemble resources recorded in the ApplicationRevision.
+// 'Assemble' means expand Application(Component and Trait) into K8s resource and get them ready to go, to be emitted
+// into K8s
+type AppManifests struct {
+	AppRevision     *v1beta1.ApplicationRevision
+	WorkloadOptions []WorkloadOption
+
+	appComponents []applicationComponent
+	components    []*v1alpha2.Component
+
+	appName        string
+	appNamespace   string
+	appLabels      map[string]string
+	appAnnotations map[string]string
+	appOwnerRef    *metav1.OwnerReference
+
+	assembledWorkloads map[string]*unstructured.Unstructured
+	assembledTraits    map[string][]*unstructured.Unstructured
+	// key is workload reference, values are the references of scopes the workload belongs to
+	referencedScopes map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference
+
+	finalized bool
+	err       error
+}
+
+// this is a helper struct to replace v1alpha2.ApplicationConfiguration for this moment
+type applicationComponent struct {
+	revisionName string
+	traits       []v1alpha2.ComponentTrait
+	scopes       []runtimev1alpha1.TypedReference
+}
+
+// WorkloadOption will be applied to each workloads AFTER it has been assembled by generic rules shown below:
+// 1) use component name as workload name
+// 2) use application namespace as workload namespace if unspecified
+// 3) set application as workload's owner
+// 4) pass all application's labels and annotations to workload's
+// Component and ComponentDefinition are enough for caller to manipulate workloads.
+// Caller can use below labels of workload to get more information:
+// - oam.LabelAppName
+// - oam.LabelAppRevision
+// - oam.LabelAppRevisionHash
+// - oam.LabelAppComponent
+// - oam.LabelAppComponentRevision
+type WorkloadOption interface {
+	ApplyToWorkload(workload *unstructured.Unstructured, comp *v1alpha2.Component, compDefinition *v1beta1.ComponentDefinition) error
+}
+
+// WithWorkloadOption add a WorkloadOption to plug in custom logic applied to each workload
+func (am *AppManifests) WithWorkloadOption(wo WorkloadOption) *AppManifests {
+	if am.WorkloadOptions == nil {
+		am.WorkloadOptions = make([]WorkloadOption, 0)
+	}
+	am.WorkloadOptions = append(am.WorkloadOptions, wo)
+	return am
+}
+
+// AssembledManifests do assemble and merge all assembled resources(except referenced scopes) into one array
+func (am *AppManifests) AssembledManifests() ([]*unstructured.Unstructured, error) {
+	if !am.finalized {
+		am.assemble()
+	}
+	if am.err != nil {
+		return nil, am.err
+	}
+	r := make([]*unstructured.Unstructured, 0)
+	for _, wl := range am.assembledWorkloads {
+		r = append(r, wl.DeepCopy())
+	}
+	for _, ts := range am.assembledTraits {
+		for _, t := range ts {
+			r = append(r, t.DeepCopy())
+		}
+	}
+	return r, nil
+}
+
+// ReferencedScopes do assemble and return workload reference and referenced scopes
+func (am *AppManifests) ReferencedScopes() (map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference, error) {
+	if !am.finalized {
+		am.assemble()
+	}
+	if am.err != nil {
+		return nil, am.err
+	}
+	r := make(map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference)
+	for k, refs := range am.referencedScopes {
+		r[k] = make([]runtimev1alpha1.TypedReference, len(refs))
+		copy(r[k], refs)
+	}
+	return r, nil
+}
+
+// GroupAssembledManifests do assemble and return all resources grouped by components
+func (am *AppManifests) GroupAssembledManifests() (
+	map[string]*unstructured.Unstructured,
+	map[string][]*unstructured.Unstructured,
+	map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference, error) {
+	if !am.finalized {
+		am.assemble()
+	}
+	if am.err != nil {
+		return nil, nil, nil, am.err
+	}
+	workloads := make(map[string]*unstructured.Unstructured)
+	for k, wl := range am.assembledWorkloads {
+		workloads[k] = wl.DeepCopy()
+	}
+	traits := make(map[string][]*unstructured.Unstructured)
+	for k, ts := range am.assembledTraits {
+		traits[k] = make([]*unstructured.Unstructured, len(ts))
+		for i, t := range ts {
+			traits[k][i] = t.DeepCopy()
+		}
+	}
+	scopes := make(map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference)
+	for k, v := range am.referencedScopes {
+		scopes[k] = make([]runtimev1alpha1.TypedReference, len(v))
+		copy(scopes[k], v)
+	}
+	return workloads, traits, scopes, nil
+}
+
+func (am *AppManifests) assemble() {
+	am.complete()
+	klog.InfoS("Assemble manifests for application", "name", am.appName, "revision", am.AppRevision.GetName())
+	if err := am.validate(); err != nil {
+		am.finalizeAssemble(err)
+		return
+	}
+	for _, ac := range am.appComponents {
+		compRevisionName := ac.revisionName
+		compName := ctrlutil.ExtractComponentName(compRevisionName)
+		commonLabels := am.generateCommonLabels(compName, compRevisionName)
+		var workloadRef runtimev1alpha1.TypedReference
+		klog.InfoS("Assemble manifests for component", "name", compName)
+		for _, comp := range am.components {
+			if comp.Name == compName {
+				wl, err := am.assembleWorkload(comp, commonLabels)
+				if err != nil {
+					am.finalizeAssemble(err)
+					return
+				}
+				am.assembledWorkloads[compName] = wl
+				workloadRef = runtimev1alpha1.TypedReference{
+					APIVersion: wl.GetAPIVersion(),
+					Kind:       wl.GetKind(),
+					Name:       wl.GetName(),
+				}
+				break
+			}
+		}
+
+		am.assembledTraits[compName] = make([]*unstructured.Unstructured, len(ac.traits))
+		for i, compTrait := range ac.traits {
+			trait, err := am.assembleTrait(compTrait, compName, commonLabels)
+			if err != nil {
+				am.finalizeAssemble(err)
+				return
+			}
+			if err := am.setWorkloadRefToTrait(workloadRef, trait); err != nil {
+				am.finalizeAssemble(errors.WithMessagef(err, "cannot set workload reference to trait %q", trait.GetName()))
+				return
+			}
+			am.assembledTraits[compName][i] = trait
+		}
+
+		am.referencedScopes[workloadRef] = make([]runtimev1alpha1.TypedReference, len(ac.scopes))
+		for i, scope := range ac.scopes {
+			am.referencedScopes[workloadRef][i] = scope
+		}
+	}
+	am.finalizeAssemble(nil)
+}
+
+func (am *AppManifests) complete() {
+	// safe to skip error-check
+	appConfig, _ := convertRawExtention2AppConfig(am.AppRevision.Spec.ApplicationConfiguration)
+	// convert v1alpha2.ApplicationConfiguration to a helper struct
+	am.appComponents = make([]applicationComponent, len(appConfig.Spec.Components))
+	for i, acc := range appConfig.Spec.Components {
+		am.appComponents[i] = applicationComponent{
+			revisionName: acc.RevisionName,
+		}
+		am.appComponents[i].traits = make([]v1alpha2.ComponentTrait, len(acc.Traits))
+		copy(am.appComponents[i].traits, acc.Traits)
+		am.appComponents[i].scopes = make([]runtimev1alpha1.TypedReference, len(acc.Scopes))
+		for j, s := range acc.Scopes {
+			am.appComponents[i].scopes[j] = s.ScopeReference
+		}
+	}
+	// Application entity in the ApplicationRevision has no metadata,
+	// so we have to get below information from AppConfig.
+	// Up-stream process must set these to AppConfig.
+	am.appName = appConfig.GetName()
+	am.appNamespace = appConfig.GetNamespace()
+	am.appLabels = appConfig.GetLabels()
+	am.appAnnotations = appConfig.GetAnnotations()
+	am.appOwnerRef = metav1.GetControllerOf(appConfig)
+
+	am.components = make([]*v1alpha2.Component, len(am.AppRevision.Spec.Components))
+	for i, rawComp := range am.AppRevision.Spec.Components {
+		// safe to skip error-check
+		comp, _ := convertRawExtention2Component(rawComp.Raw)
+		am.components[i] = comp
+	}
+
+	am.assembledWorkloads = make(map[string]*unstructured.Unstructured)
+	am.assembledTraits = make(map[string][]*unstructured.Unstructured)
+	am.referencedScopes = make(map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference)
+}
+
+func (am *AppManifests) finalizeAssemble(err error) {
+	am.finalized = true
+	if err != nil {
+		klog.ErrorS(err, "Failed assembling manifests for application", "name", am.appName, "revision", am.AppRevision.GetName())
+		am.err = errors.WithMessagef(err, "cannot assemble resources' manifests for application %q", am.appName)
+	}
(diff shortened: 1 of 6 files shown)


pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,281 @@
...
+	o.Comps = make([]*v1alpha2.Component, len(o.AppRevision.Spec.Components))
+	for i, rawComp := range o.AppRevision.Spec.Components {
+		// safe to skip error-check
+		comp, _ := convertRawExtention2Component(rawComp.Raw)
+		o.Comps[i] = comp
+	}
+	o.AppName = o.AppRevision.Spec.Application.Name
+	o.AppNamespace = o.AppRevision.Spec.Application.Namespace
+	o.AppLables = o.AppRevision.Spec.Application.Labels
+	o.AppAnnotations = o.AppRevision.Spec.Application.Annotations
```
captainroy-hy, line 159: This is a mistake, App in AppRev has no metadata.

wangyikewxgm:
Can we support a general func to generate workload name from AppRev? I think it's useful for rollout and deploy etc.

captainroy-hy:
> Can we support a general lightweight func to generate workload name from AppRev? I think it's useful for rollout and deploy etc.

Rollout, deploy or any other consumer of assemble can prepare their own naming logic and plug it in assemble through `WorkloadOption`. A sample is shown in options_test.go.

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,281 @@
...
+// workload and trait both have these annotations
+func (o *Options) setAnnotations(obj *unstructured.Unstructured) {
+	// pass application's all annotations
+	util.AddAnnotations(obj, o.AppAnnotations)
+	// remove useless annotations for workload/trait
+	util.RemoveAnnotations(obj, []string{
+		oam.AnnotationAppRollout,
+		oam.AnnotationRollingComponent,
+		oam.AnnotationInplaceUpgrade,
+	})
```
wangyikewxgm, line 255: Does the logic truly needed? These annotations are pass from AC set by appRollout. After refactor AppRollout won't set these label.

captainroy-hy: It works the same as how AppConfig works now. These annotations are just removed from workloads/traits, rollout controller will not consume them from workloads/traits directly, right?

wangyikewxgm: Yes,AppRollout control appContext templating workload by setting these annotation.After remove appContext rollout won't set them directly to workload/trait anymore.

wangyikewxgm:
Can we increase the test-coverage a bit to solve the codecov warning? And IMO this pr is generally ready to be merged. We can go head for other refactor work on this.

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,289 @@
...
+func (o *Options) WithWorkloadOption(wo WorkloadOption) *Options {
+	if o.WorkloadOptions == nil {
+		o.WorkloadOptions = make([]WorkloadOption, 0)
+	}
+	o.WorkloadOptions = append(o.WorkloadOptions, wo)
+	return o
+}
+
+// Assemble an application's manifests including workloads and traits according to a specific application revision
+func (o *Options) Assemble() (map[string]*unstructured.Unstructured, map[string][]*unstructured.Unstructured, map[string][]runtimev1alpha1.TypedReference, error) {
```
hongchaodeng, line 85: This is too tightly coupled to components and traits.
It should just take in appRev and return an array of Unstructureds.

captainroy-hy: Yes, that's truly a problem... My concern is, is it possible to have a scenario where consumer of assemble want to distinguish these resource and do something based on them? If no, I think it's ok to merge them into an array of Unstructureds.

captainroy-hy: Or provide two interfaces, one is to return all in one array, the other is to return categorized results.

hongchaodeng: Let's return an array for now. If there are requests from the community for other results, we can add them at that point.

captainroy-hy: Got it! Will do that.

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+// cluster.
+type Options struct {
+	AppRevision     *v1beta1.ApplicationRevision
+	WorkloadOptions []WorkloadOption
+
+	AppConfig      *v1alpha2.ApplicationConfiguration
+	Comps          []*v1alpha2.Component
+	AppName        string
+	AppNamespace   string
+	AppLables      map[string]string
```
wonderflow, line 49:
```suggestion
	AppLabels      map[string]string
```

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+
+// Options contains configuration to assemble resources recorded in the ApplicationRevision.
+// 'Assemble' means completing workload/trait resources and get them ready to emit into K8s
+// cluster.
+type Options struct {
+	AppRevision     *v1beta1.ApplicationRevision
+	WorkloadOptions []WorkloadOption
+
+	AppConfig      *v1alpha2.ApplicationConfiguration
+	Comps          []*v1alpha2.Component
```
wonderflow, line 46:
```suggestion
	V1alpha2AppConfig      *v1alpha2.ApplicationConfiguration
	V1alpha2Comps          []*v1alpha2.Component
```

Does it necessary？ I think it would be deprecated？ Let's make it clear if Necessary

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+	AppNamespace   string
+	AppLables      map[string]string
+	AppAnnotations map[string]string
+	AppOwnerRef    *metav1.OwnerReference
+
+	assembledWorkloads map[string]*unstructured.Unstructured
+	assembledTraits    map[string][]*unstructured.Unstructured
+	referencedScopes   map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference
+
+	assembled bool
```
wonderflow, line 57: lets use assemble state for this value, and this should be a enum variable

captainroy-hy: good idea.

wonderflow: another design logic is `frozen bool`, assembled means frozen.

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+
+// NewAssembleOptions create a Options
+func NewAssembleOptions(appRevision *v1beta1.ApplicationRevision) *Options {
+	return &Options{AppRevision: appRevision}
+}
+
+// Options contains configuration to assemble resources recorded in the ApplicationRevision.
+// 'Assemble' means completing workload/trait resources and get them ready to emit into K8s
+// cluster.
+type Options struct {
```
wonderflow, line 41: Not good name. What about `Resources` `AppResources`?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	o.WorkloadOptions = append(o.WorkloadOptions, wo)
+	return o
+}
+
+// AssembledResources do assemble and return assembled resources in one bundle except referenced scopes
+func (o *Options) AssembledResources() ([]*unstructured.Unstructured, error) {
+	if !o.assembled {
+		if err := o.assemble(); err != nil {
+			return nil, errors.WithMessagef(err, "cannot assemble manifests of application %q", o.AppName)
```
wonderflow, line 89:
```suggestion
			return nil, errors.WithMessagef(err, "cannot assemble resources' manifests for application %q", o.AppName)
```

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+// - oam.LabelAppName
+// - oam.LabelAppRevision
+// - oam.LabelAppRevisionHash
+// - oam.LabelAppComponent
+// - oam.LabelAppComponentRevision
+type WorkloadOption interface {
+	ApplyToWorkload(workload *unstructured.Unstructured, comp *v1alpha2.Component, compDefinition *v1beta1.ComponentDefinition) error
+}
+
+// WithWorkloadOption add a WorkloadOption
```
wonderflow, line 76: for what?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+	"github.com/oam-dev/kubevela/pkg/oam/util"
+)
+
+// NewAssembleOptions create a Options
+func NewAssembleOptions(appRevision *v1beta1.ApplicationRevision) *Options {
+	return &Options{AppRevision: appRevision}
+}
+
+// Options contains configuration to assemble resources recorded in the ApplicationRevision.
+// 'Assemble' means completing workload/trait resources and get them ready to emit into K8s
```
wonderflow, line 39:
```suggestion
// 'Assemble' means expand Application(Component and Trait) into K8s resource and get them ready to go, to be emitted into K8s
```

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+	referencedScopes   map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference
+
+	assembled bool
+}
+
+// WorkloadOption will be applied to each workloads AFTER it has been assembled by generic rules shown below:
+// 1) use component name as workload name
+// 2) use application namespace as workload namespace if unspecified
+// 3) set application as workload's owner
+// 4) pass all application's lables and annotations to workload's
```
wonderflow, line 64:
```suggestion
// 4) pass all application's labels and annotations to workload's
```

Please clean all the mis-spell

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
captainroy-hy: provide a way for callers to plug in custom logic to process workloads, such as ones shown in options.go file.

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+		}
+	}
+	o.assembled = true
+	return nil
+}
+
+func (o *Options) complete() {
+	// safe to skip error-check
+	o.AppConfig, _ = convertRawExtention2AppConfig(o.AppRevision.Spec.ApplicationConfiguration)
+	o.Comps = make([]*v1alpha2.Component, len(o.AppRevision.Spec.Components))
```
wonderflow, line 204: why do we still need to create these two objects in the new implementation？

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
captainroy-hy: `Resources` seems too generic. How about `Manifests` or `AppManifests`?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+			o.setOwnerReference(trait)
+			if err := o.setWorkloadRefToTrait(workloadRef, trait); err != nil {
+				return errors.WithMessagef(err, "cannot set workload reference to trait %q", trait.GetName())
+			}
+			o.assembledTraits[compName][i] = trait
+		}
+
+		o.referencedScopes[workloadRef] = make([]runtimev1alpha1.TypedReference, len(acc.Scopes))
+		for i, scope := range acc.Scopes {
+			o.referencedScopes[workloadRef][i] = scope.ScopeReference
```
wonderflow, line 194: we should write a case to cover it

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+	obj.SetOwnerReferences([]metav1.OwnerReference{*o.AppOwnerRef})
+}
+
+func (o *Options) setWorkloadRefToTrait(wlRef runtimev1alpha1.TypedReference, trait *unstructured.Unstructured) error {
+	traitType := trait.GetLabels()[oam.TraitTypeLabel]
+	traitDef := o.AppRevision.Spec.TraitDefinitions[traitType]
+	workloadRefPath := traitDef.Spec.WorkloadRefPath
+	// only add workload reference to the trait if it asks for it
+	if len(workloadRefPath) != 0 {
+		if err := fieldpath.Pave(trait.UnstructuredContent()).SetValue(workloadRefPath, wlRef); err != nil {
```
wonderflow, line 328: we should write test case for it, this is a feature which used to rely on AppConfig

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
wonderflow: AppManifests works for me

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
wonderflow: clarify it in code

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+	if err := o.validate(); err != nil {
+		return errors.WithMessagef(err, "cannot assemble manifests of application %q", o.AppName)
+	}
+	for _, acc := range o.AppConfig.Spec.Components {
+		compRevisionName := acc.RevisionName
+		compName := ctrlutil.ExtractComponentName(compRevisionName)
+		commonLables := o.generateCommonLables(compName, compRevisionName)
+		var workloadRef runtimev1alpha1.TypedReference
+		for _, comp := range o.Comps {
+			if comp.Name == compName {
```
wonderflow, line 146: please write a seprate func for this logic, e.g. `assembleWorkload`

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+					APIVersion: wl.GetAPIVersion(),
+					Kind:       wl.GetKind(),
+					Name:       wl.GetName(),
+				}
+				break
+			}
+		}
+
+		o.assembledTraits[compName] = make([]*unstructured.Unstructured, len(acc.Traits))
+		for i, compTrait := range acc.Traits {
```
wonderflow, line 175: please write a seprate func for this logic, e.g. assembleTrait

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,333 @@
...
+		oam.LabelAppRevision:          o.AppRevision.Name,
+		oam.LabelAppRevisionHash:      o.AppRevision.Labels[oam.LabelAppRevisionHash],
+		oam.LabelAppComponent:         compName,
+		oam.LabelAppComponentRevision: compRevisionName,
+	}
+	// pass application's all labels to workload/trait
+	return util.MergeMapOverrideWithDst(lables, o.AppLables)
+}
+
+func (o *Options) setWorkloadLables(wl *unstructured.Unstructured, commonLables map[string]string) {
```
wonderflow, line 265: labels

wonderflow (changes requested):
Greate work! But we MUST cover the main logic as it's really important feature which will be used not only in Application but also Rollout and others.

captainroy-hy:
> MUST cover the main logic as it's really important feature which will be used not only in Application but also Rollout and others.

@wonderflow  Got it. I will fix all comments and cover every single logic in tests.

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
captainroy-hy: I made a helper struct to replace AppConfig just in assemble pkg.  As for v1alpha2.Component and ComponentTrait, it costs a lot to replace them now. Maybe we can refactor them later.

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,400 @@
...
+	appName        string
+	appNamespace   string
+	appLabels      map[string]string
+	appAnnotations map[string]string
+	appOwnerRef    *metav1.OwnerReference
+
+	assembledWorkloads map[string]*unstructured.Unstructured
+	assembledTraits    map[string][]*unstructured.Unstructured
+	// key is workload reference, values are the references of scopes the workload belongs to
+	referencedScopes map[runtimev1alpha1.TypedReference][]runtimev1alpha1.TypedReference
```
wonderflow, line 58: why not use component name as the key?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,400 @@
...
+
+// AppManifests contains configuration to assemble resources recorded in the ApplicationRevision.
+// 'Assemble' means expand Application(Component and Trait) into K8s resource and get them ready to go, to be emitted
+// into K8s
+type AppManifests struct {
+	AppRevision     *v1beta1.ApplicationRevision
+	WorkloadOptions []WorkloadOption
+
+	appComponents []applicationComponent
+	components    []*v1alpha2.Component
```
wonderflow, line 47: what's these two difference?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,400 @@
...
+// 4) pass all application's labels and annotations to workload's
+// Component and ComponentDefinition are enough for caller to manipulate workloads.
+// Caller can use below labels of workload to get more information:
+// - oam.LabelAppName
+// - oam.LabelAppRevision
+// - oam.LabelAppRevisionHash
+// - oam.LabelAppComponent
+// - oam.LabelAppComponentRevision
+type WorkloadOption interface {
+	ApplyToWorkload(workload *unstructured.Unstructured, comp *v1alpha2.Component, compDefinition *v1beta1.ComponentDefinition) error
```
wonderflow, line 84: why `v1alpha2.Component` is still exist concept?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,400 @@
...
+
+	/* NOTE a workload has these possible labels
+	   app.oam.dev/app-revision-hash: ce053923e2fb403f
+	   app.oam.dev/appRevision: myapp-v2
+	   app.oam.dev/component: mycomp
+	   app.oam.dev/name: myapp
+	   app.oam.dev/resourceType: WORKLOAD
+	   app.oam.dev/revision: mycomp-v2
+	   workload.oam.dev/type: kube-worker
+	*/
```
wonderflow, line 349: let's make it as test case and remove it from here?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -0,0 +1,400 @@
...
+	/* NOTE a trait has these possible labels
+	   app.oam.dev/app-revision-hash: ce053923e2fb403f
+	   app.oam.dev/appRevision: myapp-v2
+	   app.oam.dev/component: mycomp
+	   app.oam.dev/name: myapp
+	   app.oam.dev/resourceType: TRAIT
+	   app.oam.dev/revision: mycomp-v2
+	   trait.oam.dev/resource: service
+	   trait.oam.dev/type: ingress // already added in render phase
+	*/
```
wonderflow, line 386: let's make it as test case and remove it from here?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble_suite_test.go
```diff
@@ -0,0 +1,159 @@
...
+		      traits:
+		        - type: ingress
+		          properties:
+		            domain: localhost
+		            http:
+		              "/": 8000
+		        - type: manualscaler
+		          properties:
+		            replicas: 3
+		*/
```
wonderflow, line 70: debug info?

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
captainroy-hy: for helm field in component, there's a WorkloadOption still need it.

wonderflow (commented):
almost good, ping @wangyikewxgm @hongchaodeng  to review

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble_suite_test.go
captainroy-hy: no, it helps to review the code because there's lot of content in test data yaml.

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
wonderflow: We should find a replace, we need to remove dependency on AppConfig/AppContext/Component

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
captainroy-hy: sure, will do that. But I want to keep the labels key here to remind developers which labels are used because these labels come from different entities(app & xxDefiniton)

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
captainroy-hy: if use comp name as the key, users have to get workloads from assembledWorkloads by the key, then construct typedReference by themselves.

wonderflow: Ok, it makes sense

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
captainroy-hy: appCompoents = appConfigComponent ([] {a comp + its traits + its scopes});
comp = raw workload + helm

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
wonderflow: We can keep it for now

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
wonderflow: understood, I think we should use something new like `models` to record this thing, and the `models` is just a memory struct that won't exist in K8s

captainroy-hy: Totally agreed!  These v1alpha2 APIs can retire now.

wonderflow (approved):
LGTM, let's move the left comments to next PR

PR #1716, opened 2021-05-26
optimize goimports

chivalryq:
Why I do this: when coding in `./reference`, `golangci-lint` in `make reviewable` just notify me some file is not `goimports-ed`, but did't fix it.

What I do: make goimports correct all golang file.It doesn't take longer time than before but make it **really** reviewable. Also, I fix all golang file's  import fault incidentally.


wonderflow (approved):
It's really cool!

PR #1731, opened 2021-05-29
fix bug: When the Component contains multiple traits of the same type, the status of the trait in the Application is reported incorrectly

yangsoon:
fix https://github.com/oam-dev/kubevela/issues/1722.

When the Component contains multiple traits of the same type, the status of the trait in the Application 
is reported incorrectly.

this PR will fix the bug

- [x] fix bug
- [x] add test


yangsoon:
this PR disables the `validateAuxiliaryNameUnique` function, which checks the uniqueness of the Auxiliary name because this function conflicts with the scene we repaired. Do we need to ensure that a Component must use different types of traits？

pkg/appfile/validate.go
```diff
@@ -56,7 +56,10 @@ func newValidationProcessContext(wl *Workload, appName, revisionName, ns string)
 	}
 	auxiliaryHooks := []process.AuxiliaryHook{
 		// add more hook funcs here to validate CUE auxiliaries
-		validateAuxiliaryNameUnique(),
+
+		// Disable verify the uniqueness of the fields in the outputs
+		// because users may want to use multiple same type traits on one Component
+		// validateAuxiliaryNameUnique(),
```
wonderflow, line 62: we can't disable it

wonderflow: This is not conflict, you can write a better test case for CUE trait

test/e2e-test/testdata/app/trait_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+      if context.outputs.config.kind == "ConfigMap" {
+        message: "configMap:" + context.outputs.config.metadata.name
+      }
+  appliesToWorkloads:
+    - deployments.apps
+  podDisruptive: true
+  schematic:
+    cue:
+      template: |-
+        outputs: config: {
```
wonderflow, line 21:
```suggestion
        outputs: parameter.configname: {
```

test/e2e-test/testdata/app/trait_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+        		metadata: name: parameter.name
+        		data: {
+        			for k, v in parameter.data {
+        				"\(k)": v
+        			}
+        		}
+        	}
+        }
+        parameter: {
+        	kind: string
```
wonderflow, line 45:
```suggestion
        	kind: string
        	configname: string
```

Try it

pkg/appfile/validate.go
yangsoon: got it

test/e2e-test/testdata/app/trait_config.yaml
yangsoon: got it

test/e2e-test/testdata/app/trait_config.yaml
yangsoon: got it

test/e2e-test/testdata/app/trait_config.yaml
yangsoon: Users maybe need use variable of `parameter` to get the resource in `context.outputs`, so we add `parameter` context for  `spec.status.customStatus`, users can get variable in `parameter` by `parameter.configname `

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -285,33 +285,32 @@ func (h *appHandler) statusAggregate(appFile *appfile.Appfile) ([]common.Applica
...
 			if err := tr.EvalContext(pCtx); err != nil {
 				return nil, false, errors.WithMessagef(err, "app=%s, comp=%s, trait=%s, evaluate context error", appFile.Name, wl.Name, tr.Name)
 			}
-		}
 
-		var traitStatusList []common.ApplicationTraitStatus
-		for _, trait := range wl.Traits {
 			var traitStatus = common.ApplicationTraitStatus{
-				Type:    trait.Name,
+				Type:    tr.Name,
```
zzxwill, line 295: Good catch.

wonderflow (approved):

PR #1732, opened 2021-05-30
refactor appfile pkg for adding workflow

hongchaodeng:
separated from https://github.com/oam-dev/kubevela/pull/1708


wonderflow (changes requested):
let's write test cases to cover the happy-path

hongchaodeng:
> let's write test cases to cover the happy-path

This PR is about refactoring. No new logic introduced to the main code path.

I will add the test in controller to cover all path as last part of https://github.com/oam-dev/kubevela/pull/1708

wonderflow (approved):
That's fine

PR #1733, opened 2021-05-31
organize registry directory

chivalryq:
1. change registry generate logic. it's now like vela-core charts: final yaml definitions are generated from cue&yaml.
2. move registry to kubevela/
3. `internal` and `registry` directory in vela-template are respectively for inners comp and traits or those will be in registry.
4. reuse mergedef.sh in gen_definitions.sh


zzxwill:
Trait hpa will be removed to the core chart `docs/examples/registry/hpa.cue → hack/vela-templates/registry/cue/hpa.cue`, so it should be deleted from the registry. Right?

chivalryq:
> Trait hpa will be removed to the core chart `docs/examples/registry/hpa.cue → hack/vela-templates/registry/cue/hpa.cue`, so it should be deleted from the registry. Right?

Actually it was move to `hack/vela-templates/registry/cue/hpa.cue` and oss only hold yaml, `hpa.cue` is just an Intermediate product

wonderflow (approved):

wonderflow (approved):
LGTM

PR #1735, opened 2021-05-31
Improve the logging system

yangsoon:
refer to https://github.com/oam-dev/kubevela/issues/378

using klog/v2 as log system which can print out the file and line numbers

- [x] add logfilepath option for helm chart
- [x] align with [Logging Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md), [Migration Instructions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#structured-logging-in-kubernetes)
- [x] set logdebug level

### Following record the step to migration log system:
1. Remove string formatting from log message. refer to [Migration Instructions-**Remove string formatting from log message**](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#remove-string-formatting-from-log-message)
    1. improve messages to comply with good practices:
        * Start from a capital letter.
        * Do not end the message with a period.
        * Use active voice. Use complete sentences when there is an acting subject ("A could not do B") or omit the subject if
        the subject would be the program itself ("Could not do B").
        * Use past tense ("Could not delete B" instead of "Cannot delete B")
        * When referring to an object, state what type of object it is. ("Deleted pod" instead of "Deleted")
     2. Name arguments in `klog.InfoS`
        * Always use lowerCamelCase

2. Using ErrorS refer to [Migration Instructions-**Using ErrorS**](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#using-errors)

    ```golang
    // ErrorS structured logs to the ERROR, WARNING, and INFO logs.
    // the err argument used as "err" field of log line.
    // The msg argument used to add constant description to the log line.
    // The key/value pairs would be join by "=" ; a newline is always appended.
    //
    // Examples:
    // >> klog.ErrorS(err, "Failed to update pod status")
    // output:
    // >> E1025 00:15:15.525108       1 controller_utils.go:114] "Failed to update pod status" err="timeout"
    func ErrorS(err error, msg string, keysAndValues ...interface{})
    ```
    1. For expected errors (errors that can happen during routine operations) use `klog.InfoS` and pass error in `err` key instead. 

3. Use klog.KObj and klog.KRef for Kubernetes objects. refer to [Migration Instructions-**Use klog.KObj and klog.KRef**](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#use-klogkobj-and-klogkref-for-kubernetes-objects)

### DIFF
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -81,7 +81,12 @@ admissionWebhooks:
 #Enable debug logs for development purpose
 logDebug: false
 
+#If non-empty, write log files in this path
 logFilePath: ""
+
+#Defines the maximum size a log file can grow to. Unit is megabytes.
+#If the value is 0, the maximum file size is unlimited.
+logFileMaxSize: 1024
 
 systemDefinitionNamespace: vela-system
 
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -81,6 +81,8 @@ admissionWebhooks:
 #Enable debug logs for development purpose
 logDebug: false
 
+logFilePath: ""
+
 systemDefinitionNamespace: vela-system
 
 applicationRevisionLimit: 10
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -61,6 +61,7 @@ var (
 func main() {
 	var metricsAddr, logFilePath, leaderElectionNamespace string
 	var enableLeaderElection, logDebug bool
+	var logFileMaxSize uint64
 	var certDir string
 	var webhookPort int
 	var useWebhook bool
@@ -80,6 +81,7 @@ func main() {
 	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "",
 		"Determines the namespace in which the leader election configmap will be created.")
 	flag.StringVar(&logFilePath, "log-file-path", "", "The file to write logs to.")
+	flag.Uint64Var(&logFileMaxSize, "log-file-max-size", 1024, "Defines the maximum size a log file can grow to, Unit is megabytes.")
 	flag.BoolVar(&logDebug, "log-debug", false, "Enable debug logs for development purpose")
 	flag.IntVar(&controllerArgs.RevisionLimit, "revision-limit", 50,
 		"RevisionLimit is the maximum number of revisions that will be maintained. The default value is 50.")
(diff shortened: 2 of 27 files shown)


wonderflow:
We can raise a contributor statute for things like the log format

cmd/core/main.go
```diff
@@ -142,93 +126,93 @@ func main() {
 		SyncPeriod:              &syncPeriod,
 	})
 	if err != nil {
-		setupLog.Error(err, "unable to create a controller manager")
+		klog.ErrorS(err, "Unable to create a controller manager")
```
zzxwill, line 129: For klog, the first letter of an error message should be capitalized?

yangsoon: Capitalize the first letter of the error message was refer to the demo in [Migration Instructions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#structured-logging-in-kubernetes), I think the k8s community wants to align with this rule in  [Migration Instructions-**Remove string formatting from log message**](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#remove-string-formatting-from-log-message)

charts/vela-core/values.yaml
```diff
@@ -81,6 +81,8 @@ admissionWebhooks:
 #Enable debug logs for development purpose
 logDebug: false
 
+logFilePath: ""
```
wonderflow, line 84: add a explanation for it

yangsoon: got it

charts/vela-core/values.yaml
```diff
@@ -81,6 +81,13 @@ admissionWebhooks:
 #Enable debug logs for development purpose
 logDebug: false
 
+#If non-empty, write log files in this path
+logFilePath: ""
+
+#Defines the maximum size a log file can grow to. Unit is megabytes.
+#If the value is 0, the maximum file size is unlimited.
+logFileMaxSize: 0
```
wonderflow, line 89: 1024

wonderflow (approved):

PR #1741, opened 2021-06-01
dispatch resources into cluster with one package

captainroy-hy:
related to #1705 

implement [dispatch pkg](https://github.com/oam-dev/kubevela/issues/1705#issuecomment-851124206)

- [x] basic functions
- [x] unit test

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
@@ -1,0 +1,265 @@
+/*
+Copyright 2021 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dispatch
+
+import (
+	"context"
+	"reflect"
+
+	"github.com/pkg/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
+	"github.com/oam-dev/kubevela/pkg/utils/apply"
+)
+
+// NewAppManifestsDispatcher creates an AppManifestsDispatcher.
+func NewAppManifestsDispatcher(c client.Client, appRev *v1beta1.ApplicationRevision) *AppManifestsDispatcher {
+	return &AppManifestsDispatcher{
+		c:          c,
+		applicator: apply.NewAPIApplicator(c),
+		appRev:     appRev,
+		gcHandler:  NewGCHandler(c, appRev.Namespace),
+	}
+}
+
+// AppManifestsDispatcher dispatch application manifests into K8s and record the dispatched manifests' references in a
+// resource tracker which is named by a particular rule: name = appRevision's Name + appRevision's namespace.
+// A bundle of manifests to be dispatched MUST come from the given application revision.
+type AppManifestsDispatcher struct {
+	c          client.Client
+	applicator apply.Applicator
+	gcHandler  GarbageCollector
+
+	appRev *v1beta1.ApplicationRevision
+	oldRT  *v1beta1.ResourceTracker
+	skipGC bool
+
+	appRevName string
+	namespace  string
+	newRTName  string
+	newRT      *v1beta1.ResourceTracker
+}
+
+// EnableGC return an AppManifestsDispatcher that always do GC after dispatching resources.
+// GC will calculate diff between the dispatched resouces and ones recorded in the given resource tracker.
+func (a *AppManifestsDispatcher) EnableGC(rt *v1beta1.ResourceTracker) *AppManifestsDispatcher {
+	if rt != nil {
+		a.oldRT = rt.DeepCopy()
+	}
+	return a
+}
+
+// EnableUpgradeAndSkipGC return an AppManifestsDispatcher that skips GC after dispatching resources.
+// For the unchanged resources, dispatcher will update their owner to the newly created resource tracker.
+// It's helpful in a rollout scenario where new revision is going to create a new workload while the old one should not
+// be deleted before rollout is terminated.
+func (a *AppManifestsDispatcher) EnableUpgradeAndSkipGC(rt *v1beta1.ResourceTracker) *AppManifestsDispatcher {
+	if rt != nil {
+		a.oldRT = rt.DeepCopy()
+		a.skipGC = true
+	}
+	return a
+}
+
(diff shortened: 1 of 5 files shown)


pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
```diff
@@ -0,0 +1,166 @@
...
+		// already exists
+		// no need to update, because the manifests from the same application revision never change
+		a.newRT = rt
+		return nil
+	}
+	if kerrors.IsNotFound(err) {
+		rt.SetName(a.newRTName)
+		// ApplicationRevision must be controlled by an Application
+		ownerRefOfApp := metav1.GetControllerOf(a.appRev)
+		rt.SetOwnerReferences([]metav1.OwnerReference{*ownerRefOfApp})
```
wangyikewxgm, line 106: Why application should own rt? As far as I can see, a namespaced scope object (app) cannot own a cluster scope object(resourceTracker) in k8s 1.20

captainroy-hy: This is not used for k8s built-in GC. I'm considering let RT controller take responsibility for GC RT while an application or AppRevision is deleted.

wangyikewxgm: I think it's a little bit hack. And I'm not sure k8s whether  truly allow this. how about set some labels in rt to do this.

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
```diff
@@ -0,0 +1,166 @@
...
+// If the latest resource tracker is offered to dispatcher, it will do GC after applying.
+func (a *AppManifestsDispatcher) Dispatch(ctx context.Context, manifests []*unstructured.Unstructured) (*v1beta1.ResourceTracker, error) {
+	a.complete()
+	if err := a.createResourceTracker(ctx, manifests); err != nil {
+		return nil, err
+	}
+	if err := a.applyManifests(ctx, manifests); err != nil {
+		return nil, err
+	}
+	if !a.skipGC && a.oldRT != nil && a.oldRT.Name != a.newRTName {
```
wangyikewxgm, line 79: Can a.oldRT be nil? I found  Line 47 in `NewAppManifestsDispatcher` has init the value with `DeepCopy`

captainroy-hy: Yes, it could be nil. Code is updated and I use another way to pass in old rt.

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
captainroy-hy: Agreed. If we can make sure the name of resource tracker ALWAYS follows the rule, `name = appRevName+namespace`, we can use its name to locate an AppRevision directly.

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
```diff
@@ -0,0 +1,262 @@
...
+
+	appRevName string
+	namespace  string
+	newRTName  string
+	newRT      *v1beta1.ResourceTracker
+}
+
+// EnableGC return an AppManifestsDispatcher that always do GC after dispatching resources.
+// GC will calculate diff between the dispatched resouces and ones recorded in the given resource tracker.
+func (a *AppManifestsDispatcher) EnableGC(rt *v1beta1.ResourceTracker) *AppManifestsDispatcher {
```
wonderflow, line 66: where is it used?

captainroy-hy: - InplaceUpgrade in app controller.  
- Before ending of rollout, v1=>2, it may need to GC v1.

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
```diff
@@ -0,0 +1,262 @@
...
+
+// Dispatch apply manifests into k8s and return a resource tracker recording applied manifests' references.
+// If GC is enabled, it will do GC after applying.
+// If 'UpgradeAndSkipGC' is enabled, it will:
+// - create new resources if not exist before
+// - update unchanged resources' owner from the old resource tracker to the new one
+// - skip deleting(GC) any resources
+func (a *AppManifestsDispatcher) Dispatch(ctx context.Context, manifests []*unstructured.Unstructured) (*v1beta1.ResourceTracker, error) {
+	if err := a.completeAndValidate(ctx); err != nil {
+		return nil, err
```
wonderflow, line 93: I don't see any log or error wraper on it, and the following lines, how do we know where the error from if we need to debug errors

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/name.go
```diff
@@ -0,0 +1,35 @@
...
+// it will create a resource tracker to record the resources emitted to K8s.
+// The resource tracker must be named by AppRevisionName and namespace joint by "-".
+
+func constructResourceTrackerName(appRevName, ns string) string {
+	return fmt.Sprintf("%s-%s", appRevName, ns)
+}
+
+func extractAppNameFromResourceTrackerName(name, ns string) string {
+	splits := strings.Split(strings.TrimSuffix(name, "-"+ns), "-")
+	return strings.Join(splits[0:len(splits)-1], "-")
```
wonderflow, line 34: please write test case

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
captainroy-hy: I have added logs or wrapped errors in those unexported methods. I think they can provide enough info for debugging.

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/gc.go
```diff
@@ -0,0 +1,95 @@
...
+			toBeDeleted.SetNamespace(oldRsc.Namespace)
+			toBeDeleted.SetName(oldRsc.Name)
+			if err := h.c.Delete(ctx, toBeDeleted); err != nil {
+				klog.ErrorS(err, "failed delete a resource", "name", oldRsc.Name, "apiVersion", oldRsc.APIVersion, "kind", oldRsc.Kind)
+				return errors.Wrapf(err, "cannot delete resource %q", oldRsc)
+			}
+			klog.InfoS("Successfully GC a resource", "name", oldRsc.Name, "apiVersion", oldRsc.APIVersion, "kind", oldRsc.Kind)
+		}
+	}
+	// TODO is it necessary to delete the old resource tracker now?
```
wonderflow, line 80: Of course, why not?

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
wonderflow: good

wonderflow (approved):
Generally LGTM, please check @wangyikewxgm

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/gc.go
captainroy-hy: In our previous talk, you mentioned that, whether it's possible to retrieve the history of app revisions' applied manifests. I think this may be an opportunity to do that. Leave the GCed resource tracker out of there and let its controller do something based on these historical resource trackers.

wonderflow: That's not the same case. When resources are deleted, the resourceTracker left is meanless.

captainroy-hy: Got it. Will modify to delete old resource tracker.

wangyikewxgm:
LGTM, great work! we can  refactor app and rollout main reconcile logic base on this.

PR #1750, opened 2021-06-02
add workflow package

hongchaodeng:


zzxwill (approved):
/lgtm

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,175 @@
...
+			UID:        obj.GetUID(),
+		},
+	}
+
+	cond, found, err := utils.GetUnstructuredObjectStatusCondition(obj, CondTypeWorkflowFinish)
+	if err != nil {
+		return nil, err
+	}
+
+	if !found || cond.Status != "True" {
```
zzxwill, line 145: Will you make `True` as a constant? And it could be a bool?

zzxwill: I've check the status of a Pod, I doubt why there are two types of `true`?

```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2021-05-26T13:20:10Z"
    reason: PodCompleted
    status: "True"
    type: Initialized
...
  initContainerStatuses:
  - containerID: docker://22e1a23c40335e7c811f869214122f1d1fa1d7bac1bd913ef0050e43bd0c3fa6
    image: busybox:latest
    imageID: docker-pullable://busybox@sha256:b5fc1d7b2e4ea86a06b0cf88de915a2c43a99a00b6b3c0af731e5f4c07ae8eff
    lastState: {}
    name: prepare-input-terraform-configurations
    ready: true
```

hongchaodeng: Pod is not a very good example... The status conditions are standardized long after it.

hongchaodeng: > Will you make True as a constant?

fixed

PR #1756, opened 2021-06-03
Add ignore feature for cue parameter

### LINKED ISSUE #1754: we should add some `keyword` such as `//+ignore` to let the user can't see the parameter and only be used by system.
we should add some `keyword` such as `//+ignore` to let the user can't see the parameter and only be used by system.

_Originally posted by @wonderflow in https://github.com/oam-dev/kubevela/pull/1748#r644563466_

wangyuan249:
When there are some cue parameters described in system level which we don't want to exposed to end user, we can use  //+ignore  to cue parameter, and through vela show we can see the corresponding field disappear.

cue template:
image

vela show result:
image
fix #1754


apis/types/capability.go
```diff
@@ -134,6 +134,7 @@ type Parameter struct {
 	Required bool        `json:"required,omitempty"`
 	Default  interface{} `json:"default,omitempty"`
 	Usage    string      `json:"usage,omitempty"`
+	Ignore   string      `json:"ignore,omitempty"`
```
zzxwill, line 137: Will changing the type to bool make it more clear? All annotation are in `key=value` format. And the default value is `false`.
`// +ignore If addRevisionLabel is true, the appRevision label will be added to the underlying pods` will be 
```
// +ignore=true
 // +usage=If addRevisionLabel is true, the appRevision label will be added to the underlying pods

@wonderflow Do you think it's a good idea?
```

charts/vela-core/templates/defwithtemplate/webservice.yaml
```diff
@@ -136,7 +136,7 @@ spec:
         	// +short=p
         	port: *80 | int
         
-        	// If addRevisionLabel is true, the appRevision label will be added to the underlying pods
+        	// +ignore If addRevisionLabel is true, the appRevision label will be added to the underlying pods
```
zzxwill, line 139: This seems a little bit weird as we normally put description in annotation `//+usage=xxx`.

apis/types/capability.go
wonderflow: mark `+ignore` here means we should ignore it in `vela show`, to avoid confusion, I think the example should in the following format

```
// +ignore
// If addRevisionLabel is true, the appRevision label will be added to the underlying pods
```

wonderflow: While I also agree @zzxwill  if we want to make the syntax to be simple and consistent.

```
// +ignore=true
// +usage=If addRevisionLabel is true, the appRevision label will be added to the underlying pods
```

Also works for me, but I think the default value is true. Nobody will mark `+ignore=false`.

zzxwill: +1

zzxwill: Oh, the default should be `false`. If the description part is spilt into `+usage`, just leaving `+ignore` as string type is okay with me.

references/plugins/references.go
```diff
@@ -558,7 +558,10 @@ func (ref *ParseReference) parseParameters(paraValue cue.Value, paramKey string,
 			if def, ok := val.Default(); ok && def.IsConcrete() {
 				param.Default = velacue.GetDefault(def)
 			}
-			param.Short, param.Usage, param.Alias = velacue.RetrieveComments(val)
+			param.Short, param.Usage, param.Alias, param.Ignore = velacue.RetrieveComments(val)
+			if param.Ignore != "" {
```
wonderflow, line 562: you should ignore in display logic and let the parse finished

pkg/cue/testdata/workloads/webservice.cue
```diff
@@ -0,0 +1,163 @@
...
+			items?: [...{
+				key:  string
+				path: string
+				mode: *511 | int
+			}]
+		}
+		if type == "emptyDir" {
+			medium: *"" | "Memory"
+		}
+	}]
```
wonderflow, line 161: these parts were not used, right? We should not add too much redundant code in unit test

wonderflow (commented):
please refine the test case, remove useless part

wonderflow (approved):

PR #1759, opened 2021-06-04
don't use appRevision as resource tracker's owner

captainroy-hy:
as title


wonderflow:
why?

wangyikewxgm (approved):
LGTM

captainroy-hy:
> why?

@wonderflow    As discuss with @wangyikewxgm , appRevision is namespace-scope while resource tracker is cluster-scope.  Such an owner relationship doesn't make sense in k8s 1.20+ . And we will use GC and finalizer to delete resource tracker.

wonderflow (approved):

PR #1763, opened 2021-06-05
remove appContext from app controller

captainroy-hy:
related to #1705 

major changes:

- get rid of appCtx in app controller
  - replace appCtx with assemble&dispatch in apply.go
  - remove resource tracker related logic 
  - always handle finalizer for resource tracker
  - modify appRevision clean up mechanism (use resource tracker to get appRevision in use)
  - fix unit tests in application pkg
- fix e2e test
  - remove appcontext_test.go file
  - remove appContext logic in al test cases
  - disable rollout related test temporarily (wait for rfc of rollout)
  - disable resource tracker test (re-write or remove latter)


wangyikewxgm:
Please going on.I'm trying my best to fix all rollout related e2e-test. Just disable rollout related e2e-test in this pr is fine. I think I can submit a separate pr for review easier. @captainroy-hy

PR #1764, opened 2021-06-05
refine  assemble and dispatch

captainroy-hy:
separated from #1763 

### major changes:
assemble pkg:
- make sure the order of resources in the assembled result is same as defined in application
- fix helmDiscovery workload option bug (assembled workload only need a name but not all spec of helm-created workload)

dispatch pkg:
- add labels(appName + namespace) to resource tracker in order to list resource trackers of a specific app
- make dispatch/applyAndRecord to incrementally record applied resources
  - scenario:  an app revision outcome 1 workload + 2 traits,  caller can dispatch the workload without traits first, until a specific condition is reached, caller then call dispatch again to dispatch traits. The resource tracker should track workload and traits.
- add exported functions about naming resource tracker in dispatch pkg
- refine and add more logs


captainroy-hy:
Close and will merge to branch rfc-remove-appcontext.

PR #1768, opened 2021-06-07
Fix invalid `log-debug` flag

kinsolee:
Now debug-level logs don't appear when vela-core is started with `log-debug=true`.

`flag.Parse` must be called before using variable `logDebug` to determine whether to set the verbosity, or `logDebug ` will always be false.


wonderflow (approved):
LGTM thanks

PR #1787, opened 2021-06-10
add environment design

hongchaodeng:
ref: https://github.com/oam-dev/kubevela/pull/1596, https://github.com/oam-dev/kubevela/pull/999, https://github.com/oam-dev/kubevela/issues/792.

This is the latest effort and thoughts to support Environment as shared base for application deployment.

### DIFF
--- a/design/vela-core/environment.md
+++ b/design/vela-core/environment.md
@@ -1,0 +1,144 @@
+# Managing Environments As Shared Bases For Application Deployment
+
+## Background
+
+A team of application development usually needs to prepare some shared environments for developers to deploy applications.
+For example, most teams will have two environments, a *dev* environment for testing their applications in live instances,
+and a *prod* environment for deploying applications to serve live requests. The environment is a logical group bundling common resources that multiple application deployments might depend on.
+
+Here are some examples of what users want to define in an environment:
+
+- K8s Cluster. *dev* environment might have smaller clusters while *prod* might use large and complex clusters.
+- System components and definitions. *dev* environment might install different Operators & CRDs from *prod* environment, e.g. *dev* might choose local logging while *prod* might use ELK logging.
+- Network boundary. *dev* and *prod* environments would have different VPC setups, gateway endpoints, and firewall settings.
+- Secret and config. *dev* might use locally-run MySQL, while *prod* might use managed RDS. The DB connection credentials would be saved differently.
+- Policies. This is more important in *prod* environment that it would usually have more restrictions on application deployments, like security scanning, config checks, SLO checks, etc.
+
+All of above examples and more are bundled as a shared base for applications to run on.
+This is usualy done by an admin for the entire application team.
+By adding Environment API, Vela aligns with user workflow better,
+provides consistent view and pluggable options for end users,
+and empowering platform builders to add modular environment capabilities. In the following, we will propose the design of API and give an implementation plan.
+
+
+## Proposal
+
+### 1. Environment CRD
+
+We propose to add an Environment CRD to describe the shared resources for applications to base on.
+
+```yaml
+kind: Environment
+metadata:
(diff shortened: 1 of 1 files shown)


design/vela-core/environment.md
```diff
@@ -0,0 +1,144 @@
...
+  environment: prod
+```
+
+### 2. EnvConfigDefinition CRD
+
+An Environment can have multiple EnvConfigs. The schema is defined in EnvConfigDefinition CRD:
+
+```yaml
+apiVersion: core.oam.dev/v1beta1
+kind: EnvConfigDefinition
```
wonderflow, line 87: Can we merge it with PolicyDefinition/WorkflowStepDefinition and call them Definition? We should reduce CRD to reduce complexity.

hongchaodeng: What's your suggestion? How can we do that if they are for different things and have different spec fields?

design/vela-core/environment.md
```diff
@@ -0,0 +1,144 @@
+# Managing Environments As Shared Bases For Application Deployment
+
+## Background
+
+A team of application development usually needs to prepare some shared environments for developers to deploy applications.
+For example, most teams will have two environments, a *dev* environment for testing their applications in live instances,
```
zzxwill, line 6: Normally we call it `testing` or `staging` environment.

design/vela-core/environment.md
```diff
@@ -0,0 +1,144 @@
...
+  name: prod
+spec:
+  # A list of pluggable EnvConfig resources.
+  # The resources will be parsed and applied when Environment CR is appled.
+  configs:
+    - type: cluster
+      properties:
+        name: prod-cluster
+        selector:
+          tier: prod
```
zzxwill, line 42: In `Env`, What we talked about most is `cluster`. I am curious about what's the definition of `cluster`. Could you please set an example for type `cluster`:)

design/vela-core/environment.md
wonderflow: Let's make a unified Object called CUEPackage, and the policy,workflow and env all rely on the CUEPackage as abstraction. And in the implementation, they're all CUE based templates.

hongchaodeng: First of all, I don't think we should only make policy, workflow, and env. If we want to make a unified object, all definitions should be there, including component and trait. We shouldn't treat policy, etc. as separate definitions.

Second, what about the specific fields that policy, workflow, or env have? They need to be in spec.

design/vela-core/environment.md
```diff
@@ -0,0 +1,144 @@
...
+
+```yaml
+kind: Environment
+metadata:
+  name: prod
+spec:
+  # A list of pluggable EnvConfig resources.
+  # The resources will be parsed and applied when Environment CR is appled.
+  configs:
+    - type: cluster
```
wonderflow, line 38: please describe the detail of cluster type.

design/vela-core/environment.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+We also need to add an optional field `environment` in Application:
+
+```yaml
+kind: Application
+spec:
+  # If set, app controller will add the AppRevision into the Environment
+  # CR spec after generating AppRevision before executing Workflow.
+  environment: prod
```
wonderflow, line 78: what if we want to deploy application to more than one environment?

design/vela-core/environment.md
wonderflow: +1

And what if it's not a K8s cluster? It's a env pointing to a cloud?

hongchaodeng:
Closing this for now.
After offline discussion, we probably do not need a new Environment API. Instead I will propose some mechanisms and implementation plans to provide environment management experience.

PR #1790, opened 2021-06-11
Only parse the parameter field when generating OpenAPI Schema

kinsolee:
fix https://github.com/oam-dev/kubevela/issues/1773


pkg/utils/common/common.go
```diff
@@ -130,8 +131,21 @@ func GenOpenAPI(inst *cue.Instance) ([]byte, error) {
 	if inst.Err != nil {
 		return nil, inst.Err
 	}
+	// Due to #1773, refine cue instance to merely include `parameter` identifier
+	paramVal := inst.LookupDef(velacue.ParameterTag)
+	sysopts := []cue.Option{cue.All(), cue.DisallowCycles(true), cue.ResolveReferences(true), cue.Docs(true)}
+	paramSyntax, err := format.Node(paramVal.Syntax(sysopts...))
```
zzxwill, line 137: `paraVal` should be checked whether it's nil ojbect.

kinsolee: Get it! I used `paramVal.Exists()` to check it.

zzxwill: I'm sorry, by looking deep into function `func (inst *Instance) LookupDef(path string) Value` again, I can confirm it won't return a nil object. So please ignore my comment and go ahead:)

kinsolee: It won't be a nil object, but you remind me that we still need to see whether there is `parameter` definition inside `cue.Instance`. So calling `Value.Exists()` should be fine here.

pkg/utils/common/common.go
```diff
@@ -140,6 +146,39 @@ func GenOpenAPI(inst *cue.Instance) ([]byte, error) {
 	return out.Bytes(), nil
 }
 
+// RefineParameterInstance refines cue instance to merely include `parameter` identifier
+func RefineParameterInstance(inst *cue.Instance) (*cue.Instance, error) {
+	paramVal := inst.LookupDef(velacue.ParameterTag)
+	if !paramVal.Exists() {
+		return nil, fmt.Errorf("cue identifier 'parameter' doesn't exists")
```
wonderflow, line 153: we should allow component or trait without parameters. In this case, we can generate an empty data.

kinsolee: It's weird that even though there is no parameter in cue.Instance, `paramVal.Exists()` still returns true. And the `Kind` of `paramVal` is `BottomKind`. 

image

I can just return an parameter with empty object here so that the outer function `utils.ConvertOpenAPISchema2SwaggerObject` can get the index from the `Schema` map: 

image


But now `utils.prepareParameterCue` returns an error when there is no parameter, do you expect that we still generate ConfigMap for those components/traits without parameter?

wonderflow: yes, we should still generate ConfigMap even no parameter, and those workload and traits are possible to have no parameters.

kinsolee: Ok, I have changed the code to make it possible to generate ConfigMap even no parameter.

pkg/controller/utils/capability.go
```diff
@@ -369,7 +369,10 @@ func getOpenAPISchema(capability types.Capability, pd *packages.PackageDiscover)
 func generateOpenAPISchemaFromCapabilityParameter(capability types.Capability, pd *packages.PackageDiscover) ([]byte, error) {
 	template, err := prepareParameterCue(capability.Name, capability.CueTemplate)
 	if err != nil {
-		return nil, err
+		// return OpenAPI with empty object parameter, making it possible to generate ConfigMap
```
wonderflow, line 372: what if other error occurs here?

pkg/utils/common/common.go
```diff
@@ -140,6 +146,39 @@ func GenOpenAPI(inst *cue.Instance) ([]byte, error) {
...
+	switch k := paramVal.IncompleteKind(); k {
+	case cue.StructKind, cue.ListKind:
+		sysopts := []cue.Option{cue.All(), cue.DisallowCycles(true), cue.ResolveReferences(true), cue.Docs(true)}
+		paramSyntax, err := format.Node(paramVal.Syntax(sysopts...))
+		if err != nil {
+			return nil, err
+		}
+		cueFormat := "#%s:{\n%s\n}"
+		if k == cue.ListKind {
+			cueFormat = "#%s: %s\n"
```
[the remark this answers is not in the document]
kinsolee: You are right! They can use the same format since there is only one child field.

pkg/utils/common/common.go
```diff
@@ -140,6 +146,39 @@ func GenOpenAPI(inst *cue.Instance) ([]byte, error) {
...
+		}
+		paramOnlyStr = fmt.Sprintf(cueFormat, velacue.ParameterTag, string(paramSyntax))
+	case cue.NumberKind, cue.StringKind, cue.FloatKind, cue.BoolKind:
+		paramOnlyStr = fmt.Sprintf("#%s: %v", velacue.ParameterTag, paramVal)
+	default:
+		return nil, fmt.Errorf("unsupport parameter kind: %s", reflect.TypeOf(k).Name())
```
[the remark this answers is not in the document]
kinsolee: Got it.

wonderflow (approved):

wonderflow:
good job

leejanee:
good job

leejanee (approved):

PR #1798, opened 2021-06-15
Remove implementation of vela install

### LINKED ISSUE #1796: remove implementation of vela install
Now `vela install` was not supported and users should use `helm install` instead. We should remove the implementation.

image

mason1kwok:
fix #1796


wonderflow (commented):
please also remove the related code:

1. `hack/chart/generate.go`
2. `references/cmd/cli/fake/chart_source.go`

wonderflow (approved):

PR #1804, opened 2021-06-16
add the option to choose whether to automatically generated workloaddef which componentdef refers to

yangsoon:
You can use the option `autoGenWorkloadDefinition` to choose whether to automatically create `WorkloadDefinition` which `ComponentDefinition` refers to. 

The default value of `autoGenWorkloadDefinition` is `true`, which means the webhook will automatically create `WorkloadDefinition`  which `ComponentDefinition` refers to , if the `WorkloadDefinition` doesn't exist.

You can set the `autoGenWorkloadDefinition` to `false` to disable this feature.

### DIFF
--- a/pkg/webhook/core.oam.dev/v1alpha2/componentdefinition/mutating_handler.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/componentdefinition/mutating_handler.go
@@ -19,6 +19,7 @@ package componentdefinition
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


pkg/webhook/core.oam.dev/v1alpha2/componentdefinition/mutating_handler.go
```diff
@@ -91,16 +93,18 @@ func (h *MutatingHandler) Mutate(obj *v1beta1.ComponentDefinition) error {
...
 		// Create workloadDefinition which componentDefinition refers to
-		workloadDef := new(v1beta1.WorkloadDefinition)
-		err = h.Client.Get(context.TODO(), client.ObjectKey{Name: defRef.Name, Namespace: obj.Namespace}, workloadDef)
-		if err != nil {
-			if apierrors.IsNotFound(err) {
-				workloadDef.SetName(defRef.Name)
-				workloadDef.SetNamespace(obj.Namespace)
-				workloadDef.Spec.Reference = defRef
-				return h.Client.Create(context.TODO(), workloadDef)
+		if h.AutoGenWorkloadDef {
```
wonderflow, line 96: we should still check before this flag, if not found , then this flag works, otherwise, we should report an error

yangsoon: got it

wonderflow (approved):
LGTM

PR #1831, opened 2021-06-22
add application  controller to caps which can be disabled and remove appdeployment

wonderflow:

This PR let application controller can be disabled by `--disable-caps=all`, so 0.3.x can be exist only for AppConfig and Component case.

And it remove the applicationdeployment controller.


hongchaodeng (approved):
lgtm

PR #1833, opened 2021-06-22
Fix capabilities publish

### LINKED ISSUE #1830: error converting YAML to JSON: yaml: line 11
**Describe the bug**

```bash

➜  ~ helm -n vela-system ls
NAME    	NAMESPACE  	REVISION	UPDATED                             	STATUS  	CHART          	APP VERSION
kubevela	vela-system	2       	2021-06-22 13:34:50.821828 +0800 CST	deployed	vela-core-1.0.6	1.0.6

kubectl vela comp --discover
Showing components from registry: oss://registry.kubevela.net
Get for-loop.yaml err: error converting YAML to JSON: yaml: line 11: did not find expected comment or line break
NAME     	REGISTRY	DEFINITION
webserver	default 	deployments.apps
```

chivalryq:

Now we use rsync to compare different and publish files to OSS registry. When we delete some wrong file `rsync` will still keep it. Maybe we should just use `cp` and `ossutil sync -f --delete` to overwrite all files in registry.

Fixes #1830 

### DIFF
--- a/.github/workflows/registry.yml
+++ b/.github/workflows/registry.yml
@@ -141,9 +141,7 @@ jobs:
         run: wget http://gosspublic.alicdn.com/ossutil/1.7.0/ossutil64 && chmod +x ossutil64 && mv ossutil64 ossutil
       - name: Configure Alibaba Cloud OSSUTIL
         run: ./ossutil --config-file .ossutilconfig config -i ${ACCESS_KEY} -k ${ACCESS_KEY_SECRET} -e ${CAPABILITY_ENDPOINT} -c .ossutilconfig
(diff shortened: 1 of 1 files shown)


zzxwill:
-1.

We should keep those legacy definitions due to forward compatibility, so I didn't implement a similar PR in my original design. As for the wrong ones, we can delete them manually. 

If manage, if a user used definition X, somehow, we removed it from our codebase, X will be removed. This is not acceptable.

chivalryq:
Well, got it. then we should remove files below: manualscale.yaml, podenv.yaml, for-loop.yaml @zzxwill

PR #1837, opened 2021-06-23
[FIX]incorrect pod-disruptive defalut configuration in label of traits

scaat:
The label of traits will update `spec.template.metadata.labels`, which will trigger pod rebuild. But in the default template, podDisruptive=true is not declared, which will cause misunderstandings by users.

Fixes #


scaat:
@wonderflow Please take a look.

wonderflow (approved):

PR #1855, opened 2021-06-29
fix invalid metadata.labels error when specify the version of the trait in the app

yangsoon:

Fixes https://github.com/oam-dev/kubevela/discussions/1810

Fix invalid metadata.labels error when specify the version of the trait in the app

### DIFF
--- a/pkg/appfile/parser.go
+++ b/pkg/appfile/parser.go
@@ -226,15 +226,20 @@ func (p *Parser) parseTrait(ctx context.Context, name string, properties map[str
 	if err != nil {
 		return nil, err
 	}
+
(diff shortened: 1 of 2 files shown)


pkg/appfile/parser.go
```diff
@@ -226,15 +226,20 @@ func (p *Parser) parseTrait(ctx context.Context, name string, properties map[str
 	if err != nil {
 		return nil, err
 	}
+
+	traitName, err := util.ConvertDefinitionRevName(name)
```
wonderflow, line 230: please write a test for this case

yangsoon: got it

wonderflow (approved):

PR #1862, opened 2021-06-30
Merge all zh-CN docs from KubeVela.io

BinaryHB0916:


wonderflow:
let's move en docs to kubevela.io

PR #1866, opened 2021-07-01
fix bug: failed to aggregate status if application use a kube/helm type component

yangsoon:

The KubeVela will failed to aggregate status if application use a kube/helm type component

Fixes https://github.com/oam-dev/kubevela/issues/1865

- [x] fix aggregate status
- [x] add test


wonderflow (approved):
LGTM

PR #1869, opened 2021-07-01
allow component have defer insert secret and config

### LINKED ISSUE #1350: [Feature] support part of components to reconcile even the whole application hit reconciling issue

For example, I have two components, `express-server` will try to consume the output secret `db-conn` generated by RDS instance of component `sample-db`. 

```
kind: Application
metadata:
  name: webapp
spec:
  components:
    - name: express-server
      type: webservice
      properties:
        image: xxx
        ports: 80
      envMapping
          # environments refer to db-conn secret
          DB_PASSWORD:
            secret: db-conn
            key: password                                    
          endpoint:
            secret: db-conn                           
          username:
            secret: db-conn

    - name: sample-db
      type: alibaba-rds
      properties:
        name: sample-db
        engine: mysql
        engineVersion: "8.0"
        instanceClass: rds.mysql.c1.large
        username: oamtest
        outputSecretName: db-conn
```

If in some situations, `express-server` wants to directly use the value of secret `db-conn` of item like `endpoint`.  Then component `sample-db` has to run first, create an RDS instance, and generate the secret. But currently, application will check both of the component are ready to run, or it will throw exception. As `express-server` will fail to retrieve the secret's value, then two component will stop to reconciling. It hangs.

So I suggest providing a way to enable some of components of an application, instead of running all of them at the same time when Application think it's ready for all of them.

**Additional context**

wonderflow:

This PR allows components won't be blocked when one of components don't have the secret and configmap ready.

Fixes #1350

### DIFF
--- a/apis/types/componentmanifest.go
+++ b/apis/types/componentmanifest.go
@@ -34,4 +34,9 @@ type ComponentManifest struct {
 	// Release, Git Repo or anything that can package and run a workload.
 	PackagedWorkloadResources []*unstructured.Unstructured
 	PackagedTraitResources    map[string][]*unstructured.Unstructured
+
+	// InsertConfigNotReady is true indicates the ComponentManifest is not ready to apply for insertSecret and configs
+	// it's possible for some of the component not ready while others are ready, we should not block all of them if only
+	// part is not ready
+	InsertConfigNotReady bool
 }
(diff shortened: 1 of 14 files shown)


captainroy-hy (approved):
LGTM

PR #1870, opened 2021-07-02
improve workflow docs and update API for flexible

wonderflow:
This PR updates the workflow API that wrap steps with a layer.


apis/core.oam.dev/v1beta1/application_types.go
```diff
@@ -71,6 +71,11 @@ type WorkflowStep struct {
 	Properties runtime.RawExtension `json:"properties,omitempty"`
 }
 
+// Workflow defines workflow steps and other attributes
+type Workflow struct {
```
hongchaodeng, line 75: Please also run `make manifests`

hongchaodeng (approved):
lgtm after nit

PR #1878, opened 2021-07-05
add observedGeneration field in status of Application

### LINKED ISSUE #1817: [Feature] Application status should have observedGeneration

The deployed resources should not be affected if it's deployed successfully, but now we don't have the mechanism to avoid it. Currently we also don't reconcile it by intend, but the controller will reconcile periodically.

1. we should add `observedGeneration` field in status of Application, after the status comes to health and it's already observed, we should not reconcile the real resources.
2. If we want to keep reconcile, we should add a policy(or a keep reconciler scope) to do that. #992

**Additional context**

Chwetion:
**What this PR does / why we need it:**

Perceive the version difference between status and spec, from #1817

**Which issue(s) this PR fixes:**

Fixes #1817 

**Special notes for your reviewer:**

### DIFF
--- a/apis/core.oam.dev/common/types.go
+++ b/apis/core.oam.dev/common/types.go
@@ -224,6 +224,10 @@ type AppStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	runtimev1alpha1.ConditionedStatus `json:",inline"`
 
+	// The generation observed by the application controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
 	Rollout AppRolloutStatus `json:"rollout,omitempty"`
 
 	Phase ApplicationPhase `json:"status,omitempty"`
--- a/pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
@@ -320,6 +320,7 @@ func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.
 }
 
 func (r *Reconciler) patchStatus(ctx context.Context, app *v1beta1.Application) error {
+	updateObservedGeneration(app)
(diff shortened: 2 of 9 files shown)


pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -110,6 +110,14 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, nil
 	}
 
+	defer func() {
+		updateObservedGeneration(app)
+		if err := r.patchStatus(ctx, app); err != nil {
+			result, returnErr = ctrl.Result{}, errors.WithMessage(err, "cannot update application status")
+			return
+		}
+	}()
```
wonderflow, line 119: can we merge it into the function `patchStatus`

wonderflow (approved):
LGTM after CI

apis/core.oam.dev/common/types.go
```diff
@@ -224,6 +224,10 @@ type AppStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	runtimev1alpha1.ConditionedStatus `json:",inline"`
 
+	// The generation observed by the application controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration"`
```
wonderflow, line 229:
```suggestion
	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -320,6 +320,7 @@ func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.
 }
 
 func (r *Reconciler) patchStatus(ctx context.Context, app *v1beta1.Application) error {
+	updateObservedGeneration(app)
```
wonderflow, line 323: please add a test in some existing case

PR #1879, opened 2021-07-05
fix rollout and only rollout the first component

wonderflow:
1. fix rolloutplan that do not limit only one component but the rollout only support one component. Make apply workload option works only for the target component
2. do not block other component to render if they're not rollout target
3. remove redundant step `templateSourceManifest`, it want's to apply the new resource with older ownerID and fails the update.


wonderflow:
prefer #1888

PR #1888, opened 2021-07-06
dont't template source in rollout and rollout first component

wangyikewxgm:
1.fix rolloutplan that do not limit only one component but the rollout only support one component. Make apply workload option works only for the target component
2.do not block other component to render if they're not rollout target
3.remove redundant step templateSourceManifest, it want's to apply the new resource with older ownerID and fails the update.

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/applicationrollout/rollout_handler.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/applicationrollout/rollout_handler.go
@@ -76,8 +76,8 @@ func (h *rolloutHandler) prepareRollout(ctx context.Context) error {
 
 	// construct a assemble manifest for targetAppRevision
 	targetAssemble := assemble.NewAppManifests(h.targetAppRevision).
-		WithWorkloadOption(rolloutWorkloadName()).
-		WithWorkloadOption(assemble.PrepareWorkloadForRollout())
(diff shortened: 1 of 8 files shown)


pkg/controller/core.oam.dev/v1alpha2/applicationrollout/rollout_handler.go
```diff
@@ -272,3 +275,49 @@ func (h *rolloutHandler) finalizeRollingSucceeded(ctx context.Context) error {
...
+			BlockOwnerDeletion: pointer.BoolPtr(true),
+		}
+		workload.SetOwnerReferences([]metav1.OwnerReference{ownerRef})
+	}
+	// after last succeed rollout finish, workload's ownerReference controller have been set true
+	// so we should disable the controller owner firstly
+	disableControllerOwner(workload)
+	if err = h.Client.Patch(ctx, workload, wlPatch, client.FieldOwner(h.appRollout.UID)); err != nil {
+		return err
+	}
```
wonderflow, line 321: 如果 wantOwner 已经存在了，为什么还要patch？

wangyikewxgm: handleSourceWorkload有两个作用，1.如果不是lagacy的workload把rt的controller owner设置为false 2.如果是lagacy除了做上面的一项，另外去掉appcontext的owner，设置成rt为他的owener

PR #1889, opened 2021-07-06
Feat: no need create appRevision if it already presents in histories

LeoLiuYan:

As the title said.

[1786](https://github.com/oam-dev/kubevela/issues/1786)
[1818](https://github.com/oam-dev/kubevela/issues/1818)

Fixes #


wonderflow (changes requested):
I think we should only list and check all revision before create. And make the new revision name use the existing revision name.  Leave other logic as it is.

Even the revision already exist, while the spec is changed from the last revision, we should trigger rolling upgrade to happen.

PR #1898, opened 2021-07-08
Bump up test environment K8s to v1.20

chivalryq:

As lots of user have been using k8s >= v1.20, it's time to bump up our test environment k8s version

none


contribute/developer-guide.md
```diff
@@ -5,7 +5,7 @@ This guide helps you get started developing KubeVela.
 ## Prerequisites
 
 1. Golang version 1.16+
-2. Kubernetes version v1.18+ with `~/.kube/config` configured.
+2. Kubernetes version v1.20+ with `~/.kube/config` configured.
```
zzxwill, line 8: v1.18+ is okay. Developers needn't have a Kubernetes cluster v1.20+.

chivalryq: Yes, but since we bump up to 1.20, we can not test what cases behave in v1.18. That may cause problem.

zzxwill: Right now, we have to support v1.18, of which version is heavily used.

wonderflow:
Currently, Let's hold it.

PR #1900, opened 2021-07-09
upgrade workflow design

leejanee:
upgrade [workflow design](https://github.com/oam-dev/kubevela/blob/master/design/vela-core/workflow_policy.md)


design/vela-core/workflow_policy.md
```diff
@@ -199,38 +201,112 @@ Here are the steps in Workflow Manager:
...
 
-- If `continue.return == true`, Task Manager will update the `observedAppGeneration` for this step in status.
+
+### Workflow Operation
+
+#### Terminate Workflow
+
+If the execution of the workflow does not meet expectations, it may be necessary to terminate the workflow
+
+There are two ways to achieve that
```
hongchaodeng, line 241:
```suggestion
There are two ways to achieve that:
```

hongchaodeng: The following two approaches are not mutual-exclusive. The op.Break can be captured and then report terminated status.

design/vela-core/workflow_policy.md
```diff
@@ -199,38 +201,112 @@ Here are the steps in Workflow Manager:
...
+```yaml
+kind: Application
+spec:
+  componnets: ...
+workflow:
+  steps:
+  - name: manual-approve
+    type: suspend   
+ ```
+  The workflow automatically pauses when the suspend-type task is started
```
hongchaodeng, line 280: How to resume?

design/vela-core/workflow_policy.md
```diff
@@ -199,38 +201,112 @@ Here are the steps in Workflow Manager:
...
+    name: foo
+  status:
+    phase: runningWorkflow
+    workflow:
+      stepIndex: 1
+      terminated: true
+      steps:
+      - name: ...
+ ```
+#### Pause & Resume Workflow
```
hongchaodeng, line 267: This section is confusing by showing two points directly. Please add some description about why you make two points.

design/vela-core/workflow_policy.md
```diff
@@ -199,38 +201,112 @@ Here are the steps in Workflow Manager:
...
+    phase: runningWorkflow
+    workflow:
+      stepIndex: 1
+      suspend: true
+      steps:
+      - name: ... 
+ ```
+#### Restart Workflow
+
+The workflow will be executed in the following two cases
```
hongchaodeng, line 298:
```suggestion
The workflow will be restarted in the following two cases:
```

design/vela-core/workflow_policy.md
```diff
@@ -199,38 +201,112 @@ Here are the steps in Workflow Manager:
...
+1. The application spec changes
```
hongchaodeng, line 299: Please describe what happens to the current state during restart

design/vela-core/workflow_policy.md
```diff
@@ -199,38 +201,112 @@ Here are the steps in Workflow Manager:
...
+2. Modify the value of the `status.phase` field to "runningWorkflow" and clear the status of the workflow
+
+```yaml
+  kind: Application
+  metadata:
+    name: foo
+  status:
+    phase: runningWorkflow
+    workflow:
```
hongchaodeng, line 308: `workflow: {}`

hongchaodeng (commented):
Besides existing comments, there are a few more issues I think we should resolve:

- The mechanism of how CUE task is executed, how to report status is not concrete.
- The limitation of reentrant tasks.
- The limitation of synchronous execution of CUE tasks is not discussed.
- The diagram is confusing:
  - What does 'Create Task' in the middle mean? => call TaskManager
  - Why are there two 'Execute CUE tasks' rectangles? => The second one should be changed to 'parse do steps'.
  - If continue.return is true, it should run next task, right?
  - How to run next task when return is true?
- Need to clarify how to execute all steps in order in a CUE Task including checking the 'do' field.
- In diagram, 'for each workflow step' should be changed to 'get current stepIndex', also need to increment stepIndex at the end.

leejanee:
@hongchaodeng please review.

hongchaodeng (approved):
lgtm

Issue #1653, opened 2021-05-13
[Feature]  implementation more workloads rollout plugin


wangyikewxgm:
**Is your feature request related to a problem? Please describe.**
Currently, our rollout plugin have supported deployment and cloneSet workload.I think implement more workload controllers such as statefulSet and daemonSet and so on
 
**Describe the solution you'd like**
1.Implement  rollout controller for statefulSet and daemonSet 
2.write unit-test integration-test and e2e-test for logic. test  coverage must reach target.

**Describe alternatives you've considered**
As a stateful workload, statefulSet's controller may have much difference with deployment and cloneSet.

IMO, this feature can be submit as a subject for OpenSource Summer Plan #1641

whichxjy:
![3](https://user-images.githubusercontent.com/33963637/125058924-539b1a00-e0dd-11eb-9cae-52402911bb5c.png)

LeoLiuYan:
Hi @wangyikewxgm ,  I'm interesting in implement the [Sidecarset](https://openkruise.io/en-us/docs/sidecarset.html) workload.

LeoLiuYan:
Hi, @wangyikewxgm. I'm interesting in implement the [SidecarSet](https://openkruise.io/en-us/docs/sidecarset.html) workload.

briankane:
This issue has been inactive for over 2 years and is scheduled for closure as part of our backlog maintenance.

  **If this issue is still relevant:**
  - Please comment below to keep it open
  - Provide updated context if the original description is outdated

  **Timeline:** If there's no activity by **Friday, October 24th**, this issue will be automatically closed.

  **Note:** Issues can always be reopened if they become relevant again. We encourage the community to reopen and reprioritize any incorrectly 
  closed issues.

briankane:
This issue has been closed following the stale issue notice. 

If this issue is still relevant and should remain open, please:
1. Reopen the issue
2. Remove the 'stale' label
3. Provide an update on the current status or importance

Thank you for your understanding and contribution to KubeVela!

PR #1901, opened 2021-07-09
Support scale controller for StatefulSet

whichxjy:
This PR adds support for the StatefulSet scale controller. It's a part of #1653.

### DIFF
--- a/pkg/controller/common/rollout/rollout_plan_controller.go
+++ b/pkg/controller/common/rollout/rollout_plan_controller.go
@@ -336,6 +336,7 @@ func (r *Controller) GetWorkloadController() (workloads.WorkloadController, erro
 	}
 
 	if r.targetWorkload.GroupVersionKind().Group == kruisev1.GroupVersion.Group {
+		// check if the target workload is CloneSet
 		if r.targetWorkload.GetKind() == reflect.TypeOf(kruisev1.CloneSet{}).Name() {
 			// check whether current rollout plan is for workload rolling or scaling
 			if r.sourceWorkload != nil {
@@ -348,6 +349,7 @@ func (r *Controller) GetWorkloadController() (workloads.WorkloadController, erro
 	}
 
 	if r.targetWorkload.GroupVersionKind().Group == apps.GroupName {
+		// check if the target workload is Deployment
 		if r.targetWorkload.GetKind() == reflect.TypeOf(apps.Deployment{}).Name() {
 			// check whether current rollout plan is for workload rolling or scaling
 			if r.sourceWorkload != nil {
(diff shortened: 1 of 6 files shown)


wangyikewxgm:
Good first work. statefulset rollout controller will contained by next pr?

whichxjy:
> Good first work. statefulset rollout controller will contained by next pr?

Yes, sure.

pkg/controller/common/rollout/rollout_plan_controller.go
```diff
@@ -357,6 +359,15 @@ func (r *Controller) GetWorkloadController() (workloads.WorkloadController, erro
...
 		}
+
+		// check if the target workload is StatefulSet
+		if r.targetWorkload.GetKind() == reflect.TypeOf(apps.StatefulSet{}).Name() {
+			if r.sourceWorkload == nil {
+				return workloads.NewStatefulSetScaleController(r.client, r.recorder, r.parentController,
+					r.rolloutSpec, r.rolloutStatus, target), nil
+			}
+		}
 	}
```
wangyikewxgm, line 370: Please delete. When next pr (`statefulset rollout controller` )ready we can add this back.

pkg/controller/common/rollout/workloads/statefulset_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	if controller := metav1.GetControllerOf(statefulSet); controller != nil &&
+		controller.Kind == v1beta1.AppRolloutKind && controller.APIVersion == v1beta1.SchemeGroupVersion.String() {
+		// it's already there
+		return true, nil
+	}
+
+	statefulSetPatch := client.MergeFrom(statefulSet.DeepCopyObject())
+
+	// add the parent controller to the owner of the StatefulSet
+	ref := metav1.NewControllerRef(c.parentController, v1beta1.AppRolloutKindVersionKind)
```
wangyikewxgm, line 51: please replace `v1beta1.AppRolloutKindVersionKind` with `c.parentController.GetObjectKind().GroupVersionKind()` we will support more kind of object  as parent

whichxjy: It seems like `c.parentController.GetObjectKind().GroupVersionKind()` would return an empty `GroupVersionKind` struct:

```
GroupVersionKind{
    Group:   "",
    Version: "",
    Kind:    "",
}
```

Is there another way to do that?

wangyikewxgm: I reproduce this issue.I think this a test-side issue, in integration-test we generate a fake appRollout that not exist in k8s . so these fields are empty. You needn't modify it , i will fix it in my pr.

wangyikewxgm (approved):
LGTM

PR #1902, opened 2021-07-09
Fix the way to ignore errors for rm command

whichxjy:
This script is piping to `rm`, a command that doesn't read from stdin.


chivalryq (approved):
Nice catch, I misspell at that time.

zzxwill:
Which word did you misspell?

chivalryq:
> Which word did you misspell?

I "set -e" to catch error on the top of the script.
"|| true" means ignore the error if command ahead fail(make the whole command return 0), I misspell as pipeline

wonderflow (approved):

PR #1920, opened 2021-07-14
fix bug: application can not create component which contains namespace resource

yangsoon:

Fix bug: Application can not create component which contains namespace resource.

Initializer rely on the app's ability to create resources, sometimes initializing an environment requires creating namespace,
this pr will fix the bug which application can not create namespace  resource.


vela-templates/addons/terraform/resource/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: tc-system
```
wonderflow, line 4:
```suggestion
  name: terraform-system
```

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -300,6 +301,11 @@ func (am *AppManifests) filterAndSetAnnotations(obj *unstructured.Unstructured)
 }
 
 func (am *AppManifests) setNamespace(obj *unstructured.Unstructured) {
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	if gvk == corev1.SchemeGroupVersion.WithKind(reflect.TypeOf(corev1.Namespace{}).Name()) {
```
wonderflow, line 305: please write a test for it

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
```diff
@@ -300,6 +301,11 @@ func (am *AppManifests) filterAndSetAnnotations(obj *unstructured.Unstructured)
 }
 
 func (am *AppManifests) setNamespace(obj *unstructured.Unstructured) {
```
chivalryq, line 303: please add comment like “setNamespace set namespace except for Namespace object"

chivalryq: to make it clearer for other developer

yangsoon: got it, thanks

vela-templates/addons/terraform/resource/namespace.yaml
yangsoon: got it

pkg/controller/core.oam.dev/v1alpha2/application/assemble/assemble.go
yangsoon: got it

wonderflow (approved):

PR #1930, opened 2021-07-18
fix bug: skip pkg discover when workload type of componentdefinition is autodetect

yangsoon:

componentdefinition controller should skip `RefreshPackageDiscover` step when the workload type of componentDefinition is `autodetects.core.oam.dev`


wonderflow (approved):
LGTM

PR #1955, opened 2021-07-22
Fix Initializer bugs & Enhance addon command 

chivalryq:

Initializer:
1. When installing dependent initializer ,controller won't update status as `CheckingDepends` ,fix that
image
image
2. After initializer patch status, reconcile return an error, that will requeue the request twice, fix that.
3. Add delete webhook, forbid deletion of initializer that are depended by other initializer. Add test for that.

vela CLI:
1. enable: continually show initializer status
image
2. disable: remove deleting ns logic(all init are now in vela-system namespace)

### DIFF
--- a/references/cli/addon.go
+++ b/references/cli/addon.go
@@ -329,7 +329,7 @@ func waitForInitializerSuccess(obj *unstructured.Unstructured) error {
 	return wait.PollImmediateInfinite(period, func() (done bool, err error) {
 		err = clt.Get(ctx, types2.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, &init)
 		if err != nil {
-			return false, nil
+			return false, client.IgnoreNotFound(err)
 		}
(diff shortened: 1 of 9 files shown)


references/cli/addon.go
```diff
@@ -311,16 +312,32 @@ func (a *Addon) enable() error {
...
+	waitForInitializerSuccess(obj)
 	return nil
 }
 
+func waitForInitializerSuccess(obj *unstructured.Unstructured) {
+	ctx := context.Background()
+	var init v1beta1.Initializer
+	stopCh := make(chan struct{})
+	period := 20 * time.Second
+	wait.Until(func() {
```
yangsoon, line 327: wait.PollImmediate() seems be better?

yangsoon:
good catch

references/cli/addon.go
chivalryq: done

wonderflow (approved):

PR #1962, opened 2021-07-22
fix app finalizer bug

captainroy-hy:

**Prblem**:

App controller uses finalizer to garbage collect resource tracker according to its status latest revision.
If an error occurs in the middle of dispatching(applying) resources, even resource tracker is created but app status cannot record its revision as the latest revision. Thus, if we delete the application now, app controller cannot GC this resource tracker.

**Solution**:
During GC, list all resource trackers related to this application and delete them.

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
@@ -38,7 +38,6 @@ import (
 	velatypes "github.com/oam-dev/kubevela/apis/types"
 	"github.com/oam-dev/kubevela/pkg/appfile"
 	core "github.com/oam-dev/kubevela/pkg/controller/core.oam.dev"
-	"github.com/oam-dev/kubevela/pkg/controller/core.oam.dev/v1alpha2/application/dispatch"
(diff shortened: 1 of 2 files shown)


pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -253,14 +252,6 @@ func (r *Reconciler) handleFinalizers(ctx context.Context, app *v1beta1.Applicat
 			klog.InfoS("Register new finalizer for application", "application", klog.KObj(app), "finalizer", resourceTrackerFinalizer)
 			return true, errors.Wrap(r.Client.Update(ctx, app), errUpdateApplicationFinalizer)
 		}
-		if appWillRollout(app) {
```
wonderflow, line 256: why this removed

captainroy-hy: Before, we use `list` to delete resource trackers only when it's created by app rollout. Now, we  use `list` to delete any resource trackers. So it's no need to handle approllout-created ones.

wonderflow (approved):

PR #1969, opened 2021-07-26
Feat: support rollout controller for StatefulSet

whichxjy:
This PR adds support for the StatefulSet rollout controller. It's a part of #1653.


wangyikewxgm (changes requested):
Please let source and target workloads be the same one

magick93:
Hello

Does this PR mean statefulsets will be supported in Kubevela? 

Currently is it possible to add support using a custom component. or perhaps some other approach?

whichxjy:
> Hello
> 
> Does this PR mean statefulsets will be supported in Kubevela?
> 
> Currently is it possible to add support using a custom component. or perhaps some other approach?

StatefulSet will be supported, but it's not ready now. When it's ready, it can be used in custom components.

wangyikewxgm (changes requested):
在本地测试发现，rollout 通过locatingTargetApp阶段之后，workload会被删除掉。请查一下这个问题 ， 代码中是有删除workload的操作吗？@whichxjy

pkg/controller/core.oam.dev/v1alpha2/applicationrollout/helper.go
```diff
@@ -49,17 +49,27 @@ func RolloutWorkloadName(rolloutComp string) assemble.WorkloadOption {
 
 		// we hard code the behavior depends on the workload group/kind for now. The only in-place upgradable resources
 		// we support is cloneset/statefulset for now. We can easily add more later.
+		supportInplaceUpgrade := false
 		if w.GroupVersionKind().Group == v1alpha1.GroupVersion.Group {
 			if w.GetKind() == reflect.TypeOf(v1alpha1.CloneSet{}).Name() ||
 				w.GetKind() == reflect.TypeOf(v1alpha1.StatefulSet{}).Name() {
```
wangyikewxgm, line 55: This line is useless, please delete.

whichxjy: The `supportInplaceUpgrade` variable will be used to determine if the workload supports in-place upgrade down below:
```go
if supportInplaceUpgrade {
	// we use the component name alone for those resources that do support in-place upgrade
	klog.InfoS("we reuse the component name for resources that support in-place upgrade",
		"GVK", w.GroupVersionKind(), "instance name", w.GetName())
	// assemble use component name as workload name by default
	// so no need to re-set name
	return nil
}
```
Should we keep this part?

wangyikewxgm: I mean only line 55 is useless.

wangyikewxgm (approved):
good work！

wonderflow (approved):
thanks for your contributions

PR #1970, opened 2021-07-27
Workflow Support Resource GC

leejanee:
Upgrade the code of the provider `kube` and call resource GC after the workflow is completed.


wonderflow (approved):
LGTM, please fix CI and increase the coverage

\cc @captainroy-hy please review

captainroy-hy (approved):
LGTM.

PR #1974, opened 2021-07-27
use corev1.ObjectReference insteadof  crossplane-runtime TypedReference

wonderflow:
And refactor the related part.

The ObjectReference from K8s core library is better than TypedReference in crossplane.

Fixes #


hongchaodeng (approved):
LGTM

zzxwill (approved):
/lgtm

PR #1985, opened 2021-07-29
upgrade K8s dependency lib to v0.21

wonderflow:
# What this PR does and Why？

1. Upgrade controller-runtime to v0.9.3 which matches the K8s v1.21.x.  Our old version(v0.2.x)  matches with K8s version v1.16 which is already not maintained any more. The community maintained K8s version is v1.19~v1.21. There're potential security risks if we don't upgrade these dependency. Fix #1972

2. Why not upgrade to controller-runtime v0.8.x which matches to K8s v1.20.x?
   * the controller-runtime didn't support K8s  v1.20 until v0.9.x https://github.com/kubernetes-sigs/controller-runtime/pull/1486 . BTW, kube-apiserver and etcd can be downloaded from https://storage.googleapis.com/kubebuilder-tools

3. Upgrade crossplane-runtime and other dependency libraries to match the new version of controller-runtime.

4. Fix a lot of incompatible changes.
   - field Object can not be decoded in runtime.RawExtension, so we need to marshal it as raw bytes into the `Raw` field.
   - longer the timeout duration to 1min for apiserver and etcd to be started
   - longer the suittest BeforeSuit function timeout to 120s

5. Upgrade github action to matches with the new version
   ```
        uses: RyanSiu1995/kubebuilder-action@v1.2
        with:
          version: 3.1.0
          kubebuilderOnly: false
          kubernetesVersion: v1.21.2
   ```


captainroy-hy:
Maybe [this tool](https://github.com/kubernetes-sigs/controller-runtime/blob/master/tools/setup-envtest/README.md#envtest-binaries-manager) can help to install evntest binaries.

pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go
```diff
@@ -327,14 +327,23 @@ type ObjectOwner interface {
 	SetOwnerReferences([]metav1.OwnerReference)
 }
 
-func setOrOverrideControllerOwner(obj ObjectOwner, controllerOwner metav1.OwnerReference) {
-	ownerRefs := []metav1.OwnerReference{controllerOwner}
+// setOrOverrideOAMControllerOwner will set the
```
captainroy-hy, line 330: typo

captainroy-hy (approved):
Great work!  LGMT

hongchaodeng (approved):
LGTM.

Please write a doc to tell contributors how to upgrade kubebuilder and test environment.

PR #1996, opened 2021-07-30
fix rollout test

wonderflow:
image

This is a mis-match, and we don't find out it because the controller-runtime don't override data when the value is empty.


leejanee:
LGTM

leejanee (approved):

PR #2009, opened 2021-08-03
bugfix: workload namespace is empty while sync appdeployment

shaowenchen:
Signed-off-by: shaowenchen mail@chenshaowen.com

What this PR does / why we need it:

It raises an error Failed to reconcile appDeployment default/cross-cluster-app: cannot get object: an empty namespace may not be set when a resource name is provided when I test case refers to https://kubevela.io/docs/v1.1/end-user/scopes/appdeploy/#appdeployment


shaowenchen:
@yangsoon Can you help me to review this pr? This problem blocks my usage.

wonderflow (approved):
good catch, thanks

PR #2011, opened 2021-08-03
Enhance policy: apply policy's workload before workflow and add context.components for policydefinition

yangsoon:

This PR enhance the policy's ability:

1. apply policy's workload before the workflow starts.
2. add `context.components` for policydefinition ,   represents the `components` field in the Application

Others:
1. add `op.#ApplyEnvBindComponent`
2. fix bug of workflow

TODO:

- [x] add test


pkg/cue/process/handle.go
```diff
@@ -134,7 +139,18 @@ func NewContextWithHooks(namespace, name, appName, appRevision string, baseHooks
 	}
 }
 
-// SetBase set templateContext base model
+// NewPolicyContext create Application Scope templateContext for Policy
+func NewPolicyContext(namespace, name, appName, appRevision string, components []common.ApplicationComponent) Context {
+	return &templateContext{
+		name:        name,
```
wonderflow, line 145: name is the component name, which should not be in context of Policy

yangsoon: this name represent the name of policy

wonderflow: ok, that's fine

wonderflow (approved):
GOOD JOB

wonderflow (approved):

PR #2012, opened 2021-08-03
disable metrics server port in tests and add log for rollout

wonderflow:

set to 0 means disable it

Fixes #1779 


hongchaodeng (approved):
lgtm

PR #2016, opened 2021-08-03
Fix vela-core-webhook namespace not configurable

tossmilestone:
Signed-off-by: Xiaoxi He <tossmilestone@gmail.com>

Fix the namespace of vela-core-webhook referenced in crd `application` is hardcoded 'vela-system', which should be '{{ .Release.Namespace }}`.

None


tossmilestone:
The solution should be using webhook-certgen to patch the namespace in crd webhook, because the crds couldn't reference the vars int `values.yaml`.

Issue #2007, opened 2021-08-02
Pass information from component into trait in the case of Terraform configurations


neinkeinkaffee:
### Discussed in https://github.com/oam-dev/kubevela/discussions/2006

<div type='discussions-op-text'>

<sup>Originally posted by **neinkeinkaffee** August  2, 2021</sup>
Hello,

I have a `ComponentDefinition` and a `TraitDefinition` that both output a Terraform configuration. The component is supposed to create an S3 bucket, while the trait is supposed to create an IAM policy and role with read access to the bucket. In order for this to work, the trait needs to know about the name of the bucket that the component it's attached to creates. I'm wondering what a good way to access that information from within the trait could be. Right now I'm just passing in the same parameter twice, to the component and the trait (see my `ApplicationDefinition` below), but I suppose there must be a better way?

This is my `ApplicationDefinition`:
```
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: some-application-with-trait
spec:
  components:
    - name: s3-output-port
      type: output-port
      properties:
        port_name: podcasts
      traits:
        - type: access-policy
          properties:
            port_name: podcasts
```

The `ComponentDefinition` looks like this:
```
apiVersion: core.oam.dev/v1beta1
kind: ComponentDefinition
metadata:
  name: output-port
  annotations:
    definition.oam.dev/description: Terraform configuration for s3 output port
    type: terraform
spec:
  workload:
    definition:
      apiVersion: terraform.core.oam.dev/v1beta1
      kind: Configuration
  schematic:
    terraform:
      configuration: |
        module "s3_bucket" {
          source = "http://my-nginx/s3.zip"

          port_name = var.port_name
        }

        output "port_bucket_name" {
          value = module.s3_bucket.bucket_name
        }

        variable "port_name" {
              description = "S3 bucket name"
              type = string
        }

```

And the `TraitDefinition` like this:
```
apiVersion: core.oam.dev/v1beta1
kind: TraitDefinition
metadata:
  name: access-policy
spec:
  schematic:
    cue:
      template: |
        import "encoding/yaml"

        parameter: {
            port_name: string
        }

        outputs: port_access: {
            apiVersion: "terraform.core.oam.dev/v1beta1"
            kind:       "Configuration"
            spec: {
              yaml.Unmarshal(hclyaml)
              variable:
                output_port: parameter.port_name
            }
        }

        hclyaml: """
          hcl: |
            module "output_port_read_access" {
[...]

PR #2030, opened 2021-08-05
Application/Controller: store component parameters in context (#2007)

neinkeinkaffee:

Currently the options for passing context information about a component to an attached trait are limited when the component is defined in Terraform/HCL.

This PR is partial quick-fix, not a fundamental solution to the problem. It effects that the `properties` configured for a given component in the `ApplicationDefinition` get stored in `context.parameters`, so that the values are accessible to a given component's traits.

**Which issue(s) this PR fixes**: #2007


wonderflow (approved):
LGTM, thanks

pkg/cue/process/handle.go
```diff
@@ -139,6 +146,11 @@ func (ctx *templateContext) SetConfigs(configs []map[string]string) {
 	ctx.configs = configs
 }
 
+// SetParameters sets templateContext parameters
+func (ctx *templateContext) SetParameters(params map[string]interface{}) {
+	ctx.parameters = params
+}
```
wonderflow, line 152: can you help add tests to cover these lines

neinkeinkaffee: Ah, interesting! I would have to think of a way. I just wonder why they're not covered by the fact that SetParameters is invoked in [pkg/cue/process/handle_test.go](https://github.com/neinkeinkaffee/kubevela/blob/d4bbd139dfbedf25990536191673900c5c7698d0/pkg/cue/process/handle_test.go#L83)...?

neinkeinkaffee: I can't see another test where the corresponding methods of the Context interface are covered, apart from handle_test.go. Could you suggest a good place where I might place an additional test for `SetParameters`?

leejanee:
good job

pkg/cue/process/handle.go
wonderflow: Oh, then it's ok, let me merge it

PR #2043, opened 2021-08-06
Specified component revision name

### LINKED ISSUE #1884: [Feature] Allow users to specify component revision name

Now our component revision is automatically generated with incremental number. It has some drawbacks:

1. Real users has their own version standard, some version number is business related. They hope the version in kubvela can match with their own revision. So they can have end to end consistency.
2. Users need to apply application first to know the revision, and list/get the created revision, and then specify revision in workflow. If they can specify a revision, the steps can merge to one.

Add an externalRevision field in component.

```
kind: Application
spec:
  components:
  - name: mycomp
    type: webservice
    externalRevision: my-revision-v1
    properties:
      ...
```

1. If externalRevision specified, generate the controller revision with the specified name.
2. Add webhook validation, if the `externalRevision` already exist, check their spec hash, if not equal, report an error.
3. Don't create or update component revision when specified externalRevision already exist.
4. If  externalRevision not specified，just work as it is now.

**Additional context**

Somefive:

Continue #1929
Fixes #1884

### DIFF
--- a/pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler_test.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler_test.go
@@ -20,9 +20,12 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	admissionv1 "k8s.io/api/admission/v1"
+	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/oam-dev/kubevela/pkg/oam"
 )
 
 var _ = Describe("Test Application Validator", func() {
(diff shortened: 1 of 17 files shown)


pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler_test.go
```diff
@@ -164,8 +164,7 @@ var _ = Describe("Test Application Validator", func() {
 				},
 			},
 			Data: runtime.RawExtension{
-				Raw: []byte(`
-{"apiVersion":"core.oam.dev/v1beta1",
+				Raw: []byte(`{"apiVersion":"core.oam.dev/v1beta1",
```
LeoLiuYan, line 167: It's strange that this bug not reproducted in my  computor.   How did you find it? 

@Somefive

Somefive: I tried to recreate a new ControllerRevision and found the format matter. I guess the reason might be the first byte in Raw should not be \n otherwise something weird will happen.

Somefive:
@LeoLiuYan You can continue with this fix by pull this branch to your branch. The current issue here is the coverage to the code is a bit lower than the acceptance threshold. It seems that error handling is not well tested. It would be better to add one or two more test to simulate some boundary scenes where errors might be raised. As this feature is on our roadmap to v1.1, we are looking forward to get this MR passed in two or three more days. Do you have time to do that? :)

pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler_test.go
LeoLiuYan: The APIServer returns confused message  ` INTERNAL_ERROR (0x2)`.  
Thanks for you let me  learn some new knowledge.

LeoLiuYan:
> @LeoLiuYan You can continue with this fix by pulling this branch to your branch. The current issue here is the coverage to the code is a bit lower than the acceptance threshold. It seems that error handling is not well tested. It would be better to add one or two more test to simulate some boundary scenes where errors might be raised. As this feature is on our roadmap to v1.1, we are looking forward to get this MR passed in two or three more days. Do you have time to do that? :)

OK.   I will add some more Unit Test in this afternoon or evening.

wonderflow:
closing as  it's merged in #1929

PR #2049, opened 2021-08-09
Add add-on input go-template implementation

chivalryq:

Now, add-on are fixed and can not be customize. We should make it extensible.

This PR add an go-template implementation for add-on to receive input.

Here is an example. One can define a addon by using `{{ index .Args foo}}` to indicate that This string is to be fill in by user.

User can use command like below to pass parameters to add-on.
```shell
$ addon enable <add-on name> foo1=bar1 foo2=bar2
```

Note that `index` and `.Args` are **fixed**. 

Fixes # 2048 

### DIFF
--- a/references/cli/addon.go
+++ b/references/cli/addon.go
@@ -17,8 +17,11 @@ limitations under the License.
 package cli
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"strings"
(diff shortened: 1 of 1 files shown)


wonderflow (commented):
please add test

wonderflow (approved):

PR #2059, opened 2021-08-10
fix metadata.name to lower case

zou2699:
when kubectl apply -f definition.yaml, the error is occurred
```
The ComponentDefinition "singletonServer" is invalid: metadata.name: Invalid value: "singletonServer": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

```

Fixes #

### DIFF
--- a/docs/examples/workflow/apply-remaining/app.yaml
+++ b/docs/examples/workflow/apply-remaining/app.yaml
@@ -6,9 +6,9 @@ metadata:
 spec:
   components:
     - name: "server1"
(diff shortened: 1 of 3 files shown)


wonderflow (approved):
good catch， please sign off

PR #2074, opened 2021-08-12
vela runtime cluster charts (contains rollout)

wangyikewxgm:

let rollout controller  can be installed in runtime cluster 

Fixes #


wangyikewxgm:
for some reason have delete the origin repository. will submit another pr contains these changes.

PR #2077, opened 2021-08-12
fix template bug

leejanee:
`patchKey` used wrong in `configmap` trait definition


charts/vela-core/templates/defwithtemplate/configmap.yaml
```diff
@@ -15,8 +15,8 @@ spec:
       template: |
         patch: spec: template: spec: {
         	containers: [{
+                // +patchKey=name
```
hongchaodeng, line 18: indent

hongchaodeng (approved):
lgtm

PR #2080, opened 2021-08-12
fix configmap patchkey bug

Somefive:

`patchKey` used wrong in `configmap` trait definition
Same with [#2077](https://github.com/oam-dev/kubevela/pull/2077) 
close #2077

Fixes #


charts/vela-core/templates/defwithtemplate/annotations.yaml
```diff
@@ -1,4 +1,4 @@
-# Code generated by KubeVela templates. DO NOT EDIT.
+# Code generated by KubeVela templates. DO NOT EDIT. Please edit the original cue file.
```
wonderflow, line 1: 加个路径

wonderflow (approved):
cool

PR #2084, opened 2021-08-12
enable vela def to use import decl

### LINKED ISSUE #2068: vela def apply should support importing k8s package
**Describe the bug**

Now `vela def apply` command can apply an cue file into a k8s cr, but doesn't support importing k8s package yet.

**To Reproduce**

Let's say we import "kube/v1" at the beginning of the official `webservice` cue file. Then  run  `vela def apply vela-templates/definitions/internal/webservice.cue` from the root path.

```
import "kube/v1"

webservice: {
	...
}
...
```

 The output throws an arror:
```
Error: failed to parse CUE: package "kube/v1" not found (and 1 more errors)
```

**Expected behavior**

It should apply a cue file correctly into a definition resource.

**Screenshots**

**KubeVela Version**
Version: master
GitRevision: git-abe299f2
GolangVersion: go1.16.4

Somefive:

Previous `vela def` do not support internal package import such as `vela/op` or `kube/v1`. This PR enables it.

Fixes #2068

Notice that to support internal package use, users need to have connections with Kubernetes api-server, as the OpenAPI Schema is required. While no Kubernetes cluster is connected, the vela def will not support using internal package.

### DIFF
--- a/references/common/definition_test.go
+++ b/references/common/definition_test.go
@@ -49,6 +49,7 @@ func TestDefinitionBasicFunctions(t *testing.T) {
 		t.Fatalf("set type should failed due to invalid type, but got no error")
 	}
 	def.Object["spec"] = GetDefinitionDefaultSpec("TraitDefinition")
+	_ = unstructured.SetNestedField(def.Object, "patch: metadata: labels: \"KubeVela-test\": parameter.tag\nparameter: tag: string\n", "spec", "schematic", "cue", "template")
 	cueString, err := def.ToCUEString()
 	if err != nil {
 		t.Fatalf("unexpected error when getting to cue: %v", err)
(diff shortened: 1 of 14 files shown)


references/common/definition_test.go
```diff
@@ -59,23 +59,29 @@ func TestDefinitionBasicFunctions(t *testing.T) {
...
 	parts := strings.Split(cueString, "template: ")
-	if err = def.FromCUEString(parts[0]); err == nil {
+	if err = def.FromCUEString(parts[0], nil); err == nil {
 		t.Fatalf("should encounter no template found error but not found error")
 	}
-	if err = def.FromCUEString("import \"strconv\"\n" + cueString); err == nil {
+	if err = def.FromCUEString("template:"+parts[1], nil); err == nil {
+		t.Fatalf("should encounter no metadata found error but not found error")
+	}
+	if err = def.FromCUEString("import \"strconv\"\n"+cueString, nil); err == nil {
```
wonderflow, line 75: Just check the function not return error maybe is not enough? Can we have any way to check the parse/format result?

Somefive: Extra check is added.

wonderflow (approved):
LGTM

PR #2089, opened 2021-08-13
chore: replace github.com/ghodss/yaml with sigs.k8s.io/yaml

hezhizhen:

Since sigs.k8s.io/yaml is widely used in Kubernetes repositories and it's a permanent fork of github.com/ghodss/yaml, I think it's better to use sigs.k8s.io/yaml only.

Fixes #


hongchaodeng (approved):
LGTM.
Nice work. Thanks for your contribution

wonderflow (approved):
LGTM thanks

PR #2120, opened 2021-08-18
fix(CI): use buildkit as docker build engine to avoid 401 unauthorized

wonderflow:
ref to https://github.com/docker/buildx/issues/327

same case: https://github.com/jetstack/jetstack-secure/pull/227


zzxwill (approved):
/lgtm

PR #2126, opened 2021-08-19
fix external revision bug

wangyikewxgm:

fix bug: support  use context.revisionName when specify externalRevision

Fixes #


pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -391,6 +391,13 @@ func (h *AppHandler) handleComponentRevisionNameSpecified(ctx context.Context, c
 
 	comp.RevisionHash = cr.GetLabels()[oam.LabelComponentRevisionHash]
 	comp.RevisionName = revisionName
+
+	for _, trait := range comp.Traits {
+		if err := replaceComponentRevisionContext(trait, comp.RevisionName); err != nil {
+			return err
+		}
+	}
+
```
wonderflow, line 400: please add a test

wonderflow (approved):

Issue #2165, opened 2021-08-26
Feat[Application]: Make Application run as DAG workflow 


leejanee:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

TODO LIST:
- [x] Modify Application CRD, add `input\output`
- [x] Actively generate workflowSteps in `Appfile`
- [x] Add Builtin Workflow Step Definition to apply component
- [ ]  Make workflow support dag mode

PR #2171, opened 2021-08-26
Fix(workflow): fix apply component workflow step definition

FogDong:
fix apply component workflow step definition

Ref #2165

/cc @leejanee


FogDong:
Merged in #2236

PR #2174, opened 2021-08-27
[Feature] Support using code-generator to generate client

majian159:
- Add client generator code comment
- Add AddToScheme to register.go
- Sync generated client to kubevela-core-api

using code-generator to generate client

Fixes https://github.com/oam-dev/kubevela-core-api/issues/2


wonderflow (approved):
awesome, sorry for the late review

PR #2181, opened 2021-08-29
Fix: align all CUE template keyword to use parameter

wonderflow:

Now we have `parameter` and `parameters` used in CUE template, and the `parameters` keyword is introduced in https://github.com/oam-dev/kubevela/pull/2030 

In this PR, I aligned the keyword, we MUST use the same keyword and always keep the consistency. So it should have some break change if you use `context.parameters`, you should change to `context.parameter`.

I don't make the code compatible because I think it's not late for change, it's a small use case and only released for one week, we can fix it. If we don't change it, it will keep confusing everyone.

This PR also fix some places that misspell the `parameter`  by `parameters`. 


hongchaodeng (approved):
lgtm

pkg/appfile/appfile.go
```diff
@@ -23,6 +23,8 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/oam-dev/kubevela/pkg/cue/model"
+
 	"github.com/oam-dev/kubevela/pkg/workflow/providers"
 	"github.com/oam-dev/kubevela/pkg/workflow/providers/kube"
```
leejanee, line 30: please group import statements

Somefive (approved):

leejanee (approved):

PR #2184, opened 2021-08-29
Feat: add more registry traits as internal ones

wonderflow:
Fix and add more registry traits as internal ones


Somefive (approved):

hongchaodeng (approved):
lgtm

PR #2190, opened 2021-08-30
Feat(workflow): add depends-on-app workflow step definition

yangsoon:
This PR add `depends-on-app` workflow step definition, replace the Initializer's dependsOn ability

Fixes #


vela-templates/definitions/internal/depends-on-app.cue
```diff
@@ -0,0 +1,64 @@
...
+					dependsOn.value
+				}
+			}
+		}
+	}
+
+	status: load.apply.value.status
+
+	wait: op.#ConditionalWait & {
+		continue: status.status == "running"
```
hongchaodeng, line 55: Why is `status.status`.... Shouldn't it rename to `phase`?

hongchaodeng (approved):
lgtm

Left one question

vela-templates/definitions/internal/depends-on-app.cue
yangsoon: good suggestion, I will fix it

wonderflow (approved):

Somefive (approved):

PR #2195, opened 2021-08-30
Fix: error doc link in README

arcosx:

old doc link 404, fix it.
image

Fixes #


wonderflow (approved):
good catch, thanks

Issue #2208, opened 2021-08-31
vela show service-binding can't display information inside map


wonderflow:
```
$ vela show service-binding
# Properties
+-------------+------------------------------------------------+------------------+----------+---------+
|    NAME     |                  DESCRIPTION                   |       TYPE       | REQUIRED | DEFAULT |
+-------------+------------------------------------------------+------------------+----------+---------+
| envMappings | The mapping of environment variables to secret | map[string]{...} | true     |         |
+-------------+------------------------------------------------+------------------+----------+---------+
```

The expected output is the type.

PR #2209, opened 2021-08-31
Fix: modify service-binding with more accurate type

wonderflow:
as title said, please note it was not fixing #2208 , but only make it more clear.


hongchaodeng (approved):
lgtm

Somefive (approved):

PR #2222, opened 2021-09-01
Fix(webhook): only check the uniqueness of component's name under the same namespace

yangsoon:
…same namespace

fix the bug, only check the uniqueness of component's name under the same namespace

Fixes #


wonderflow (approved):
good catch ,thanks

PR #2226, opened 2021-09-02
Fix(addons): remove internal krusie addon

yangsoon:
1. we only keep terrafrom and fluxcd as internal addons (In order to ensure the installation order, we use terrform's dependency ability to install fluxcd first)
2. add helm hook `post-install`, `post-upgrade`, `pre-delete` for addons

Fixes #


vela-templates/gen_addons.go
```diff
@@ -66,7 +66,7 @@ const (
 )
 
 // DefaultEnableAddons is default enabled addons
-var DefaultEnableAddons = []string{"fluxcd", "kruise", "terraform"}
+var DefaultEnableAddons = []string{}
```
hongchaodeng, line 69: Do we still need this feature? Why not remove the relevant code as well?

wonderflow: Yes, we need it.

Somefive (approved):

wonderflow (approved):

PR #2264, opened 2021-09-10
Fix:delete useless rollouttrait

### LINKED ISSUE #2263: [Chore] deprecate rollouttrait and remove the related code from vela-core
We have rollout trait and rollouttraits, we should remove the `rollouttraits`.

wangyikewxgm:
fixes #2263


hongchaodeng (approved):
lgtm

PR #2269, opened 2021-09-11
remove podspecworkload controller and CRD

### LINKED ISSUE #2258: [Chore] remove podspecworkload out from vela-core
podspecworkload is not the core controller of KubeVela for a long time, and we have already make a standalone chart about it. https://github.com/oam-dev/catalog/tree/master/workloads/podspecworkload/charts/podspecworkload

As a result, we should remove it from vela core, so we can reduce the CRD installed by default.

Things that possible to be removed:

* podspecworkload CRD
* podspecworkload controller
* related test code/configurations

wonderflow:
fix #2258


hongchaodeng (approved):
lgtm

wangyikewxgm (approved):
LGTM

PR #2270, opened 2021-09-11
Fix: remove appcontext CRD and controller

### LINKED ISSUE #2259: [Chore] Remove ApplicationContext CRD and it's Controller
It's a temporary solution for KubeVela using AppConfig, but it's not used in kubevela now and it's all compatible. So we should delete the legacy code.

wonderflow:
fix #2259


hongchaodeng (approved):
lgtm

PR #2271, opened 2021-09-11
Add more options for leader election configuration to avoid pressure …

### LINKED ISSUE #2266: [Feature] Add more options for leader election configuration to avoid pressure on kube-apiserver
Now we have a bootstrap parameter to swith on-off leader-election, that is not enough. We should also make these leader-election configuration configurable:

- --leader-election-resource-lock string     default value："configmapsleases"

```
       // LeaderElectionResourceLock determines which resource lock to use for leader election,
	// defaults to "configmapsleases". Change this value only if you know what you are doing.
	// Otherwise, users of your controller might end up with multiple running instances that
	// each acquired leadership through different resource locks during upgrades and thus
	// act on the same resources concurrently.
	// If you want to migrate to the "leases" resource lock, you might do so by migrating to the
	// respective multilock first ("configmapsleases" or "endpointsleases"), which will acquire a
	// leader lock on both resources. After all your users have migrated to the multilock, you can
	// go ahead and migrate to "leases". Please also keep in mind, that users might skip versions
	// of your controller.
	//
	// Note: before controller-runtime version v0.7, the resource lock was set to "configmaps".
	// Please keep this in mind, when planning a proper migration path for your controller.
	LeaderElectionResourceLock string
```

- --leader-election-lease-duration duration     default: 15s

```
	// LeaseDuration is the duration that non-leader candidates will
	// wait to force acquire leadership. This is measured against time of
	// last observed ack. Default is 15 seconds.
	LeaseDuration *time.Duration
```

- --leader-election-renew-deadline duration      default: 10s

```
// RenewDeadline is the duration that the acting controlplane will retry
	// refreshing leadership before giving up. Default is 10 seconds.
	RenewDeadline *time.Duration
```

- --leader-elect-retry-period duration      default: 2s
```
	// RetryPeriod is the duration the LeaderElector clients should wait
	// between tries of actions. Default is 2 seconds.
	RetryPeriod *time.Duration
```

LiuYuHui:
Fixes #2266

### DIFF
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -75,6 +75,10 @@ func main() {
 	var qps float64
 	var burst int
 	var pprofAddr string
+	var leaderElectionResourceLock string
+	var leaseDuration time.Duration
+	var renewDeadline time.Duration
+	var retryPeriod time.Duration
 
 	flag.BoolVar(&useWebhook, "use-webhook", false, "Enable Admission Webhook")
 	flag.StringVar(&certDir, "webhook-cert-dir", "/k8s-webhook-server/serving-certs", "Admission webhook cert/key dir.")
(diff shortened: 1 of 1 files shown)


cmd/core/main.go
```diff
@@ -112,6 +116,13 @@ func main() {
 	flag.StringVar(&controllerArgs.OAMSpecVer, "oam-spec-ver", "v0.3", "oam-spec-ver is the oam spec version controller want to setup, available options: v0.2, v0.3, all")
 	flag.StringVar(&pprofAddr, "pprof-addr", "", "The address for pprof to use while exporting profiling results. The default value is empty which means do not expose it. Set it to address like :6666 to expose it.")
 	flag.BoolVar(&commonconfig.PerfEnabled, "perf-enabled", false, "Enable performance logging for controllers, disabled by default.")
+	flag.StringVar(&leaderElectionResourceLock, "leader-election-resource-lock", "configmapsleases", "The resource lock to use for leader election")
+	flag.DurationVar(&leaseDuration, "leader-election-lease-duration", 15*time.Second,
+		"The duration that non-leader candidates will wait to force acquire leadership")
+	flag.DurationVar(&renewDeadline, "leader-election-renew-deadline", 10*time.Second,
+		"The duration that the acting controlplane will retry refreshing leadership before giving up")
+	flag.DurationVar(&retryPeriod, "leader-elect-retry-period", 2*time.Second,
```
wonderflow, line 124:
```suggestion
	flag.DurationVar(&retryPeriod, "leader-election-retry-period", 2*time.Second,
```

wonderflow (approved):
lgtm after nit, thanks for your contributions

PR #2276, opened 2021-09-12
Feat: istio addon and use case demo

wangyikewxgm:
Feat: istio addon and canary rollout use case

** must block until pr:https://github.com/oam-dev/kubevela/pull/2274 merged **

Fixes #


docs/examples/canary-rollout-use-case/rollout-v2.yaml
```diff
@@ -0,0 +1,112 @@
...
+  name: book-info
+  namespace: default
+spec:
+  components:
+    - name: reviews
+      type: webservice
+      properties:
+        image: docker.io/istio/examples-bookinfo-reviews-v2:1.16.2
+        port: 9080
+        volumes:
```
hongchaodeng, line 13: What are the volumes used for?

hongchaodeng (approved):
LGTM!
Would be better to provide a README.

docs/examples/canary-rollout-use-case/rollout-v2.yaml
wangyikewxgm: storage the log, the container will `crash` if without this volume

wangyikewxgm:
> LGTM!
> Would be better to provide a README.

Will add a README doc

wonderflow (approved):

PR #2280, opened 2021-09-13
Fix (CUE): patch any key using `retainKeys` strategy

hongchaodeng:
Fix https://github.com/oam-dev/kubevela/issues/2144

This will enable users to patch any keys, not just list. For example, patch metadata.name, or even kind.

@tatianguiqu @kinsolee


kinsolee:
This will be a great fix.  

We can write cleaner codes in definition files now. 

I have been suffering from cumbersome codes like [this](https://github.com/oam-dev/kubevela/issues/2144#issuecomment-906190059) for a long time. You will save me a lot of time.

Thanks for your contribution!

leejanee (approved):
/lgtm

PR #2293, opened 2021-09-14
Fix: fix alibaba cloud rds module

zzxwill:
Enable users to set `db_instance_storage_type`. This can fix the issue
happended since Sept. 14, 2021.
{"Code":"InvalidNetworkTypeClassicWhenCloudStorage","HostId":
"rds.aliyuncs.com","Message":"The Specified InstanceNetworkType
value Classic is not valid when choose cloud storage type."

Fixes #

### DIFF
--- a/charts/vela-core/templates/definitions/terraform-alibaba-rds.yaml
+++ b/charts/vela-core/templates/definitions/terraform-alibaba-rds.yaml
@@ -16,10 +16,11 @@ spec:
     terraform:
       configuration: |
         module "rds" {
-          source = "terraform-alicloud-modules/rds/alicloud"
+          source = "github.com/zzxwill/terraform-alicloud-rds"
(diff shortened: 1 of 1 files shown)


charts/vela-core/templates/definitions/terraform-alibaba-rds.yaml
```diff
@@ -16,10 +16,11 @@ spec:
     terraform:
       configuration: |
         module "rds" {
-          source = "terraform-alicloud-modules/rds/alicloud"
+          source = "github.com/zzxwill/terraform-alicloud-rds"
```
zzxwill, line 19: @wonderflow I will revert this PR once https://github.com/terraform-alicloud-modules/terraform-alicloud-rds/pull/23 is merged.

wonderflow (approved):

PR #2311, opened 2021-09-15
feat(workflow): add op.#Patch to allow patch any object within workflow

kinsolee:

This allows we patch components after op.#LoadComponent, so that we could make changes to the component's properties before applying it.


kinsolee:
Duplicated with https://github.com/oam-dev/kubevela/pull/2304

PR #2318, opened 2021-09-17
Fix: fix cluster-gateway image tag in chart

Somefive:

Previously, using `latest` tag for cluster-gateway will be automatically set to the tag of chart when release pipeline is triggered by [Github Action](https://github.com/oam-dev/kubevela/blob/8407c0bf4d01db70f580910c03c7792c2dd6798c/.github/workflows/release.yml#L29). This binds the KubeVela version to cluster-gateway.

Currently, as a quick fix, we retag the cluster-gateway:latest to cluster-gateway:v1.1.1. But in long term, they should be decoupled.  

Fixes #

Related to https://github.com/oam-dev/cluster-gateway/issues/4

### DIFF
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -117,7 +117,7 @@ multicluster:
     port: 9443
     image:
       repository: oamdev/cluster-gateway
-      tag: latest
+      tag: v1.1.1
       pullPolicy: Always
     resource:
       limits:


hongchaodeng (approved):

hongchaodeng:
LGTM
Let me rerun the e2e rollout test... It is still flaky.

wonderflow (approved):

PR #2328, opened 2021-09-18
Fix(helm): Do not install fluxcd and terraform by default

yangsoon:

Fixes #


vela-templates/addons/terraform/definitions/terraform-alibaba-rds.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: core.oam.dev/v1alpha2
 kind: ComponentDefinition
 metadata:
   name: alibaba-rds
-  namespace: {{.Values.systemDefinitionNamespace}}
+  namespace: vela-system
```
wonderflow, line 5: 这个一定要固定namespace吗

wonderflow (approved):

PR #2332, opened 2021-09-20
Fix: add comments and adjust helm typed component's spec

zzxwill:
Co-authored-by: Hongchao Deng <hongchaodeng1@gmail.com>

Fixes #

### DIFF
--- a/charts/vela-core/templates/addons/observability.yaml
+++ b/charts/vela-core/templates/addons/observability.yaml
@@ -115,14 +115,6 @@ data:
                   tag: v2.1.0
               version: 3.4.1
             type: helm
-          - name: kube-prometheus-stack
-            properties:
-              chart: kube-prometheus-stack
-              repoType: helm
-              targetNamespace: observability
-              url: https://charts.kubevela.net/addons
-              version: 18.0.0
-            type: helm
         status:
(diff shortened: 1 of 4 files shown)


charts/vela-core/templates/addons/observability.yaml
```diff
@@ -115,14 +115,6 @@ data:
                   tag: v2.1.0
               version: 3.4.1
             type: helm
-          - name: kube-prometheus-stack
```
zzxwill, line 118: After more verification, this component is not necessary.

hongchaodeng (approved):

docs/examples/obervability/initializer-observability.yaml
```diff
@@ -54,7 +54,8 @@ spec:
           properties:
             chart: grafana
             version: 6.14.1
-            type: helm
+            repoType: helm
```
zzxwill, line 57: Latest `helm` ComponentDefinition uses the spec filed `repoType` instead of `type` https://github.com/oam-dev/kubevela/blob/master/vela-templates/addons/fluxcd/definitions/helm-release.yaml#L109.

Somefive (approved):

PR #2334, opened 2021-09-22
Make ignore tag work when generating openapi json schemas for form rendering

kinsolee:

We should remove the fields with ignore tag in openapi json schemas. This PR is to filter them out.

Fixes #


wonderflow:
can you give a real case that we should ignore

kinsolee:
> can you give a real case that we should ignore
There are some parameters we'd like to fill in by scripts automatically throughout the whole CICD lifecycle, they should be hidden from users when it comes to UI form rendering.

leejanee:
@kinsolee thanks for your contribution.

But , generally the whole app is the `ui` for end users,  so there are no parameters that should  be hidden from users.

I think your appeal can be classified as `vela api-server`. Vela api-server will provide an http interface closer to the business scenario.

kinsolee:
Hi @leejanee  

According to the [official document](https://kubevela.io/docs/platform-engineers/openapi-v3-json-schema/), schema based on parameters is stored in a ConfigMap and we can render it into a form. 

I am not sure if there are other functionalities(e.g. webhook validation) use this configmap. But it seems to be the only place we can get `openapi schema` to render forms.

Now my team has developed an UI dashboard based on it and we'd like to hide some parameters from platform users. 

Can we take it as a workaround before `vela api-server` provides further features?

wonderflow (approved):
I'm not opinionated on this feature, it's fine to approve if your use case real need it. 

ping @leejanee @hongchaodeng

leejanee:
@kinsolee parameter json schema is not just for end users or platform users. Avoid to make special treatment. The solution I suggest is that judge whether to display on your web server according to the field description.

kinsolee:
@leejanee  Thanks to your reply. Now that `openapi schema` has other usages. I will filter the fields outside configmap.

PR #2343, opened 2021-09-23
Feat: add revisionHistoryLimit to helm chart

### LINKED ISSUE #2342: [Feature] Set revisionHistoryLimit in helm chart for Certificates

If I try to install the helm chart for vela-core in a cluster which runs OPA with a policy that requires users to explicit set revisionHistoryLimit on the Certificate object, the helm install fails.

I would like to be able to specify a value for the revisionHistoryLimit in the values file I supply to the helm command.

Otherwise I will need to unpack the chart and manually edit the templates.

**Context**

The reason our cluster enforce revisionHistoryLimit is because if its not specified, the cluster is polluted with revisions since we deploy pods quite frequently and certificates ttl is about a week.

haugom:

When using cert-manager certificates it is good practice to set revisionHistoryLimit. Today you have to unpack the chart and edit the template.

It should be possible to set the value.

Fixes #2342

### DIFF
--- a/charts/oam-runtime/templates/certmanager.yaml
+++ b/charts/oam-runtime/templates/certmanager.yaml
@@ -18,6 +18,7 @@ metadata:
 spec:
   secretName: {{ template "oam-runtime.fullname" . }}-root-cert
   duration: 43800h # 5y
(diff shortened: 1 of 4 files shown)


wonderflow (approved):
thanks

PR #2355, opened 2021-09-26
Fix: revision GC in workflow mode

### LINKED ISSUE #2316: [BUG] Inconsistent workload between  garbageCollection and applyComponent
**Describe the bug**

Thanks to https://github.com/oam-dev/kubevela/pull/2304, now we could rewrite component's properties in the workflow. 

But application_controller.garbageCollection may fail if we rewrite component's properties in `workflowstepdefinition`.  

Failure will happen if we try to refer fields from `context.parameter` in `traitdefinition` but no such fields in component's properties.

**To Reproduce**

Let's say we have a trait to output a cronjob. Here we'd like to add a `version` label passed from `context.parameter.version`.
```
outputs: cronjob: {
    apiVersion: "batch/v1beta1"
    kind:       "cronjob"
    metadata: {
        namespace: context.namespace
        labels: {
	        app: "\(context.name)-cronjob"
                // Read from component's parameter
                version: "\(context.parameter.version)"
        }
    }
...
```
And we definitely declare the `version` field in component's parameter.
```
parameter: {
    cmd?: [...string]

    // +ignore
    version: *"" | string
}
```
When we go through the workflowstep below, it works fine because we declare the version field.
```
deploy: op.#Steps & {
    load: op.#Load @step(1)
    apply: op.#Steps & {
	    for compName, comp in load.value {
		    applyComp: op.#ApplyComponent & {
			    value: comp & {
				    properties: {
                                         version: "v1.1.1"
				    }
			    }
		    }
	    }
    } @step(2)
}
```

But vela-core will output error log likes below:
```
E0917 15:13:49.770521   29433 instance.go:81] "failed to have the workload/trait unstructured" err="metadata.labels.version: reference \"context\" not found"
```

Both `revision.cleanUpComponentRevision` and `application.AppHandler.applyComponentFunc` call `appfile.Appfile.GenerateComponentManifest` to generate ComponentManifest. 

But the `Appfile.Workloads` among them are different as they read from different source. 

One is from Applicaiton.Spec.Components https://github.com/oam-dev/kubevela/blob/8407c0bf4d01db70f580910c03c7792c2dd6798c/pkg/appfile/parser.go#L204,
another is from workflowstep's component with rewritten properties.
https://github.com/oam-dev/kubevela/blob/8407c0bf4d01db70f580910c03c7792c2dd6798c/pkg/workflow/providers/oam/apply.go#L57

**Expected behavior**
As we might apply multiple times from workflowstep. We should persist the written component properties into apprevision for consistency. 

**Screenshots**

**KubeVela Version**

Version: master
GitRevision: git-941c599a
GolangVersion: go1.17

**Cluster information**

**Additional context**

FogDong:
Fix revision GC in workflow mode

Fixes #2316

/cc @leejanee

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/revision.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/revision.go
@@ -19,6 +19,7 @@ package application
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"sort"
 	"strings"
@@ -33,6 +34,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	ktypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -816,6 +818,66 @@ func (h historiesByRevision) Less(i, j int) bool {
 }
 
 func cleanUpComponentRevision(ctx context.Context, h *AppHandler) error {
+	if appWillRollout(h.app) {
+		return cleanUpRollOutComponentRevision(ctx, h)
+	}
+	return cleanUpWorkflowComponentRevision(ctx, h)
+}
+
+func cleanUpWorkflowComponentRevision(ctx context.Context, h *AppHandler) error {
```
wonderflow, line 827: please add tests

wonderflow (approved):

leejanee (approved):

PR #2368, opened 2021-09-27
Feat: bootstrap multicluster testing

Somefive:

1. Bootstrap new CI action to run e2e test for multi-cluster.
2. Multi-cluster e2e test will start 2 kind clusters while running. One cluster (Hub) will install KubeVela as normal. The other one will be an empty cluster.
3. Multi-cluster e2e test use simplified setup process which only install vela-core without oam-runtime or any addon (like flux or terraform).
4. The e2e test coverage report will be uploaded so that multi-cluster related Vela CLI test could be implemented and covered.

Fixes #

### DIFF
--- a/.github/workflows/e2e-multicluster-test.yml
+++ b/.github/workflows/e2e-multicluster-test.yml
@@ -1,0 +1,102 @@
+name: E2E MultiCluster Test
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+  workflow_dispatch: {}
+  pull_request:
+    branches:
+      - master
+      - release-*
+
+env:
+  # Common versions
+  GO_VERSION: '1.16'
+  GOLANGCI_VERSION: 'v1.38'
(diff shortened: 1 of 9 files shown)


.github/workflows/e2e-multicluster-test.yml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Detect No-op Changes
+        id: noop
+        uses: fkirc/skip-duplicate-actions@v3.3.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          paths_ignore: '["**.md", "**.mdx", "**.png", "**.jpg"]'
+          do_not_skip: '["workflow_dispatch", "schedule", "push"]'
+          concurrent_skipping: false
+
+  e2e-tests:
```
wonderflow, line 36:
```suggestion
  e2e-multi-cluster-tests:
```

wonderflow (approved):

PR #2382, opened 2021-09-28
Feat: use #ApplyComponent for EnvBinding

Somefive:

Currently, EnvBinding policy is implemented in the following procedure:
1. [ApplicationController] Generate a EnvBinding CR in k8s.
2. [EnvBindingController] Calculate target clusters for Envs.
3. [EnvBindingController] Apply patch and component selector for application in different Envs.
4. [EnvBindingController] Render patched applications into resources and save in configmap.
5. [ApplicationController] Run workflow to read configmap and apply resources.

The above procedure, although works well for applications in multiple clusters, is not well aligned with the implementation in ApplicationController which leads to the following problems:
1. The workflow DAG mode cannot be used with EnvBinding.
2. The component revision is not handled in EnvBinding controller, the workload generation is not skipped for managed workloads and therefore it cannot be used together with Rollout.
3. Extra configmap is generated for internal use, which is not an elegant implementation.

It would be better if the EnvBindingController could do its job in ApplicationController. As workflow is capable of handling these things, we should move the implementation for EnvBinding into WorkflowSteps.

This PR detachs the Step 4 described above from EnvBindingController. Instead, it is now implemented with the WorkflowStep `#ApplyComponent`. In this way, we can solve Problem 2 and pave the way for Problem 1 & 3.

Fixes #

* The GC problem for ComponentRevision in sub-cluster is not handled now. It will be addressed together with Rollout.


Somefive:
cc @FogDong @wangyikewxgm @leejanee

cmd/core/main.go
```diff
@@ -30,6 +30,7 @@ import (
 	"strings"
 	"time"
 
+	v1 "k8s.io/api/apps/v1"
```
hongchaodeng, line 33: rename to appsv1?

Somefive: ok

PR #2389, opened 2021-09-29
Feat: add vela minimal in make manifests

FogDong:
 add vela minimal in make manifests

Fixes #

/cc @wonderflow


vela-templates/gen_definitions.sh
```diff
@@ -54,10 +55,28 @@ function render {
...
+  retVal=$?
+  if [ $retVal -ne 0 ]; then
+    echo -ne "${RED}Failed. Exit code: ${retVal}.${NC}\n"
+    exit $retVal
+  fi
+
+  rm -f $outputDir/env-binding.yaml
+  rm -f $outputDir/deploy2env.yaml  
+}
+
```
Somefive, line 74: This part is duplicated. Maybe we do not need to generate MINIMAL definitions but only need to copy them from INTERNAL.

Somefive (changes requested):

Somefive (approved):

wonderflow (approved):

PR #2585, opened 2021-10-29
Feat: allow import package in custom status cue template

Chwetion:
Fixes #2578

PR #2596, opened 2021-11-01
Feat: add create cloud cluster interface

Somefive:

Add create cloud cluster API and support ACK.
+ Create ACK Cluster
+ Get creation status
+ List all creation
+ Cancel creation

I have:

This part of API does not have unit-test since creating cluster is heavy and is not suitable for being used in test. It has been tested manually.

**Implementation**:

Since currently the terraform part (terraform-controller:v0.2.6) do not support multi-tenant natively, the implementation of these APIs are directly working with underlying Terraform Configuration CRDs.

To create an ACK cluster, the creation API executes the following steps:
1. APIServer creates a secret containing AK/SK for providers to use.
2. APIServer creates a Terraform Provider to use the AK/SK.
3. APIServer creates a Terraform Configuration which leverages the provider established in step 2.
4. Terraform Controller will observe the Terraform Configuration and create a job to apply the resource through Terraform.

Step 1 & 2 uses the hash of [AK;SK] as its key, so that later requests with same AK/SK can reuse it.
Configuration created by Step 3 can be manipulated through GET/LIST/DELETE APIs. The DELETE API will trigger the deletion process of the target ACK cluster.

Several things to notice which are not addressed in this PR:
1. There is no AK/SK GC.
2. There is no paging for the list API.
3. The creation process could hang due to invalid resource request, which in return will block the DELETE as well. See #2598.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,10 +7,10 @@ require (
 	github.com/AlecAivazis/survey/v2 v2.1.1
 	github.com/Masterminds/sprig v2.22.0+incompatible
 	github.com/Netflix/go-expect v0.0.0-20180615182759-c93bf25de8e8
+	github.com/agiledragon/gomonkey/v2 v2.3.0
 	github.com/alibabacloud-go/cs-20151215/v2 v2.4.5
 	github.com/alibabacloud-go/darabonba-openapi v0.1.4
 	github.com/alibabacloud-go/tea v1.1.15
-	github.com/agiledragon/gomonkey/v2 v2.3.0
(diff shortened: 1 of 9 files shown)


barnettZQG:
No dependencies on the Application?

Somefive:
> No dependencies on the Application?

@barnettZQG Yes. Decoupled with Application since current `alibaba-ack` trait definition cannot support the multi-user scenario here, so using native solution. See #2585 for more details.

barnettZQG (approved):

PR #2629, opened 2021-11-04
Fix: patch status retry while conflict happens

Somefive:

Fix patchStatus in application controller. Sometimes, users may modify or delete application during reconcile, which will lead to patchStatus failure. However, our controller relies on the status to work properly. For example, if the workflow is deploying resources and user delete the app at this time, the controller would not be able to add newly created resources in status and the gc will skip them.

I have:

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
@@ -23,6 +23,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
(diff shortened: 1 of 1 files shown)


pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -353,7 +354,19 @@ func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.
 func (r *Reconciler) patchStatus(ctx context.Context, app *v1beta1.Application, phase common.ApplicationPhase) error {
```
wonderflow, line 354: I think we should write another function to mark as `patchStatusRetry`.

Not all patchStatus should retry. I'm very concern about this as it will case conflicts and inconsistent.

Somefive: The change here will only retry when conflict happens and it has a default maximal retry times as 5.
I agree that maybe not all patchStatus should retry. How about enforce all patchStatus related to workflow to retryOnConflict? /cc @leejanee @wonderflow

wonderflow: Yes, workflow related patch can use this , the rest should keep as it is. We should avoid risks.

leejanee (approved):
lgtm

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -342,18 +343,48 @@ func (r *Reconciler) handleFinalizers(ctx context.Context, app *v1beta1.Applicat
...
 		return ctrl.Result{}, errors.WithMessage(err, "cannot update application status")
 	}
 	return ctrl.Result{}, fmt.Errorf("object level reconcile error, type: %q, msg: %q", string(condition.Type), condition.Message)
 }
 
+func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.Application, condition condition.Condition, phase common.ApplicationPhase) (ctrl.Result, error) {
+	return r._endWithNegativeCondition(ctx, app, condition, phase, false)
+}
+
+func (r *Reconciler) endWithNegativeConditionWithRetry(ctx context.Context, app *v1beta1.Application, condition condition.Condition, phase common.ApplicationPhase) (ctrl.Result, error) {
```
wonderflow, line 362: please add a note here:

```
// Note: Only operations that must override the status should use this function, it only contains workflow operations by now.
```

wonderflow (approved):

PR #2675, opened 2021-11-10
Fix: cluster url automatically patch port and scheme

Somefive:

Using `vela cluster join` now only support kubeconfig with full apiserver url such as https://127.0.0.1:6443. It must include `scheme`, `host`, `port` three parts to work. This PR allows user only specify `host` and the added function will automatically add `scheme` and `port` to it.

I have:

### DIFF
--- a/pkg/utils/url.go
+++ b/pkg/utils/url.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2021 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


pkg/utils/url.go
```diff
@@ -0,0 +1,59 @@
...
+		return "", fmt.Errorf("invalid endpoint url: %s", server)
+	}
+	var scheme, port, host string
+	results := r.FindStringSubmatch(server)
+	for i, name := range r.SubexpNames() {
+		if name == "scheme" {
+			scheme = results[i]
+		}
+		if name == "host" {
+			host = results[i]
```
zzxwill, line 38: Is better to convert three `if` to a switch control?

zzxwill (approved):

PR #2688, opened 2021-11-11
Feat: change model name

barnettZQG:


pkg/apiserver/rest/apis/v1/types.go
```diff
@@ -638,15 +636,41 @@ type ApplicationDeployResponse struct {
...
+// UpdateApplicationTrait update application trait req
+type UpdateApplicationTrait struct {
+	Alias       *string `json:"alias,omitempty" validate:"checkalias" optional:"true"`
+	Description *string `json:"description,omitempty" optional:"true"`
+	Properties  *string `json:"properties"`
+}
+
+// ApplicationTrait application trait
+type ApplicationTrait struct {
+	Name        string `json:"name"`
```
wangyikewxgm, line 670: trait have no `name` filed, only `type` filed is enough.

PR #2719, opened 2021-11-15
Fix: Align -n command argument

chivalryq:

Align commands to use `-n` means namespace

Fixes #

I have:

### DIFF
--- a/references/appfile/api/appfile.go
+++ b/references/appfile/api/appfile.go
@@ -31,7 +31,6 @@ import (
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/common"
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha2"
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
-	"github.com/oam-dev/kubevela/apis/types"
 	"github.com/oam-dev/kubevela/pkg/builtin"
 	"github.com/oam-dev/kubevela/pkg/oam"
 	cmdutil "github.com/oam-dev/kubevela/pkg/utils/util"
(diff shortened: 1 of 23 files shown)


Somefive (approved):
lgtm

references/cli/def.go
```diff
@@ -47,22 +47,6 @@ import (
...
-	// FlagOutput command flag to specify which file to save
-	FlagOutput = "output"
-	// FlagMessage command flag to specify which file to save
-	FlagMessage = "message"
-	// FlagType command flag to specify which definition type to use
-	FlagType = "type"
-	// FlagNamespace command flag to specify which namespace to use
-	FlagNamespace = "namespace"
-	// FlagInteractive command flag to specify the use of interactive process
-	FlagInteractive = "interactive"
```
wonderflow, line 65: why they're deleted?

references/appfile/api/appfile.go
```diff
@@ -142,7 +141,7 @@ func (app *AppFile) ExecuteAppfileTasks(io cmdutil.IOStreams) error {
 }
 
 // BuildOAMApplication renders Appfile into Application, Scopes and other K8s Resources.
-func (app *AppFile) BuildOAMApplication(env *types.EnvMeta, io cmdutil.IOStreams, tm template.Manager, silence bool) (*v1beta1.Application, []oam.Object, error) {
```
wonderflow, line 145: we should keep the env concept and add `namespace` args as override. Keep the compatibility

chivalryq: I think we shouldn't keep env concept any more. There are three reasons 
1. It's originally for create namespace. If we keep 2 concept, users will be confused.
2. As we all focus on release 1.2 now. There are already a lot concepts about envrionment. We shouldn't keep this as another environment.
3. It's easy to swtch from -e to -n.

wonderflow: 如果要删就要整体都删，不能半吊子

chivalryq: 我遗留的部分是exec/port-forward/logs，在 @Somefive 的工作流程中，他合并后我会继续把这个概念清理干净

wonderflow: 那先把 vela env 命令清理掉

wonderflow (changes requested):
we should keep the compatibility with `vela env` until we have alternative plan

wonderflow (approved):
good job

PR #2734, opened 2021-11-17
Feat: envbinding support cloud resource deploy and share

Somefive:

Support CloudResource deploy and bind in env-binding. 
- **deploy-cloud-resource**: works as an alternative to **deploy2env**. **deploy2env** applies component into child clusters while **deploy-cloud-resource** apply terraform component in the manage cluster and only dispatch cloud resource connection secret (such as RDS) into child clusters.
- **share-cloud-resource**: after deploying cloud resources, this WorkflowStep can help propagate connection secrets to other clusters, so that other clusters can share the same cloud resource. 

A detailed example is enclosed [here](https://github.com/oam-dev/kubevela/blob/99dffe1aa94938995e675bab37c95a3831428b1b/docs/examples/envbinding/deploy-and-bind-cloud-resource.yaml). In this case, the application will create two databases. One for the Hangzhou cluster to use and the other for the HongKong cluster. Additionally, the first one will also be shared to Beijing cluster and the second one will be shared across different namespaces in the HongKong cluster.   

I have:

Known issue: The update of Cloud Resource will not automatically update the secrets in sub-clusters. This will be implemented with further works.


wonderflow (approved):
LGTM after tests added

charts/vela-minimal/templates/defwithtemplate/bind-cloud-resource.yaml
```diff
@@ -0,0 +1,39 @@
+# Code generated by KubeVela templates. DO NOT EDIT. Please edit the original cue file.
+# Definition source cue file: vela-templates/definitions/internal/bind-cloud-resource.cue
+apiVersion: core.oam.dev/v1beta1
+kind: WorkflowStepDefinition
+metadata:
+  annotations:
+    definition.oam.dev/description: Sync secrets created by terraform component to runtime clusters
+  name: bind-cloud-resource
```
zzxwill, line 8: In cloud resource `Provision and Binding`, `bind` means consuming/using the cloud resources by upper business application.  So the step name sounds a little bit confusing. The original `sync` is great, I think.

hongchaodeng (approved):

docs/examples/envbinding/deploy-and-bind-cloud-resource.yaml
```diff
@@ -0,0 +1,78 @@
...
+                    # region: hongkong
+                    instance_name: hongkong_db
+                    writeConnectionSecretToRef:
+                      name: hk-project-rds-credential
+
+  workflow:
+    steps:
+      # 部署 RDS 给杭州区用
+      - name: deploy-hangzhou-rds
+        type: deploy-cloud-resource
```
zzxwill, line 54: In our last discussion, `deploy-cloud-resource` should not exist. Without specifying a cluster, `deploy2env` means deploying the workload in the managed cluster. Right?

Somefive: Right. But that is **deploy2env**, not **deploy-cloud-resource**, they are different things. **deploy-cloud-resource** only deploy terraform configuration in manage cluster. Only the secret will be dispatched to child clusters.

Somefive: If you want to dispatch terraform configuration into child clusters, you can switch **deploy-cloud-resource** to **deploy2env**.

charts/vela-minimal/templates/defwithtemplate/bind-cloud-resource.yaml
Somefive: I agree that `bind` is not a perfect name here but the `sync-cloud-resource` is also not very appropriate I think. How about `share-cloud-resource`?

zzxwill: +1

docs/examples/envbinding/deploy-and-bind-cloud-resource.yaml
zzxwill: image

It would be simple and clear to end-users if we invent fewer idioms. In this decision we made above, we can still use `deploy2env` without setting parameter `env`, which means not to dispatch Configuration to child clusters.

```  workflow:
    steps:
      - name: deploy-test-server
        type: deploy2env
        properties:
          policy: env
          env: test
```

A Terraform application is special, but we are aiming to treat it as a normal application as possible.

wonderflow: `deploy-cloud-resource` is correct for me. They have different logic. `deploy-cloud-resource` only deploy the secret into child cluster while `deploy2env` will deploy the whole cloud resource object.

FogDong (approved):

PR #2739, opened 2021-11-18
Feat: EnableAddon supports runtime cluster

hongchaodeng:
If use runtime cluster mode, the definitions will be applied to control plane k8s directly, not included in the Application object.

### DIFF
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -4,12 +4,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	v1 "k8s.io/api/core/v1"
 	"net/url"
 	"path"
 	"path/filepath"
 	"strings"
 	"time"
+
+	v1 "k8s.io/api/core/v1"
 
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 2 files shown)


pkg/addon/addon.go
```diff
@@ -362,26 +363,60 @@ func RenderApplication(addon *types.Addon, args map[string]string) (*v1beta1.App
...
+}
+
+func isDeployToRuntimeOnly(addon *types.Addon) bool {
+	if addon.DeployTo == nil {
+		return false
+	}
+	if !addon.DeployTo.RuntimeCluster {
+		return false
+	}
+	return true
```
chivalryq, line 409: just `return addon.DeployTo.RuntimeCluster`

pkg/addon/addon.go
```diff
@@ -362,26 +363,60 @@ func RenderApplication(addon *types.Addon, args map[string]string) (*v1beta1.App
...
+			app.Spec.Components = append(app.Spec.Components, *comp)
 		}
-		app.Spec.Components = append(app.Spec.Components, *comp)
 	}
 
-	return app, nil
+	return app, defObjs, nil
+}
+
+func isDeployToRuntimeOnly(addon *types.Addon) bool {
```
chivalryq, line 402: Why is `isDeployToRuntimeOnly` , why can't flux be installed both runtime and control cluster?

chivalryq (changes requested):
I just feel a bit confused

pkg/addon/addon.go
hongchaodeng: Yes. It will..
The runtime clusters include the control plane.

PR #2746, opened 2021-11-19
Feat: support update component and query app statistics info

barnettZQG:


pkg/addon/addon.go
```diff
@@ -499,6 +500,7 @@ func RenderArgsSecret(addon *types.Addon, args map[string]string) *v1.Secret {
 	return &sec
 }
 
+// Convert2SecName conver secret name
 func Convert2SecName(name string) string {
```
hongchaodeng, line 504:
```suggestion
func MakeSecName(name string) string {
```

pkg/apiserver/rest/apis/v1/types.go
```diff
@@ -261,6 +264,14 @@ type ApplicationStatusResponse struct {
 	Status  *common.AppStatus `json:"status"`
 }
 
+// ApplicationStatisticsResponse application statistics response body
+type ApplicationStatisticsResponse struct {
+	EnvNumber            int64 `json:"envNumber"`
```
hongchaodeng, line 269: What does "Number" mean?

hongchaodeng (approved):
lgtm after nit

pkg/apiserver/rest/apis/v1/types.go
hongchaodeng: Looks like it should be named "XXXCount"

PR #2749, opened 2021-11-19
Chore: Adding more chart values for set up cluster-gateway

yue9944882:


Somefive (approved):
lgtm

hongchaodeng:
Please run `make reviewable` and fix lint errors

yue9944882:
for clarification, the original intention of this pr is to provide an optional integration w/ [konnectivity traffic-egress](https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/) for the cluster-gateway (note that it's not coupled w/ OCM or OCM's [cluster-proxy](https://github.com/oam-dev/cluster-gateway) addon, konnectivity is just a standalone proxy tunnel that applies to any L4 traffic). however, due to the fact that the cluster-gateway is installed together w/ all the other core components such as vela controllers/webhooks, we dont have the flexibility of updating the args of gateway on the fly after the vela control plane is set. e.g. for the cases where the konnectivity services are installed after the vela, i presume these helm chart variables will not be helpful b/c it's a second-time installation. 

on the other hand, for a vela admin who's trying to build a vela environment for the first time, mostly he wont be even aware that he needs the konnectivity until running into network isolation issues. so as workaround i opened https://github.com/oam-dev/cluster-gateway/pull/20 which is a typical operator that helps us to manage the full lifecycle of the gateway including (1) updating the args (2) upgrading the images. etc. likewise for OCM, the konnectivity service is pluggable as an OCM addon so over there we need the cluster-gateway to flexibly degrade to non-proxy mode when the addon is off.

FWIW i think these additional helm vars are no longer needed if we do the first-time installation of the gateway via helm and defer the enabling/disabling of the konnectivity integration to the operator that seamlessly adopts the gateway instances.

yue9944882:
in addition to above, the current konnectivity integration of cluster-gateway doesn't apply to any arbitrarily installed konnectivity services. the implicit assumption is that the konnectivity agent deployed in the managed clusters are registered w/ `destHost` properties and the konnectivity service should be explicitly enable the hostname based routing where the hostname has to be the cluster name of the `ClusterGateway` resource. otherwise the integration will not be working. so there're actually a bunch of requirements before we enable the integration, which basically means that the konnectivity has to be installed via OCM's [cluster-proxy](https://open-cluster-management.io/getting-started/integration/cluster-proxy/) (the mgmt of konnectivity per se is complex), so i suppose we'd better not reflect the integration in kubevela's original helm chart. just hiding the internals as an optional operator is just practical.

wonderflow:
@yue9944882 I see, thanks for your contribution. Let's close this PR.

PR #2751, opened 2021-11-19
Feat: Alternatively using the original apiserver endpoint from the hub kubeconfig as registry endpoint

yue9944882:
> Signed-off-by: yue9944882 <291271447@qq.com>

the server endpoint stored in-cluster cluster-info will be an internal ip/host address e.g. on the ACK provider. adding a `--in-cluster-bootstrap=[true|false]` flag (which is defaulted to true considering backward compatibility) will help those admin to alternatively let the registration agent adopt the original endpoint (which is mostly an external hostname) as the registration entry. otherwise the registration will be constantly failing to network/VPC isolation.


yangsoon (approved):
lgtm

Somefive:
Looks good. Run `make reviewable` to format the code.

Somefive (approved):

PR #2761, opened 2021-11-21
Fix: fix several addon issue

wangyikewxgm:
Fix:
    
    1. load component in arrary, so apply them in order addon's needNamespace will be apply firstly
    2. apply application  in controle plane will be first workflowStep
    3. bigger application reconcile timeout context get avoid of time out

### Description of your changes

Fixes #

I have:

- [ ] Read and followed KubeVela's [contribution process](https://github.com/oam-dev/kubevela/blob/master/contribute/create-pull-request.md).
- [ ] [Related Docs](https://github.com/oam-dev/kubevela.io) updated properly. In a new feature or configuration option, an update to the documentation is necessary. 
- [ ] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

PR #2781, opened 2021-11-23
Fix: add back missed commit and add tests

wangyikewxgm:
add back missed commit and add tests

some logic in #2761   have been lost 

Fixes #

I have:


hongchaodeng (approved):

zzxwill (approved):

references/cli/cluster.go
```diff
@@ -155,6 +155,20 @@ func ensureResourceTrackerCRDInstalled(c client.Client, clusterName string) erro
 	return nil
 }
 
+func ensureVelaSystemNamespaceInstalled(c client.Client, clusterName string) error {
+	ctx := context.Background()
+	remoteCtx := multicluster.ContextWithClusterName(ctx, clusterName)
+	if err := c.Get(remoteCtx, types2.NamespacedName{Name: types.DefaultKubeVelaNS}, &v1.Namespace{}); err != nil {
```
Somefive, line 161: I suggest find the KubeVelaNS in management cluster first instead of directly using the default value.

wangyikewxgm: How can I find the default namespace in hub cluster? The reason why should create this namespace is our most addons application in vela-system namespace, so the I think this namespace cannot be modified easily.

Somefive: > How can I find the default namespace in hub cluster? The reason why should create this namespace is our most addons application in vela-system namespace, so the I think this namespace cannot be modified easily.

User can set the installed namespace by configuring it while installing helm. See https://github.com/oam-dev/kubevela/blob/3d2fcacb5a98af841b4b69995a0be204f6062c81/charts/vela-core/templates/kubevela-controller.yaml#L81. It is not a hard job.

Therefore, in my opinion, it is necessary to pre-check the installed namespace, since the namespace should be aligned with the kubevela system's namespace.

However, I found that for our addons like [fluxcd](https://github.com/oam-dev/kubevela/blob/3d2fcacb5a98af841b4b69995a0be204f6062c81/charts/vela-core/templates/addons/fluxcd.yaml#L12), the `vela-system` is fixed in the namespace configuration, which I do not think is a well-considered design. In the case where user install vela-core in another namespace, the addons will not work properly. But following the principle that different modules should be decoupled, I prefer not let this problem happens to multi-cluster module again, since sometimes users do not want to use addons together with multi-cluster.

I notice that even finding the namespace of vela-core is not enough since fluxcd addons do not use that. I suggest add another flag `--create-namespace=xxx` to `vela cluster join` to leave an option for users and let the default value to be `vela-system`.

Somefive (approved):
LGTM

wonderflow (approved):

leejanee (approved):

hongchaodeng (approved):

PR #2849, opened 2021-12-01
Feat: ResourceTracker new architecture

Somefive:
This PR is proposed to rework ResourceTracker and upgrade the current resource management architecture for KubeVela controller.

# Major Changes
1. Remove OwnerReference in managed resources. This simplifies the create and update for resources.
2. Record **rendered** Resource in ResourceTracker.Spec.ManagedResource. This allows StateKeep after workflow so that configuration drift for applied resources can be prevented.
3. ResourceTracker deletion will use application controller to reconcile.
4. Resources in ManagedCluster are directly recorded in ResourceTrackers in Hub Cluster, which means we do not need RT in Managed Clusters anymore.
5. Add GarbageCollect Policy and ApplyOnce Policy.
  a. The former one allows difference resources have various life-cycle. Usually, resources are deleted as soon as the latest application does not use them anymore. Some resources need to be kept until the application is gone. Some might need be kept even if the application is gone. 
  b. The latter one allows us to prevent configuration-drift by running StateKeep after workflow suspended/terminated/succeed. 

# Minor Changes
1. ControllerRevisions are directly recorded in ResourceTrackers.
2. Addon Application use ApplyOnce by default.
3. MultiCluster do not need extra gc logic.

# Deprecation
1. AppRollout is not supported anymore. Another PR is called to remove it permanently.
2. ReosurceTracker.Status is deprecated. 

# Implementation
1. Main changes are implemented in `pkg/resourcekeeper/` where resource recording, garbage collecting and state keeping are supported. *It is recommended for code reviewers to review from here, since other changes are all about making this package work properly.*
2. Lots of tests are fixed and added.

I have:

- [x] Add more e2e-test.


apis/core.oam.dev/v1alpha1/garbagecollect_types.go
```diff
@@ -0,0 +1,81 @@
...
+// GarbageCollectPolicyRule defines a single garbage-collect policy rule
+type GarbageCollectPolicyRule struct {
+	Selector GarbageCollectPolicyRuleSelector `json:"selector"`
+	Strategy GarbageCollectStrategy           `json:"strategy"`
+}
+
+// GarbageCollectPolicyRuleSelector select the targets of the rule
+type GarbageCollectPolicyRuleSelector struct {
+	TraitTypes []string `json:"traitTypes"`
+}
```
wonderflow, line 50: can they be omit-empty?

apis/core.oam.dev/v1beta1/resourcetracker_types.go
```diff
@@ -0,0 +1,274 @@
...
+	if !in.ClusterObjectReference.Equal(r.ClusterObjectReference) {
+		return false
+	}
+	if !in.OAMObjectReference.Equal(r.OAMObjectReference) {
+		return false
+	}
+	if in.Data != nil && r.Data != nil {
+		return bytes.Equal(in.Data.Raw, r.Data.Raw)
+	}
+	return (in.Data != nil) == (r.Data != nil)
```
wonderflow, line 93: do you mean you just want to check the Data is all nil or non-nil? Do you want to check their data equal?

apis/core.oam.dev/v1beta1/resourcetracker_types.go
```diff
@@ -0,0 +1,274 @@
...
+			Cluster: oam.GetCluster(rsc),
+		},
+		OAMObjectReference: common.NewOAMObjectReferenceFromObject(rsc),
+		Deleted:            false,
+	}
+	if !metaOnly {
+		bs, err := json.Marshal(rsc)
+		if err != nil {
+			return err
+		}
```
wonderflow, line 208: does it necessary to marshel it?

Why not just store the object in `&runtime.RawExtension{Object: rsc}`

pkg/resourcetracker/app.go
```diff
@@ -0,0 +1,129 @@
...
+	if err := rt.AddManagedResource(manifest, metaOnly); err != nil {
+		return err
+	}
+	return cli.Update(ctx, rt)
+}
+
+// DeletedManifestInResourceTracker marks resources as deleted in resourcetracker, if remove is true, resources will be removed from resourcetracker
+func DeletedManifestInResourceTracker(ctx context.Context, cli client.Client, rt *v1beta1.ResourceTracker, manifest *unstructured.Unstructured, remove bool) error {
+	rt.DeleteManagedResource(manifest, remove)
+	return cli.Update(ctx, rt)
```
wonderflow, line 128: where do you delete the ownerRef in resourceTracker for compatibility？

pkg/resourcetracker/app.go
```diff
@@ -0,0 +1,129 @@
...
+	Finalizer = "resourcetracker.core.oam.dev/finalizer"
+)
+
+func getRootResourceTrackerName(app *v1beta1.Application) string {
+	return fmt.Sprintf("%s-%s", app.Name, app.Namespace)
+}
+
+func getCurrentResourceTrackerName(app *v1beta1.Application) string {
+	return fmt.Sprintf("%s-v%d-%s", app.Name, app.GetGeneration(), app.Namespace)
+}
```
leejanee, line 42: Should consider a situation that user specifies the publish version

apis/core.oam.dev/v1alpha1/garbagecollect_types.go
Somefive: rules can be empty, selector and strategy cannot, fixed it.

pkg/resourcekeeper/statekeep.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
+	"github.com/oam-dev/kubevela/pkg/multicluster"
+	"github.com/oam-dev/kubevela/pkg/utils/apply"
+)
+
+// StateKeep run this function to keep resources up-to-date
+func (h *resourceKeeper) StateKeep(ctx context.Context) error {
```
wonderflow, line 30: you should write a user facing doc to explain this policy and recommand the best practice.

apis/core.oam.dev/v1beta1/resourcetracker_types.go
Somefive: Yes, need to check their data equal. The function here is not in-use now. Shall we remove it?

pkg/resourcekeeper/statekeep.go
Somefive: Ok.

pkg/resourcetracker/app.go
Somefive: The compatibility code has not been added yet. Will be added together with upgrade tests.

pkg/resourcetracker/app.go
Somefive: Ok.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -435,6 +450,22 @@ func isHealthy(services []common.ApplicationComponentStatus) bool {
...
+			UpdateFunc: func(updateEvent ctrlEvent.UpdateEvent, limitingInterface workqueue.RateLimitingInterface) {
+				handleResourceTracker(updateEvent.ObjectNew, limitingInterface)
+			},
+			DeleteFunc: func(deleteEvent ctrlEvent.DeleteEvent, limitingInterface workqueue.RateLimitingInterface) {
+				handleResourceTracker(deleteEvent.Object, limitingInterface)
+			},
+			GenericFunc: func(genericEvent ctrlEvent.GenericEvent, limitingInterface workqueue.RateLimitingInterface) {
+				handleResourceTracker(genericEvent.Object, limitingInterface)
+			},
+		}).
```
wonderflow, line 468: what are these handler for?

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -294,7 +275,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 	}
 
-	if err := garbageCollection(ctx, handler); err != nil {
+	if err := handler.resourceKeeper.StateKeep(ctx); err != nil {
+		logCtx.Error(err, "Failed to run prevent-configuration-drift")
+		r.Recorder.Event(app, event.Warning(velatypes.ReasonFailedStateKeep, err))
+		return r.endWithNegativeCondition(logCtx, app, condition.ReconcileError(err), phase)
```
wonderflow, line 281: What will happend In this case? I think the application is already running well here,and the health check should report error first if not? 

So , does it necessary to report an error? But a event even change the condition is necessary

apis/core.oam.dev/v1beta1/resourcetracker_types.go
wonderflow: Of course you should remove them. Dead code can cause confusion

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
Somefive: These handlers are added for application controller to response ResourceTracker change events. Designed by @leejanee .

apis/core.oam.dev/v1beta1/resourcetracker_types.go
Somefive: Right, will fix it.

apis/core.oam.dev/v1beta1/resourcetracker_types.go
Somefive: refactored to `reflect.DeepEqual`

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
Somefive: If StateKeep failed (failed to get/update target resource and is not caused by not found or conflict, such as cluster disconnect), it will trigger the next reconcile of StateKeep due to the returned error.

Somefive: I think at least a warning should be reported for users, since a failure is encountered during ensuring application managed resources and bringing state back.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
wonderflow: What's the real affect? Who will change RT besides app controller?

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
wonderflow: Yes， it deserves a warning. But return error just cause a new reconcile which doesn't help in this case.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
Somefive: In the mode of KeepLegacyResource, history versioned ResourceTrackers will not be automatically deleted by application controller. Instead, users have the control to select which history version to keep or discard. In this case, user will delete RT.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
Somefive: In some cases, it is worth a retry. For example, if the child cluster apiserver is busy temporarily, then the resource check might timeout which will cause a StateKeep error here. However, the child cluster apiserver might recover by itself later, then a retry will help controller do the StateKeep. Generally, it would be better to have more details for the returned error and handle it more elegantly. I agree that currently we do not necessarily need an error to be returned here. Will fix it.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
wonderflow: I got it. That means app controller is also the controller of RT

wonderflow (approved):
LGTM, great work

leejanee (approved):
/lgtm

apis/core.oam.dev/common/types.go
```diff
@@ -476,6 +481,48 @@ const (
 	WorkflowResourceCreator ResourceCreatorRole = "workflow"
 )
 
+// OAMObjectReference defines the object reference for an oam resource
+type OAMObjectReference struct {
+	Component string `json:"component,omitempty"`
```
zzxwill, line 486: Is it based on `one application--one component--one trait` module?

apis/core.oam.dev/common/types.go
```diff
@@ -476,6 +481,48 @@ const (
...
+	Trait     string `json:"trait,omitempty"`
+	Env       string `json:"env,omitempty"`
+}
+
+// Equal check if two references are equal
+func (in OAMObjectReference) Equal(r OAMObjectReference) bool {
+	return in.Component == r.Component && in.Trait == r.Trait && in.Env == r.Env
```
zzxwill, line 493: Can `reflect.DeepEqual` do the comparasion?

apis/core.oam.dev/common/types.go
Somefive: Not exactly. This reference is for one target resource, for example, a deployment or a service. Each resource is always belonging to `1 app - 1 comp (- 1 trait, optional)`.

apis/core.oam.dev/common/types.go
Somefive: Yes, there are several unit-tests can ensure that.

zzxwill:
I noticed there are some changes on `apis/core.oam.dev/v1alpha1`. Will it lead to any compatibilities problems?

Somefive:
> I noticed there are some changes on `apis/core.oam.dev/v1alpha1`. Will it lead to any compatibilities problems?

Yes, there will be compatibility problems. ResourceTracker will face an upgrade concerning existing KubeVela system. Some compatibility code and tests will be added.

apis/core.oam.dev/common/types.go
Somefive: Yes. It can. It is also possible to do the equality check with `reflect.DeepEqual`.

zzxwill (approved):

PR #2875, opened 2021-12-03
Fix: OSS read data race

chivalryq:

Fix several problem.
1. When get `bucket.endpoint?prefix=terraform`, files in `terraform-alibaba` will be listed.(They all have "terraform" prefix) This will cause race. 
2. AddonRegistryMeta's Git or OSS can be set and should be access by SourceOf function.
3. Unify the interface

Fixes #

I have:

### DIFF
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -102,12 +102,12 @@ func GetAddonsFromReader(r AsyncReader, opt ListOptions) ([]*types.Addon, error)
 	if err != nil {
 		return nil, err
 	}
+	var l sync.Mutex
 	for _, subItem := range items {
 		if subItem.GetType() != "dir" {
 			continue
 		}
 		wg.Add(1)
-		var l sync.Mutex
 		go func(item Item) {
 			defer wg.Done()
 			ar := r.WithNewAddonAndMutex()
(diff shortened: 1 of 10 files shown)


hongchaodeng (approved):

pkg/addon/addon.go
```diff
@@ -102,12 +102,12 @@ func GetAddonsFromReader(r AsyncReader, opt ListOptions) ([]*types.Addon, error)
...
 	}
+	var l sync.Mutex
 	for _, subItem := range items {
 		if subItem.GetType() != "dir" {
 			continue
 		}
 		wg.Add(1)
-		var l sync.Mutex
 		go func(item Item) {
 			defer wg.Done()
```
wonderflow, line 112: the `errChan` is going to block the whole function if error occurs more than twice:

image

wonderflow (changes requested):
goroutine, for-loop, channels, it's becoming too complex 

I suggest you just use channel only, and there can be different type of channel vaulues which and identify error、success or completed.

Please make it simpler.

pkg/addon/addon.go
chivalryq: fixed

wonderflow (approved):
let's merge for rc preview version, pelase refine the code in the following PR

barnettZQG (approved):

PR #2900, opened 2021-12-08
Chore: OCM addon image version

yue9944882:
> Signed-off-by: yue9944882 <291271447@qq.com>

### DIFF
--- a/charts/vela-core/templates/addons/ocm-cluster-manager.yaml
+++ b/charts/vela-core/templates/addons/ocm-cluster-manager.yaml
@@ -417,9 +417,9 @@ data:
           metadata:
             name: cluster-manager-hub
           spec:
-            placementImagePullSpec: quay.io/open-cluster-management/placement
-            registrationImagePullSpec: quay.io/open-cluster-management/registration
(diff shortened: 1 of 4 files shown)


charts/vela-core/templates/cluster-gateway.yaml
```diff
@@ -4,6 +4,8 @@ kind: Deployment
 metadata:
   name: {{ .Release.Name }}-cluster-gateway
   namespace: {{ .Release.Namespace }}
+  annotations:
+    helm.sh/resource-policy: keep
```
zzxwill, line 8: Why do we choose to keep this resource after vela-core is uninstalled? https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

yue9944882: the other way round, the idea is to make the chart installation compatible w/ existing cluster-gateway instances installed via OCM. anyway, i stripped the chart changes from this thread.

Somefive (approved):

zzxwill (approved):

wonderflow (approved):

PR #2901, opened 2021-12-08
Fix: delete invisible addon feature

wangyikewxgm:

delete invisible addon feature

I have:


zzxwill:
I think we should keep this feature.

For example, we have released an addon x, but somehow, we have found a very serious issue. Then we need to hide it temporarily. If we disable this feature, we have to manually move it to somewhere else.

PR #3008, opened 2021-12-27
Feat: reset re-sync time

Somefive:
Signed-off-by: Yin Da <yd219913@alibaba-inc.com>

The default 1h re-sync time is a bit long for state-keep. In community version, users can experience the state-keep feature within 5 min now.

I have:

### DIFF
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -62,6 +62,9 @@ affinity: {}
 rbac:
   create: true
 
+controllerArgs:
+  reSyncInterval: 5m
+
 admissionWebhooks:
   enabled: true
   failurePolicy: Fail
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -63,7 +63,7 @@ rbac:
   create: true
 
 controllerArgs:
-  reSyncInterval: 5m
+  reSyncPeriod: 5m
 
 admissionWebhooks:
   enabled: true
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -74,7 +74,6 @@ func main() {
 	var healthAddr string
 	var disableCaps string
 	var storageDriver string
-	var syncPeriod time.Duration
 	var applyOnceOnly string
 	var qps float64
 	var burst int
@@ -110,8 +109,8 @@ func main() {
 		"For the purpose of some production environment that workload or trait should not be affected if no spec change, available options: on, off, force.")
 	flag.StringVar(&disableCaps, "disable-caps", "", "To be disabled builtin capability list.")
 	flag.StringVar(&storageDriver, "storage-driver", "Local", "Application file save to the storage driver")
-	flag.DurationVar(&syncPeriod, "informer-re-sync-interval", 5*time.Minute,
-		"controller shared informer lister full re-sync period")
+	flag.DurationVar(&commonconfig.ApplicationReSyncPeriod, "application-re-sync-period", 5*time.Minute,
+		"Re-sync period for application to re-sync, also known as the state-keep interval.")
 	flag.DurationVar(&commonconfig.ReconcileTimeout, "reconcile-timeout", time.Minute*3,
 		"the timeout for controller reconcile")
 	flag.StringVar(&oam.SystemDefinitonNamespace, "system-definition-namespace", "vela-system", "define the namespace of the system-level definition")
@@ -205,7 +204,6 @@ func main() {
 		Port:                       webhookPort,
 		CertDir:                    certDir,
 		HealthProbeBindAddress:     healthAddr,
-		SyncPeriod:                 &syncPeriod,
 		LeaderElectionResourceLock: leaderElectionResourceLock,
 		LeaseDuration:              &leaseDuration,
 		RenewDeadline:              &renewDeadline,
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -208,7 +208,12 @@ func main() {
 		LeaseDuration:              &leaseDuration,
 		RenewDeadline:              &renewDeadline,
 		RetryPeriod:                &retryPeriod,
-		ClientDisableCacheFor:      []client.Object{&appsv1.ControllerRevision{}},
+		// SyncPeriod is configured with default value, aka. 10h. First, controller-runtime does not
+		// recommend use it as a time trigger, instead, it is expected to work for failure tolerance
+		// of controller-runtime. Additionally, set this value will affect not only application
+		// controller but also all other controllers like definition controller. Therefore, for
+		// functionalities like state-keep, they should be invented in other ways.
+		ClientDisableCacheFor: []client.Object{&appsv1.ControllerRevision{}},
 	})
 	if err != nil {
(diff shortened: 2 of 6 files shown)


cmd/core/main.go
```diff
@@ -205,7 +204,6 @@ func main() {
 		Port:                       webhookPort,
 		CertDir:                    certDir,
 		HealthProbeBindAddress:     healthAddr,
-		SyncPeriod:                 &syncPeriod,
```
wonderflow, line 208: we'd better add an annotation here to explain why we don't use it.

charts/vela-core/values.yaml
```diff
@@ -62,6 +62,9 @@ affinity: {}
 rbac:
   create: true
 
+controllerArgs:
+  reSyncPeriod: 5m
```
wonderflow, line 66: can we make it more fast? such as 2m ? how did we get the time result?

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -101,7 +101,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		if !kerrors.IsNotFound(err) {
 			logCtx.Error(err, "get application")
 		}
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		return ctrl.Result{RequeueAfter: common2.ApplicationReSyncPeriod}, client.IgnoreNotFound(err)
```
wonderflow, line 104: let's add a commno value for return.

For example:  `SuccessRequeueResult`

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -292,34 +292,34 @@ func (r *Reconciler) handleFinalizers(ctx monitorContext.Context, app *v1beta1.A
...
 }
 
 func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.Application, condition condition.Condition, phase common.ApplicationPhase) (ctrl.Result, error) {
 	app.SetConditions(condition)
 	if err := r.patchStatus(ctx, app, phase); err != nil {
-		return ctrl.Result{}, errors.WithMessage(err, "cannot update application status")
+		return ctrl.Result{RequeueAfter: common2.ApplicationReSyncPeriod}, errors.WithMessage(err, "cannot update application status")
 	}
-	return ctrl.Result{}, fmt.Errorf("object level reconcile error, type: %q, msg: %q", string(condition.Type), condition.Message)
+	return ctrl.Result{RequeueAfter: common2.ApplicationReSyncPeriod}, fmt.Errorf("object level reconcile error, type: %q, msg: %q", string(condition.Type), condition.Message)
```
wonderflow, line 322: we shouldn't return both requeue and erorr. That'll cause redundant reconcile.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -292,34 +292,34 @@ func (r *Reconciler) handleFinalizers(ctx monitorContext.Context, app *v1beta1.A
...
 			ctx.Info("Register new finalizer for application", "finalizer", resourceTrackerFinalizer)
-			return true, ctrl.Result{}, errors.Wrap(r.Client.Update(ctx, app), errUpdateApplicationFinalizer)
+			return true, ctrl.Result{RequeueAfter: common2.ApplicationReSyncPeriod}, errors.Wrap(r.Client.Update(ctx, app), errUpdateApplicationFinalizer)
 		}
 	} else {
 		if meta.FinalizerExists(app, resourceTrackerFinalizer) {
 			rootRT, currentRT, historyRTs, cvRT, err := resourcetracker.ListApplicationResourceTrackers(ctx, r.Client, app)
 			if err != nil {
-				return true, ctrl.Result{}, err
+				return true, ctrl.Result{RequeueAfter: common2.ApplicationReSyncPeriod}, err
```
wonderflow, line 301: same here

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -279,9 +279,9 @@ func (r *Reconciler) gcResourceTrackers(logCtx monitorContext.Context, handler *
 	}
 	logCtx.Info("GarbageCollected resourcetrackers")
 	if phase == common.ApplicationRendering {
-		return ctrl.Result{}, r.updateStatus(logCtx, handler.app, common.ApplicationRunningWorkflow)
+		return ctrl.Result{RequeueAfter: common2.ApplicationReSyncPeriod}, r.updateStatus(logCtx, handler.app, common.ApplicationRunningWorkflow)
 	}
-	return ctrl.Result{}, r.patchStatus(logCtx, handler.app, phase)
```
wonderflow, line 284: can we check whether we have changed the status? if changed, we shouldn't add a requeue.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
Somefive: Agree.

charts/vela-core/values.yaml
Somefive: 2m is a bit aggressive I think. For example, with 2c4g the standard application supported in 1 minute is 200~300 (without advanced optimization). If the reSyncPeriod is set to 2m, it means with 2c4g we can support 400~600 applications. Therefore, for the default helm install configuration (0.5c1g), we can only support 100~150 simple applications. For 5m, we can have 250+ applications.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
Somefive: Yes. It's a mistake.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
Somefive: Yes. There should be some better check here.

charts/vela-core/values.yaml
wonderflow: This data should be documented well.

wonderflow (approved):

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -275,13 +275,41 @@ func (r *Reconciler) gcResourceTrackers(logCtx monitorContext.Context, handler *
...
+	return r
+}
+
+func (r *reconcileResult) ret() (ctrl.Result, error) {
+	if r.Duration.Seconds() != 0 {
+		return ctrl.Result{RequeueAfter: r.Duration}, r.err
+	} else if r.err != nil {
+		return ctrl.Result{}, r.err
+	}
+	return ctrl.Result{RequeueAfter: common2.ApplicationReSyncPeriod}, nil
```
huanghantao, line 303: Hello, I wonder why there is no `Result.Requeue` false here. It currently seems that even if the application is completely cleaned up, an Reconcile will still be triggered every 5 minutes.

PR #3009, opened 2021-12-27
Fix: pending deprecated test

Somefive:
Signed-off-by: Yin Da <yd219913@alibaba-inc.com>

Previous Helm Chart type is deprecated and replaced by newer implementation. Old test has the following problem:
- While dispatching helm chart component and apply patch trait to it, the deployment created by fluxcd can be relatively slow.
- Workflow has backoff wait but when the deployment is not created, error is returned, so backoff has a maximum time at 10. Therefore, after about 1 minute, if the deployment has not been created, the application will enter suspending status and will not retry anymore.
- In this case, the patch trait cannot work so e2e-test will fail.

Since this feature is not going to be supported and deprecated, the e2e-test can be pending for now.

If we would like to support this feature and re-enable this e2e-test, we need to fix the backoff problem.

I have:

### DIFF
--- a/test/e2e-test/definition_revision_test.go
+++ b/test/e2e-test/definition_revision_test.go
@@ -352,7 +352,7 @@ var _ = Describe("Test application of the specified definition version", func() 
 		}, 30*time.Second, 3*time.Second).Should(Succeed())
 	})
 
-	It("Test deploy application which containing helm module", func() {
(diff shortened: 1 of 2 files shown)


wonderflow (approved):
Please add a comment on the test case pended

leejanee (approved):

PR #3051, opened 2022-01-06
Fix: generate docs for Terraform ComponentDefinition

zzxwill:
Generated kubevela.io docs for Terraform typed ComponentDefinition
with `vela def gen-doc xxx` cli.

Fixes #

I have:

### DIFF
--- a/pkg/controller/utils/capability.go
+++ b/pkg/controller/utils/capability.go
@@ -190,9 +190,16 @@ func GetOpenAPISchemaFromTerraformComponentDefinition(configuration string) ([]b
 // GetTerraformConfigurationFromRemote gets Terraform Configuration(HCL)
 func GetTerraformConfigurationFromRemote(name, remoteURL, remotePath string) (string, error) {
 	tmpPath := filepath.Join("./tmp/terraform", name)
+	// Check if the directory exists. If yes, remove it.
+	if _, err := os.Stat(tmpPath); err == nil {
+		err := os.RemoveAll(tmpPath)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to remove the directory")
+		}
+	}
 	_, err := git.PlainClone(tmpPath, false, &git.CloneOptions{
 		URL:      remoteURL,
-		Progress: os.Stdout,
+		Progress: nil,
 	})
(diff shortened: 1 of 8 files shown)


references/cli/def.go
```diff
@@ -325,6 +327,40 @@ func NewDefinitionGetCommand(c common.Args) *cobra.Command {
 	return cmd
 }
 
+// NewDefinitionGenDocCommand create the `vela def gen-doc` command to generate documentation of definitions
+func NewDefinitionGenDocCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "gen-doc NAME",
+		Short: "Generate documentation of definitions (Only Terraform typed definitions are supported)",
```
wonderflow, line 334: can we also support other cue based definition?

zzxwill: Yes, it's in my plan. This feature is the official feature by CLI to generate docs for KubeVela definitions (this was supported by https://github.com/oam-dev/kubevela/blob/master/hack/references/generate.go). I tried to generate all definitions (including workflowstep), but 1) there were some blockers to run the tool, 2) the generated docs differs from thoes hosted in kubevela.io. So I plan to support Terraform typed definition first, and then for all components, traits, policy and workflow. Is that okay?

pkg/controller/utils/capability.go
```diff
@@ -190,9 +190,16 @@ func GetOpenAPISchemaFromTerraformComponentDefinition(configuration string) ([]b
 // GetTerraformConfigurationFromRemote gets Terraform Configuration(HCL)
 func GetTerraformConfigurationFromRemote(name, remoteURL, remotePath string) (string, error) {
 	tmpPath := filepath.Join("./tmp/terraform", name)
+	// Check if the directory exists. If yes, remove it.
```
zzxwill, line 193: @FogDong Here are two modifications of this method. Please help check whether they are okay with you.

references/cli/def.go
wonderflow: sure,thanks

wonderflow (approved):

Somefive (approved):

PR #3066, opened 2022-01-10
Feat: addon enable support local dir for install offline

wangyikewxgm:
Signed-off-by: wangyike <wangyike_wyk@163.com>

support install an addon by local dir

addon-local

Fixes #

I have:

### DIFF
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -109,6 +109,8 @@ const (
 	ObservabilityAddonEndpointComponent = "grafana"
 	// ObservabilityAddonDomainArg is the domain argument name of the observability addon
 	ObservabilityAddonDomainArg = "domain"
+	// LocalAddonRegistryName is the addon-registry name for those installed by local dir
+	LocalAddonRegistryName = "local"
 )
 
 // ObservabilityEnvironment contains the Observability addon's domain for each cluster
@@ -949,6 +951,8 @@ func (h *Installer) dispatchAddonResource(addon *InstallPackage) error {
 		return err
 	}
 	app.Name = appName
+
+	app.SetLabels(util.MergeMapOverrideWithDst(app.GetLabels(), map[string]string{oam.LabelAddonRegistry: h.r.Name}))
 
 	defs, err := RenderDefinitions(h.addon, h.config)
 	if err != nil {
--- a/pkg/addon/reader_local.go
+++ b/pkg/addon/reader_local.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2021 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package addon
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
+)
+
+type localReader struct {
+	dir  string
+	name string
+}
+
+func (l localReader) ListAddonMeta() (map[string]SourceMeta, error) {
+	metas := SourceMeta{Name: l.name}
+	if err := recursiveFetchFiles(l.dir, &metas); err != nil {
+		return nil, err
+	}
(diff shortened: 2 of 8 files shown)


pkg/addon/addon.go
```diff
@@ -109,6 +109,8 @@ const (
 	ObservabilityAddonEndpointComponent = "grafana"
 	// ObservabilityAddonDomainArg is the domain argument name of the observability addon
 	ObservabilityAddonDomainArg = "domain"
+	// LocalAddonRegistryName is the addon-registry name for those installed by local dir
+	LocalAddonRegistryName = "local"
```
zzxwill, line 113: What if a user already used `local` as his/her registry name?

references/cli/addon.go
```diff
@@ -314,6 +362,7 @@ func listAddons(ctx context.Context, clt client.Client, registry string) error {
 	if err != nil {
 		return err
 	}
+	onlineAddon := map[string]bool{}
```
zzxwill, line 365: If an add-on registry is valid, it's online?

references/cli/addon.go
```diff
@@ -255,6 +288,10 @@ func NewAddonStatusCommand(c common.Args, ioStream cmdutil.IOStreams) *cobra.Com
 	}
 }
 
+func parseAddonArgsFromFlag(cmd *cobra.Command) {
+	cmd.Flags().StringP(addonLocalDir, "f", "", "enable addon with a local dir path")
```
zzxwill, line 292:
```suggestion
	cmd.Flags().StringP(addonLocalDir, "d", "", "enable addon with a local dir path")
```
Is `-d` or `-l` better?

pkg/addon/addon.go
wangyikewxgm: Cannot got the reason why user name an addon-registry with `local`.

references/cli/addon.go
wangyikewxgm: Just a local variable in this func， won't show this info for user.

zzxwill (approved):

references/cli/addon.go
wonderflow: yes, it's not a file

wonderflow (changes requested):

references/cli/addon.go
wonderflow: the best user expericnce is helm like:


```
vela addon enable <addonname>
vela addon enable <local-dir>
```

The args can be local dir or name, you can check it in code.

pkg/addon/reader_local.go
```diff
@@ -0,0 +1,67 @@
...
+		if file.IsDir() {
+			if err := recursiveFetchFiles(fmt.Sprintf("%s/%s", path, file.Name()), metas); err != nil {
+				return err
+			}
+		} else {
+			metas.Items = append(metas.Items, OSSItem{tp: "file", path: fmt.Sprintf("%s/%s", path, file.Name()), name: file.Name()})
+		}
+	}
+	return nil
+}
```
wonderflow, line 67: please write a test to cover it.

references/cli/addon.go
wangyikewxgm: fixed

wonderflow (approved):
good job

references/cli/addon.go
```diff
@@ -171,25 +185,42 @@ func NewAddonUpgradeCommand(c common.Args, ioStream cmdutil.IOStreams) *cobra.Co
...
-				return errors.Wrapf(err, "cannot fetch addon related addon %s", name)
-			}
 			addonArgs, err := parseToMap(args[1:])
 			if err != nil {
 				return err
 			}
-			err = enableAddon(ctx, k8sClient, config, name, addonArgs)
-			if err != nil {
-				return err
+			addonOrDir := args[0]
```
chivalryq, line 192: what if the local directory has the same name as an addon in the registry? maybe we should add a bool flag to mark. How about `--dev` or `--local`

references/cli/addon.go
chivalryq: > the best user expericnce is helm like:
> 
> ```
> vela addon enable <addonname>
> vela addon enable <local-dir>
> ```
> 
> The args can be local dir or name, you can check it in code.

What if there is conflict? IMO a flag is needed.

references/cli/addon.go
wangyikewxgm: Have considered this situation. But IMO this is fine. The main logic is  if local truly have this path will install it from local dir otherwise will install from remote registry. Refer to the implementation of helm https://github.com/helm/helm/blob/292368a64d6c9794b4573afa5a8dc6e940a21045/pkg/action/install.go#L668

PR #3069, opened 2022-01-10
Fix: recover filtered event by re-sync

Somefive:

Previous resync events are filtered by the equality check. Thus the logic will make StateKeep failed.


pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -393,6 +393,10 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 			// filter the changes in workflow status
 			// let workflow handle its reconcile
 			UpdateFunc: func(e ctrlEvent.UpdateEvent) bool {
+				// if completely equal, this update is triggered by re-sync and should not be filtered
+				if reflect.DeepEqual(e.ObjectNew, e.ObjectOld) {
+					return true
+				}
```
wonderflow, line 399: if so, does L407~409 necessary？

Somefive: Right, this PR is not needed. I neglect L407~409. Close it.

PR #3086, opened 2022-01-13
Feat: add more performance optimization and prometheus metrics for controller 

yangsoon:

This PR proposed several optimization techniques to enhance the performance of KubeVela controller and improve the exposure of metrics to monitor system running state.

1. Use customized `ControllerClient` as the underlying client instead of the default controller-runtime client. The benefits includes:
  - Enable detailed metrics for client. The default controller-runtime metrics are unable to get the latency while calling cache, only requests from client-go can be accounted.
  - Enable customized cache options for unstructured resources (and disable it for child clusters which will lead to chaos). By default, controller-runtime client does not provide a configurable to do this.
  - For ResourceTracker List by label operation (which will be done almost every time in application_controller reconcile), add customized index to improve the cache read performance. This can prevent the List operation latency linearly increases when application number goes up.
2. Add configurations and more metrics for system observability. Metrics includes:
  - Times for each reconcile stage such as executing workflow and garbage collecting.
  - Times for each reconcile loop grouping by phase transition. This can help us identify high latency reconcile loop reason.  Various transition requires highly different execute time. For example, running workflow can be much more slower than simply patching finalizer. So average reconcile time or p99 metrics cannot accurately reflect the details of reconcile time.
  - Times for each ControllerClient call. This can distinguish function calls by client-go and cache reader. 
3. Add performance optimization methods as options to use. By default, **optimize-resource-tracker-list-op**, **optimize-mark-with-prob** and **optimize-enable-resource-tracker-delete-only-trigger** are enabled since their side-effect can almost be ignored (the scenario for side effect occurs in rare scenario).
  - **optimize-cached-gvks**: Optimize controller client cache with specified gvks. If this option is set, target resources with specified gvk will be cached even if they are unstructured types.
  - **optimize-resource-tracker-list-op**: Optimize ResourceTracker List Op by adding index. This will increase the use of memory and accelerate the list operation of ResourceTracker.
  - **optimize-controller-reconcile-loop-reduction**: Optimize ApplicationController reconcile by reducing the number loops to reconcile application. With this option enabled, application reconcile can reduce the number of reconcile loops by half approximately. It fuses several application phase transitions into one. The side effect is that if errors occurs during the phase transition, the restart process might be time-consuming. It is encouraged to be enabled if all application workflow steps are guaranteed to be idempotent.
  - **optimize-mark-with-prob**: Optimize ResourceTracker GC by only run mark stage with probability. When using `keep-legacy-resource` during gc, by default, the mark stage will loop over all versions of legacy resources and identify which ones are outdated. However, this can be time-consuming if too much (like 20+) versions are kept. Set this option to a float number like `0.1` will only do this check with probability. It is encouraged to be enabled when a lot of versions of applications are expected to be kept.
  - **optimize-disable-component-revision**: Optimize ComponentRevision by disabling the creation and gc. If all applications do not use rollout, this option can be enabled to reduce time and space.
  - **optimize-disable-application-revision**: Optimize ApplicationRevision by disabling the creation and gc. If application rollback is not needed, this option can be enabled to reduce time and space.
  - **optimize-disable-workflow-recorder**: Optimize workflow recorder by disabling the creation and gc. VelaUX relies on this to get workflow running history.
  - **optimize-enable-in-memory-workflow-context**: Optimize workflow by use in-memory context. By default, workflow state are kept in ConfigMaps for persistency. Enabling this option can accelerate the retrieval and storing of workflow state. However, if KubeVela Controller is shutdown or restarted, all application workflow running states will be lost. Therefore, application will restart workflow. Ensure workflow steps are idempotent while enabling this.
  - **optimize-disable-resource-apply-double-check**: Optimize workflow by ignoring resource double check after apply. While enabling this optimization, the ApplyComponent operation will skip the check the result of resource creation.
  - **optimize-enable-resource-tracker-delete-only-trigger**: Optimize ResourceTracker by only trigger reconcile when ResourceTracker is deleted. While enabling this optimization, manual edit to ResourceTrackers will not be handled immediately.

I have:


charts/oam-runtime/templates/oam-runtime-controller.yaml
```diff
@@ -92,6 +92,10 @@ spec:
     metadata:
       labels:
     {{- include "oam-runtime.selectorLabels" . | nindent 8 }}
+      annotations:
+          prometheus.io/path: /metrics
+          prometheus.io/port: "8080"
+          prometheus.io/scrape: "true"
```
wonderflow, line 98: not necessary for this old version

cmd/core/main.go
```diff
@@ -131,6 +131,8 @@ func main() {
 	flag.DurationVar(&retryPeriod, "leader-election-retry-period", 2*time.Second,
 		"The duration the LeaderElector clients should wait between tries of actions")
 	flag.BoolVar(&enableClusterGateway, "enable-cluster-gateway", false, "Enable cluster-gateway to use multicluster, disabled by default.")
+	flag.StringVar(&ctrlClient.CachedGVKs, "cached-gvks", "", "Types of resources to be cached. For example, --cached-resources=Deployment.v1.apps,Job.v1.batch")
```
wonderflow, line 134: all optimization args should have a common prefix, lets name it "optimize-cache-gvks"

pkg/optimize/flags.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package optimize
+
+import (
+	"flag"
+)
+
+// AddFlags add flags
+func AddFlags() {
```
wonderflow, line 24:
```suggestion
func AddOptimizeFlags() {
```

wonderflow:
The PR description should be one of the document in kubevela.io \cc @yangsoon @Somefive

charts/oam-runtime/templates/oam-runtime-controller.yaml
Somefive: Solved.

cmd/core/main.go
Somefive: Solved.

pkg/optimize/flags.go
Somefive: The package name is optimize. Do we need to rename it?

Somefive:
> The PR description should be one of the document in kubevela.io \cc @yangsoon @Somefive

Right.

apis/core.oam.dev/common/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
 // +build !ignore_autogenerated
```
wonderflow, line 2: why add a duplicate line?

leejanee (commented):
optimize目录出现的很突兀,最好还是把代码整合到各模块下

cmd/core/main.go
```diff
@@ -131,6 +131,8 @@ func main() {
 	flag.DurationVar(&retryPeriod, "leader-election-retry-period", 2*time.Second,
 		"The duration the LeaderElector clients should wait between tries of actions")
 	flag.BoolVar(&enableClusterGateway, "enable-cluster-gateway", false, "Enable cluster-gateway to use multicluster, disabled by default.")
+	flag.StringVar(&ctrlClient.CachedGVKs, "optimize-cached-gvks", "", "Types of resources to be cached. For example, --cached-resources=Deployment.v1.apps,Job.v1.batch")
```
wonderflow, line 134:
```suggestion
	flag.StringVar(&ctrlClient.CachedGVKs, "optimize-cached-gvks", "", "Types of resources to be cached. For example, --optimize-cached-gvks=Deployment.v1.apps,Job.v1.batch")
```

please make sure it's worked

pkg/client/controller_client.go
```diff
@@ -0,0 +1,118 @@
...
+var (
+	// CachedGVKs identifies the GVKs of resources to be cached during dispatching
+	CachedGVKs = ""
+
+	rtCount = 0
+	lock    = sync.Mutex{}
+)
+
+// DefaultNewControllerClient function for creating controller client
+func DefaultNewControllerClient(cache cache.Cache, config *rest.Config, options client.Options, uncachedObjects ...client.Object) (c client.Client, err error) {
```
wonderflow, line 48: Do we fork the client code from controller-runtime?

pkg/controller/core.oam.dev/v1alpha2/application/generator.go
```diff
@@ -181,6 +187,9 @@ func (h *AppHandler) applyComponentFunc(appParser *appfile.Parser, appRev *v1bet
 		if !isHealth {
 			return nil, nil, false, nil
 		}
+		if optimize.WorkflowOptimizer.DisableResourceApplyDoubleCheck {
+			return readyWorkload, readyTraits, true, nil
+		}
```
wonderflow, line 192: as we disable the double check, can we still ensure the resources is created in health check?

pkg/optimize/flags.go
```diff
@@ -0,0 +1,39 @@
...
+	"flag"
+)
+
+// AddOptimizeFlags add flags
+func AddOptimizeFlags() {
+	// optimize client
+	flag.BoolVar(&ResourceTrackerOptimizer.OptimizeListOp, "optimize-resource-tracker-list-op", false, "Optimize ResourceTracker List Op by adding index. This will increase the use of memory and accelerate the list operation of ResourceTracker.")
+
+	// optimize controller reconcile loop
+	flag.BoolVar(&ControllerOptimizer.EnableReconcileLoopReduction, "optimize-controller-reconcile-loop-reduction", false, "Optimize ApplicationController reconcile by reducing the number loops to reconcile application.")
```
wonderflow, line 29: you should declare which ops was reduced here

pkg/optimize/flags.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// optimize functions
+	flag.Float64Var(&ResourceTrackerOptimizer.MarkWithProbability, "optimize-mark-with-prob", 0.0, "Optimize ResourceTracker GC by only run mark with probability. Default to 0.0 means not enable it. Side effect: outdated ResourceTracker might not be able to be removed immediately.")
```
wonderflow, line 32: can we open this optimization by default? it seems not a big issue if just delete rt with a latency.

pkg/optimize/flags.go
```diff
@@ -0,0 +1,39 @@
...
+	flag.BoolVar(&RevisionOptimizer.DisableAllComponentRevision, "optimize-disable-component-revision", false, "Optimize ComponentRevision by disabling the creation and gc. Side effect: rollout cannot be used.")
+	flag.BoolVar(&RevisionOptimizer.DisableAllApplicationRevision, "optimize-disable-application-revision", false, "Optimize ApplicationRevision by disabling the creation and gc. Side effect: application cannot rollback.")
+	flag.BoolVar(&WorkflowOptimizer.DisableRecorder, "optimize-disable-workflow-recorder", false, "Optimize workflow recorder by disabling the creation and gc. Side effect: workflow will not record application after finished running.")
```
wonderflow, line 35: what's the real side effect, e.g. workflow may lose input/output data and need to run again from scratch if controller restarted.

pkg/optimize/flags.go
```diff
@@ -0,0 +1,39 @@
...
+	flag.BoolVar(&WorkflowOptimizer.EnableInMemoryContext, "optimize-enable-in-memory-workflow-context", false, "Optimize workflow by use in-memory context. Side effect: controller crash will lead to mistakes in workflow inputs/outputs.")
```
wonderflow, line 36:
```suggestion
	flag.BoolVar(&WorkflowOptimizer.EnableInMemoryContext, "optimize-enable-in-memory-workflow-context", false, "Optimize workflow by use in-memory context. Side effect: controller crash will lead to workflow run again from scratch and possible to cause mistakes in workflow inputs/outputs.")
```

pkg/optimize/flags.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package optimize
+
+import (
+	"flag"
+)
+
+// AddOptimizeFlags add flags
+func AddOptimizeFlags() {
```
wonderflow, line 24: every flags here should add some scenarios that users can use this optimization.

pkg/optimize/flags.go
```diff
@@ -0,0 +1,39 @@
...
+	// optimize client
+	flag.BoolVar(&ResourceTrackerOptimizer.OptimizeListOp, "optimize-resource-tracker-list-op", false, "Optimize ResourceTracker List Op by adding index. This will increase the use of memory and accelerate the list operation of ResourceTracker.")
+
+	// optimize controller reconcile loop
+	flag.BoolVar(&ControllerOptimizer.EnableReconcileLoopReduction, "optimize-controller-reconcile-loop-reduction", false, "Optimize ApplicationController reconcile by reducing the number loops to reconcile application.")
+
+	// optimize functions
+	flag.Float64Var(&ResourceTrackerOptimizer.MarkWithProbability, "optimize-mark-with-prob", 0.0, "Optimize ResourceTracker GC by only run mark with probability. Default to 0.0 means not enable it. Side effect: outdated ResourceTracker might not be able to be removed immediately.")
+	flag.BoolVar(&RevisionOptimizer.DisableAllComponentRevision, "optimize-disable-component-revision", false, "Optimize ComponentRevision by disabling the creation and gc. Side effect: rollout cannot be used.")
```
wonderflow, line 33:
```suggestion
	flag.BoolVar(&RevisionOptimizer.DisableAllComponentRevision, "optimize-disable-component-revision", false, "Optimize ComponentRevision by disabling the creation and gc. Side effect: rollout cannot be used. If you don't rely on componnet revision, you can use this optimization.")
```

wonderflow (commented):
generally LGTM after commented issues

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -487,3 +515,12 @@ func handleResourceTracker(obj client.Object, limitingInterface workqueue.RateLi
 		}
 	}
 }
+
```
FogDong, line 518: Filter `ResourceVersion` in `func filterManagedFieldChangesUpdate`

cmd/core/main.go
yangsoon: it's typo, fixed

pkg/client/controller_client.go
yangsoon: There is no fork code, here is some custom configuration

pkg/resourcekeeper/gc.go
```diff
@@ -145,6 +159,9 @@ func (h *gcHandler) scan(ctx context.Context) (inactiveRTs []*v1beta1.ResourceTr
 	} else {
 		if h.cfg.passive {
 			inactiveRTs = []*v1beta1.ResourceTracker{}
+			if optimize.ResourceTrackerOptimizer.MarkWithProbability > 0 && rand.Float64() > optimize.ResourceTrackerOptimizer.MarkWithProbability { //nolint
```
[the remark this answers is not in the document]
Somefive: @sonatype-lift ignore

pkg/controller/core.oam.dev/v1alpha2/application/generator.go
Somefive: No, if health policy is configured, we can. Otherwise not.

pkg/optimize/flags.go
Somefive: This optimization will not work for common cases where users do not use `keepLegacyResource`.

pkg/optimize/flags.go
Somefive: This recorder is mainly used in VelaUX. The ConfigMap recorder should be removed in the future.

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
Somefive: Fixed in #3008.

pkg/optimize/flags.go
Somefive: OK.

pkg/optimize/flags.go
Somefive: OK.

pkg/optimize/flags.go
Somefive: OK.

pkg/optimize/flags.go
wonderflow: then it should be enabled by default

pkg/optimize/flags.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// optimize controller reconcile loop
+	flag.BoolVar(&ControllerOptimizer.EnableReconcileLoopReduction, "optimize-controller-reconcile-loop-reduction", false, "Optimize ApplicationController reconcile by reducing the number loops to reconcile application. In detail, reconciles after finalizer patching and workflow finished will not return immediately but will continue running.")
+
+	// optimize functions
+	flag.Float64Var(&ResourceTrackerOptimizer.MarkWithProbability, "optimize-mark-with-prob", 0.0, "Optimize ResourceTracker GC by only run mark with probability. Default to 0.0 means not enable it. Side effect: outdated ResourceTracker might not be able to be removed immediately.")
+	flag.BoolVar(&RevisionOptimizer.DisableAllComponentRevision, "optimize-disable-component-revision", false, "Optimize ComponentRevision by disabling the creation and gc. Side effect: rollout cannot be used. If you don't rely on component revision, you can use this optimization.")
+	flag.BoolVar(&RevisionOptimizer.DisableAllApplicationRevision, "optimize-disable-application-revision", false, "Optimize ApplicationRevision by disabling the creation and gc. Side effect: application cannot rollback.")
```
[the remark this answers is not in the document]
Somefive: ok

pkg/optimize/flags.go
Somefive: Ok.

pkg/optimize/flags.go
Somefive: ok.

pkg/optimize/flags.go
Somefive: Ok.

wonderflow (approved):

FogDong (approved):

PR #3126, opened 2022-01-19
Fix: add NotUpdateRenderHashEqual option to statekeep apply

kinsolee:

Statekeep should have same apply options with Dispatch to skip update when the render-hash doesn't change.

Fixes https://github.com/oam-dev/kubevela/issues/3123

I have:


wonderflow:
please help rebase the PR and trigger CI。

Somefive:
This option is designed to be ignore in StateKeep. The render hash will only change when the rendered part is changed. It is compared with the hash label instead of recomputing the retrieved object. Therefore, if the target resource is changed, the render hash check will still pass. Then the StateKeep won't be working.

PR #3135, opened 2022-01-20
Fix: prioritize namespace flag for `vela up`

devholic:

Currently, CLI applies application to `default` namespace when there is no explicit namespace in application spec, even if the namespace flag is set.

This commit fixes issue by overriding the namespace if the namespace flag is set.

I have:

  - ~Q. How can I add a `backport` label?~

```sh
# Permanent link for vela.yaml in master branch as of 2022-01-20T08:30:42Z
export APP_SPEC_URL=https://raw.githubusercontent.com/oam-dev/kubevela/947455adb998426d56139f128b3a9cd7100e1d35/docs/examples/vela.yaml

# This will be applied to default namespace, since there is no explicit namespace in spec
vela up -f "${APP_SPEC_URL}"

# This will be applied to default namespace even if we override namespace to `some-other-namespace`
vela up -f "${APP_SPEC_URL}" -n some-other-namespace

# New binary should apply application to `some-other-namespace`
vela up -f "${APP_SPEC_URL}" -n some-other-namespace
```

I think overriding namespace in `NewUpCommand ` is better than extending `ApplyApplication` since it might be ambiguous  when taking it as a parameter. WDYT?

### DIFF
--- a/references/cli/up.go
+++ b/references/cli/up.go
@@ -68,6 +68,10 @@ func NewUpCommand(c common2.Args, order string, ioStream cmdutil.IOStreams) *cob
 			if err != nil {
 				return errors.Wrap(err, "File format is illegal, only support vela appfile format or OAM Application object yaml")
 			}
+			// override namespace if namespace flag is set
+			if namespace != "" {
+				app.SetNamespace(namespace)
+			}
 			err = common.ApplyApplication(app, ioStream, kubecli)
 			if err != nil {
 				return err
--- a/references/cli/up.go
+++ b/references/cli/up.go
@@ -68,8 +68,10 @@ func NewUpCommand(c common2.Args, order string, ioStream cmdutil.IOStreams) *cob
 			if err != nil {
 				return errors.Wrap(err, "File format is illegal, only support vela appfile format or OAM Application object yaml")
 			}
-			// override namespace if namespace flag is set
-			if namespace != "" {
+
+			// Override namespace if namespace flag is set. We should check if namespace is `default` or not
+			// since GetFlagNamespaceOrEnv returns default namespace when failed to get current env.
+			if namespace != "" && namespace != types.DefaultAppNamespace {
 				app.SetNamespace(namespace)
 			}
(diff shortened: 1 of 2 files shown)


wonderflow:
image

wonderflow:
The idea is fine, LGTM

wonderflow:
let's add a test case about it.

wonderflow (approved):

devholic:
I can't add label through sidebar since I have no permission..., BTW thanks for adding it 👍 

image

wonderflow:
> I can't add label through sidebar since I have no permission..., BTW thanks for adding it 👍
> 
> image

Can you add label when create PR?

devholic:
No, I cannot add label while creating PR like below:

image

wonderflow:
> No, I cannot add label while creating PR like below:
> 

Thanks, seems we need to refine the precess.

references/cli/up.go
```diff
@@ -68,6 +68,12 @@ func NewUpCommand(c common2.Args, order string, ioStream cmdutil.IOStreams) *cob
 			if err != nil {
 				return errors.Wrap(err, "File format is illegal, only support vela appfile format or OAM Application object yaml")
 			}
+
+			// Override namespace if namespace flag is set. We should check if namespace is `default` or not
+			// since GetFlagNamespaceOrEnv returns default namespace when failed to get current env.
+			if namespace != "" && namespace != types.DefaultAppNamespace {
```
devholic, line 74: I found an edge case that `GetFlagNamespaceOrEnv` returns `default` namespace if failed to call `GetCurrentEnv`. So I have to add `default` namespace check.

Could you double-check for this?

https://github.com/oam-dev/kubevela/blob/947455adb998426d56139f128b3a9cd7100e1d35/references/cli/env.go#L243-L249

zzxwill (approved):

wonderflow (approved):
LGTM, it makes sense! Thanks!

PR #3138, opened 2022-01-21
Feat: extend gateway trait to set class in spec

devholic:

[`kubernetes.io/ingress.class` annotation is deprecated](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation) in favor of `.spec.ingressClassName`. However, there is no way to set it through the gateway trait for now.

This commit extends the gateway trait by adding `classInSpec` to parameter. Forcing the use of `.spec.ingressClassName` makes sense, but some ingress controllers (including old versions) may not recognize this field. Therefore, set default value of `classInSpec` to `false` for backward compatibility.

I have:

  - https://github.com/oam-dev/kubevela.io/pull/474

```yaml
kind: Application
metadata:
  name: first-vela-app
spec:
  components:
    - name: express-server
      type: webservice
      properties:
        image: crccheck/hello-world
        port: 8000
      traits:
        - type: gateway
          properties:
            class: contour-private
            # set to `false` or remove for checking current behavior
            # or set to `true` to check ingress class is set to `.spec.ingressClassName`
            classInSpec: false
            domain: testsvc.example.com
            http:
              "/": 8000
```

None

### DIFF
--- a/charts/vela-minimal/templates/defwithtemplate/gateway.yaml
+++ b/charts/vela-minimal/templates/defwithtemplate/gateway.yaml
@@ -67,7 +67,7 @@ spec:
         	// +usage=Specify the class of ingress to use
         	class: *"nginx" | string
 
-        	// +usage=Set ingress class to '.spec.ingressClassName' instead of 'kubernetes.io/ingress.class' annotation.
+        	// +usage=Set ingress class in '.spec.ingressClassName' instead of 'kubernetes.io/ingress.class' annotation.
         	classInSpec: *false | bool
         }
   status:
--- a/charts/vela-minimal/templates/defwithtemplate/gateway.yaml
+++ b/charts/vela-minimal/templates/defwithtemplate/gateway.yaml
@@ -32,21 +32,30 @@ spec:
         	kind:       "Ingress"
         	metadata: {
         		name: context.name
-        		annotations: "kubernetes.io/ingress.class": parameter.class
+        		annotations: {
+        			if !parameter.classInSpec {
+        				"kubernetes.io/ingress.class": parameter.class
+        			}
+        		}
         	}
-        	spec: rules: [{
-        		host: parameter.domain
-        		http: paths: [
-        			for k, v in parameter.http {
-        				path:     k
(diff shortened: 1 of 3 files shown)


charts/vela-minimal/templates/defwithtemplate/gateway.yaml
```diff
@@ -57,6 +66,10 @@ spec:
 
         	// +usage=Specify the class of ingress to use
         	class: *"nginx" | string
+
+        	// +usage=Set ingress class to 'kubernetes.io/ingress.class' annotation instead of spec.
```
FogDong, line 70: Please make sure the comment is on one lone so that it can be recognized.

charts/vela-minimal/templates/defwithtemplate/gateway.yaml
```diff
@@ -57,6 +66,10 @@ spec:
...
+        	// Set to false if ingress controller supports '.spec.ingressClassName'.
+        	useClassAnnotation: *true | bool
```
FogDong, line 72: I think we can change `useClassAnnotation` to `classInSpec`, so the default value can be `false` which is more clear. WDYT?

wonderflow: classInSpec is better

devholic: I also think `classInSpec` is better. Thanks for your suggestion 🙂

wonderflow (approved):
good job， thanks

devholic:
@wonderflow Thanks for the quick review 👍 
Could you add `backport release-1.2` label to this PR? (I have no permission to do so 😢 - [reference](https://github.com/oam-dev/kubevela/pull/3135#issuecomment-1017320342))

wonderflow:
/backport 1.2


Maybe this command can also work without higher permission.

devholic:
> /backport 1.2
> 
> Maybe this command can also work without higher permission.

Thanks for letting me know. I'll try this at further PRs.

PR #3163, opened 2022-01-25
Fix: Use systemDefinitionNamespace core Helm Chart

o-otte:
Use value in Addon Registry CM and the Test Application

Some vela-core helm chart templates did not use the value `systemDefinitionNamespace`. This leads to issues when using a different namespace than vela-system. I search all charts for those hardcoded references and make them configurable via the value.

I didn't create an Issue for this yet. Should I? Seems like a really minor change.

I have:

Did `helm template vela-test . --set=systemDefinitionNamespace=sys-kubevela | grep vela-system` see that no vela-system namespace is used anymore and a `helm lint`


wonderflow (approved):
good catch！thanks

PR #3192, opened 2022-02-07
Feat: application support controller requirement

leejanee:
Signed-off-by: Jian.Li <lj176172@alibaba-inc.com>

The purpose is to realize the grayscale upgrade scheme based on the application. You can choose  the controller with a specific version  to handle your application. 

For example, the application as follow require that the version of the controller to handle must be `v1.2.0`. If the controller with version  not equal to `v1.2.0`,  it will ignore the application.
```
metadata:
  annotations:
    app.oam.dev/controller-version-require: v1.2.0
```

I have:

Notice:
1. The controller must be upgraded to a version that includes this pr. Only after that, the grayscale release of the controller should be performed.
2. When there are multiple versions of the controller in the cluster, make sure that only one version of the controller has the command line parameter `ignore-app-without-ctrl-req` turned off( be false).

### DIFF
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -130,7 +130,7 @@ func main() {
 		"The duration the LeaderElector clients should wait between tries of actions")
 	flag.BoolVar(&enableClusterGateway, "enable-cluster-gateway", false, "Enable cluster-gateway to use multicluster, disabled by default.")
 	flag.BoolVar(&controllerArgs.EnableCompatibility, "enable-asi-compatibility", false, "enable compatibility for asi")
-	flag.BoolVar(&controllerArgs.IgnoreAppWithoutControllerRequirement, "ignore-app-without-ctrl-req", false, "If true, application controller will not process the app without 'app.oam.dev/controller-version-require' annotation")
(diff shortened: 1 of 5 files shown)


cmd/core/main.go
```diff
@@ -130,6 +130,7 @@ func main() {
 		"The duration the LeaderElector clients should wait between tries of actions")
 	flag.BoolVar(&enableClusterGateway, "enable-cluster-gateway", false, "Enable cluster-gateway to use multicluster, disabled by default.")
 	flag.BoolVar(&controllerArgs.EnableCompatibility, "enable-asi-compatibility", false, "enable compatibility for asi")
+	flag.BoolVar(&controllerArgs.IgnoreAppWithoutControllerRequirement, "ignore-app-without-ctrl-req", false, "If true, application controller will not process the app without 'app.oam.dev/controller-version-require' annotation")
```
wonderflow, line 133:
```suggestion
	flag.BoolVar(&controllerArgs.IgnoreAppWithoutControllerRequirement, "ignore-app-without-controller-version", false, "If true, application controller will not process the app without 'app.oam.dev/controller-version-require' annotation")
```

wonderflow (commented):
We should add a doc about how to rollout the controller with this feature, it's better if we have CLI command for this.

我们要加个文档告知用户怎么去做KubeVela控制器的灰度发布，最好要有个 CLI 命令。

wonderflow (approved):

PR #3223, opened 2022-02-10
Feat: support standalone style multi-cluster

Somefive:

Several features are supported in this PR:
- **ref-objects typed component**: User can refer external objects as the component to be applied. This can be useful if user want to promote resources in staging cluster to production cluster. Both name and labelSelector are supported as follows.
```yaml
kind: Application
metadata:
  name: podinfo
  namespace: demo
spec:
  components:
    - name: podinfo
      type: ref-objects
      properties:
        objects:
          - apiVersion: apps/v1
            kind: Deployment
            name: podinfo
          - apiVersion: v1
            kind: ConfigMap
            selector:
              app: podinfo
```
- **new multi-cluster policy and workflow step**: The new `topology` policy, `override` policy and `deploy` WorkflowStep reworks the original `env-binding` policy and `deploy2env` WorkflowStep. The advantages of the new policies includes
  - Better decouple of configuration override and cluster specification. Configuration override can be split into different parts and references by `deploy` WorkflowStep concurrently. 
  - Cluster specification supports select clusters by labels. (But do not support namespace in this PR, which will be implemented soon.) 

>  Notice that previous env-binding policy requires traits in patch need to be pre-declared in the original base component but now this is not required. 

  See examples with the external policy & workflow in the next section.
- **external policy & workflow**: Users can write policy and workflow spec outside application as individual objects and reference them in application spec (can be used combined with policies inside application) as follows.
```yaml
kind: Application
metadata:
  name: podinfo
  namespace: demo
spec:
  components:
    - name: podinfo
      type: ref-objects
      properties:
        objects:
          - apiVersion: apps/v1
            kind: Deployment
            name: podinfo
  policies:
    - type: topology
      name: topology-beijing
      properties:
        clusters: ["beijing-1"]
  workflow:
    ref: deploy-podinfo
---
apiVersion: core.oam.dev/v1alpha1
kind: Workflow
metadata:
  name: deploy-podinfo
  namespace: demo
steps:
  - type: deploy
    name: deploy-beijing
    properties:
      policies: ["topology-beijing"]
  - type: deploy
    name: deploy-us-west
    properties:
      auto: false
      policies: ["override-high-availability", "topology-us-west"]
---
apiVersion: core.oam.dev/v1alpha1
kind: Policy
metadata:
  name: override-high-availability
  namespace: demo
type: override
properties:
  components:
    - name: podinfo
      traits:
      - type: scaler
        properties:
          replicas: 5
---
apiVersion: core.oam.dev/v1alpha1
kind: Policy
metadata:
  name: topology-us-west
  namespace: demo
type: topology
properties:
  clusterSelector:
    region: us-west
```
- **admission control for dispatching resources**: Now KubeVela controller can set two admission check during dispatching resources. One is namespace check and the other is resource type check.
  - If `--allow-cross-namespace-resource=false` is added to KubeVela controller bootstrap parameters, resources outside the application's namespace will be forbidden.
  - If `--allow-resource-types=whitelist:Deployment.v1.apps,ConfigMap.v1` is added to KubeVela controller bootstrap parameters, only Deployment and ConfigMap can be dispatched.
  - If `--allow-resource-types=blacklist:Deployment.v1.apps,ConfigMap.v1` is added to KubeVela controller bootstrap parameters, Deployment and ConfigMap will not be able to dispatch.
- **extension to cue patch**: Currently, the patch for cue values can override existing field only if `patchStrategy=retainKeys` be declared on this field. This will make recursive override hard. This PR support a new strategy named open, which indicates that all fields can be overrode. Examples like patch trait (which can be used in override policy) is as follows
```cue
patch: {
	type: "trait"
	annotations: {}
	labels: {}
	description: "Patch the output directly."
	attributes: {
		podDisruptive: true
		appliesToWorkloads: ["*"]
	}
}
template: {
	parameter: {...}
	// +patchStrategy=open
	patch: parameter
}
```

**Minor changes**: Previous PolicyDefinition uses the short name of `policy`, which is conflict with the new external policy CRD. Change it to `def-policy`.

I have:


charts/vela-core/crds/core.oam.dev_policydefinitions.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     listKind: PolicyDefinitionList
     plural: policydefinitions
     shortNames:
-    - policy
+    - policy-def
```
wonderflow, line 18:
```suggestion
    - def-policy
```

Start from def is better

pkg/workflow/providers/oam/apply.go
```diff
@@ -146,10 +206,82 @@ func lookUpValues(v *value.Value) (*common.ApplicationComponent, *value.Value, s
 	return comp, patcher, clusterName, overrideNamespace, env, nil
 }
 
+func (p *provider) loadDynamicComponent(comp *common.ApplicationComponent) (*common.ApplicationComponent, error) {
+	if comp.Type == "ref-objects" {
```
wonderflow, line 210: generally, we return the smaller logic first to avoid complex, so here we recommand:

```
if comp.Type!="ref-objects"{
  return ...
}

_comp:==comp.DeepCopy()
...
```

Somefive: OK

apis/core.oam.dev/v1beta1/application_types.go
```diff
@@ -68,6 +68,7 @@ type WorkflowStep struct {
 
 // Workflow defines workflow steps and other attributes
 type Workflow struct {
+	Ref   string         `json:"ref,omitempty"`
 	Steps []WorkflowStep `json:"steps,omitempty"`
```
wonderflow, line 72: have we checked these two fields are exclusive with each other?

apis/core.oam.dev/v1beta1/policy_definition.go
```diff
@@ -61,7 +61,7 @@ func (d *PolicyDefinition) GetCondition(conditionType condition.ConditionType) c
 // +kubebuilder:object:root=true
 
 // PolicyDefinition is the Schema for the policydefinitions API
-// +kubebuilder:resource:scope=Namespaced,categories={oam},shortName=policy
+// +kubebuilder:resource:scope=Namespaced,categories={oam},shortName=def-policy
```
wonderflow, line 64: please make this change in PR description.

pkg/appfile/appfile.go
```diff
@@ -212,14 +212,6 @@ func (af *Appfile) PrepareWorkflowAndPolicy(ctx context.Context) ([]*unstructure
...
 	}
 
-	af.WorkflowSteps, err = step.NewChainWorkflowStepGenerator(
-		&step.Deploy2EnvWorkflowStepGenerator{},
-		&step.ApplyComponentWorkflowStepGenerator{},
-	).Generate(af.app, af.WorkflowSteps)
-
-	if err != nil {
-		return nil, err
-	}
```
wonderflow, line 222: why these code removed?

pkg/workflow/providers/oam/apply.go
```diff
@@ -90,6 +97,11 @@ func (p *provider) ApplyComponent(ctx wfContext.Context, v *value.Value, act wfT
 		return err
 	}
 
+	if mu != nil {
```
barnettZQG, line 100: Is the Cue not thread-safe?

pkg/workflow/providers/oam/apply.go
```diff
@@ -117,6 +129,54 @@ func (p *provider) ApplyComponent(ctx wfContext.Context, v *value.Value, act wfT
...
+	err = components.StepByFields(func(name string, in *value.Value) (bool, error) {
+		ch <- struct{}{}
+		wg.Add(1)
+		go func(_name string, _in *value.Value) {
+			defer func() {
+				wg.Done()
+				<-ch
+			}()
+			if err := p.applyComponent(ctx, _in, act, mu); err != nil {
+				mu.Lock()
```
barnettZQG, line 163: Why do we need a lock here?

pkg/appfile/appfile.go
Somefive: They are moved to parser.go:115 so that external policy and referenced override component/trait policy can be counted in application revision.

pkg/workflow/providers/oam/apply.go
Somefive: No, it is not thread safe.

pkg/workflow/providers/oam/apply.go
Somefive: It is not thread safe. Without lock, data race could happen.

charts/vela-core/templates/defwithtemplate/ref-objects.yaml
```diff
@@ -0,0 +1,25 @@
...
+      template: |
+        output: parameter.objects[0]
+        outputs: {
+        	for i, v in parameter.objects {
+        		if i > 0 {
+        			"objects-\(i)": v
+        		}
+        	}
+        }
+        parameter: objects: [...{}]
```
wonderflow, line 22: we should probably use something like K8sObject to do some validation:

```
parameter: objects: [...#K8sObject]
#K8sObject :{
   apiVersion: string
   kind: string
   ...
}
```

pkg/resourcekeeper/admission_test.go
```diff
@@ -0,0 +1,81 @@
...
+			},
+		},
+	}}
+	AllowResourceTypes = "blacklist:Service.v1,Secret.v1"
+	err := (&ResourceTypeAdmissionHandler{}).Validate(context.Background(), objs)
+	r.NotNil(err)
+	r.Contains(err.Error(), "forbidden resource")
+	AllowResourceTypes = "blacklist:ConfigMap.v1,Deployment.v1.apps"
+	r.NoError((&ResourceTypeAdmissionHandler{}).Validate(context.Background(), objs))
+	AllowResourceTypes = "whitelist:ConfigMap.v1,Deployment.v1.apps"
```
wonderflow, line 75: usually the k8s community uses `group/version.Kind` 

e.g. `apps/v1.Deployment`

pkg/resourcekeeper/admission_test.go
```diff
@@ -0,0 +1,81 @@
...
+		Object: map[string]interface{}{
+			"apiVersion": "v1",
+			"kind":       "Secret",
+			"metadata": map[string]interface{}{
+				"name":      "demo",
+				"namespace": "demo",
+			},
+		},
+	}}
+	AllowResourceTypes = "blacklist:Service.v1,Secret.v1"
```
wonderflow, line 69: the `blacklist` and `whitelist` seems too long, what about `white`/ `black`, or `pass`/`forbidden`, or `w`/`b`

pkg/workflow/providers/multicluster/multicluster.go
```diff
@@ -159,6 +161,84 @@ func (p *provider) ListClusters(ctx wfContext.Context, v *value.Value, act wfTyp
 	return v.FillObject(clusters, "outputs", "clusters")
 }
 
+func (p *provider) ExpandTopology(ctx wfContext.Context, v *value.Value, act wfTypes.Action) error {
+	policiesRaw, err := v.LookupValue("inputs", "policies")
```
wonderflow, line 165: let's use const for these string

pkg/resourcekeeper/admission_test.go
Somefive: I think making it shorten might be a bit unclear. With a growing number of complex configurations, I think one of the solution is to use a separate configuration file, but might not necessary now in this PR?

pkg/workflow/providers/multicluster/multicluster.go
Somefive: It is a style problem for all cue providers. If const should be used here, we might need to change all the cue providers to use the same style.

pkg/resourcekeeper/admission_test.go
Somefive: This is following the [apimachinery](https://github.com/kubernetes/apimachinery/blob/2936d3f03931b9c06a641799605d8e806cb2a58b/pkg/runtime/schema/group_version.go#L44) process style. This configuration uses the same format with the previous `--optimized-cached-gvks` parameter, which is handled [here](https://github.com/oam-dev/kubevela/blob/342b85331553520f4d70e37ff80b393f606c71fa/pkg/client/controller_client.go#L72). If `apps/v1.Deployment` needs to be used, we should change all of them.

pkg/workflow/providers/oam/apply.go
```diff
@@ -117,6 +129,54 @@ func (p *provider) ApplyComponent(ctx wfContext.Context, v *value.Value, act wfT
...
+	parallelism, err := v.GetInt64("parallelism")
+	if err != nil {
+		return err
+	}
+	if parallelism <= 0 {
+		return errors.Errorf("parallelism cannot be smaller than 1")
+	}
+	mu := &sync.Mutex{}
+	var wg sync.WaitGroup
+	ch := make(chan struct{}, parallelism)
```
wonderflow, line 152: why we need using both this `channel` and `waitGroup`

pkg/workflow/providers/oam/apply.go
```diff
@@ -117,6 +129,54 @@ func (p *provider) ApplyComponent(ctx wfContext.Context, v *value.Value, act wfT
...
+		wg.Add(1)
+		go func(_name string, _in *value.Value) {
+			defer func() {
+				wg.Done()
+				<-ch
+			}()
+			if err := p.applyComponent(ctx, _in, act, mu); err != nil {
+				mu.Lock()
+				errs = append(errs, errors.Wrapf(err, "failed to apply component %s", _name))
+				mu.Unlock()
```
wonderflow, line 165: you can using a channel to send signal for error or success, then you can avoid using waitGroup and this mutex. Just receive the signal outside.

Somefive: This mutex is not only used in L163 and L165. It is also used inside applyComponent while handling ResourceTracker. I think channel can handle the return of error but not be able to handle the concurrency problem inside applyComponent.

pkg/workflow/step/generator.go
```diff
@@ -54,6 +57,27 @@ func NewChainWorkflowStepGenerator(generators ...WorkflowStepGenerator) Workflow
...
+	context.Context
+	client.Client
+}
+
+// Generate generate workflow steps
+func (g *RefWorkflowStepGenerator) Generate(app *v1beta1.Application, existingSteps []v1beta1.WorkflowStep) (steps []v1beta1.WorkflowStep, err error) {
+	if len(existingSteps) > 0 {
+		return existingSteps, nil
+	}
+	if app.Spec.Workflow != nil && app.Spec.Workflow.Ref != "" {
```
wonderflow, line 71:
```suggestion
        if app.Spec.Workflow == nil || app.Spec.Workflow.Steps != nil {
             return existingSteps, nil
        } 
	if  app.Spec.Workflow.Ref == "" {
	    return errors.New("either workflow steps or ref should not be empty")
	}
```

you should not judge by if there's existingSteps, you should judge by if the app.Spec.Workflow is not nil.

pkg/workflow/step/generator.go
```diff
@@ -102,3 +126,58 @@ func (g *Deploy2EnvWorkflowStepGenerator) Generate(app *v1beta1.Application, exi
...
 }
+
+// DeployWorkflowStepGenerator generate deploy workflow steps for all topology & override in the application
+type DeployWorkflowStepGenerator struct{}
+
+// Generate generate workflow steps
+func (g *DeployWorkflowStepGenerator) Generate(app *v1beta1.Application, existingSteps []v1beta1.WorkflowStep) (steps []v1beta1.WorkflowStep, err error) {
+	if len(existingSteps) > 0 {
+		return existingSteps, nil
+	}
```
wonderflow, line 137: why? Should we append deploy steps at the end?

pkg/resourcekeeper/admission_test.go
wonderflow: well, this is also fine.

pkg/resourcekeeper/admission_test.go
wonderflow: it's fine for me.

pkg/workflow/providers/multicluster/multicluster.go
wonderflow: yes, we can do that in the following PRs, and this is necessary

pkg/workflow/providers/oam/apply.go
wonderflow: at lease we don't need the channel here now.

pkg/workflow/step/generator.go
Somefive: Not exactly. If ref and steps are both null, we will generate workflow steps according to Components and Policies.

Somefive: But we should add logic to avoid user setting workflow ref and steps at the same time here. I add it now.

pkg/workflow/providers/oam/apply.go
wonderflow: 1. In my view, we don't need to limmit the concurrency in the control plane, this won't be large throughput. You can leave it to the go runtime.
2. If you really want to control the parallelism，you can just use a channel, like below:

```
// ApplyComponents apply components in parallel.
func (p *provider) ApplyComponents(ctx wfContext.Context, v *value.Value, act wfTypes.Action) error {
	components, err := v.LookupValue("components")
	if err != nil {
		return err
	}
	parallelism, err := v.GetInt64("parallelism")
	if err != nil {
		return err
	}
	if parallelism <= 0 {
		return errors.Errorf("parallelism cannot be smaller than 1")
	}
	mu := &sync.Mutex{}
	type sig struct {
		err error
		msg string
	}
	ch := make(chan sig, parallelism+1)
	go func() {
		err = components.StepByFields(func(name string, in *value.Value) (bool, error) {
			go func(_name string, _in *value.Value) {
				subErr := p.applyComponent(ctx, _in, act, mu)
				ch <- sig{
					err: subErr,
					msg: "component: "+ _name,
				}
			}(name, in)
			return false, nil
		})
		if err != nil {
			ch <- sig{
				err: err,
				msg: "failed to looping over components",
			}
		} else {
			ch <- sig{msg: "main"}
		}
	}()
for _,sig:=range ch{
// collect error and wait for signal here	
}
	return nil
}
```

Somefive: There will be another PR to handle the concurrent execution framework problem.

wonderflow (approved):

FogDong (approved):

pkg/workflow/providers/oam/apply.go
```diff
@@ -146,10 +206,82 @@ func lookUpValues(v *value.Value) (*common.ApplicationComponent, *value.Value, s
...
 
+func (p *provider) loadDynamicComponent(comp *common.ApplicationComponent) (*common.ApplicationComponent, error) {
+	if comp.Type != "ref-objects" {
+		return comp, nil
+	}
+	_comp := comp.DeepCopy()
+	props := &struct {
+		Objects []struct {
+			APIVersion string            `json:"apiVersion"`
+			Kind       string            `json:"kind"`
```
yue9944882, line 217: `Kind` is not the proper resource identifier, it's used for serialize/deserialize general kubernetes resources. we dont involve codec if we're merely referencing them. in addition to that, the version is neglectable, b/c api-discovery will suggest the preferred version. we should use `group` + `resource`(i.e. plural lowercased form) + `name` + `namespace` (optional) to identify the resources

pkg/workflow/providers/oam/apply.go
```diff
@@ -146,10 +206,82 @@ func lookUpValues(v *value.Value) (*common.ApplicationComponent, *value.Value, s
...
+			Name       string            `json:"name,omitempty"`
+			Selector   map[string]string `json:"selector,omitempty"`
```
yue9944882, line 219: obscure naming, is it field-selector or the label? either has built-in support in kubernetes

pkg/workflow/providers/oam/apply.go
```diff
@@ -146,10 +206,82 @@ func lookUpValues(v *value.Value) (*common.ApplicationComponent, *value.Value, s
 	return comp, patcher, clusterName, overrideNamespace, env, nil
 }
 
+func (p *provider) loadDynamicComponent(comp *common.ApplicationComponent) (*common.ApplicationComponent, error) {
+	if comp.Type != "ref-objects" {
```
yue9944882, line 210: nit: conventionally all the types should be camel-cased, we dont mix camel case and snake-case in one type

pkg/workflow/providers/oam/apply.go
```diff
@@ -146,10 +206,82 @@ func lookUpValues(v *value.Value) (*common.ApplicationComponent, *value.Value, s
...
+		un.SetResourceVersion("")
+		un.SetGeneration(0)
+		un.SetOwnerReferences(nil)
+		un.SetDeletionTimestamp(nil)
+		un.SetManagedFields(nil)
+		un.SetUID("")
+		unstructured.RemoveNestedField(un.Object, "metadata", "creationTimestamp")
+		unstructured.RemoveNestedField(un.Object, "status")
+		// TODO(somefive): make the following logic more generalizable
+		if un.GetKind() == "Service" && un.GetAPIVersion() == "v1" {
```
yue9944882, line 236: NB: in addition to "Service", any built-in or custom kubernetes types w/ defaulting or immutability will need the hacks below, that's why i was proposing using work api to deploy resource (in another place)

pkg/workflow/providers/oam/apply.go
yue9944882: golang doesnt support "union" primitives, however in kubernetes the conventional way of defining union types is combing a union discriminator and serveral pointers to structs, e.g.:

```
struct {
  Type SomeType # [Foo|Bar]
  Foo *Foo
  Bar *Bar
}
```

see the original proposal about api-union at: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1027-api-unions

the implementation is gonna cover CRD in the foreseeable future, currently all the built-in types in kubernetes has conformed w/ the union structure

Somefive: snake-case for component(or trait/workflow-step) type is a history design in KubeVela. So using camel case here will cause inconsistency.

apis/core.oam.dev/v1alpha1/policy_types.go
```diff
@@ -0,0 +1,36 @@
...
+	// TopologyPolicyType refers to the type of topology policy
+	TopologyPolicyType = "topology"
+	// OverridePolicyType refers to the type of override policy
+	OverridePolicyType = "override"
+)
+
+// TopologyPolicySpec defines the spec of topology policy
+type TopologyPolicySpec struct {
+	Clusters        []string          `json:"clusters,omitempty"`
+	ClusterSelector map[string]string `json:"clusterSelector,omitempty"`
```
yue9944882, line 29: the name of selector also looks obscure to me, is it the label selecting or the claim selecting? the latter is accepted by sig-mc as a standard

apis/core.oam.dev/v1alpha1/policy_types.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// OverridePolicySpec defines the spec of override policy
+type OverridePolicySpec struct {
+	Components []EnvComponentPatch `json:"components,omitempty"`
+	Selector   []string            `json:"selector,omitempty"`
```
yue9944882, line 35: ditto

fourierr:
是否考虑保留原来policy里component selector的功能呢，比如在一个workflowstep中只想选中单个component部署到cluster1、cluster2，而不是将所有component一块部署

apis/core.oam.dev/v1alpha1/external_types.go
```diff
@@ -0,0 +1,75 @@
...
+// +kubebuilder:printcolumn:name="TYPE",type=string,JSONPath=`.type`
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type Policy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Type string `json:"type"`
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Properties *runtime.RawExtension `json:"properties,omitempty"`
```
yue9944882, line 40: ^^^ see the ref'd desgin about api-unions in my previous comment (https://github.com/oam-dev/kubevela/pull/3223/files#r819299546) the raw-extension is not designed for this, see `+kubebuilder:validation:EmbeddedResource` for the proper usage

fourierr:
是否考虑保留原来policy里component selector的功能呢，比如我要在一个workflowstep中只想选中部分component部署到cluster1、cluster2，而不是一块将所有component一块部署，

apis/core.oam.dev/v1alpha1/external_types.go
```diff
@@ -0,0 +1,75 @@
...
+// Workflow is the Schema for the policy API
+// +kubebuilder:storageversion
+// +kubebuilder:resource:categories={oam}
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type Workflow struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Steps []v1beta1.WorkflowStep `json:"steps,omitempty"`
```
yue9944882, line 64: cross version type imports should be prohibitated

apis/core.oam.dev/v1alpha1/policy_types.go
```diff
@@ -0,0 +1,36 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+const (
+	// TopologyPolicyType refers to the type of topology policy
+	TopologyPolicyType = "topology"
```
yue9944882, line 21: nit: stick w/ single casing

apis/core.oam.dev/v1alpha1/policy_types.go
Somefive: It is label selecting, which is a direct way for user to use. One of the lack here is we still do not have convenient `vela cluster label` command to help user manage cluster labels.

Somefive:
> 是否考虑保留原来policy里component selector的功能呢，比如在一个workflowstep中只想选中单个component部署到cluster1、cluster2，而不是将所有component一块部署

有的，这个能力保留了，这个 Override Policy 本质上就是原来 EnvBinding 的 Patch，底层是一致的。

yue9944882 (commented):
a few outstanding comments about the API types. i wasnt able to look into the controller implementation, but generally it's recommended to split the big PR into two: one for the API-changes, another enfleshes the implementaion.

apis/core.oam.dev/v1alpha1/external_types.go
Somefive: Yes. It is currently hard to let the new CRD use the name Policy & Workflow but still keep compatibility for the previous code. I agree we should not do this.

pkg/workflow/providers/oam/apply.go
Somefive: The component type here is not a literal type. Except the new added `ref-objects`, other component type matches a ComponentDefinition where cue is used inside to describe the details of rendering component into resources. So here it is not a union type design.

fourierr:
> > 是否考虑保留原来policy里component selector的功能呢，比如在一个workflowstep中只想选中单个component部署到cluster1、cluster2，而不是将所有component一块部署
> 
> 有的，这个能力保留了，这个 Override Policy 本质上就是原来 EnvBinding 的 Patch，底层是一致的。

我的意思是，EnvBinding里placement是否支持component selector呢，也就是通过placement指定某个集群，通过component selector选择某一个组件，

PR #3285, opened 2022-02-17
Fix(charts): concurrent reconciles configurable

jrkeen:
Signed-off-by: jrkeen <jrkeen@hotmail.com>

Fixes #

I have:


jrkeen:
cc @wonderflow

charts/vela-core/templates/kubevela-controller.yaml
```diff
@@ -135,6 +135,7 @@ spec:
             - "--enable-cluster-gateway"
             {{ end }}
             - "--application-re-sync-period={{ .Values.controllerArgs.reSyncPeriod }}"
+            - "--concurrent-reconciles={{ .Values.concurrentReconciles }}"
```
wonderflow, line 138: So, this is a mistake that we forget to set args while we have this value in chart?  @Somefive @yangsoon

Somefive: Seems like a bug. We have this parameter in values.yaml but not use it. This change looks reasonable for me. /cc @yangsoon

Somefive (commented):
Recommend to apply this fix to vela-minimal as well.

charts/vela-core/templates/kubevela-controller.yaml
yangsoon: this seems forget to set, LGTM

jrkeen:
> Recommend to apply this fix to vela-minimal as well.

PTAL

wonderflow (approved):
good catch, thanks

PR #3359, opened 2022-03-03
Feat: enhance container related traits

Somefive:

The current env trait can only **patch** environment variables to the container with the name of current component.
This PR extends this trait and make it support the following features:
1. User can specify which container to patch by setting `name`.
2. User can clean up all existing environment variables by setting `replace: true`.
3. User can unset some existing environment variables by setting `unset: ["existing-env"]`
4. User can set environment variables to multiple containers by setting `containers: [...]`

Similarly, the `container-image` trait and `command` trait are developed that allows user to modify existing image/imagePullPolicy/command/args in workload.

See docs/examples/traits/ for full examples.

Another feature introduce in this PR is that now the `errs` field output by trait can be exposed to user directly (instead of relying on the cue error, which is often hard to understand). For example, in the env trait, if the container name is set to a non-existing container, the error in the Application will report:

`step apply: run step(provider=oam,do=component-apply): GenerateComponentManifest: evaluate template trait=env app=busybox: container busybox-bad not found`.

BTW, to achieve the `replace` effect, a new PatchStrategy **replace** is added to support the override of list.

I have:


charts/vela-core/templates/defwithtemplate/env.yaml
```diff
@@ -15,18 +15,88 @@ spec:
...
+        				PatchContainer & {_args: c}
+        			}
+        		}]
+        	}
+        }
+        parameter: {
+        	#PatchArgs
+        	containers?: [...#PatchArgs]
         }
-        parameter: env: [string]: string
```
FogDong, line 31: It is best to keep the previous definition and make it deprecated and then create a new one, otherwise, it will affect the users who use the old definition.

FogDong: Ah, I see, the previous format is still supported.

vela-templates/definitions/internal/trait/env.cue
```diff
@@ -8,18 +8,90 @@ env: {
...
+	#PatchArgs: {
+		name:    *"" | string
+		replace: *false | bool
+		env: [string]: string
+		unset: *[] | [...string]
+	}
+	PatchContainer: {
+		_args: #PatchArgs
+		name:  string
+		if _args.name != "" {
```
FogDong, line 20: Our schema generator may not support arguments in this format currently.

charts/vela-core/templates/defwithtemplate/env.yaml
Somefive: Right, I was also thinking of making this trait `container-env`, which is more accurate. But it might be better to keep them converge together.

FogDong (approved):

wonderflow (approved):

PR #3393, opened 2022-03-08
Feat: support gitee repository in addon registry

fourierr:
Signed-off-by: fourierr <maxiangboo@163.com>

support gitee repository in addon registry, and the url pattern supports two valid format:
`https://gitee.com/<owner>/<repo>/tree/<branch>/<path-to-dir>`
`https://gitee.com/<owner>/<repo>/<path-to-dir>`

I have:


pkg/addon/addon.go
```diff
@@ -418,15 +424,74 @@ func createGitHelper(content *utils.Content, token string) *gitHelper {
...
+	}
+	response, err := c.Client.Do(req.WithContext(ctx))
+	defer func() {
+		if err := response.Body.Close(); err != nil {
+			fmt.Println(err)
+		}
+	}()
+	if err != nil {
+		return nil, nil, err
+	}
```
wonderflow, line 479:
```suggestion
	response, err := c.Client.Do(req.WithContext(ctx))
	if err != nil {
		return nil, nil, err
	}
	defer response.Body.Close()
```

Usually we check the error first, if the err is not nil, the response can be nil. Then just defer close the body is fine, the error can be ignored and should not print out.

pkg/addon/addon.go
```diff
@@ -418,15 +424,74 @@ func createGitHelper(content *utils.Content, token string) *gitHelper {
...
+	content, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, nil, err
+	}
+	fileUnmarshalError := json.Unmarshal(content, &fileContent)
+	if fileUnmarshalError == nil {
+		return fileContent, nil, nil
```
wonderflow, line 486: what does it mean when the unmarshal failed?  why not return error? let's add some explanation in the code

pkg/addon/addon.go
```diff
@@ -418,15 +424,74 @@ func createGitHelper(content *utils.Content, token string) *gitHelper {
...
+	}
+	directoryUnmarshalError := json.Unmarshal(content, &directoryContent)
+	if directoryUnmarshalError == nil {
+		return nil, directoryContent, nil
+	}
```
wonderflow, line 491: maybe it's better to make these lines in a single function, and write some unit tests for it.

pkg/addon/reader_gitee_test.go
```diff
@@ -0,0 +1,112 @@
...
+// baseURLPath is a non-empty Client.BaseURL path to use during tests,
+// to ensure relative URLs are used for all endpoints. See issue #752.
+//baseURLPath = "/api-v3"
+)
+
+//var (
+//	//go:embed testdata
+//	testdata       embed.FS
+//	testdataPrefix = "testdata"
+//)
```
wonderflow, line 44: the useless code should be deleted.

pkg/addon/addon.go
fourierr: ok, I will add the boundary error case

zzxwill:
@wangyikewxgm Why do we still need to support Gitee as we have supported git protocol?

wonderflow (approved):
great job!

wonderflow:
@zzxwill  the git protocal only cover github for now

barnettZQG (approved):
LGTM

PR #3409, opened 2022-03-10
Feat: enhance topology policy

Somefive:

This PR enhances topology policy through several aspects.
1. Allow local cluster as VirtualCluster. Now user can use local cluster in topology policy like previous env-binding policy.
2. Allow specifying namespace in topology policy. Now user can use a different namespace other than the original application namespace like previous env-binding policy. Notice that this will only work when bootstrap parameter `--allow-cross-namespace-resource` is set.
3. Allow directly using `targets` in topology policy. User can directly use targets in topology policy. *Target reference is not included in this PR.* /cc @barnettZQG 
4. Rename `clusterSelector` to `clusterLabelSelector` to make it more clear. `clusterSelector` is treated as an alias to `clusterLabelSelector`. /cc @yue9944882 

Example:
```yaml
kind: Application
metadata:
  name: example
  namespace: example-ns
spec:
  components:
    - name: nginx
      type: webservice
      properties:
        image: nginx
  policies:
    - name: topology-beijing-demo
      type: topology
      properties:
        clusterLabelSelector:
          region: beijing
        namespace: demo
    - name: topology-local
      type: topology
      properties:
        targets: ["local/demo", "local/example-demo"]
  workflow:
    steps:
      - type: deploy
        name: deploy-all
        properties:
          policies: ["topology-local", "topology-beijing-demo"]
```

I have:

### DIFF
--- a/apis/core.oam.dev/v1alpha1/policy_types.go
+++ b/apis/core.oam.dev/v1alpha1/policy_types.go
@@ -25,8 +25,24 @@ const (
 
 // TopologyPolicySpec defines the spec of topology policy
 type TopologyPolicySpec struct {
-	Clusters        []string          `json:"clusters,omitempty"`
+	// Clusters directly specify which cluster to use
+	Clusters []string `json:"clusters,omitempty"`
+	// ClusterLabelSelector label selector for cluster
+	ClusterLabelSelector map[string]string `json:"clusterLabelSelector,omitempty"`
+	// ClusterSelector is an alias to clusterLabelSelector
 	ClusterSelector map[string]string `json:"clusterSelector,omitempty"`
+	// Namespace specify the namespace for override, used along which clusters/clusterLabelSelector/clusterSelector
+	Namespace string `json:"namespace,omitempty"`
+	// Targets specify the delivery targets, used alone
+	Targets []string `json:"targets,omitempty"`
+}
+
+// GetClusterLabelSelector primary use of clusterLabelSelector
+func (in *TopologyPolicySpec) GetClusterLabelSelector() map[string]string {
+	if in.ClusterLabelSelector != nil {
+		return in.ClusterLabelSelector
+	}
+	return in.ClusterSelector
 }
 
 // OverridePolicySpec defines the spec of override policy
--- a/pkg/multicluster/virtual_cluster.go
+++ b/pkg/multicluster/virtual_cluster.go
@@ -34,6 +34,8 @@ import (
 )
 
 const (
+	// CredentialTypeInternal identifies the virtual cluster from internal kubevela system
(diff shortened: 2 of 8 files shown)


wonderflow (approved):

apis/core.oam.dev/v1alpha1/policy_types.go
```diff
@@ -25,8 +25,24 @@ const (
...
+	// ClusterSelector is an alias to clusterLabelSelector
 	ClusterSelector map[string]string `json:"clusterSelector,omitempty"`
+	// Namespace specify the namespace for override, used along which clusters/clusterLabelSelector/clusterSelector
+	Namespace string `json:"namespace,omitempty"`
+	// Targets specify the delivery targets, used alone
+	Targets []string `json:"targets,omitempty"`
+}
+
+// GetClusterLabelSelector primary use of clusterLabelSelector
+func (in *TopologyPolicySpec) GetClusterLabelSelector() map[string]string {
```
yue9944882, line 41: keep the api struct neat and stateless, offload the member function to a static helper and move it to the helper package, e.g.:

```
func GetClusterLabelSelectorWithCompatibility(in *TopologyPolicySpec) map[string]string {
}
```

pkg/multicluster/virtual_cluster.go
```diff
@@ -34,6 +34,8 @@ import (
 )
 
 const (
+	// CredentialTypeInternal identifies the virtual cluster from internal kubevela system
+	CredentialTypeInternal v1alpha1.CredentialType = "Internal"
```
yue9944882, line 38: nit: move enum type definitions to the api package

pkg/multicluster/virtual_cluster_test.go
```diff
@@ -108,7 +108,7 @@ var _ = Describe("Test Virtual Cluster", func() {
 
 		vcs, err := ListVirtualClusters(ctx, k8sClient)
 		Expect(err).Should(Succeed())
-		Expect(len(vcs)).Should(Equal(3))
+		Expect(len(vcs)).Should(Equal(4))
```
yue9944882, line 111: this is suspicious

yue9944882 (commented):
follow-up: plz gate the compatibility w/ feature-gates, just copy/pasting the package from https://github.com/oam-dev/cluster-gateway/blob/master/pkg/featuregates/featue_gate.go will work

pkg/multicluster/virtual_cluster_test.go
Somefive: It is because we add `local` cluster as a default cluster for control plane, so the number of all virtual clusters is increased by 1.

Somefive:
This PR is broken in to #3436 and #3438

PR #3423, opened 2022-03-11
Feat: support query applied resources by velaQL

barnettZQG:

1. new cue provider function `ql.#ListAppliedResources`, list applied resources of an application, they read from ResourceTracker.
2. new velaQL view  `service-applied-resources-view`.
3. support filter service endpoints by component name. 

I have:

### DIFF
--- a/pkg/velaql/providers/query/handler.go
+++ b/pkg/velaql/providers/query/handler.go
@@ -107,6 +107,29 @@ func (h *provider) ListResourcesInApp(ctx wfContext.Context, v *value.Value, act
 	return v.FillObject(appResList, "list")
 }
 
+// ListAppliedResources list applied resource from tracker
+func (h *provider) ListAppliedResources(ctx wfContext.Context, v *value.Value, act types.Action) error {
+	val, err := v.LookupValue("app")
+	if err != nil {
+		return err
+	}
+	opt := Option{}
+	if err = val.UnmarshalTo(&opt); err != nil {
+		return err
+	}
+	collector := NewAppCollector(h.cli, opt)
+	app := new(v1beta1.Application)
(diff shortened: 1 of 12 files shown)


pkg/velaql/providers/query/handler.go
```diff
@@ -107,6 +107,29 @@ func (h *provider) ListResourcesInApp(ctx wfContext.Context, v *value.Value, act
 	return v.FillObject(appResList, "list")
 }
 
+// ListAppliedResources list applied resource from tracker
+func (h *provider) ListAppliedResources(ctx wfContext.Context, v *value.Value, act types.Action) error {
```
Somefive, line 111: I think we can use ListApplicationResources. Since the `appliedResources` usually refers to the .Status.appliedResource in application, the query here is finding resources in ResourceTracker, it is a bit confusing to use `ListAppliedResources` here.

barnettZQG: This function's purpose is to replace `.Status. appliedResource`. so, I used the same name.

Somefive: But we still need .Status.appliedResource outside VelaQL right? If you can completely remove the existence of appliedResource, it would be ok to reuse the name. Actually, when I first saw this function, I was guessing this VelaQL provider is trying to load resources from .Status.appliedResources. It is a bit ambiguous. On the other hand, since this function is not exposed to user, the ambiguity only exists for KubeVela developers. With the comments above the function, it is ok to keep this name.

Somefive (approved):

wonderflow (approved):

leejanee (approved):

PR #3434, opened 2022-03-13
Feat: reconcile app with scoped permissions

### LINKED ISSUE #3432: [Feature] Reconcile Application with scoped permissions

Currently, all `Application` resources are reconciled by the Roles bound to the controller service account. This behavior gives us the power to manage resources across multiple namespaces.

However, this behavior can be problematic in the soft-multitenancy environment (e.g. Namespaces as a Service or shared cluster across teams in a single organization). For example,

- Team `A` has permission on namespace `team-a-dev` (through RBAC)
- Team `B` has permission on namespace `team-b-dev` and `team-b-staging` (through RBAC)

Team `A` can deploy to or read resources in all namespaces (including non-authorized namespaces - `team-b-dev`, `team-b-staging`) by creating an `Application` resource in `team-a-dev`.

I would like to suggest adding the `serviceAccountName` property to `ApplicationSpec`.

When the user specifies `serviceAccountName`, the controller act as a given service account so that RBAC can be performed by Kubernetes.

#3434 is an implementation for this solution.

Add `AppPolicy` to check if components can be deployed or not.

However, it might be complicated because:

- We should implement our own RBAC checker in cuelang (or another component)
  - I think utilizing Kubernetes native feature is safer because we don't have to validate it for every Kubernetes release.
- Policy management tool (e.g. Kyverno) or admission webhook is required to ensure that each `Application` has the `AppPolicy`

**Drawbacks**

This proposal cannot cover the multi-cluster scenarios since either `Application` namespace or `ServiceAccount` might not exist in other clusters. (We cannot guarantee that resources mirrored)

This drawback can be resolved if `Cluster` resources are managed per namespaces, not cluster-wide. Before then, I think it would be better to skip impersonating in multi-cluster scenarios.

**Open Questions**

- We may set the `default` service account when the service account name is omitted. However, this can break existing applications, so I think it would be better to give users a migration term. Should we give a migration term or take this breaking change?

**Additional context**

- [Three Tenancy Models For Kubernetes](https://kubernetes.io/blog/2021/04/15/three-tenancy-models-for-kubernetes/)

devholic:

Resolves #3432

I have:

    - https://github.com/oam-dev/kubevela.io/pull/542

- Creating an Application without `.annotations["app.oam.dev/service-account-name"]` (for testing backward compatibility)
- Creating an Application with `.annotations["app.oam.dev/service-account-name"]`
  - Existing ServiceAccount
  - Existing ServiceAccount without Roles to create rendered components
  - Non-existing ServiceAccount

N/A

### DIFF
--- a/pkg/auth/round_trippers.go
+++ b/pkg/auth/round_trippers.go
@@ -1,0 +1,69 @@
+/*
+
+ Copyright 2021 The KubeVela Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+
+*/
+
+package auth
+
+import (
+	"net/http"
+
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/client-go/transport"
+
+	"github.com/oam-dev/kubevela/pkg/multicluster"
+	oamutil "github.com/oam-dev/kubevela/pkg/oam/util"
+	"github.com/oam-dev/kubevela/pkg/utils"
+)
+
+var _ utilnet.RoundTripperWrapper = &impersonatingRoundTripper{}
+
+type impersonatingRoundTripper struct {
+	rt http.RoundTripper
+}
+
+// NewImpersonatingRoundTripper will add an ImpersonateUser header to a request
+// if the context has a specific user whom to act-as.
+func NewImpersonatingRoundTripper(rt http.RoundTripper) http.RoundTripper {
+	return &impersonatingRoundTripper{
+		rt: rt,
+	}
+}
+
+func (rt *impersonatingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+
+	// Skip impersonation on non-local cluster requests
+	if !multicluster.IsInLocalCluster(ctx) {
+		return rt.rt.RoundTrip(req)
+	}
+
+	sa := oamutil.GetServiceAccountInContext(ctx)
+	if sa == "" {
+		return rt.rt.RoundTrip(req)
+	}
+	req = req.Clone(req.Context())
+	req.Header.Set(transport.ImpersonateUserHeader, sa)
+	return rt.rt.RoundTrip(req)
+}
+
+func (rt *impersonatingRoundTripper) CancelRequest(req *http.Request) {
+	utils.TryCancelRequest(rt.WrappedRoundTripper(), req)
+}
+
+func (rt *impersonatingRoundTripper) WrappedRoundTripper() http.RoundTripper {
+	return rt.rt
+}
--- a/pkg/resourcekeeper/delete.go
+++ b/pkg/resourcekeeper/delete.go
@@ -87,7 +87,7 @@ func (h *resourceKeeper) delete(ctx context.Context, manifest *unstructured.Unst
 	}
 	// 2. delete manifests
 	deleteCtx := multicluster.ContextWithClusterName(ctx, oam.GetCluster(manifest))
-	deleteCtx = oamutil.SetServiceAccountInContext(deleteCtx, h.app.Namespace, oam.GetServiceAccountName(h.app))
+	deleteCtx = oamutil.SetServiceAccountInContext(deleteCtx, h.app.Namespace, oam.GetServiceAccountNameFromAnnotations(h.app))
 	if err = h.Client.Delete(deleteCtx, manifest); err != nil && !kerrors.IsNotFound(err) {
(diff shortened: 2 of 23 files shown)


Somefive:
Great work! The implementation of ServiceAccount in Application is amazing. Generally looks good to me.

As @wonderflow said in #3432, we are also planing to implement such features except two extra features: the UID admission webhook and the multi-cluster ServiceAccount. For the former one, I think this PR is a first step towards the target. For the second one, there might be some extension to proposed RoundTripper, where some ServiceAccount mapping might be done to retrieve the ServiceAccount in worker clusters. But in my opinion, these could be implemented in another PR, not necessarily in this PR.

pkg/workflow/providers/kube/handle.go
```diff
@@ -85,6 +88,9 @@ func (h *provider) Apply(ctx wfContext.Context, v *value.Value, act types.Action
 		return err
 	}
 	deployCtx := multicluster.ContextWithClusterName(context.Background(), cluster)
+	if h.app != nil {
```
FogDong, line 91: Nit: maybe we can wrap the context handle logic into one function in this file.

devholic:
Rebased for resolving conflicts

pkg/auth/round_trippers.go
```diff
@@ -0,0 +1,69 @@
...
+	return &impersonatingRoundTripper{
+		rt: rt,
+	}
+}
+
+func (rt *impersonatingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+
+	// Skip impersonation on non-local cluster requests
+	if !multicluster.IsInLocalCluster(ctx) {
```
barnettZQG, line 50: I think there is no necessary limit, when the user config the ServiceAccount, we can assume that it already exists in the target cluster.

devholic: Is it okay to assume that both Namespace and ServiceAccount exist in the target cluster? (I was not sure about the existence of the Application namespace and logical equality of the namespace. For example, application namespace `app-ns` can be used in the target cluster for the other purpose)

wonderflow: It's not safe for now, if we want to free this limit, we should have mechanism to keep it safe.

barnettZQG: @devholic Currently the namespace of the target cluster may be inconsistent with the app.

devholic: @barnettZQG If so, keeping this limit seems better before implementing OCM Executor-like approaches. (https://github.com/oam-dev/kubevela/issues/3432#issuecomment-1066264448)

wonderflow (approved):
LGTM， great job!

wonderflow:
ping @Somefive @FogDong @leejanee @barnettZQG

FogDong (approved):
LGTM, thx!

pkg/resourcekeeper/delete.go
```diff
@@ -85,7 +86,10 @@ func (h *resourceKeeper) delete(ctx context.Context, manifest *unstructured.Unst
 		}
 	}
 	// 2. delete manifests
-	if err = h.Client.Delete(multicluster.ContextWithClusterName(ctx, oam.GetCluster(manifest)), manifest); err != nil && !kerrors.IsNotFound(err) {
+	deleteCtx := multicluster.ContextWithClusterName(ctx, oam.GetCluster(manifest))
+	deleteCtx = oamutil.SetServiceAccountInContext(
```
barnettZQG, line 90:
```suggestion
	deleteCtx = oamutil.SetServiceAccountInContext(deleteCtx, h.app)
```

devholic:
I'm not sure why e2e tests are skipped, but it passed before I rebased 🤔

pkg/resourcekeeper/delete.go
barnettZQG: The logic of getting the ServiceAccount name and namespace converges in the method.

devholic: I think passing app directly is less extensible than passing the namespace / serviceAccountName pair. Therefore, I added `GetServiceAccountName`, which extracts ServiceAccount name from annotation like below:

https://github.com/oam-dev/kubevela/pull/3434/commits/78bc21b943410ccaeb4d955d146ea3adb730b4f7#diff-6740e9863579149c3533908aaffc41f207ae7a1031759e1ac075c703f74b74ec

and make all callers use it. WDYT?

barnettZQG: But, the namespace is also not fixed to `app.namespace`

devholic: > But, the namespace is also not fixed to app.namespace

Could you give me more details or additional context for this comment?

barnettZQG: We need to get the ServiceAccount name and namespace from the application, Currently from annotations, and possibly from the spec in the future. 

In multi-cluster mode, the name and namespace of ServiceAccount are both possible to define in a spec.

devholic: Thanks for the kind explanation 🙂 

If so, both `GetServiceAccountName` and passing application might be ambiguous since multi-cluster mode may contain multiple ServiceAccount names per cluster.

For making the context clear, how about changing `GetServiceAccountName` to `GetServiceAccountNameFromAnnotation` and letting method callers select which ServiceAccount or namespace to use?

barnettZQG: Agree to this change.

Somefive (approved):
LGTM

barnettZQG (approved):
LGTM

barnettZQG:
@devholic Thanks your PR.

PR #3446, opened 2022-03-15
Feat: rework ref objects

Somefive:

1. Referred objects are loaded during the AppFile parsing stage now, instead of during executing workflow step. This change is for future application version to store referred objects. (This PR implements the storing process but not implement the loading process.)
2. ref-objects spec is refactored, following the advice by @yue9944882 . Now user can use `resource` to specify the target type of Kubernetes objects. `APIVersion/Kind` is also supported for backward-compatibility when feature gate `LegacyObjectTypeIdentifier=true` is set.
3. The label selector field in ref-objects now renamed from `selector` to `labelSelector` for clarity. (This old one is kept for compatibility)
4. Cross namespace and cross cluster object reference is supported, when `--ref-objects-available-scope=global` is added to the bootstrap parameter.

```yaml
kind: Application
metadata:
  name: podinfo
  namespace: example
spec:
  components:
    - name: podinfo
      type: ref-objects
      properties:
        objects:
          - resource: deployment
            name: podinfo
            namespace: debug
  policies:
    - type: topology
      name: topo
      properties:
        clusters: ["local"]
        namespace: "rev"
```

I have:


pkg/component/ref-objects.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return selector.DeprecatedLabelSelector
+}
+
+// GetGroupVersionKindFromRefObjectSelector extract GroupVersionKind by Resource if provided, otherwise, extract from APIVersion and Kind directly
+func GetGroupVersionKindFromRefObjectSelector(mapper meta.RESTMapper, selector v1alpha1.ObjectReferrer) (schema.GroupVersionKind, error) {
+	if selector.Resource != "" {
+		return mapper.KindFor(schema.GroupVersionResource{Group: selector.Group, Resource: selector.Resource})
+	}
+	if selector.APIVersion != "" && selector.Kind != "" {
```
yue9944882, line 45: gate the legacy fields w/ the feature-gate

pkg/component/ref-objects.go
```diff
@@ -0,0 +1,127 @@
...
+	un.SetResourceVersion("")
+	un.SetGeneration(0)
+	un.SetOwnerReferences(nil)
+	un.SetDeletionTimestamp(nil)
+	un.SetManagedFields(nil)
+	un.SetUID("")
+	unstructured.RemoveNestedField(un.Object, "metadata", "creationTimestamp")
+	unstructured.RemoveNestedField(un.Object, "status")
+	// TODO(somefive): make the following logic more generalizable
+	if un.GetKind() == "Service" && un.GetAPIVersion() == "v1" {
```
yue9944882, line 78: consider copy/paste the following code which is proven working correct for yrs.

https://github.com/openshift/library-go/blob/d48a1fb9b7c2c89f5914cc670a4056bcccdd0a34/pkg/operator/resource/resourceapply/core.go#L130-L177

yue9944882 (commented):
in general, it's not recommended to the client-side defaulting/validating in the controller, ideally they should be done in an admission webhook. b/c client-side validating doesn't secure/sanitize the persisted data-plane for the cluster --- the invalid objects should be validated/rejected __before they're placed into storage__. as for cilent-side defaulting, if we're doing it in the controller, we have to keep it in-memory defaulting otherwise writing the defaulted object (which is a non-status update) back to storage will trigger the generation change which makes the generation field on the custom resource meaningless/usable

pkg/component/ref-objects.go
```diff
@@ -0,0 +1,127 @@
...
+	if selector.LabelSelector != nil {
+		return selector.LabelSelector
+	}
+	return selector.DeprecatedLabelSelector
+}
+
+// GetGroupVersionKindFromRefObjectSelector extract GroupVersionKind by Resource if provided, otherwise, extract from APIVersion and Kind directly
+func GetGroupVersionKindFromRefObjectSelector(mapper meta.RESTMapper, selector v1alpha1.ObjectReferrer) (schema.GroupVersionKind, error) {
+	if selector.Resource != "" {
+		return mapper.KindFor(schema.GroupVersionResource{Group: selector.Group, Resource: selector.Resource})
```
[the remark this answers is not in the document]
Somefive: OK.

pkg/component/ref-objects.go
Somefive: OK.

apis/core.oam.dev/v1alpha1/component_types.go
```diff
@@ -0,0 +1,70 @@
...
+	// ObjectSelector select object by name or labelSelector
+	ObjectSelector `json:",inline"`
+}
+
+// ObjectTypeIdentifier identifies the scheme of Kubernetes object
+type ObjectTypeIdentifier struct {
+	// Resource is the resource name of the Kubernetes object.
+	Resource string `json:"resource,omitempty"`
+	// Group is the API Group of the Kubernetes object.
+	Group string `json:"group,omitempty"`
```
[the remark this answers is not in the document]
Somefive: OK.

pkg/component/ref-objects.go
Somefive: @yue9944882 If I'm understanding it correct, .spec.ports is not copied in this implementation? Seems only .spec.type and .spec.selector are copied.

Somefive:
> in general, it's not recommended to the client-side defaulting/validating in the controller, ideally they should be done in an admission webhook. b/c client-side validating doesn't secure/sanitize the persisted data-plane for the cluster --- the invalid objects should be validated/rejected **before they're placed into storage**. as for cilent-side defaulting, if we're doing it in the controller, we have to keep it in-memory defaulting otherwise writing the defaulted object (which is a non-status update) back to storage will trigger the generation change which makes the generation field on the custom resource meaningless/usable

In ApplicationValidatingWebhook, we will parse the submitted Application before processing it in the application controller. Although this parsing logic is duplicated in both validating webhook and controller, it is necessary to do so since there are times users want to deploy KubeVela without admission webhook.

leejanee (approved):

PR #3462, opened 2022-03-17
Feat: support the mountPath of trait-storage-secret is optional

fourierr:
Signed-off-by: maxiangbo maxiangboo@cmbchina.com

Support the mountPath parameter of trait-storage-secret is optional.
by this feature, we can valueFrom secret in container env, without mounting the secret to the container at the same time.

I have:

### DIFF
--- a/vela-templates/definitions/internal/trait/storage.cue
+++ b/vela-templates/definitions/internal/trait/storage.cue
@@ -19,7 +19,7 @@ template: {
 	] | []
 
 	configMapVolumesList: *[
-				for v in parameter.configMap {
+				for v in parameter.configMap if v.mountPath != _|_ {
 			{
 				name: "configmap-" + v.name
 				configMap: {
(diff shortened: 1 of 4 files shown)


vela-templates/definitions/internal/trait/storage.cue
```diff
@@ -279,7 +279,7 @@ template: {
 				envName:   string
 				secretKey: string
 			}
-			mountPath:   string
+			mountPath?:  string
```
FogDong, line 282: We should also change configmap's mountPath to optional since env can get value from configmap.

fourierr: > We should also change configmap's mountPath to optional since env can get value from configmap.

ok, I will add it

fourierr: > We should also change configmap's mountPath to optional since env can get value from configmap.

a new commit added cm's mountPath to be optional

FogDong (approved):
LGTM, thx for the contribution!

PR #3476, opened 2022-03-21
Feat: ctrl-plane command to start vela control plane in one step

chivalryq:
Signed-off-by: qiaozp <chivalry.pp@gmail.com>

This PR mainly do two things

1. provide a CLI command `vela ctrl-plane`, which automatically install k3s and vela-core in it. Because of the new multicluster pattern used in KubeVela (mainly based on `ref-object` type component). This command will install k3s without workload controllers (Deployment, Job, etc) by defualt. This conmmand only work in Linux for now.

```shell
$ vela ctrl-plane -h
Quickly setup a KubeVela control plane, using K3s and only for Linux now

Usage:
  vela ctrl-plane [command]

Available Commands:
  install     Quickly setup a KubeVela control plane
  kubeconfig  print kubeconfig to access control plane
  uninstall   uninstall control plane

Flags:
  -h, --help   help for ctrl-plane

Global Flags:

Use "vela ctrl-plane [command] --help" for more information about a command.
```

2. Now our Helm chart contains several Deployment/Job manifests. We want to use vela-core's multicluster in control plane standalone mode, which means install the chart in a cluster whose controller-manager is incomplete. This PR provide a parameter: podOnly in vela-core chart. It won't affect the default install behavior. When set podOnly=true. All Deployments will be replaced with Pod with the same spec. Here is the replaced manifest list :

  - kubevela-controller
  - cluster-gateway
  - kubevela-vela-core-admission-create
  - kubevela-vela-core-admission-patch

I have:


Somefive (approved):

wonderflow (changes requested):
This change will make the chart hardly to maintain. Is there any alternative solution?

I think we can create a pod spec yaml and use the same values.yaml?

chivalryq:
>

chivalryq:
> This change will make the chart hardly to maintain. Is there any alternative solution?

> I think we can create a pod spec yaml and use the same values.yaml?

Do you mean helm chart is like below?

```
#deploy.yaml
{{ if not podOnly }}
Deployment
{{ end }}

# pod.yaml
{{ if .Values.PodOnly }}
Pod
{{ end }}
```

This creates a duplicate copy.

chivalryq:
ping @wonderflow please review again.

wonderflow:
let's move this feature into a seprate project https://github.com/oam-dev/velad

charts/vela-core/values.yaml
```diff
@@ -210,3 +210,6 @@ admissionWebhooks:
 kubeClient:
   qps: 50
   burst: 100
+
+## @param podOnly Use pod instead of Deployment to deploy vela-core
+podOnly: false
```
wonderflow, line 215:
```suggestion
deployByPod: false
```

PR #3490, opened 2022-03-22
Fix: fix bugs in vela def doc-gen from local file

Nicola115:
…rseLocalFile

Signed-off-by: Nicola115 <2225992901@qq.com>

1. add readFrom instead of namespace to pass either namespace or localFilePath to `GenerateRefeenceDocs`. 
2. set ref.DefinitionName as fileName if the args[0] is a filePath. 
3. change yaml.Unmarsh in ParseLocalFile to converting yaml to json using k8s pkg first.

Fixes #
  
1. using local filepath as generated markdown filename. 
4. failed to unmarshal parts of yaml data into ComponentDefinitions due to lack of yaml tags.  
5. wrong unit tests. 
  

I have:

1. pass TestParseLocalFile. 
2. I build the local vela and test `vela def doc-gen <localFilePath>` command.

### DIFF
--- a/references/cli/def.go
+++ b/references/cli/def.go
@@ -457,36 +457,36 @@ func NewDefinitionGenDocCommand(c common.Args) *cobra.Command {
 				return fmt.Errorf("please specify definition name or a definition file")
 			}
 
-			var readFrom string
 			ref := &plugins.MarkdownReference{}
 			if strings.HasSuffix(args[0], ".yaml") {
+				// read from local file
 				localFilePath := args[0]
 				fileName := filepath.Base(localFilePath)
 				if !strings.HasSuffix(fileName, ".yaml") {
 					return fmt.Errorf("invalid local file path `%s`", localFilePath)
 				}
 				ref.DefinitionName = strings.TrimSuffix(fileName, ".yaml")
-				readFrom = localFilePath
+				ref.Local = &plugins.Local{Path: localFilePath}
 			} else {
 				namespace, err := cmd.Flags().GetString(FlagNamespace)
 				if err != nil {
 					return errors.Wrapf(err, "failed to get `%s`", Namespace)
 				}
 				ref.DefinitionName = args[0]
-				readFrom = namespace
+				ref.Remote = &plugins.Remote{Namespace: namespace}
 			}
 
 			ctx := context.Background()
 			pathEn := plugins.KubeVelaIOTerraformPath
 			ref.I18N = plugins.En
-			if err := ref.GenerateReferenceDocs(ctx, c, pathEn, readFrom); err != nil {
+			if err := ref.GenerateReferenceDocs(ctx, c, pathEn); err != nil {
 				return errors.Wrap(err, "failed to generate reference docs")
 			}
 			cmd.Printf("Generated docs in English for %s in %s/%s.md\n", args[0], pathEn, ref.DefinitionName)
 
 			pathZh := plugins.KubeVelaIOTerraformPathZh
 			ref.I18N = plugins.Zh
-			if err := ref.GenerateReferenceDocs(ctx, c, pathZh, readFrom); err != nil {
+			if err := ref.GenerateReferenceDocs(ctx, c, pathZh); err != nil {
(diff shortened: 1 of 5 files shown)


references/plugins/reference_test.go
```diff
@@ -651,9 +650,10 @@ func TestParseLocalFile(t *testing.T) {
 			if err != nil {
 				t.Errorf("ParseLocalFile(...): -want: %v, got error: %s\n", tc.want, err)
 			}
-			if reflect.DeepEqual(*lc, tc.want) {
+			if !reflect.DeepEqual(*lc, tc.want) {
 				t.Errorf("ParseLocalFile(...): -want: %v, got: %v\n", tc.want, *lc)
 			}
 		})
+		break
```
zzxwill, line 657: Why a `break` is added?

references/plugins/references.go
```diff
@@ -557,25 +557,26 @@ func setDisplayFormat(format string) {
 }
 
 // GenerateReferenceDocs generates reference docs
-func (ref *MarkdownReference) GenerateReferenceDocs(ctx context.Context, c common.Args, baseRefPath string, namespace string) error {
+func (ref *MarkdownReference) GenerateReferenceDocs(ctx context.Context, c common.Args, baseRefPath string, readFrom string) error {
```
zzxwill, line 560: `readFrom` seems odd.

Nicola115: as it could be namespace or local filepath, i think it's not appropriate to be named as namespace any more.

zzxwill: You can refer to the refactoring.

```
type Remote struct {
	DefinitionName string `json:"definitionName"`
	Namespace string `json:"namespace"`
}

type Local struct {
	Path string `json:"path"`
}

// MarkdownReference is the struct for capability information in
type MarkdownReference struct {
	Remote Remote `json:"remote"`
	Local Local `json:"local"`
	ParseReference
}
```

We would also remove `namespace` parameter from the function `GenerateReferenceDocs`.

It seems more clear. What's your idea?

Nicola115:
```go
type Remote struct {
	Namespace string `json:"namespace"`
}

type Local struct {
	Path string `json:"path"`
}

// MarkdownReference is the struct for capability information in
type MarkdownReference struct {
	Remote Remote `json:"remote"`
	Local Local `json:"local"`
        DefinitionName string `json:"definitionName"`
	ParseReference
}
```

I've set DefinitionName as an attribute of MarkdownReference cause it's used in both local and remote ways to generate docs.

references/cli/def.go
```diff
@@ -457,31 +457,39 @@ func NewDefinitionGenDocCommand(c common.Args) *cobra.Command {
...
-			var namespace string
-			if !strings.HasSuffix(args[0], ".yaml") {
-				var err error
-				namespace, err = cmd.Flags().GetString(FlagNamespace)
+			ref := &plugins.MarkdownReference{}
+			if strings.HasSuffix(args[0], ".yaml") {
+				// read from local file
+				localFilePath := args[0]
+				fileName := filepath.Base(localFilePath)
+				if !strings.HasSuffix(fileName, ".yaml") {
```
zzxwill, line 465: Is the purpose of this line to check the existence of the input file?

Nicola115: check the input file format as yaml and extract the fileName from filePath.

wonderflow (approved):
LGTM after comments from @zzxwill  resolved

references/cli/def.go
zzxwill: `if strings.HasSuffix(args[0], ".yaml") {` means the fileName is ends with `.yaml`.

Nicola115: yes, if it doesn't have yaml suffix, it will return error. Otherwise, it will call trimSuffix

zzxwill: I mean L465 is duplicate of L461.

Nicola115: I'm afraid there might be invalid output of filepath.BaseName. if filepath is not a valid path, then it will output ".".

Somefive (approved):

PR #3512, opened 2022-03-25
Fix: helm install cluster-gateway be compatible w/ existing APIService w/o annoation

yue9944882:


wonderflow (approved):
good catch, thanks

Somefive (approved):

PR #3513, opened 2022-03-25
Fix: add missing field in op.#Component

### LINKED ISSUE #3510: Cannot use `dependsOn` in `components` when `deploy2env` exists
**Describe the bug**

Cannot use `dependsOn` in `components` when `deploy2env` exists

**To Reproduce**

I am following the How-to Guides [Cloud Resources - Provision a Database and Import a SQL File for initialization](https://kubevela.io/docs/end-user/components/cloud-services/provision-and-initiate-database#!), and intentionally add `dependsOn` within the `web` component like below:

```cue
kind: Application
metadata:
  name: favorite-links
spec:
  components:
    - name: web
      type: webservice
      properties:
        image: oamdev/nodejs-mysql-links:v0.0.1
        port: 4000
      #################
      # MY CHANGE HERE #
      #################
      dependsOn:
      - db
      traits:
# ...
```

Then I apply this YAML, but get error below:

```bash
$ vela status favorite-links

About:

  Name:      	favorite-links
  Namespace: 	default
  Created at:	2022-03-25 14:25:38 +0800 CST
  Status:    	runningWorkflow

Workflow:

  mode: StepByStep
  finished: false
  Suspend: false
  Terminated: false
  Steps
  - id:ul2if3351w
    name:deploy
    type:deploy2env
    phase:failed
    message:step app: step apply: step outputs: _|_ // field "dependsOn" not allowed in closed struct
: app.apply.outputs: field "dependsOn" not allowed in closed struct

Services:
```

**Expected behavior**

No error, `dependsOn` should not be denied.

**Screenshots**

image

**KubeVela Version**

CLI Version: v1.2.4
Core Version: 1.2.4
GitRevision: git-f6032f3
GolangVersion: go1.17.7

**Cluster information**

v1.22.3

**Additional context**

huanggze:
Signed-off-by: huanggze <gzhuang2014@gmail.com>

Fixes #3510 CUE

I have:

### DIFF
--- a/pkg/stdlib/pkgs/multicluster.cue
+++ b/pkg/stdlib/pkgs/multicluster.cue
@@ -24,6 +24,7 @@
 		properties: {...}
 	}]
 	externalRevision?: string
+	dependsOn?: [...string]
 }
 
 #ReadPlacementDecisions: {


pkg/stdlib/pkgs/multicluster.cue
```diff
@@ -24,6 +24,7 @@
 		properties: {...}
 	}]
 	externalRevision?: string
+	dependsOn?: [...string]
```
FogDong, line 27: `Inputs` & `Outputs` can not be used in `deploy2env` for the same reason.
Maybe you can just add `...` here for more parameters, but this may cause some confusion.

/cc @Somefive

Somefive: It is ok to add dependsOn here. The `deploy2env` does not support dependsOn but we are planning to fix it in the `deploy` WorkflowStep.

Somefive: Adding `...` also works. I can accept both solution.

Somefive (approved):

Somefive (changes requested):
I realize that you want to fix the problem of "`deploy2env` not support depends on". But this fix is far from to achieve that.

huanggze:
Thanks for your reply. I wonder if there is any way to apply components in order explicitly? @Somefive

Somefive:
> Thanks for your reply. I wonder if there is any way to apply components in order explicitly, while still use `deploy2env ` at the same time? @Somefive

This is not supported now. We are planning to implement it but might not be able to do it immediately. One way to work around is to write `deploy2env` with component selector and then make different component applied in several individual WorkflowSteps.

Somefive:
Let's discuss it in #3510.

PR #3542, opened 2022-03-29
Fix: addon support empty components

Somefive:

Support empty components in application for addon.


wonderflow (approved):
thanks

PR #3546, opened 2022-03-29
Feat: enhance storage trait to support multi-mountToEnv config

ZhongsJie96:
Signed-off-by: Shijie Zhong <zhongsjie@cmbchina.com>

Support the multi-mountToEnv parameter of storage trait in secret and configmap. By this feature, we can mount multi-env in container env, without creating multi-secret or multi-configmap resrouces

I have:

### DIFF
--- a/charts/vela-core/templates/defwithtemplate/storage.yaml
+++ b/charts/vela-core/templates/defwithtemplate/storage.yaml
@@ -77,12 +77,12 @@ spec:
         	},
         ] | []
         configMapEnvMountsList: *[
-        			for v in parameter.configMap if v.mountToEnv != _|_ {
+        			for v in parameter.configMap if v.mountToEnv != _|_ for k in v.mountToEnv {
         		{
-        			name: v.mountToEnv.envName
+        			name: k.envName
         			valueFrom: configMapKeyRef: {
         				name: v.name
-        				key:  v.mountToEnv.configMapKey
+        				key:  k.configMapKey
         			}
         		}
         	},
@@ -96,12 +96,12 @@ spec:
         	},
         ] | []
         secretEnvMountsList: *[
-        			for v in parameter.secret if v.mountToEnv != _|_ {
+        			for v in parameter.secret if v.mountToEnv != _|_ for k in v.mountToEnv {
         		{
-        			name: v.mountToEnv.envName
+        			name: k.envName
         			valueFrom: secretKeyRef: {
         				name: v.name
-        				key:  v.mountToEnv.secretKey
+        				key:  k.secretKey
         			}
         		}
         	},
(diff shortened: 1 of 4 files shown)


wonderflow (approved):

FogDong (commented):
Great, thanks for the contribution!

Please also update the examples here: https://github.com/oam-dev/kubevela/blob/master/docs/examples/core-definitions/trait/storage.yaml

FogDong (changes requested):
After changing `mountToEnv` to the array, the users who use the former version will be affected. 

In fact, we need to have definition revisions first to solve this problem.🤔

ZhongsJie96:
> Great, thanks for the contribution!
> 
> Please also update the examples here: https://github.com/oam-dev/kubevela/blob/master/docs/examples/core-definitions/trait/storage.yaml

ZhongsJie96:
> Great, thanks for the contribution!
> 
> Please also update the examples here: https://github.com/oam-dev/kubevela/blob/master/docs/examples/core-definitions/trait/storage.yaml

ok，I will add it

ZhongsJie96:
> After changing `mountToEnv` to the array, the users who use the former version will be affected.
> 
> In fact, we need to have definition revisions first to solve this problem.🤔

As you said, this feature will affect the users who used it before. What should I do to define the revision? At the same time, I think it is a reasonable feature to support some special needs. Sorry, I accidentally closed this pr.

Somefive:
I think you can have two ways to keep compatibility. The first one is to rename this trait to a new trait such as `storages` or `multi-storage`. The second one is try to move the multi-settings in the propose changes in another field. You can see `container-image`, `env`, or `command` traits. /cc @FogDong

anselmhou:
Looking for this PR to be merge.  I have a use case where I have Kubernetes secrete that are defined externally with multiple key and value pairs, and those value pairs to the application as environment variables.

I can work around with `service-bindings` but since it is marked as deprecated so wishing to move to storage traits.  But in order to do and without the support of multiple `mountToEnv` it seems a blocker to me.  I cannot use a different name since it references the same secret in Kubernetes.

So thank you for the contribution and looking forward to this release and hoping to see it in the near future release of the project.

FogDong:
> Looking for this PR to be merge. I have a use case where I have Kubernetes secrete that are defined externally with multiple key and value pairs, and those value pairs to the application as environment variables.
> 
> I can work around with `service-bindings` but since it is marked as deprecated so wishing to move to storage traits. But in order to do and without the support of multiple `mountToEnv` it seems a blocker to me. I cannot use a different name since it references the same secret in Kubernetes.
> 
> So thank you for the contribution and looking forward to this release and hoping to see it in the near future release of the project.

Since definition revision is a large and complex feature, maybe we can add `mountToEnvs` first to support array in this definition. 
WDYT? /cc @ZhongsJie96

PR #3587, opened 2022-04-05
Feat: support sequential gc after application deletion

### LINKED ISSUE #2449: The GC after application deletion is not sequential
**Describe the bug**

The GC after application deletion is not sequential.

If we have a controller and CR in an application, when we delete the application, the controller will be deleted first. In this case, the finalizer managed by the controller will be added to the CR. As a result, the deletion of the CR fails.

**To Reproduce**

```yaml
kind: Application
metadata:
  name: mysql
  namespace: default
spec:
  components:
    - name: mysql-controller
      type: helm
      properties:
        repoType: helm
        url: https://presslabs.github.io/charts
        chart: mysql-operator
        version: "0.4.0"
    - name: mysql-cluster
      type: raw
      dependsOn:
        - mysql-controller
      properties:
        apiVersion: mysql.presslabs.org/v1alpha1
        kind: MysqlCluster
        metadata:
          name: mysql-cluster
        spec:
          replicas: 1
          secretName: mysql-secret
```

Apply the yaml above and delete the app later.

**Expected behavior**

**Screenshots**

**KubeVela Version**

**Cluster information**

**Additional context**

fourierr:
Fixes #2449

The GC after application deletion is sequential, accoding to the reversal order of the creation. 
There is an example for this feature, we have five component A,B,C,D,E the relationship is as follows
`A dependsOn _`
`B dependsOn _`
`C dependsOn A, B`
`D dependsOn C`
`E dependsOn C, D`
the order of creatation is that `A, B -> C -> D -> E`
the order of gc is that `E -> D -> C -> A, B`

I have:


pkg/resourcekeeper/cache.go
```diff
@@ -101,3 +101,13 @@ func (cache *resourceCache) get(ctx context.Context, mr v1beta1.ManagedResource)
 	}
 	return entry
 }
+
+func (cache *resourceCache) setNotExist(mr v1beta1.ManagedResource) {
```
wonderflow, line 105:
```suggestion
func (cache *resourceCache) initializeFor(mr v1beta1.ManagedResource) {
```

pkg/resourcekeeper/cache.go
```diff
@@ -101,3 +101,13 @@ func (cache *resourceCache) get(ctx context.Context, mr v1beta1.ManagedResource)
...
+	key := mr.ResourceKey()
+	entry, cached := cache.m[key]
+	if !cached {
+		entry = &resourceCacheEntry{obj: mr.ToUnstructured(), mr: mr}
+	}
+	entry.exists = false
```
wonderflow, line 111: why we don't set `deleted=true` here?

Somefive: deleted=true is another situation. deleted=true means the resource should be deleted. exists=false means it does not exist. So exists!=deleted means everything ok, but exists==deleted means either resource should be deleted or state kept.

apis/core.oam.dev/v1beta1/resourcetracker_types.go
```diff
@@ -81,6 +81,8 @@ type ManagedResource struct {
 	Data *runtime.RawExtension `json:"raw,omitempty"`
 	// Deleted marks the resource to be deleted
 	Deleted bool `json:"deleted,omitempty"`
+	// DependsOn records the component that was dependent
+	DependsOn []string `json:"dependsOn,omitempty"`
```
wonderflow, line 85:
```suggestion
	// DependsOn records the component name that was dependent
	DependsOnComponents []string `json:"dependsOn,omitempty"`
```

wonderflow: what if the dependency is recorded as inputs/outputs? \cc @FogDong

pkg/resourcekeeper/resourcekeeper.go
```diff
@@ -37,7 +37,7 @@ import (
 
 // ResourceKeeper handler for dispatching and deleting resources
 type ResourceKeeper interface {
-	Dispatch(context.Context, []*unstructured.Unstructured, ...DispatchOption) error
+	Dispatch(context.Context, []string, []*unstructured.Unstructured, ...DispatchOption) error
```
Somefive, line 40: Make the `dependsOn` a DispatchOption, not a parameter.

pkg/resourcekeeper/gc.go
```diff
@@ -245,23 +245,71 @@ func (h *gcHandler) Sweep(ctx context.Context) (finished bool, waiting []v1beta1
 }
 
 func (h *gcHandler) recycleResourceTracker(ctx context.Context, rt *v1beta1.ResourceTracker) error {
```
wonderflow, line 247: can we make the recycle functions to work per component? That can be more clear like:

```
func recycleComponents(compnents ..){
    for _, comp:=range components{
     if checkDependsOnComponents(comp){
         deleteDependsOnComponents(comp)
     }
    delete(comp)
   }
}
```

pkg/resourcekeeper/gc.go
```diff
@@ -245,23 +245,71 @@ func (h *gcHandler) Sweep(ctx context.Context) (finished bool, waiting []v1beta1
...
+		}
+		if len(pendingDelMRs) > 0 {
+			return h.recycleResourceTracker(ctx, rt)
 		}
 	}
 	return nil
 }
 
+func (h *gcHandler) checkPending(ctx context.Context, mr v1beta1.ManagedResource, rt *v1beta1.ResourceTracker, todoDelMRs []*v1beta1.ManagedResource, pendingDelMRs []*v1beta1.ManagedResource) ([]*v1beta1.ManagedResource, []*v1beta1.ManagedResource, error) {
+	// if the ManagedResource is not exist so return directly
```
Somefive, line 286: if the ManagedResource does not exist, then return directly

pkg/resourcekeeper/gc.go
```diff
@@ -245,23 +245,71 @@ func (h *gcHandler) Sweep(ctx context.Context) (finished bool, waiting []v1beta1
...
+	if getEntry.err != nil {
+		return nil, nil, getEntry.err
+	}
+	if !getEntry.exists {
+		return todoDelMRs, pendingDelMRs, nil
+	}
+	// iterate RT's resource to find the dependent
+	for _, managedResource := range rt.Spec.ManagedResources {
+		for _, dependsOnComp := range managedResource.DependsOn {
+			if mr.Component == dependsOnComp {
```
Somefive, line 297: What will happen in multi-cluster scenario? I think checking component equality is not enough. Will the component in cluster A depends on the component in cluster B? What if the component is dispatched in both cluster A namespace X and cluster A namespace Y? So if you want to check the dependent, you cannot only validate the name of the component.

pkg/workflow/tasks/custom/task.go
```diff
@@ -137,6 +137,7 @@ func (t *TaskLoader) makeTaskGenerator(templ string) (wfTypes.TaskGenerator, err
 			if err := json.Unmarshal(bt, &params); err != nil {
 				return nil, err
 			}
+			params["dependsOn"] = wfStep.DependsOn
```
Somefive, line 140: So here it is injecting dependsOn to the parameters of the WorkflowStep? What is it for?

pkg/resourcekeeper/cache.go
Somefive, line 105: I'm a bit confused of what is this function for.

Somefive (changes requested):

apis/core.oam.dev/v1beta1/resourcetracker_types.go
FogDong: We can record both depends on and inputs/outputs in this field.

pkg/workflow/tasks/custom/task.go
FogDong: I'm also confused here. Is there any handle for `inputs/outputs` btw?

fourierr: the dependsOn is declared in compoment, and add it to param, so as to the provider handler `component-apply` can get the dependsOn from params

pkg/resourcekeeper/cache.go
fourierr: When gc by the reversal order of the creation, if the comp's resource is already gc, i want to mark it in the resourceCache.That's why we have this function

fourierr: OK, i rename it

pkg/resourcekeeper/gc.go
fourierr: The function `deleteDependsOnComponents`  also will be a iterate function to check the relation of dependsOn. And current gc is not able to recycle at the compnent level.

pkg/resourcekeeper/gc.go
fourierr: yes, the ManagedResource does not exist means that gc for the resource is done

apis/core.oam.dev/v1beta1/resourcetracker_types.go
fourierr: > We can record both depends on and inputs/outputs in this field.

yes, the  inputs/outputs can be recorded in this field

pkg/resourcekeeper/gc.go
fourierr: In multi-cluster scenario, as long as there is an equal to the mr.component, this manageResource will be add to `pendingDelMRs`

pkg/workflow/tasks/custom/task.go
FogDong: If so, we should handle `inputs/outputs` here as well.

pkg/resourcekeeper/gc.go
Somefive: I mean the comment has some grammar error :)

wonderflow:
closing as it's done in #3701

PR #3687, opened 2022-04-15
Fix: set provider name as the config name

zzxwill:
- For VelaUX, users don't need to manually set the provider name. Used
the application/component name (config name) to be the provider name.
- Store the description of a config to the annotation of the config application

Fixes #

I have:

### DIFF
--- a/apis/types/types.go
+++ b/apis/types/types.go
@@ -73,8 +73,8 @@ const (
 	LabelConfigSyncToMultiCluster = "config.oam.dev/multi-cluster"
 	// LabelConfigIdentifier is the label for config identifier
 	LabelConfigIdentifier = "config.oam.dev/identifier"
-	// LabelConfigDescription is the label for config description
-	LabelConfigDescription = "config.oam.dev/description"
+	// AnnotationConfigDescription is the annotation for config description
+	AnnotationConfigDescription = "config.oam.dev/description"
 	// AnnotationConfigAlias is the annotation for config alias
 	AnnotationConfigAlias = "config.oam.dev/alias"
 )
@@ -147,4 +147,9 @@ const (
 	ImageRegistry = "config-image-registry"
 	// HelmRepository is the config type for Helm chart repository
 	HelmRepository = "config-helm-repository"
+)
+
+const (
+	// TerrfaormComponentPrefix is the prefix of component type of terraform-xxx
+	TerrfaormComponentPrefix = "terraform-"
 )
--- a/pkg/apiserver/rest/usecase/config.go
+++ b/pkg/apiserver/rest/usecase/config.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	set "github.com/deckarep/golang-set"
 	"github.com/pkg/errors"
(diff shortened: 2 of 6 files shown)


Somefive (approved):

apis/types/types.go
```diff
@@ -148,3 +148,8 @@ const (
 	// HelmRepository is the config type for Helm chart repository
 	HelmRepository = "config-helm-repository"
 )
+
+const (
+	// TerrfaormComponentPrefix is the prefix of component type of terraform-xxx
+	TerrfaormComponentPrefix = "terraform-"
```
wonderflow, line 154:
```suggestion
	TerrfaormComponentPrefix = "tf-"
```

K8s name has a limit of 64 char, we should make the common prefix short

wonderflow (approved):
generally LGTM

pkg/apiserver/rest/usecase/config.go
```diff
@@ -134,12 +135,27 @@ func (u *configUseCaseImpl) GetConfigType(ctx context.Context, configType string
 }
 
 func (u *configUseCaseImpl) CreateConfig(ctx context.Context, req apis.CreateConfigRequest) error {
+	p := req.Properties
+	// If the component is Terraform type, set the provider name same as the application name and the component name
+	if strings.HasPrefix(req.ComponentType, types.TerrfaormComponentPrefix) {
```
barnettZQG, line 140: we should handle it in the cue script. such as:

```
...
provider_name: context.name
...
```

there is no need to add a field named `name`.

apis/types/types.go
zzxwill: They are names used as labels to mark the application which works as the controller to create a Terraform provider config.

pkg/apiserver/rest/usecase/config.go
zzxwill: For cli users, it's used to indicate the names of the provider, and it has the default name for a specific provider like https://github.com/oam-dev/catalog/blob/master/addons/terraform-baidu/definitions/baidu-provider.cue#L64. `context.name` could not express the default value.

zzxwill: And the default provider name is to be compatible with previous addons, within which an provider will be created.

barnettZQG: @zzxwill For the CLI users, what is the use case like?

barnettZQG: If he must create a component, it must have a component name. What is the starting point for making the component name and provider name inconsistent here?

zzxwill: For cli users, we create a provider whose name is decided by `--name`.
```
➜  kubevela git:(terraform-ui) ✗ vela provider add terraform-alibaba -h
Authenticate Terraform Cloud Provider terraform-alibaba by creating a credential secret and a Terraform Controller Provider

Usage:
  vela provider add terraform-alibaba [flags]

Examples:
vela provider add terraform-alibaba

Flags:
      --ALICLOUD_ACCESS_KEY string   Get ALICLOUD_ACCESS_KEY per this guide https://help.aliyun.com/knowledge_detail/38738.html
      --ALICLOUD_REGION string       Get ALICLOUD_REGION by picking one RegionId from Alibaba Cloud region list https://www.alibabacloud.com/help/doc-detail/72379.htm
      --ALICLOUD_SECRET_KEY string   Get ALICLOUD_SECRET_KEY per this guide https://help.aliyun.com/knowledge_detail/38738.html
  -h, --help                         help for terraform-alibaba
      --name default                 The name of Terraform Provider for Alibaba Cloud, default is default (default "default")

Global Flags:
```

There won't be any inconsistencies as the default name is the provider name, like https://github.com/oam-dev/catalog/blob/master/addons/terraform-baidu/definitions/baidu-provider.cue#L21.

barnettZQG: @zzxwill So, both the CLI and the API will create an application that includes a component. you could set the component name by value from `--name`. no need set a parameter with the name `name`.

For the default value:

* API: no need set it, users must set a name for the configuration.
* CLI: we could set a default value for the different provider kinds.

zzxwill: @barnettZQG Correct! For CLI users, we can set the default value for difficult providers.
```
switch {
  case "terraform-alibaba":
    name := "default"
  case "terraform-aws":
    name := "aws"
...
}
```
In this case, the implementation for VelaUX is simpler, and the parameter `name` in the definition is not needed. In the original case, the implementation for CLI is simpler.

barnettZQG: About the default name, is there any other reason besides compatibility?

barnettZQG: default name == provider name?

zzxwill: @barnettZQG 

> About the default name, is there any other reason besides compatibility?

No. All definitions also uses the default name in the mainifests.

zzxwill: > default name == provider name?

Correct.

barnettZQG: Define the UI schemas for all providers, and hide the named field in UI.

barnettZQG (approved):
lgtm

PR #3710, opened 2022-04-20
Fix: fix configuration not found health check error

FogDong:

fix configuration not found health check error

I have:

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/apply.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/apply.go
@@ -26,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/common"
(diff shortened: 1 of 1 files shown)


pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -231,7 +232,8 @@ func (h *AppHandler) collectHealthStatus(ctx context.Context, wl *appfile.Worklo
 	if wl.CapabilityCategory == types.TerraformCategory {
 		var configuration terraformapi.Configuration
 		if err := h.r.Client.Get(ctx, client.ObjectKey{Name: wl.Name, Namespace: namespace}, &configuration); err != nil {
-			return nil, false, errors.WithMessagef(err, "app=%s, comp=%s, check health error", appName, wl.Name)
+			klog.ErrorS(err, "failed to get configuration", "app name", appName, "comp name", wl.Name)
+			return nil, false, nil
```
wonderflow, line 236: why don't return the error here?

anoop2811: Same question as @wonderflow . Maybe I am missing what the fix here is as it only seems to log the error.

FogDong: If we return the error here, in the application status, we can see the `configuration not found error...`. but this error is caused by we failed to get configuration when we check the health.

wonderflow: the root cause is why we will miss the configuration here? Will we periodically check the existence of configuration?

PR #3713, opened 2022-04-20
Feat: multi-cluster authentication

Somefive:

Support authenticate application. By default, disabled.

1. All request from vela-core controller to KAS (kube-apiserver) will now impersonate as User `vela-core` with Group `core.oam.dev`. Group "core.oam.dev" is binded to ClusterRole `cluster-admin`.
2. When user create or modify application, the request will be processed by the application mutating webhook. If the authentication is enabled, the webhook will inject the user info into the application annotations. *Notice that the requests carries `core.oam.dev` group will not inject user info. These requests are sent by vela-core controller.* Only **Groups** with specified pattern (`kubevela:*` by default) will be injected and User is not. This can be modified by setting bootstrap parameter `--authentication-with-user` and `--authentication-group-pattern`. 
3. When dispatching resources or deleting resources in application controller, the user info in application annotations will be passed into the context. The context will then be consumed by the impersonate roundTripper, where user info will be converted into impersonation headers. *To let all requests be able to be delegated to ClusterGateway, the `cluster-gateway-accessor` group is automatically added to the impersonation group.*
4. The request, carrying impersonation headers, will be handled by KAS and after impersonated, it will then delegated to ClusterGateway.
5. ClusterGateway will recognize the impersonated user and re-carries these information into headers when request is directed to KAS in managed cluster.

Step 4 and 5 is done by KAS and ClusterGateway automatically. This PR mainly handles Step 1~3.

This PR only implements the authentication part and has nothing to do with authorization. We might need convenient authorization tools to assign roles in managed clusters in Vela CLI.

This PR is compatible with the previous ServiceAccount impersonation in #3432 & #3434.
When authentication is disabled, the application's ServiceAccount annotation will be recognized as the impersonation target and reuse the impersonation. The only break change is that in multi-cluster scenario, the ServiceAccount impersonation will also be passed into managed clusters.

I have:

- [x] Add Test.

### DIFF
--- a/charts/vela-core/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
+++ b/charts/vela-core/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
@@ -125,6 +125,32 @@ webhooks:
       service:
         name: {{ template "kubevela.name" . }}-webhook
         namespace: {{ .Release.Namespace }}
+        path: /mutating-core-oam-dev-v1beta1-applications
+    {{- if .Values.admissionWebhooks.patch.enabled }}
+    failurePolicy: Ignore
+    {{- else }}
+    failurePolicy: Fail
+    {{- end }}
+    name: mutating.core.oam.dev.v1beta1.applications
+    admissionReviewVersions:
+      - v1beta1
+      - v1
+    sideEffects: None
+    rules:
+      - apiGroups:
+          - core.oam.dev
+        apiVersions:
+          - v1beta1
+        operations:
+          - CREATE
+          - UPDATE
+        resources:
+          - applications
+  - clientConfig:
+      caBundle: Cg==
+      service:
+        name: {{ template "kubevela.name" . }}-webhook
+        namespace: {{ .Release.Namespace }}
         path: /mutating-core-oam-dev-v1beta1-componentdefinitions
     {{- if .Values.admissionWebhooks.patch.enabled  }}
     failurePolicy: Ignore
--- a/charts/vela-minimal/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
+++ b/charts/vela-minimal/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
@@ -97,6 +97,32 @@ webhooks:
       service:
         name: {{ template "kubevela.name" . }}-webhook
         namespace: {{ .Release.Namespace }}
+        path: /mutating-core-oam-dev-v1beta1-applications
+    {{- if .Values.admissionWebhooks.patch.enabled }}
+    failurePolicy: Ignore
+    {{- else }}
+    failurePolicy: Fail
+    {{- end }}
+    name: mutating.core.oam.dev.v1beta1.applications
+    admissionReviewVersions:
+      - v1beta1
+      - v1
+    sideEffects: None
+    rules:
+      - apiGroups:
+          - core.oam.dev
+        apiVersions:
+          - v1beta1
+        operations:
+          - CREATE
+          - UPDATE
+        resources:
+          - applications
+  - clientConfig:
+      caBundle: Cg==
+      service:
+        name: {{ template "kubevela.name" . }}-webhook
+        namespace: {{ .Release.Namespace }}
         path: /mutating-core-oam-dev-v1beta1-componentdefinitions
     {{- if .Values.admissionWebhooks.patch.enabled  }}
     failurePolicy: Ignore
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -36,6 +36,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
+	apicommon "github.com/oam-dev/kubevela/apis/core.oam.dev/common"
+	"github.com/oam-dev/kubevela/apis/types"
 	"github.com/oam-dev/kubevela/pkg/auth"
 	ctrlClient "github.com/oam-dev/kubevela/pkg/client"
 	standardcontroller "github.com/oam-dev/kubevela/pkg/controller"
@@ -50,6 +52,7 @@ import (
 	"github.com/oam-dev/kubevela/pkg/oam"
 	"github.com/oam-dev/kubevela/pkg/oam/discoverymapper"
 	"github.com/oam-dev/kubevela/pkg/resourcekeeper"
+	pkgutils "github.com/oam-dev/kubevela/pkg/utils"
 	"github.com/oam-dev/kubevela/pkg/utils/common"
 	"github.com/oam-dev/kubevela/pkg/utils/system"
 	"github.com/oam-dev/kubevela/pkg/utils/util"
@@ -57,10 +60,6 @@ import (
 	"github.com/oam-dev/kubevela/pkg/workflow"
 	"github.com/oam-dev/kubevela/pkg/workflow/tasks/custom"
 	"github.com/oam-dev/kubevela/version"
-)
-
-const (
-	kubevelaName = "kubevela"
 )
 
 var (
@@ -202,10 +201,22 @@ func main() {
 	klog.InfoS("Vela-Core init", "definition namespace", oam.SystemDefinitonNamespace)
 
 	restConfig := ctrl.GetConfigOrDie()
-	restConfig.UserAgent = kubevelaName + "/" + version.GitRevision
+	restConfig.UserAgent = types.KubeVelaName + "/" + version.GitRevision
 	restConfig.QPS = float32(qps)
 	restConfig.Burst = burst
 	restConfig.Wrap(auth.NewImpersonatingRoundTripper)
+	restConfig.Impersonate.UserName = types.VelaCoreName
+	if sub := pkgutils.GetServiceAccountSubjectFromConfig(restConfig); sub != "" {
+		restConfig.Impersonate.UserName = sub
+	}
+	restConfig.Impersonate.Groups = []string{apicommon.Group}
+	klog.InfoS("Kubernetes Config Loaded",
+		"UserAgent", restConfig.UserAgent,
+		"QPS", restConfig.QPS,
+		"Burst", restConfig.Burst,
+		"Impersonate-User", restConfig.Impersonate.UserName,
+		"Impersonate-Group", strings.Join(restConfig.Impersonate.Groups, ","),
+	)
 
 	// wrapper the round tripper by multi cluster rewriter
 	if enableClusterGateway {
@@ -225,7 +236,7 @@ func main() {
 	}
 	ctrl.SetLogger(klogr.New())
 
-	leaderElectionID := util.GenerateLeaderElectionID(kubevelaName, controllerArgs.IgnoreAppWithoutControllerRequirement)
+	leaderElectionID := util.GenerateLeaderElectionID(types.KubeVelaName, controllerArgs.IgnoreAppWithoutControllerRequirement)
 	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
 		Scheme:                     scheme,
 		MetricsBindAddress:         metricsAddr,
--- a/pkg/auth/round_trippers.go
+++ b/pkg/auth/round_trippers.go
@@ -22,11 +22,15 @@ import (
 	"net/http"
 
 	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/client-go/transport"
 
-	"github.com/oam-dev/kubevela/pkg/multicluster"
-	oamutil "github.com/oam-dev/kubevela/pkg/oam/util"
+	"github.com/oam-dev/kubevela/apis/types"
 	"github.com/oam-dev/kubevela/pkg/utils"
+)
+
+const (
+	impersonateKey = "impersonate"
 )
 
 var _ utilnet.RoundTripperWrapper = &impersonatingRoundTripper{}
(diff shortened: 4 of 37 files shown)


charts/vela-minimal/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
```diff
@@ -92,6 +92,32 @@ webhooks:
...
+    name: validating.core.oam.dev.v1beta1.applications
+    admissionReviewVersions:
+      - v1beta1
+      - v1
+    sideEffects: None
+    rules:
+      - apiGroups:
+          - core.oam.dev
+        apiVersions:
+          - v1beta1
```
yue9944882, line 115: missed v1alpha1?

Somefive: The webhook is applied to v1beta1.Application which is already in version v1beta1.

pkg/auth/round_trippers_test.go
```diff
@@ -76,12 +87,13 @@ func TestImpersonatingRoundTripper(t *testing.T) {
...
 			require.NoError(t, err)
-			if ts.expected == "" {
+			if ts.expectedUser == "" {
 				_, ok := rt.Request.Header[transport.ImpersonateUserHeader]
 				require.False(t, ok)
 				return
 			}
-			require.Equal(t, ts.expected, rt.Request.Header.Get(transport.ImpersonateUserHeader))
+			require.Equal(t, ts.expectedUser, rt.Request.Header.Get(transport.ImpersonateUserHeader))
+			require.Equal(t, ts.expectedGroup, rt.Request.Header.Values(transport.ImpersonateGroupHeader))
```
devholic, line 96: It would be great if we check group values even if the `expectedUser` is blank for [this](https://github.com/oam-dev/kubevela/pull/3713/files#diff-d23fc29a3ee18f5d0c238a21f6cfe8769843e53a3b5108b1920b0249a59115bdR66-R67) test case.

pkg/auth/userinfo.go
```diff
@@ -0,0 +1,81 @@
...
+		name = user.Anonymous
+	}
+
+	return &user.DefaultInfo{
+		Name: name,
+		Groups: slices.Filter(
+			[]string{},
+			strings.Split(annotations[oam.AnnotationApplicationGroup], groupSeparator),
+			func(s string) bool {
+				return s != ""
```
devholic, line 78: We may filter out if the given group is blank or not
```suggestion
				return len(strings.TrimSpace(s)) > 0
```

pkg/webhook/core.oam.dev/v1alpha2/application/mutating_handler.go
```diff
@@ -0,0 +1,79 @@
...
+	if slices.Contains(req.UserInfo.Groups, common.Group) {
+		return admission.Patched("")
+	}
+
+	app := &v1beta1.Application{}
+	if err := h.Decoder.Decode(req, app); err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	auth.SetUserInfoInAnnotation(&app.ObjectMeta, req.UserInfo)
```
devholic, line 58: Q. So if the feature gate enabled, does it always overwrite the user info annotation with the last-modified user?

devholic (commented):
Thanks for your work 👍

pkg/auth/round_trippers.go
```diff
@@ -45,18 +49,20 @@ func NewImpersonatingRoundTripper(rt http.RoundTripper) http.RoundTripper {
...
-		return rt.rt.RoundTrip(req)
+	req = req.Clone(ctx)
+	userInfo, exists := request.UserFrom(ctx)
+	if exists && userInfo != nil {
+		if name := userInfo.GetName(); name != "" {
+			req.Header.Set(transport.ImpersonateUserHeader, name)
+			req.Header.Set(transport.ImpersonateGroupHeader, clusterGatewayAccessorGroup)
+			for _, group := range userInfo.GetGroups() {
+				req.Header.Add(transport.ImpersonateGroupHeader, group)
+			}
```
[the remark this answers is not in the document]
Somefive: Right.

pkg/auth/round_trippers.go
```diff
@@ -22,13 +22,17 @@ import (
...
 	"k8s.io/client-go/transport"
 
-	"github.com/oam-dev/kubevela/pkg/multicluster"
-	oamutil "github.com/oam-dev/kubevela/pkg/oam/util"
 	"github.com/oam-dev/kubevela/pkg/utils"
 )
 
+const (
+	impersonateKey              = "impersonate"
+	clusterGatewayAccessorGroup = "cluster-gateway-accessor"
```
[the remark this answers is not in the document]
Somefive: Good idea. `ClusterGatewayAccessorGroup` moved to package apis/types.

pkg/auth/round_trippers_test.go
Somefive: Good idea. Another test is added. The current test is to verify empty application in ContextWithInfo as well.

pkg/auth/userinfo.go
Somefive: Nice idea.

pkg/webhook/core.oam.dev/v1alpha2/application/mutating_handler.go
Somefive: Yes, it will. Another feature is introduced now. By default, the User will not be injected into Application. Only Groups will specified pattern will be injected. If --authentication-with-user flag is used in controller bootstrap parameter, it will do so.

Somefive: BTW, even if this feature gate is not enabled, the previous ServiceAccount authentication will still work as well. The major changes made by this feature gate is:
1. Use the MutatingWebhook to inject user info into Application.
2. Dispatching resources will always send requests with impersonation in avoid of using the controller's identity. This is done by automatically adding Impersonate-User if user not exists in application.
So when the feature gate is not enabled, if the Application carries ServiceAccount annotation, the Impersonate-User will be set to the SA and not be overriden by the default user.

cmd/core/main.go
```diff
@@ -206,6 +208,8 @@ func main() {
 	restConfig.QPS = float32(qps)
 	restConfig.Burst = burst
 	restConfig.Wrap(auth.NewImpersonatingRoundTripper)
+	restConfig.Impersonate.UserName = controllerName
```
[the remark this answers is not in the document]
Somefive: `username` in `restConfig` is used for basic auth, which is not the actual User for the client. I add an auxiliary function to extract ServiceAccount username from the decoded JWT token and use the `vela-core` as the fallback name.

charts/vela-core/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
```diff
@@ -120,6 +120,32 @@ webhooks:
...
+    rules:
+      - apiGroups:
+          - core.oam.dev
+        apiVersions:
+          - v1beta1
+        operations:
+          - CREATE
+          - UPDATE
+        resources:
+          - applications
```
wonderflow, line 148: we already have a validating webhook for application in the file `validatingWebhookConfiguration`

wonderflow (approved):
generally LGTM

charts/vela-core/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
Somefive: Yes. This handler is doing a mutating job.

pkg/utils/k8s.go
```diff
@@ -106,3 +107,10 @@ func UpdateNamespace(ctx context.Context, kubeClient client.Client, name string,
 	}
 	return kubeClient.Update(ctx, &namespace)
 }
+
+// GetServiceAccountSubjectInToken extract ServiceAccount subject from token
+func GetServiceAccountSubjectInToken(token string) string {
```
devholic, line 112: Nit: How about moving extracting logic to `jwt.go` and calling it in this function? Extracting subject from JWTs can be used in non-kubernetes usecases too.

```go
func GetTokenSubject(token string) string { ... }

func GetSubjectFromConfig(cfg rest.Config) string {
    return GetTokenSubject(cfg.BearerToken)
}
```

devholic (approved):
LGTM 👍

pkg/utils/k8s.go
```diff
@@ -106,3 +107,10 @@ func UpdateNamespace(ctx context.Context, kubeClient client.Client, name string,
...
+	claims := jwt.MapClaims{}
+	_, _ = jwt.ParseWithClaims(token, claims, nil)
+	return claims["sub"].(string)
```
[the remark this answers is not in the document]
Somefive: Great suggestion.

pkg/utils/k8s.go
Somefive: Nice. Edited now.

charts/vela-core/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
Somefive: Fixed the validating typo.

devholic (approved):

pkg/auth/userinfo.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	re := regexp.MustCompile(strings.ReplaceAll(AuthenticationGroupPattern, "*", ".*"))
+	var groups []string
+	for _, group := range userInfo.Groups {
+		if re.MatchString(group) {
+			groups = append(groups, group)
+		}
+	}
+	metav1.SetMetaDataAnnotation(obj, oam.AnnotationApplicationGroup, strings.Join(groups, groupSeparator))
+	util.RemoveAnnotations(obj, []string{oam.AnnotationApplicationServiceAccountName})
```
Somefive, line 65: Here a critical change is added. @devholic 
I realize that when in the mode that User is not attached to the application annotation, if the ServiceAccount annotation is set by user, it could lead to the leak of permissions.
For example, one user X only have a small scope of permissions but it added a service account name in the application annotation, which contains larger scope of permissions, the application will work with more permissions it should not have.
So the change here means, when enabling authentication in KubeVela, the ServiceAccount annotation will not work anymore. But in this case, setting up `--authentication-with-user` and adding ServiceAccount token to the creation process of Application will also work.

devholic: Thanks for pointing this out 👍 

I think returning an error instead of mutating the request is better since it can give more information to the users and prevent them from unexpected behaviors. (Anyway, mutating sounds reasonable to me 🙂)

---

Maybe, we can enhance this feature by letting users annotate `allowed users` on the ServiceAccount, and checking it in the validating webhook. WDYT? (This is not important for now, so we may discuss this in the other issue.)

Somefive: @devholic Thanks for the reply :)

> I think returning an error instead of mutating the request is better since it can give more information to the users and prevent them from unexpected behaviors. (Anyway, mutating sounds reasonable to me 🙂)

I agree with that. So it's like returning BadRequest error from the mutating webhook when it saw the ServiceAccount annotation exists and the authentication option enabled?  

> Maybe, we can enhance this feature by letting users annotate `allowed users` on the ServiceAccount, and checking it in the validating webhook. WDYT? (This is not important for now, so we may discuss this in the other issue.)

I'm a bit confused with the `allowed users`. Is it an option in the validating webhook configured by the system maintainer? Not sure if I'm understanding it correctly. If so, I think it is useful for the non-authentication impersonation scene (which is the case without authentication switched on and using the ServiceAccount impersonation).

pkg/utils/jwt.go
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+package utils
+
+import "github.com/form3tech-oss/jwt-go"
+
+// GetTokenSubject extract the subject field from the jwt token
+func GetTokenSubject(token string) string {
+	claims := jwt.MapClaims{}
+	_, _ = jwt.ParseWithClaims(token, claims, nil)
```
FogDong, line 24: Should we ignore the error here?

leejanee (approved):
LGTM

pkg/utils/jwt.go
Somefive: OK

pkg/auth/userinfo.go
Somefive: > I think returning an error instead of mutating the request is better since it can give more information to the users and prevent them from unexpected behaviors. (Anyway, mutating sounds reasonable to me 🙂)

I moved this logic in mutating handler to validate the service account annotation non-existence when authentication enabled.

> Maybe, we can enhance this feature by letting users annotate allowed users on the ServiceAccount, and checking it in the validating webhook. WDYT? (This is not important for now, so we may discuss this in the other issue.)

We can reserve this enhancement discussion in future PR. :)

FogDong (approved):

pkg/auth/userinfo.go
devholic: Sorry for the late reply 🙇 

> I'm a bit confused with the allowed users. Is it an option in the validating webhook configured by the system maintainer? Not sure if I'm understanding it correctly. If so, I think it is useful for the non-authentication impersonation scene (which is the case without authentication switched on and using the ServiceAccount impersonation).

My suggestion was that the system maintainer may specify allowed users for the ServiceAccount. So if the allowed user tries to edit ServiceAccount-bound Application, there is no need to remove ServiceAccount annotation from the Application.

> We can reserve this enhancement discussion in future PR. :)

👍

Somefive: I see. We can do it in further PR. 👍🏼

PR #3768, opened 2022-04-27
Feat: add apiextensionserver (solve cluster-scoped ResourceTracker)

Somefive:

1. Add APIExtensionServer, a k8s native extension API server, like cluster-gateway. (which is different from vela apiserver)
2. The APIExtensionServer can provider web interfaces that conform the Kubernetes Standard API, which means the APIExtensionServer web interface can be accessed by clients like client-go or kubectl.
3. In this PR, the ApplicationResourceTracker is proposed. This object works as a delegator to the ResourceTracker but is namespace-scoped. Users can access the ApplicationResourceTracker by `kubectl get apprt -n <namespace>` and the APIExtensionServer will find the corresponding underlying ResourceTracker. Therefore, we can restrict the user access to ResourceTracker within specified namespace by using the Kubernetes RBAC authorization. 
4. The APIExtensionServer is not installed by default since the ApplicationResourceTracker proxy is only useful in some special scenarios where we need to restrict the user's permission. But it would be an interesting step to align more apis used in VelaUX and vela-core, such as DeliveryTarget / VirtualCluster / Project...

ApplicationResourceTracker Usage:
- Run `kubectl get apprt -n example` will retrieve all the ResourceTracker targeting applications in example namespace.
- Run `kubectl get apprt -A` will retrieve all ResourceTrackers, same as `kubectl get rt`.
- Run `kubectl get apprt app-v1 -n default` will retrieve ResourceTracker `app-v1-default`.

No additional storage or CRD is needed.

FYI, this PR also bumps multiple core dependencies like controller-runtime and k8s.io/xxx.

I have:

This work will not affect the controller or existing CLI / velaux-apiserver logic.

This APIExtensionServer is tested by integration test. There is no special e2e test for it.

The auto-build for APIExtensionServer image is not added by now. We will work on it later.


charts/vela-core/README.md
```diff
@@ -145,6 +145,22 @@ helm install --create-namespace -n vela-system kubevela kubevela/vela-core --wai
 | `authentication.groupPattern` | Application authentication will impersonate as the request Group that matches the pattern                                  | `kubevela:*`         |
 
 
+### APIExtensionServer parameters
+
+| Name                                         | Description                          | Value                               |
+| -------------------------------------------- | ------------------------------------ | ----------------------------------- |
+| `apiextensionserver.enabled`                 | Whether to enable APIExtensionServer | `false`                             |
```
wonderflow, line 152:
```suggestion
| `apiext.enabled`                 | Whether to enable APIExtensionServer | `false`                             |
```

let's make the configuration simpler

charts/vela-core/values.yaml
```diff
@@ -242,3 +242,31 @@ authentication:
...
+## @param apiextensionserver.resources.limits.memory APIExtensionServer memory limit
+## @param apiextensionserver.secureTLS.enabled Whether to enable secure TLS
+## @param apiextensionserver.secureTLS.certPath Path to the certificate file
+apiextensionserver:
+  enabled: false
+  replicaCount: 1
+  port: 9443
+  image:
+    repository: oamdev/vela-apiextensionserver
+    tag: v1.3.2
```
wonderflow, line 264: shall we inject this version automatically with the release?

pkg/resourcetracker/app.go
```diff
@@ -100,17 +106,42 @@ func CreateComponentRevisionResourceTracker(ctx context.Context, cli client.Clie
...
 		oam.LabelAppName:      app.Name,
 		oam.LabelAppNamespace: app.Namespace,
-	}); err != nil {
+	})
+	if err == nil {
+		return rts.Items, nil
+	}
+	if !kerrors.IsForbidden(err) && !kerrors.IsUnauthorized(err) {
+		return nil, err
+	}
```
wonderflow, line 120:
```suggestion
	}
	
	// we will check the apiextension server for unauthorized apps
```

Let's add some explanation here for more clear

Dockerfile.apiextensionserver
```diff
@@ -0,0 +1,54 @@
...
+WORKDIR /
+
+ARG TARGETARCH
+COPY --from=upx /workspace/apiextensionserver-${TARGETARCH}-upx /usr/local/bin/apiextensionserver
+
+COPY entrypoint.sh /usr/local/bin/
+
+ENTRYPOINT ["entrypoint.sh"]
+
+CMD ["apiextensionserver"]
```
wonderflow, line 54: can we give a better name for this server?  How about `velaguard`？

Somefive:
This PR is not needed anymore due to #3824 & [vela-prism](https://github.com/kubevela/vela-prism).

PR #3824, opened 2022-05-07
Feat: add support for ApplicationResourceTracker

Somefive:

Support use ApplicationResourceTracker for list resource trackers when users do not have permissions to access cluster-scoped ResourceTracker.

This PR is a cherry-pick for #3768.
The server part is moved to [vela-prism](https://github.com/kubevela/prism).

Fixes #

I have:

### DIFF
--- a/pkg/resourcetracker/app.go
+++ b/pkg/resourcetracker/app.go
@@ -20,20 +20,33 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/oam-dev/kubevela/pkg/monitor/metrics"
-
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/pkg/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/common"
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
+	"github.com/oam-dev/kubevela/pkg/monitor/metrics"
 	"github.com/oam-dev/kubevela/pkg/oam"
+	velaerrors "github.com/oam-dev/kubevela/pkg/utils/errors"
 )
 
 const (
 	// Finalizer for resourcetracker to clean up recorded resources
 	Finalizer = "resourcetracker.core.oam.dev/finalizer"
+)
+
+var (
+	applicationResourceTrackerGroupVersionKind = schema.GroupVersionKind{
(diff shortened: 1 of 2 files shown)


pkg/resourcetracker/app.go
```diff
@@ -20,22 +20,35 @@ import (
...
 )
 
 const (
 	// Finalizer for resourcetracker to clean up recorded resources
 	Finalizer = "resourcetracker.core.oam.dev/finalizer"
 )
 
+var (
+	applicationResourceTrackerGroupVersionKind = schema.GroupVersionKind{
+		Group:   "apiextensions.core.oam.dev",
```
wonderflow, line 46:
```suggestion
		Group:   "prism.core.oam.dev",
```

pkg/resourcetracker/app.go
```diff
@@ -100,17 +113,62 @@ func CreateComponentRevisionResourceTracker(ctx context.Context, cli client.Clie
...
+	if !kerrors.IsForbidden(err) && !kerrors.IsUnauthorized(err) {
+		return nil, err
+	}
+	appRts := &unstructured.UnstructuredList{}
+	appRts.SetGroupVersionKind(applicationResourceTrackerGroupVersionKind)
+	if err = cli.List(ctx, appRts, client.MatchingLabels{
+		oam.LabelAppName: app.Name,
+	}, client.InNamespace(app.Namespace)); err != nil {
+		if velaerrors.IsCRDNotExists(err) {
+			return nil, errors.New("no permission for ResourceTracker and apiextensions are not enabled")
```
wonderflow, line 149: let's return the former err in L133 here

pkg/resourcetracker/app.go
```diff
@@ -100,17 +113,62 @@ func CreateComponentRevisionResourceTracker(ctx context.Context, cli client.Clie
...
+		}
+		return nil, err
```
wonderflow, line 151: can we check if the prism not installed/availabile and nofity user to check that?

Somefive: @wonderflow Actually, L148 is checking for that.

pkg/resourcetracker/app.go
Somefive: @wonderflow How about "prism.oam.dev"? Will it be confusing?

wonderflow: prism.oam.dev is fine

pkg/resourcetracker/app.go
wonderflow: it's only check CRD exists, does it mean the server is running?

Somefive: If the server is not running, the APIService will not work properly, then it will return CRD not exists. The underlying logic is okay.

pkg/resourcetracker/app.go
Somefive: OK.

pkg/resourcetracker/app.go
wonderflow: ok

pkg/resourcetracker/app.go
Somefive: OK.

wonderflow (approved):

PR #3828, opened 2022-05-08
Fix: refactors lint issues in missing err in `json` return value

afzal442:

Fixes #

I have:


pkg/apiserver/datastore/kubeapi/kubeapi.go
```diff
@@ -79,7 +79,10 @@ func generateName(entity datastore.Entity) string {
 }
 
 func (m *kubeapi) generateConfigMap(entity datastore.Entity) *corev1.ConfigMap {
-	data, _ := json.Marshal(entity)
+	data, err := json.Marshal(entity)
+	if err != nil {
+		return nil
```
wonderflow, line 84: what does it mean here by return nil but found err

pkg/apiserver/sync/worker.go
```diff
@@ -58,7 +58,10 @@ func startAppSyncing(ctx context.Context, factory dynamicinformer.DynamicSharedI
 	informer := factory.ForResource(v1beta1.SchemeGroupVersion.WithResource("applications")).Informer()
 	getApp := func(obj interface{}) *v1beta1.Application {
 		app := &v1beta1.Application{}
-		bs, _ := json.Marshal(obj)
+		bs, err := json.Marshal(obj)
+		if err != nil {
+			logrus.Fatal(err)
+		}
 		_ = json.Unmarshal(bs, app)
```
wonderflow, line 65: why we need to check the marshal error but can ignore the unmarshal error

wonderflow (changes requested):
this PR is confusing me in many error check places.

pkg/apiserver/datastore/kubeapi/kubeapi.go
afzal442: Sorry @wonderflow, I'll change it. Thanks.

afzal442: hey @wonderflow , if I make it like below. what you would think.
```go
	if err != nil {
		logrus.Fatal(err)
		return nil
	}
```

wonderflow: no, we shouldn't make it a fatal error as it didn't affect the main work

afzal442: so, I'm skipping this err!

pkg/cue/process/handle.go
```diff
@@ -185,12 +185,18 @@ func (ctx *templateContext) BaseContextFile() string {
 	buff += fmt.Sprintf(model.ContextPublishVersion+": \"%s\"\n", ctx.publishVersion)
 
 	if ctx.appLabels != nil {
-		bt, _ := json.Marshal(ctx.appLabels)
+		bt, err := json.Marshal(ctx.appLabels)
+		if err != nil {
+			return fmt.Sprint(err)
```
FogDong, line 190: We should add `error` to the return value of the function instead of returning error string.

pkg/apiserver/sync/worker.go
```diff
@@ -58,8 +58,14 @@ func startAppSyncing(ctx context.Context, factory dynamicinformer.DynamicSharedI
 	informer := factory.ForResource(v1beta1.SchemeGroupVersion.WithResource("applications")).Informer()
 	getApp := func(obj interface{}) *v1beta1.Application {
 		app := &v1beta1.Application{}
-		bs, _ := json.Marshal(obj)
-		_ = json.Unmarshal(bs, app)
+		bs, err := json.Marshal(obj)
+		if err != nil {
+			logrus.Fatal(err)
```
wonderflow, line 63:
```suggestion
			klog.Errorf("failed to convert obj %v to application with marshal err %v", obj, err)
			return nil
```

the same below, some places we shouldn't panic the process, just report the log is fine.

afzal442: Thanks @wonderflow !

pkg/cue/process/handle.go
```diff
@@ -235,21 +253,29 @@ func (ctx *templateContext) BaseContextFile() string {
...
+		}
 		buff += fmt.Sprintf("\n %s", structMarshal(string(d)))
 	}
 
-	return fmt.Sprintf("context: %s", structMarshal(buff))
+	return fmt.Sprintf("context: %s", structMarshal(buff)), nil
 }
 
 // ExtendedContextFile return cue format string of templateContext and extended secret context
 func (ctx *templateContext) ExtendedContextFile() string {
```
wonderflow, line 267: we should also return an error here if so

afzal442: If I do so, there will be other refactors required like in `template.go, handle_test.go, task.go`. Let me know if I still need to return an error here.

wonderflow: then, let's revert this part of the chage, I think it's fine to omit the error here.

afzal442: hmm! do you mean reverting the changes on `BaseContextFile()` method? 😕

wonderflow: yes

wonderflow: I'm afrait we will cause confusion if we return error info as the CUE context string back. Actually we must return a clear error type if we define err here.

afzal442: >  Actually we must return a clear error type if we define err here.

Sorry Jianbo! I couldn't get that. Also let me tell you as @FogDong suggested me returning `error` value for BaseContextFile() method.

wonderflow: Sorry, let me discuss with Fog.

FogDong: This PR will not return error info as the CUE context string, instead it returns err as an error.

afzal442: Sorry again! I'm still confused. Let me know if I'm wrong here;
```go
func (ctx *templateContext) BaseContextFile() (string, error) {
```
The above method is called by `context, err := ctx.BaseContextFile()` Then if error is not nil, then it returns `return fmt.Sprintf("failed to convert data to application with marshal err %v", err)`.

So, what's the logic of returning error info as the CUE context string? Thanks

afzal442: I think I got that. this `fmt.Sprintf("failed to convert data to application with marshal err %v", err)` should be  in the format of CUE ctx string.

FogDong: If there's err, the string should be `""`.
In your example, if there's err in `context, err := ctx.BaseContextFile()`, it should return `"", err`

pkg/cue/process/handle.go
```diff
@@ -235,21 +253,29 @@ func (ctx *templateContext) BaseContextFile() string {
...
-	context := ctx.BaseContextFile()
-
+	context, err := ctx.BaseContextFile()
+	if err != nil {
+		return fmt.Sprintf("failed to convert data to application with marshal err %v", err)
```
FogDong, line 270: That is to say, it's better if we return `"", err` here.  Thin function should can be `func (ctx *templateContext) ExtendedContextFile() (string, error)`

afzal442: Thanks much! @FogDong for your kind suggestion!

wonderflow (approved):
GOOD JOB!

PR #3849, opened 2022-05-10
Fix: use vela prism for accessing clusters as alternative

Somefive:

`vela status --tree` requires the cluster alias mapping. The cluster mapping, previously, requires the access to the cluster secret, which is kind of a credential leak for the clusters.

With the help of `vela-prism`, the vela CLI do not need to access the cluster secret to retrieve alias metadata now, but instead, through accessing `clusters.prism.oam.dev` object. If `vela-prism` is not installed, the retrieval logic will fallback to the old logic. 

> NOTICE: This PR introduces importing `vela-prism`. Following PRs will be made to propagate the usage of prism Cluster/ApplicationResourceTracker.

FYI, this PR also upgrades the `cluster-gateway` library and therefore some migration of cluster-gateway types or constants is made.

I have:


go.mod
```diff
@@ -41,30 +40,30 @@ require (
 	github.com/hashicorp/hcl/v2 v2.9.1
 	github.com/hinshun/vt10x v0.0.0-20180616224451-1954e6464174
 	github.com/imdario/mergo v0.3.12
+	github.com/kubevela/prism v0.0.0-20220510071112-8a71d6d2d142
```
wonderflow, line 43: we can make a fixed version for prism

go.mod
```diff
@@ -41,30 +40,30 @@ require (
...
 	github.com/kyokomi/emoji v2.2.4+incompatible
 	github.com/mitchellh/hashstructure/v2 v2.0.1
-	github.com/oam-dev/cluster-gateway v1.1.6
+	github.com/oam-dev/cluster-gateway v1.3.3-0.20220509095841-4272c540e1e9
```
wonderflow, line 46: release some version?

wonderflow (approved):
good job

go.mod
Somefive: Later will release official versions, and then update this dependency.

go.mod
Somefive: Similar to above.

zzxwill (approved):
Great.

barnettZQG (approved):

FogDong (approved):

PR #3850, opened 2022-05-10
Fix: there are probability to report error when controlPlaneOnly=true

fourierr:
when controlPlaneOnly=true, There are probability to report error "resource not found"

After applycomponent, when waiting for health, the trace still checks the CR of the corresponding trace from the managedcluster corresponding to comp. At this time, the status of the CR should be checked from local

I have:


wonderflow (approved):
good catch! ping @Somefive

Somefive (approved):

PR #1704, opened 2021-05-24
Support patch data-pass object in CUE

cnbailian:
feature #1685

PR #3854, opened 2022-05-10
Fix: delete context when patch outputs in trait

### LINKED ISSUE #1851: [Feature] Add support for a trait to patch to any outputs defined before it and not just the output defined by the corresponding component.
Currently, traits can patch the workload of the corresponding components only. It is not possible for traits to patch workloads of other traits. This tends to make a trait overly complex when its targeting a fairly complex resource. For e.g. in Istio, a `DestinationRule` is responsible for several related activities like loadbalancing and circuit breaking. However, I would like to expose these as two separate traits. So now I'm in a scenario where a two traits are attempting to work on a single resource which isn't the one described by the corresponding component.

There should be a field in a CUE part of the trait which indicated which resource is it trying to patch. So a trait can patch the resource defined by the component or any other `outputs` which were defined before it.

The only way to solve this right now is to make a separate component for `DestinationRule` all together and attach traits to that instead which doesn't really sound like a good idea.

**Additional context**

FogDong:

Fixes #1851

I have:

/cc @wonderflow

### DIFF
--- a/pkg/cue/definition/template.go
+++ b/pkg/cue/definition/template.go
@@ -43,6 +43,8 @@ const (
 	OutputsFieldName = model.OutputsFieldName
 	// PatchFieldName is the name of the struct contains the patch of CR data
 	PatchFieldName = "patch"
+	// PatchOutputsFieldName is the name of the struct containts the patch of outputs CR data
+	PatchOutputsFieldName = "patchOutputs"
 	// CustomMessage defines the custom message in definition template
 	CustomMessage = "message"
 	// HealthCheckPolicy defines the health check policy in definition template
@@ -344,8 +346,8 @@ func (td *traitDef) Complete(ctx process.Context, abstractTemplate string, param
 	}
 
 	patcher := inst.Lookup(PatchFieldName)
+	base, auxiliaries := ctx.Output()
 	if patcher.Exists() {
-		base, auxiliaries := ctx.Output()
 		p, err := model.NewOther(patcher)
 		if err != nil {
 			return errors.WithMessagef(err, "invalid patch of trait %s", td.name)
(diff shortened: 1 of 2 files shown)


wonderflow (approved):
please make sure it won't break any existing use case

oiar:
> 

No effect.

oiar:
> ping @[oiar](https://github.com/oiar), Will this changes affect the patch trait that we use?

No effect.

FogDong:
> Details

It may affect the users who're using `context: outputs: resource` to patch traits in trait definition, but those users are also affected by this bug: the context is overridden.

wonderflow:
> > please make sure it won't break any existing use case
> 
> It may affect the users who're using `context: outputs: resource` to patch traits in trait definition, but those users are also affected by this bug: the context is overridden.

Can you make some clarification about this bug fix and change? I think users will refer to this PR in release not.

wonderflow:
I'll leave this feature in v1.4 and not backport to 1.3. Because it may cause some incompatible.

FogDong:
Though this PR fixes the override context problem raised by #1704 , still, if we want to patch outputs and output at the same time, we'll patch outputs into the output.
We may need to change the code here to avoid it:
1. We can delete the `outputs` part in `patch` when patching to workload. But if we do so, users can not patch the outputs in the workload struct.
2. Or we can delete the original code and add some new field like `patchTrait` to patch traits.

/cc @wonderflow @leejanee @fourierr

wonderflow:
@FogDong Please give some examples

FogDong:
> @FogDong Please give some examples

For example, if we patch trait and workload at the same time in trait, the definition will like:

```yaml
  patch: {
      outputs: {
        service: {
          spec: {
            // +patchKey=port
            // +patchStrategy=retainKeys
            ports: [{
              port: 2333
              targetPort: 2333
              name: "test"
            }]
          }
        }
      }
      spec: template: spec: {
          containers: [{
              name: context.name
              // +patchKey=configMapRef.name,secretRef.name
              // +patchStrategy=retainKeys
              envFrom: [
                  {
                    configMapRef: {
                      name: parameter.cmname
                    }
                  },
                  {
                    secretRef: {
                      name: parameter.sename
                    }
                  }
              ]
          }]
      }
  }
```

After patching, the workload will like:

```yaml
// original workload struct
...

outputs: {
....
}
```

We can delete the `outputs` part when trying to patch the worload, but it'll make `outputs` our built-in field in the workload and users can not use them.

We can also change it to:

```yaml
patch: {
// patch to workload
}

patchTraits: {
// patch to traits
}
```

wonderflow:
Yes, use a separate keyword can make the things easier:

```
patch: {
// patch to workload
}

patchOutputs: {
// patch to traits or outputs from workload
}
```

leejanee (approved):
LGTM

fourierr:
lgtm

PR #3876, opened 2022-05-12
Feat: add more detailed output for `vela addon status`

### LINKED ISSUE #3860: [Feature] vela addon status shows more detail info.

The command `addon status` showing info  is too simple which only shows the addon whether  has been enabled. We can show more detail info about this addon.

These info must be showed.

1. The addon's readme witch tell user how to use this addon.
2. All available versions of this addon.
3. Dependency addon.
4. Available parameter of this addon.
5. If this addon is enabled show the setup parameter.

**Additional context**

charlie0129:

Fixes #3860

`vela addon status` now shows more detailed info.

##### Before

Screen Shot 2022-05-12 at 11 12 48 PM

##### After

- If an addon is installed, and it also exists in the registry:

  - non-verbose output
Screen Shot 2022-05-19 at 2 56 00 PM
  - verbose output
Screen Shot 2022-05-19 at 2 56 09 PM

- If an addon is not installed, but it does exist in the registry:

  - non-verbose output
Screen Shot 2022-05-19 at 2 56 23 PM
  - verbose output
Screen Shot 2022-05-19 at 2 56 36 PM

- If an addon is installed, but it does not exist in the registry:

  - non-verbose output
Screen Shot 2022-05-19 at 2 56 59 PM
  - verbose output
Screen Shot 2022-05-19 at 2 57 11 PM

- If an addon is not installed, nor does it exist in the registry:
  - non-verbose output (since it does not connect to network when `-v/--verbose` is off, it is not possible to know whether this addon is in registry, so, only `disabled` is shown)
Screen Shot 2022-05-19 at 2 57 49 PM
  - verbose output (now we check the registry to know that this addon does not exist, an error will be returned. It is handled by upstream error handlers, and return code will not be 0, indicating an error:
Screen Shot 2022-05-19 at 2 57 41 PM

I have:

Relevant unit tests have been added, which will test the four situations above.

Unit tests for helper functions have been created as well.


pkg/addon/helper.go
```diff
@@ -127,15 +128,59 @@ func EnableAddonByLocalDir(ctx context.Context, name string, dir string, cli cli
 	return nil
 }
 
-// GetAddonStatus is genrall func for cli and apiServer get addon status
+// GetAddonStatus is a wrapper for GetAddonStatusDetailed, to provide backwards compatibility.
+// It is faster than GetAddonStatusDetailed, and should be used in most cases.
```
zzxwill, line 132: Why is it faster?

charlie0129: Because it is equivalent to `GetAddonStatusDetailed(ctx, cli, name, false)`, notice the last argument.

Setting the last argument `detailed` to `true` will add `AddonPackage` (the addon install package, which includes parameter descriptions, default vaues and other information) to the return value. These information will be be fetched from the registry, which will take some time. 

So, if `AddonPackage` is not needed by the caller, setting the last argument `detailed` to `false` will skip fetching addon install packages, to speed things up

Because this function is called many times when listing addons `addon list` (we need to check each addon is enabled), it makes sense to make this faster.

charlie0129: Of course, it also makes sense to take the logic to get the addon install package outside of `func GetAddonStatus`. Then we don't really need these *faster* equivalents. Just use a separate function when needed.

It is probably a better way? I will refactor the code later.

wonderflow:
> If an addon is not installed, nor does it exist in the registry:

Why not show the addon is not existed in this case?

pkg/addon/helper.go
```diff
@@ -243,11 +278,104 @@ func GetObservabilityAccessibilityInfo(ctx context.Context, k8sClient client.Cli
...
+			vr := BuildVersionedRegistry(r.Name, r.Helm.URL, &common.HTTPOption{Username: r.Helm.Username, Password: r.Helm.Password})
+			for _, addonName := range addonNames {
+				wholePackage, err := vr.GetAddonWholePackage(ctx, addonName, "")
+				if err != nil {
+					continue
+				}
+				merge(wholePackage)
+			}
+		} else {
+			meta, err := r.ListAddonMeta()
```
wonderflow, line 336: will it use cache here? I'm not sure if the performance will be very bad here. \cc @wangyikewxgm 

Maybe, we can downgrade the experience to the old mode if the registry is not versioned.

pkg/addon/versioned_registry.go
```diff
@@ -36,6 +36,7 @@ type VersionedRegistry interface {
 	ListAddon() ([]*UIData, error)
 	GetAddonUIData(ctx context.Context, addonName, version string) (*UIData, error)
 	GetAddonInstallPackage(ctx context.Context, addonName, version string) (*InstallPackage, error)
+	GetAddonWholePackage(ctx context.Context, addonName, version string) (*WholeAddonPackage, error)
```
wonderflow, line 39:
```suggestion
	GetDetailedAddon(ctx context.Context, addonName, version string) (*WholeAddonPackage, error)
```

references/cli/addon.go
```diff
@@ -396,12 +396,13 @@ func statusAddon(name string, ioStreams cmdutil.IOStreams, cmd *cobra.Command, c
...
 	}
-	status, err := pkgaddon.GetAddonStatus(context.Background(), k8sClient, name)
+
+	status, str, err := generateAddonInfo(k8sClient, name)
 	if err != nil {
 		return err
 	}
 
-	fmt.Print(generateAddonInfo(name, status))
+	fmt.Print(str)
```
wonderflow, line 405: shall we remove this line? It's more like a debug message here.

wonderflow (approved):
generally LGTM, ping @wangyikewxgm  for review, mainly about the performance.

We should keep the `vela addon status` command response quickly.

charlie0129:
> Why not show the addon is not existed in this case?

Great suggestion, which will make the output more informative to the user. 

Will implement it!

references/cli/addon.go
charlie0129: It will print the actual outputs of `vela addon status`, which is returned as a string in `str`. I will refactor the code to make it clearer.

wangyikewxgm:
@wonderflow  The main costs is fetching  whole addon package. Because user cannot get  knowledge of the parameter info about this addon by vela cli before, so I design this feature. Please note that the `GetAddonStatus` in helper will be used in apiserver and cli, it must be quickly @charlie0129  . So I suggest add `--verbose` flag in `vela addon status`, if true cli will fetch whole package and show the parameter info, otherwise won't. The apiserver calls this func needn't fetch the whole package.

charlie0129:
> @wonderflow The main costs is fetching whole addon package. Because user cannot get knowledge of the parameter info about this addon by vela cli before, so I design this feature. Please note that the `GetAddonStatus` in helper will be used in apiserver and cli, it must be quickly @charlie0129 . So I suggest add `--verbose` flag in `vela addon status`, if true cli will fetch whole package and show the parameter info, otherwise won't. The apiserver calls this func needn't fetch the whole package.

Yes, that is why I worried about the performance issue at first. But now `GetAddonStatus` doesn't get the whole package. It just checks current parameters from secrets, which (I think) is relatively fast. The actual process of getting the whole package is in `generateAddonInfo`, which is called only when `vela addon status` is executed.

wangyikewxgm:
> > @wonderflow The main costs is fetching whole addon package. Because user cannot get knowledge of the parameter info about this addon by vela cli before, so I design this feature. Please note that the `GetAddonStatus` in helper will be used in apiserver and cli, it must be quickly @charlie0129 . So I suggest add `--verbose` flag in `vela addon status`, if true cli will fetch whole package and show the parameter info, otherwise won't. The apiserver calls this func needn't fetch the whole package.
> 
> Yes, that is why I worried about the performance issue at first. But now `GetAddonStatus` doesn't get the whole package. It just checks current parameters from secrets, which (I think) is relatively fast. The actual process of getting the whole package is in `generateAddonInfo`, which is called only when `vela addon status` is executed.

Please be aware that if the network is too slow  to get the whole addon package from registry, user should still check the addon status quickly, this is why I suggest show detail info only when `verbose flag` is true.

charlie0129:
> So I suggest add `--verbose` flag in `vela addon status`

Good idea. Will be implemented!

wonderflow (approved):
great job!

pkg/addon/helper.go
charlie0129: it is only called when 
1. cli 
2. verbose is on
so it think it is not a big deal

Somefive (approved):
Fantastic work!

wangyikewxgm (approved):
LGTM

PR #3848, opened 2022-05-10
Feat(charts): add cert-manager support for "cluster-gateway" component in chart "vela-core"

cnfatal:
### Description of your changes

Fixes #3847 

I have:

- [x] Read and followed KubeVela's [contribution process](https://github.com/kubevela/kubevela/blob/master/contribute/create-pull-request.md).
- [x] [Related Docs](https://github.com/kubevela/kubevela.io) updated properly. In a new feature or configuration option, an update to the documentation is necessary. 
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### Special notes for your reviewer

should review per commit to find out changes instead review the whole changes together.

PR #3895, opened 2022-05-16
Feat: cluster gateway incorrect chart params

Somefive:

The cluster-gateway deployment in #3848 is modified to use incorrect parameters which will cause the mounted configmap not be properly used. This PR fixes it.

NOTICES: Without this PR, the default installation "helm install" for v1.4.0-alpha.3 will not be successfully finished. 

I have:

### DIFF
--- a/charts/vela-core/templates/cluster-gateway/cluster-gateway.yaml
+++ b/charts/vela-core/templates/cluster-gateway/cluster-gateway.yaml
@@ -33,8 +33,7 @@ spec:
             - "--secret-namespace={{ .Release.Namespace }}"
             - "--feature-gates=APIPriorityAndFairness=false"
             {{- if .Values.multicluster.clusterGateway.secureTLS.enabled }}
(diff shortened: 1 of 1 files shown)


wonderflow (approved):

barnettZQG (approved):

Somefive:
This issue needs further investigation.

Somefive:
The case here is
- #3848 changes the secret from apiserver.crt/apiserver.key to tls.crt/tls.key since the cert-manager uses tls.crt/tls.key.
- Existing installed KubeVela will encounter problems when upgrading to newer version (>=v1.4.0-alpha.3), since the secret will not automatically recreated. Therefore, tls.crt/tls.key is not embedded in the secret. However, since the cluster-gateway deployment changes its bootstrap parameter configuration, it will try to find tls.crt/tls.key but will fail to do that.

The manual solution is to delete the secret in the vela-system.

PR #3925, opened 2022-05-18
Feat: gateway trait support https protocol (#3832)

### LINKED ISSUE #3832: [Feature] gateway trait should support https protocol

The ingress behind of the gateway trait support  tls configuration but we didn't expose the arguments.

https://kubernetes.io/docs/concepts/services-networking/ingress/#tls

We should support it to support https route.

**Additional context**

oiar:
The gateway trait support https protocol (#3832)

Fixes #3832

I have:

### DIFF
--- a/charts/vela-core/templates/defwithtemplate/gateway.yaml
+++ b/charts/vela-core/templates/defwithtemplate/gateway.yaml
@@ -44,6 +44,14 @@ spec:
         		if parameter.classInSpec {
         			ingressClassName: parameter.class
         		}
+        		if parameter.secretName != _|_ {
+        			tls: [{
+        				hosts: [
+        					parameter.domain,
+        				]
+        				secretName: parameter.secretName
+        			}]
+        		}
         		rules: [{
         			host: parameter.domain
         			http: paths: [
@@ -71,6 +79,9 @@ spec:
 
         	// +usage=Set ingress class in '.spec.ingressClassName' instead of 'kubernetes.io/ingress.class' annotation.
         	classInSpec: *false | bool
+
+        	// +usage=Specify the secret name you want to quote.
+        	secretName?: string
         }
   status:
     customStatus: |-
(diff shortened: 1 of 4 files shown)


fourierr:
please run make renewable locally and add DCO before push

charts/vela-core/templates/defwithtemplate/gateway.yaml
```diff
@@ -44,6 +44,16 @@ spec:
         		if parameter.classInSpec {
         			ingressClassName: parameter.class
         		}
+        		if parameter.tlsInSpec {
```
wonderflow, line 47: what about use `secretName` to check if we use tls or not, then we don't need a redundant `tlsInSpec` field.

```suggestion
        		if parameter.secretName != _|_ {
```

fourierr:
ltgm, 
- if we extend tls to slice, the parameter.secretName also need to extend to slice, 
- the users who use the former version will be affected. 
- To keep compatibility,we can extend it after cue definition revisions
@wonderflow @oiar

fourierr (approved):

oiar:
> generally LGTM. @oiar you can refer to @fourierr suggustion, but we should keep the user experience clear

The rules only support one, now need to support multiple tls? The hosts in the tls section need to explicitly match the host in the rules section. Or do we need to modify the rules to be multiple?

pkg/controller/core.oam.dev/v1alpha2/application/application_controller_test.go
```diff
@@ -2664,6 +2689,59 @@ var _ = Describe("Test Application Controller", func() {
...
+			return k8sClient.Get(ctx, client.ObjectKey{Name: app.Name + "-v1", Namespace: app.GetNamespace()}, appRev)
+		}, 10*time.Second, 500*time.Millisecond).Should(Succeed())
+
+		By("Check ingress Created with the expected trait-gateway spec")
+		ingress := &v1beta12.Ingress{}
+		Expect(k8sClient.Get(ctx, client.ObjectKey{
+			Namespace: ns.Name,
+			Name:      app.Spec.Components[0].Name + "-ingress",
+		}, ingress)).Should(BeNil())
+
```
wonderflow, line 2740: why don't check the ingress has the secret and tls fields here?  It's the main goal for the long test, but you ignored.

PR #3939, opened 2022-05-20
Fix: remove last-applied-config annotation for configmap and secret

### LINKED ISSUE #3938: vela reports error when apply configmap: annotations: Too long: must have at most 262144 bytes
**Describe the bug**

> Vela在apply一个Grafana面板的configmap的时候，有“metadata.annotations: Too long: must have at most 262144 bytes”的报错；这个用kubectl提交可以用kubectl replace去workaround

When submit a grafana configmap by using kubevela, it reports “metadata.annotations: Too long: must have at most 262144 bytes”.

This is also a common issue in kubectl, but we can use `kubectl replace` to workaround, in kubevela, how can I fix that?

**To Reproduce**

Write an application with long configmap nested by using `ref-objects` component type.

**Expected behavior**

We can have any work around to update the object.

**Screenshots**

**KubeVela Version**

**Cluster information**

**Additional context**

wonderflow:
Signed-off-by: Jianbo Sun <jianbo.sjb@alibaba-inc.com>

This PR remove the last-applied-config annotation for configmap and secret, and allow users to disable it by fill `-` or `skip` into this annotation `app.oam.dev/last-applied-configuration`.

You can disable it by adding a annotation triat, or manuelly add this annotation on it.

```
kind: Application
metadata:
  name: first-vela-app
spec:
  components:
    - name: express-server
      type: webservice
      properties:
        image: oamdev/hello-world
        ports:
         - port: 8000
           expose: true
      traits:
        - type: annotations
          properties:
            app.oam.dev/last-applied-configuration: skip
```

Then the deployment won't have the `app.oam.dev/last-applied-configuration` filled:

image

Fixes #3938

I have:

### DIFF
--- a/pkg/utils/apply/apply.go
+++ b/pkg/utils/apply/apply.go
@@ -20,19 +20,19 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
-	"github.com/oam-dev/kubevela/pkg/controller/utils"
-	"github.com/oam-dev/kubevela/pkg/oam/util"
-
-	"github.com/oam-dev/kubevela/pkg/oam"
-
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
+	"github.com/oam-dev/kubevela/pkg/controller/utils"
+	"github.com/oam-dev/kubevela/pkg/oam"
+	"github.com/oam-dev/kubevela/pkg/oam/util"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/utils/apply/apply.go
```diff
@@ -107,13 +107,40 @@ func loggingApply(msg string, desired client.Object) {
...
+}
+
 // Apply applies new state to an object or create it if not exist
 func (a *APIApplicator) Apply(ctx context.Context, desired client.Object, ao ...ApplyOption) error {
 	_, err := generateRenderHash(desired)
 	if err != nil {
 		return err
 	}
-	applyAct := &applyAction{updateAnnotation: true}
+	fmt.Println("XXXXXXXXXXXXX", filterRecordForSpecial(desired))
```
leejanee, line 142: debug print？

leejanee (approved):

pkg/utils/apply/apply.go
wonderflow: good catch, thanks

pkg/utils/apply/apply.go
```diff
@@ -107,13 +107,39 @@ func loggingApply(msg string, desired client.Object) {
...
+}
+
 // Apply applies new state to an object or create it if not exist
 func (a *APIApplicator) Apply(ctx context.Context, desired client.Object, ao ...ApplyOption) error {
 	_, err := generateRenderHash(desired)
 	if err != nil {
 		return err
 	}
-	applyAct := &applyAction{updateAnnotation: true}
+	applyAct := &applyAction{updateAnnotation: filterRecordForSpecial(desired)}
```
barnettZQG, line 142: There is an ApplyOption with the name DisableUpdateAnnotation, can we use this option?

wonderflow: I know that option, but this is an initial and default behavior in this case

barnettZQG (approved):

PR #3945, opened 2022-05-21
Feat: Refactoring the API Server module for better layering

barnettZQG:

As the business logic increases, the previous architecture is no longer able to better fulfill the requirements. I have completed the reconstruction of the API module, as follows:

# APIServer Module Architecture

image

## Interfaces

Typically, it includes user interfaces, web services, and so on.

* API

Define the Restful API and perform basic validation on user input parameters.

* DTO

The carrier of data transmission. Here does not have any business logic inside, and the internal domain models can be isolated from the outside world through DTO.

* Assembler

Realize the mutual conversion and data exchange between DTO and DO.

## Event

Asynchronous task module. All workers under this module only run on the master node. The domain layer can be called.

## Domain

Code related to core business logic.

* Service

A domain service is a piece of business logic composed of multiple entities. You can put all domain services in the aggregate in one domain service class, or you can design each domain service as a class.

* Repository

Usually includes repository interface and repository implementation method. In order to facilitate the splitting and combination of aggregates, the principle is set: one aggregate corresponds to one repository. For example, save user information to a database.

* Model

Database model entity.

## Infrastructure

Provide general technical basic services for other layers, such as:

* Database
* Cache
* Kube Client
* MQ

I have:


cmd/apiserver/main.go
```diff
@@ -77,14 +78,13 @@ func main() {
 		return
 	}
 
-	srvc := make(chan struct{})
+	errChan := make(chan error, 1)
```
wonderflow, line 81: why we need a buffer here?

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/alibabacloud-go/darabonba-openapi v0.1.4
 	github.com/alibabacloud-go/tea v1.1.15
 	github.com/aryann/difflib v0.0.0-20210328193216-ff5ff6dc229b
+	github.com/barnettZQG/inject v0.0.1
```
wonderflow, line 15: why not fork it in kubevela org

pkg/apiserver/README.md
```diff
@@ -0,0 +1,48 @@
+# APIServer Module Architecture
```
wonderflow, line 1: we have some developer guide here, make sure this docs can be seen by add some reference, or we can directly put the docs there.

https://kubevela.io/docs/contributor/code-contribute#run-velaux-locally

wonderflow (approved):

wonderflow:
please fix the lint errors reported by the bot @barnettZQG

Somefive (approved):
Generally LGTM. Great Job!
P.S. Some of the names for the package I think is still a bit confusing. But generally I can accept this naming, better than before. 
Things that might be a bit inappropriate or confusing:
- Package `interface` in Golang sometimes refers to the abstraction layer of the code structure, but the `interface` here refers to `REST Interface` specifically, which is okay but might be a bit ambiguous. ServiceProvider/ServiceHandler could be more dedicated?
- Model & Service & Infrastructure are nice.
- Domain is generally okay but Probably `Resource` is a more native choice for Kubernetes? Although, `Resource` should be used in a more limited scenario.
- Repository is relatively confusing. It is hard for me to understand what does it really mean and why env/target/workflow should be separated from other model/service. If it is necessary for aggregation, then I can accept the separation here but the name `repository` might need more detailed explanation or be renamed in the future?

Somefive:
> Generally LGTM. Great Job! P.S. Some of the names for the package I think is still a bit confusing. But generally I can accept this naming, better than before. Things that might be a bit inappropriate or confusing:
> 
> * Package `interface` in Golang sometimes refers to the abstraction layer of the code structure, but the `interface` here refers to `REST Interface` specifically, which is okay but might be a bit ambiguous. ServiceProvider/ServiceHandler could be more dedicated?
> * Model & Service & Infrastructure are nice.
> * Domain is generally okay but Probably `Resource` is a more native choice for Kubernetes? Although, `Resource` should be used in a more limited scenario.
> * Repository is relatively confusing. It is hard for me to understand what does it really mean and why env/target/workflow should be separated from other model/service. If it is necessary for aggregation, then I can accept the separation here but the name `repository` might need more detailed explanation or be renamed in the future?

Seems like the overall arch is following [Domain Driven Design](https://en.wikipedia.org/wiki/Domain-driven_design) pattern.

yangsoon (approved):
lgtm

PR #3949, opened 2022-05-22
Feat: support download kustomize in local folder

JarHMJ:
Signed-off-by: huangminjie <minjie.huang@daocloud.io>

When I run `make kustomize`, I get the following err:
```
installing kustomize-v4.5.4      into /Users/jar/go/bin
/Users/jar/go/bin/kustomize exists. Remove it first.
make: *** [makefiles/dependency.mk:64: kustomize] Error 1
```

The reason is that I already have another version kustomize in my GOBIN.
So I think it is better to download the kustomize in project folder rather than GOBIN.

Fixes #

I have:

### DIFF
--- a/makefiles/dependency.mk
+++ b/makefiles/dependency.mk
@@ -56,20 +56,19 @@ else
 CUE=$(shell which cue)
 endif
 
-KUSTOMIZE_VERSION ?= 4.5.4	
-
+KUSTOMIZE_VERSION ?= 4.5.4
+KUSTOMIZE = $(shell pwd)/bin/kustomize
 .PHONY: kustomize
 kustomize:
-ifeq (, $(shell kustomize version | grep $(KUSTOMIZE_VERSION)))
+ifeq (, $(shell $(KUSTOMIZE) version | grep $(KUSTOMIZE_VERSION)))
 	@{ \
 	set -eo pipefail ;\
-	echo 'installing kustomize-v$(KUSTOMIZE_VERSION) into $(GOBIN)' ;\
-	curl -sS https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh | bash -s $(KUSTOMIZE_VERSION) $(GOBIN);\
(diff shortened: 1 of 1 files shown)


wonderflow:
Will the one in the folder override?

JarHMJ:
> Will the one in the folder override?

yes, the old will be removed and the new one will be downlod

makefiles/dependency.mk
```diff
@@ -56,20 +56,19 @@ else
...
 kustomize:
-ifeq (, $(shell kustomize version | grep $(KUSTOMIZE_VERSION)))
+ifeq (, $(shell $(KUSTOMIZE) version | grep $(KUSTOMIZE_VERSION)))
 	@{ \
 	set -eo pipefail ;\
-	echo 'installing kustomize-v$(KUSTOMIZE_VERSION) into $(GOBIN)' ;\
-	curl -sS https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh | bash -s $(KUSTOMIZE_VERSION) $(GOBIN);\
+    echo "installing kustomize-v$(KUSTOMIZE_VERSION) into $(shell pwd)/bin" ;\
+    mkdir -p $(shell pwd)/bin ;\
+    rm -f $(KUSTOMIZE) ;\
```
barnettZQG, line 68: Remove the old version after downloading successfully.

wonderflow:
@JarHMJ What if `bin/` of the project home directory is not added to the system `PATH`?

makefiles/dependency.mk
JarHMJ: I think it's okay to remove the old version  first, because the old version  is useless.

JarHMJ:
> @JarHMJ What if `bin/` of the project home directory is not added to the system `PATH`?

the KUSTOMIZE path is absolute，so the `bin` folder does not need to be added to the system `PATH`.

wonderflow:
I got your idea, that's good!

wonderflow (approved):

makefiles/dependency.mk
wonderflow: Yes, users can re-run this command for a new download.

PR #3963, opened 2022-05-24
Fix: add error message when the image doesn't exist

zzxwill:
Do not return error directly when the image doesn't exist

Fixes #

I have:

### DIFF
--- a/pkg/apiserver/domain/service/project.go
+++ b/pkg/apiserver/domain/service/project.go
@@ -657,12 +657,15 @@ func validateImage(ctx context.Context, k8sClient client.Client, project, imageN
 		}
 	}
 
+	var errMsg string
 	existed, err := image.IsExisted(username, password, imageName)
 	if err != nil {
-		return nil, err
+		errMsg = err.Error()
 	}
+
 	return &apisv1.ImageResponse{
 		Existed: existed,
 		Secret:  imagePullSecret,
+		Message: errMsg,
 	}, nil
 }
--- a/pkg/apiserver/interfaces/api/dto/v1/types.go
+++ b/pkg/apiserver/interfaces/api/dto/v1/types.go
@@ -214,6 +214,7 @@ type Config struct {
 type ImageResponse struct {
 	Existed bool   `json:"existed"`
 	Secret  string `json:"secret"`
+	Message string `json:"message"`
 }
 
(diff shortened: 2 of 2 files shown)


pkg/apiserver/domain/service/project.go
```diff
@@ -657,12 +657,15 @@ func validateImage(ctx context.Context, k8sClient client.Client, project, imageN
 		}
 	}
 
+	var errMsg string
 	existed, err := image.IsExisted(username, password, imageName)
 	if err != nil {
-		return nil, err
+		errMsg = err.Error()
```
barnettZQG, line 663: Is this message clear for users?

zzxwill: So are these what you are expecting?
- The image tag `latest1` is not valid
- The registry of the image `registry2.cn-beijing.aliyuncs.com` is not valid

barnettZQG: > Yes, the error of the registry library is good. For example:
> 
> * image tag is not valid
> 
> > GET https://registry.cn-beijing.aliyuncs.com/v2/vela/nginx/manifests/latest1: MANIFEST_UNKNOWN: manifest unknown; map[Tag:latest1]
> 
> * registry is not valid
> 
> > "Get "https://registry2.cn-beijing.aliyuncs.com/v2/": dial tcp: lookup registry2.cn-beijing.aliyuncs.com on 100.100.2.138:53: no such host"

These messages are too technicalization, follows are good to me:

* The image name or tag maybe does not exist.     (manifest not found)
* The image registry is private, you must set the correct secret first.  (403 error)
* The image registry can not be accessed.  (others network error)

BTW, you also can return the business code for every kind of error, the front end handles user interaction information.

zzxwill: Think about what we are doing at the moment. We are not providing an API to validate a docker image for the universe, instead, we are providing the API for KubeVela end-users who is about to create their OWN BUSINESS application. Are they just getting lost for the image they should run the business, and are we holding up all road signs to help them get home - find the right docker image?

zzxwill: And here is an agreement in the DingTalk group it won't provide any bcode for errors.

FogDong (approved):

yangsoon (approved):

Issue #4017, opened 2022-05-27
repeated application of the same resource results in delete deployment


clarechu:
**Describe the bug**
When I updated the application resource of vela, I found that the deployment resource application was deleted by vela-core.

**To Reproduce**

```yaml
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: xxxx
  namespace: xxxx
spec:
  components:
    - name: xxxxxx
      type: webservice
      dependsOn: []
      properties:
        exposeType: NodePort
        image: xxxxxxx
        imagePullSecrets:
          - xxx-secret
        ports:
          - expose: true
            port: 80
      traits:
        - properties:
            domain: xxxxxxx
            class: nginx
            classInSpec: true
            http:
              /: 80
          type: gateway
```

1. apply application
```bash
$ kubectl apply -f demo.yaml

application.core.oam.dev/xxxx created
```

2. get pod

```bash
➜  ~ k get po -w
NAME                              READY   STATUS    RESTARTS   AGE
xxx-6cb857dc5f-nhnmt   0/1     Pending   0          0s
xxx-6cb857dc5f-nhnmt   0/1     Pending   0          0s
xxx-6cb857dc5f-nhnmt   0/1     ContainerCreating   0          0s
xxx-6cb857dc5f-nhnmt   0/1     ContainerCreating   0          3s
xxx-6cb857dc5f-nhnmt   1/1     Running             0          4s
```

3. Apply the same again yaml

```bash
kubectl apply -f demo.yaml

application.core.oam.dev/xxxx configured
```

4. get po

```bash

➜  ~ k get po -w
NAME                              READY   STATUS    RESTARTS   AGE
xxxx-6cb857dc5f-nhnmt   1/1     Terminating         0          8m35s
xxxx-6cb857dc5f-nhnmt   1/1     Terminating         0          8m36s
xxxx-6cb857dc5f-nhnmt   0/1     Terminating         0          8m37s

```

**Expected behavior**

**Screenshots**

**KubeVela Version**
vela-core : v1.3.2

**Cluster information**

Kubernetes Version [1.20.8]

chivalryq:
This is quite strange. Is this app still reconciling? And the new Deployment haven't been created?

clarechu:
> This is quite strange. Is this app still reconciling? And the new Deployment haven't been created?

yes， Very strange

chivalryq:
[...]

PR #4053, opened 2022-05-30
dependsOn cannot be an empty array

clarechu:
Signed-off-by: clarechu [106218615@qq.com](mailto:106218615@qq.com)

When I updated the application resource of vela, I found that the deployment resource application was deleted by vela-core.

Fixes #

I have:

```yaml
kubectl apply -f -<<EOF
kind: Application
metadata:
  name: app1
spec:
  components:
    - name: comp-1
      type: webservice
      dependsOn: []
      properties:
        image: nginx
        ports:
          - expose: true
            port: 80
EOF
```

Response result is

```bash
Error from server: error when creating "STDIN": admission webhook "validating.core.oam.dev.v1beta1.applications" denied the request: field "components[0].dependsOn": Invalid value error encountered, empty array not allowed.
```

### DIFF
--- a/pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler_test.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/application/validating_handler_test.go
@@ -372,5 +372,25 @@ var _ = Describe("Test Application Validator", func() {
 		}
 		resp = handler.Handle(ctx, req)
 		Expect(resp.Allowed).Should(BeFalse())
+
+		By("dependsOn cannot be an empty array")
+		req = admission.Request{
+			AdmissionRequest: admissionv1.AdmissionRequest{
+				Operation: admissionv1.Create,
+				Resource:  metav1.GroupVersionResource{Group: "core.oam.dev", Version: "v1beta1", Resource: "applications"},
+				Object: runtime.RawExtension{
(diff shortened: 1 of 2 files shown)


wonderflow:
Why not omit the dependsOn field in this case?

clarechu:
Will cause this problem
https://github.com/kubevela/kubevela/issues/4017

wonderflow:
> Will cause this problem #4017


I'm not quite understand the relationships. I mean why not using app like this in your scenario:

```
kubectl apply -f -<<EOF
kind: Application
metadata:
  name: app1
spec:
  components:
    - name: comp-1
      type: webservice
      properties:
        image: nginx
        ports:
          - expose: true
            port: 80
EOF
```

barnettZQG:
@clarechu I think we should allow the dependsOn is an empty array.

clarechu:
indeed, What is the meaning if dependson is empty? 
Just don't do it?

wonderflow:
To be more clear:

1. I'm just curious why we need to add an empty slice instead of omit the field?
2. If we will add the empty sclice of `dependsOn`, we should support this case instead of prevent it.

clarechu:
ok ! thanks @wonderflow

PR #4108, opened 2022-06-03
Feat: add message for `vela addon enable pyroscope`

fourierr:

Add message for `vela addon enable pyroscope`

I have:

### DIFF
--- a/references/cli/addon.go
+++ b/references/cli/addon.go
@@ -209,6 +209,14 @@ func AdditionalEndpointPrinter(ctx context.Context, c common.Args, k8sClient cli
 		fmt.Println(`Select "Cluster: local | Namespace: vela-system | Kind: Service | Name: velaux" from the prompt.`)
 		fmt.Println(`Please refer to https://kubevela.io/docs/reference/addons/velaux for more VelaUX addon installation and visiting method.`)
 	}
+	if name == "pyroscope" {
+		fmt.Println(`Get the application URL by running these commands:`)
+		fmt.Println(`    export POD_NAME=$(kubectl get pods --namespace pyroscope-system -l "app.kubernetes.io/name=pyroscope,app.kubernetes.io/instance=pyroscope" -o jsonpath="{.items[0].metadata.name}")`)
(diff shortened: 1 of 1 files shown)


wonderflow (changes requested):
In this case, we will harccode for every add-on, that's a disaster

fourierr:
> In this case, we will harccode for every add-on, that's a disaster

- I proposal that we need to add a field `message` in add-on's metadata.yaml, e.g
```yaml
name: pyroscope
version: 0.17.1
description: Pyroscope is an open source platform, consisting of server and agent. It allows the user to collect, store, and query the profiling data in a CPU and disk efficient way.
url: https://github.com/pyroscope-io/pyroscope
icon: https://raw.githubusercontent.com/pyroscope-io/pyroscope/main/webapp/images/favicon.ico

tags:
  - profiling tool

deployTo:
  control_plane: true

dependencies:
  - name: fluxcd

needNamespace:
  - pyroscope-system

invisible: false

meaaasge:
 - first-msg
 - second-msg
```
- Print the message when using `vela addon enbale xxx`.
@wonderflow

wonderflow:
> I proposal that we need to add a field message in add-on's metadata.yaml.

Yes, we need something like this, can you propose a more detailed design? I think the message is not only fixed string, we will need some dynamically runtime variables here. \cc @wangyikewxgm

fourierr:
What are the runtime variables? Can you give an example.

wonderflow:
> What are the runtime variables? Can you give an example.

We should align with the helm chart Note.txt

barnettZQG:
Let's add a 'note' field in the metadata file, and print it after the addon is installed.

fourierr:
I will give a new feature request for detailed https://github.com/kubevela/kubevela/issues/4114

Somefive:
Should be able to add customized message with NOTES.cue as mentioned https://kubevela.io/docs/platform-engineers/addon/addon-cue#notes-info-notescue .

Feel free to reopen it if it does not solve the issue.

PR #3644, opened 2022-04-13
Feat(delay suspend): delayDuration in suspend step properties

ArenaSu:
### Description of your changes

Add **waitDuration** properties in suspend workflow step, with **waitDuration** , app can wait for some time(such as 30s), and then continue to run workflow step.It works ok in  either  StepByStep or DAG mode.

```
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: delay-suspend-test
spec:
  components:
    - name: delay-suspend-comp
      type: webservice
      properties:
        image: nginx
        port: 80
  policies:
    - name: local-deploy
      type: topology
      properties:
        clusters: [ "local" ]
        namespace: default
  workflow:
    steps:
      - name: suspend-test
        type: suspend
        properties:
          waitDuration: 30s
      - name: deploy-to-local
        type: deploy
        properties:
          policies: [ "local-deploy" ]
```

Fixes #3504

I have:

- [x] Read and followed KubeVela's [contribution process](https://github.com/oam-dev/kubevela/blob/master/contribute/create-pull-request.md).
- [ ] [Related Docs](https://github.com/oam-dev/kubevela.io) updated properly. In a new feature or configuration option, an update to the documentation is necessary. 
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

ginkgo -v  --focus="Test wait suspend" ./test/e2e-test

PR #4121, opened 2022-06-06
Feat: add timeout in workflow step

### LINKED ISSUE #4020: [Feature] Support timeout in workflow step

If the workflow step is `wait healthy`, in some cases we don't want it wait forever, we can add `timeout` in that step like:

```yaml
steps:
   - name: a
      timeout: 2m
```

**Additional context**

FogDong:

Fixes #4020

This PR adds timeout in workflow step, you can use it like:
```yaml
  workflow:
    steps:
    - name: suspend
      type: suspend
      timeout: 15s
```

After 15s, if the workflow is still not resumed, then the status will be like:
```yaml
steps:
    - name: suspend
       type: suspend
       phase: failed
       reason: Timeout
```

**Note**: In the past version, the `suspend` step succeeded when executed. After this PR is merged, the `suspend` step' status is `running` when it was executed, and become `succeeded` if it's resumed.

I have:

/cc @leejanee @wonderflow 

This PR also made some optimization to #3644 /cc @ArenaSu

### DIFF
--- a/apis/core.oam.dev/common/types.go
+++ b/apis/core.oam.dev/common/types.go
@@ -349,6 +349,8 @@ type WorkflowStep struct {
 
 	If string `json:"if,omitempty"`
 
+	Timeout string `json:"timeout,omitempty"`
+
 	DependsOn []string `json:"dependsOn,omitempty"`
 
 	Inputs StepInputs `json:"inputs,omitempty"`
@@ -367,6 +369,8 @@ type WorkflowSubStep struct {
 	Properties *runtime.RawExtension `json:"properties,omitempty"`
 
 	If string `json:"if,omitempty"`
+
+	Timeout string `json:"timeout,omitempty"`
 
 	DependsOn []string `json:"dependsOn,omitempty"`
 
--- a/pkg/apiserver/domain/service/workflow.go
+++ b/pkg/apiserver/domain/service/workflow.go
@@ -44,6 +44,7 @@ import (
 	utils2 "github.com/oam-dev/kubevela/pkg/utils"
 	"github.com/oam-dev/kubevela/pkg/utils/apply"
 	"github.com/oam-dev/kubevela/pkg/workflow/tasks/custom"
+	wfTypes "github.com/oam-dev/kubevela/pkg/workflow/types"
 )
 
 // WorkflowService workflow manage api
@@ -561,8 +562,7 @@ func (w *workflowServiceImpl) ResumeRecord(ctx context.Context, appModel *model.
 		return err
 	}
 
-	oamApp.Status.Workflow.Suspend = false
-	if err := w.KubeClient.Status().Patch(ctx, oamApp, client.Merge); err != nil {
+	if err := ResumeWorkflow(ctx, w.KubeClient, oamApp); err != nil {
 		return err
 	}
 	if err := w.syncWorkflowStatus(ctx, oamApp, recordName, oamApp.Name); err != nil {
@@ -588,6 +588,26 @@ func (w *workflowServiceImpl) TerminateRecord(ctx context.Context, appModel *mod
 	return nil
 }
 
+// ResumeWorkflow resume workflow
+func ResumeWorkflow(ctx context.Context, kubecli client.Client, app *v1beta1.Application) error {
+	app.Status.Workflow.Suspend = false
+	steps := app.Status.Workflow.Steps
+	for i, step := range steps {
+		if step.Type == wfTypes.WorkflowStepTypeSuspend && step.Phase == common.WorkflowStepPhaseRunning {
+			steps[i].Phase = common.WorkflowStepPhaseSucceeded
+		}
+		for j, sub := range step.SubStepsStatus {
+			if sub.Type == wfTypes.WorkflowStepTypeSuspend && sub.Phase == common.WorkflowStepPhaseRunning {
+				steps[i].SubStepsStatus[j].Phase = common.WorkflowStepPhaseSucceeded
+			}
+		}
+	}
+	if err := kubecli.Status().Patch(ctx, app, client.Merge); err != nil {
+		return err
+	}
+	return nil
+}
+
 // TerminateWorkflow terminate workflow
 func TerminateWorkflow(ctx context.Context, kubecli client.Client, app *v1beta1.Application) error {
 	// set the workflow terminated to true
@@ -596,7 +616,7 @@ func TerminateWorkflow(ctx context.Context, kubecli client.Client, app *v1beta1.
 	for i, step := range steps {
 		switch step.Phase {
 		case common.WorkflowStepPhaseFailed:
-			if step.Reason != custom.StatusReasonFailedAfterRetries {
+			if step.Reason != custom.StatusReasonFailedAfterRetries && step.Reason != custom.StatusReasonTimeout {
 				steps[i].Reason = custom.StatusReasonTerminate
 			}
 		case common.WorkflowStepPhaseRunning:
@@ -607,7 +627,7 @@ func TerminateWorkflow(ctx context.Context, kubecli client.Client, app *v1beta1.
 		for j, sub := range step.SubStepsStatus {
 			switch sub.Phase {
 			case common.WorkflowStepPhaseFailed:
-				if sub.Reason != custom.StatusReasonFailedAfterRetries {
+				if sub.Reason != custom.StatusReasonFailedAfterRetries && sub.Reason != custom.StatusReasonTimeout {
 					steps[i].SubStepsStatus[j].Phase = custom.StatusReasonTerminate
 				}
 			case common.WorkflowStepPhaseRunning:
--- a/pkg/apiserver/domain/service/workflow.go
+++ b/pkg/apiserver/domain/service/workflow.go
@@ -612,6 +612,8 @@ func ResumeWorkflow(ctx context.Context, kubecli client.Client, app *v1beta1.App
 func TerminateWorkflow(ctx context.Context, kubecli client.Client, app *v1beta1.Application) error {
 	// set the workflow terminated to true
 	app.Status.Workflow.Terminated = true
+	// set the workflow suspend to false
+	app.Status.Workflow.Suspend = false
 	steps := app.Status.Workflow.Steps
 	for i, step := range steps {
 		switch step.Phase {
--- a/pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
@@ -19,6 +19,7 @@ package application
 import (
 	"context"
 	"fmt"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -38,16 +39,32 @@ func (h *ValidatingHandler) ValidateWorkflow(ctx context.Context, app *v1beta1.A
 		stepName := make(map[string]interface{})
 		for _, step := range app.Spec.Workflow.Steps {
 			if _, ok := stepName[step.Name]; ok {
-				errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps"), step.Name, "duplicate step name"))
+				errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps"), step.Name, "duplicated step name"))
 			}
 			stepName[step.Name] = nil
+			if step.Timeout != "" {
+				errs = append(errs, h.ValidateTimeout(step.Name, step.Timeout)...)
+			}
 			for _, sub := range step.SubSteps {
 				if _, ok := stepName[sub.Name]; ok {
-					errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps", "subSteps"), sub.Name, "duplicate step name"))
+					errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps", "subSteps"), sub.Name, "duplicated step name"))
 				}
 				stepName[sub.Name] = nil
+				if step.Timeout != "" {
+					errs = append(errs, h.ValidateTimeout(step.Name, step.Timeout)...)
+				}
 			}
 		}
+	}
+	return errs
+}
+
+// ValidateTimeout validates the timeout of steps
+func (h *ValidatingHandler) ValidateTimeout(name, timeout string) field.ErrorList {
+	var errs field.ErrorList
+	_, err := time.ParseDuration(timeout)
+	if err != nil {
+		errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps", "timeout"), name, "invalid timeout, please use the format of timeout like 1s, 1m, 1h or 1d"))
 	}
 	return errs
 }
--- a/pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
+++ b/pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
@@ -31,8 +31,29 @@ import (
 	"github.com/oam-dev/kubevela/pkg/oam/util"
 )
 
-// ValidateCreate validates the Application on creation
-func (h *ValidatingHandler) ValidateCreate(ctx context.Context, app *v1beta1.Application) field.ErrorList {
+// ValidateWorkflow validates the Application workflow
+func (h *ValidatingHandler) ValidateWorkflow(ctx context.Context, app *v1beta1.Application) field.ErrorList {
+	var errs field.ErrorList
+	if app.Spec.Workflow != nil {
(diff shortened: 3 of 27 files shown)


pkg/workflow/workflow.go
```diff
@@ -609,13 +639,74 @@ func (e *engine) steps(taskRunners []wfTypes.TaskRunner, dag bool) error {
...
+				default:
+					// TODO:(fog) support more if cases
+					return &wfTypes.PreCheckResult{Skip: false}, nil
+				}
+			},
+			func(step oamcore.WorkflowStep) (*wfTypes.PreCheckResult, error) {
+				status := e.stepStatus[step.Name]
+				if e.parentRunner != "" {
+					if status, ok := e.stepStatus[e.parentRunner]; ok && status.Phase == common.WorkflowStepPhaseFailed && status.Reason == custom.StatusReasonTimeout {
+						return &wfTypes.PreCheckResult{Timeout: true}, nil
```
leejanee, line 680: 是否意味着父步骤timeout，那么所有子步骤都状态变为超时了?

leejanee (commented):
现在是否要求整个流水线中 位于不同stepGroup中的子步骤名字不能重复？

pkg/workflow/workflow.go
FogDong: > 现在是否要求整个流水线中 位于不同stepGroup中的子步骤名字不能重复？
是的，子步骤名字不能重复，我加个 webhook 校验下。

> 是否意味着父步骤timeout，那么所有子步骤都状态变为超时了?
是的，子步骤中还处于 running 的步骤会变成超时

FogDong:
> Why does timeout not be configured in properties for suspend step ? Will it be common property for all steps ?

Yes, every step can have a `timeout` filed.

leejanee (approved):
LGTM

pkg/apiserver/domain/service/workflow.go
```diff
@@ -588,15 +588,37 @@ func (w *workflowServiceImpl) TerminateRecord(ctx context.Context, appModel *mod
...
 	// set the workflow terminated to true
 	app.Status.Workflow.Terminated = true
+	// set the workflow suspend to false
+	app.Status.Workflow.Suspend = false
 	steps := app.Status.Workflow.Steps
 	for i, step := range steps {
 		switch step.Phase {
 		case common.WorkflowStepPhaseFailed:
-			if step.Reason != custom.StatusReasonFailedAfterRetries {
+			if step.Reason != custom.StatusReasonFailedAfterRetries && step.Reason != custom.StatusReasonTimeout {
```
wonderflow, line 621: why we need to add this check？

pkg/workflow/workflow.go
```diff
@@ -270,79 +290,88 @@ func (w *workflow) Trace() error {
 	return recorder.With(w.cli, w.app).Save("", data).Limit(10).Error()
 }
 
-func (w *workflow) GetBackoffWaitTime() time.Duration {
-	nextTime, ok := w.wfCtx.GetValueInMemory(wfTypes.ContextKeyNextExecuteTime)
-	if !ok {
-		return time.Second
+func (w *workflow) GetSuspendBackoffWaitTime() time.Duration {
+	if w.app.Status.Workflow == nil {
+		return 0
```
wonderflow, line 295: what does this case mean?

pkg/workflow/workflow.go
```diff
@@ -255,6 +265,16 @@ func newEngine(ctx monitorContext.Context, wfCtx wfContext.Context, w *workflow,
 		rk:            w.rk,
 		stepStatus:    stepStatus,
 		stepDependsOn: stepDependsOn,
+		stepTimeout:   make(map[string]time.Time),
+	}
+}
+
+func setStepStatus(statusMap map[string]common.StepStatus, status common.StepStatus) {
```
wonderflow, line 272:
```suggestion
func copyStepStatus(statusMap map[string]common.StepStatus, status common.StepStatus) {
```

wonderflow: set means you're changing something, copy is better here that you're just copy and override

pkg/workflow/workflow.go
```diff
@@ -270,79 +290,88 @@ func (w *workflow) Trace() error {
...
+		for _, sss := range ss.SubStepsStatus {
+			setStepStatus(stepStatus, sss.StepStatus)
+		}
 	}
-	next := time.Unix(unix, 0)
-	if next.After(time.Now()) {
-		return time.Until(next)
+	max := time.Duration(1<<63 - 1)
+	min := max
+	if w.app.Spec.Workflow != nil {
```
wonderflow, line 306: we should check and return the nil case first to avoid complex.

```suggestion
	if w.app.Spec.Workflow == nil {
	    return 0
	}
```

pkg/apiserver/domain/service/workflow.go
FogDong: When you try to terminate a workflow, if the step is already `failedAfterRetries` or `timeout`, then its reason shouldn't be `terminate` since its already finished.

pkg/workflow/workflow.go
```diff
@@ -270,79 +290,88 @@ func (w *workflow) Trace() error {
...
-			return
-		}
-
-		if d {
-			doWaiting = d
-			remainingDuration := wd - time.Since(stepStatus.FirstExecuteTime.Time)
-			if remainingDuration <= 0 {
-				w.app.Status.Workflow.Steps[i].Phase = common.WorkflowStepPhaseSucceeded
+func handleSuspendBackoffTime(step oamcore.WorkflowStep, status common.StepStatus, min time.Duration) time.Duration {
+	if status.Phase == common.WorkflowStepPhaseRunning {
```
wonderflow, line 330: please return the result first to avoid code block complex:

```
if status.Phase != common.WorkflowStepPhaseRunning {
   return min
}
...
// the rest code can avoid indentation
```

pkg/workflow/workflow.go
```diff
@@ -270,79 +290,88 @@ func (w *workflow) Trace() error {
...
-
-	return time.Second
+	if min == max {
+		return 0
+	}
+	return min
 }
 
-func (w *workflow) HandleSuspendWait(ctx monitorContext.Context) (doWaiting bool, durationWaiting time.Duration, errRet error) {
-	ctx.Info("handle suspend wait")
```
wonderflow, line 291: shall we keep this kind of info？I don't know if it's helpful, we should clean them if they're useless.

pkg/workflow/tasks/discover.go
```diff
@@ -290,25 +288,25 @@ func NewViewTaskDiscover(pd *packages.PackageDiscover, cli client.Client, cfg *r
 }
 
 // GetSuspendStepDurationWaiting get suspend step wait duration
-func GetSuspendStepDurationWaiting(step v1beta1.WorkflowStep) (bool, builtintime.Duration, error) {
+func GetSuspendStepDurationWaiting(step v1beta1.WorkflowStep) (time.Duration, error) {
```
wonderflow, line 291: why we omit the bool return value?

pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
```diff
@@ -31,8 +31,29 @@ import (
 	"github.com/oam-dev/kubevela/pkg/oam/util"
 )
 
-// ValidateCreate validates the Application on creation
-func (h *ValidatingHandler) ValidateCreate(ctx context.Context, app *v1beta1.Application) field.ErrorList {
+// ValidateWorkflow validates the Application workflow
+func (h *ValidatingHandler) ValidateWorkflow(ctx context.Context, app *v1beta1.Application) field.ErrorList {
+	var errs field.ErrorList
+	if app.Spec.Workflow != nil {
```
wonderflow, line 37: return the nil case first

pkg/workflow/tasks/discover.go
FogDong: The bool value is redundant, `time.Duration(0)` can replace it.

pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
```diff
@@ -31,8 +31,29 @@ import (
...
+			stepName[step.Name] = nil
+			for _, sub := range step.SubSteps {
+				if _, ok := stepName[sub.Name]; ok {
+					errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps", "subSteps"), sub.Name, "duplicate step name"))
+				}
+				stepName[sub.Name] = nil
+			}
+		}
+	}
+	return errs
```
wonderflow, line 52: Shall we parse the timeout format here? BTW, parse a string is not efficient, I think the duration can be a number with the unit `s` to avoid this check.

apis/core.oam.dev/common/types.go
```diff
@@ -349,6 +349,8 @@ type WorkflowStep struct {
 
 	If string `json:"if,omitempty"`
 
+	Timeout string `json:"timeout,omitempty"`
```
wonderflow, line 352: As I commented below, it'll be better if it's a int here.

pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
```diff
@@ -31,8 +31,29 @@ import (
...
-// ValidateCreate validates the Application on creation
-func (h *ValidatingHandler) ValidateCreate(ctx context.Context, app *v1beta1.Application) field.ErrorList {
+// ValidateWorkflow validates the Application workflow
+func (h *ValidatingHandler) ValidateWorkflow(ctx context.Context, app *v1beta1.Application) field.ErrorList {
+	var errs field.ErrorList
+	if app.Spec.Workflow != nil {
+		stepName := make(map[string]interface{})
+		for _, step := range app.Spec.Workflow.Steps {
+			if _, ok := stepName[step.Name]; ok {
+				errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps"), step.Name, "duplicate step name"))
```
wonderflow, line 41:
```suggestion
				errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps"), step.Name, "duplicated step name"))
```

pkg/webhook/core.oam.dev/v1alpha2/application/validation.go
```diff
@@ -31,8 +31,29 @@ import (
...
+			}
+			stepName[step.Name] = nil
+			for _, sub := range step.SubSteps {
+				if _, ok := stepName[sub.Name]; ok {
+					errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps", "subSteps"), sub.Name, "duplicate step name"))
```
wonderflow, line 46:
```suggestion
					errs = append(errs, field.Invalid(field.NewPath("spec", "workflow", "steps", "subSteps"), sub.Name, "duplicated step name"))
```

pkg/workflow/tasks/discover.go
```diff
@@ -193,34 +195,29 @@ func (tr *stepGroupTaskRunner) Pending(ctx wfContext.Context, stepStatus map[str
...
+		result, err := hook(tr.step)
+		if err != nil {
+			return common.StepStatus{}, nil, errors.WithMessage(err, "do preCheckHook")
+		}
+		if result.Skip {
+			status.Phase = common.WorkflowStepPhaseSkipped
+			status.Reason = custom.StatusReasonSkip
+			options.StepStatus[tr.step.Name] = status
+			break
+		} else if result.Timeout {
```
wonderflow, line 215:
```suggestion
		}
		if result.Timeout {
```

We don't need to use else as the former case will always break

pkg/workflow/workflow.go
```diff
@@ -270,79 +290,88 @@ func (w *workflow) Trace() error {
...
+	}
+	return min
 }
 
-func (w *workflow) getWorkflowStepByName(name string) oamcore.WorkflowStep {
-	for _, s := range w.app.Spec.Workflow.Steps {
-		if s.Name == name {
-			return s
+func (w *workflow) GetBackoffWaitTime() time.Duration {
+	nextTime, ok := w.wfCtx.GetValueInMemory(wfTypes.ContextKeyNextExecuteTime)
```
wonderflow, line 358: what will happend if the controller crash and restared?

pkg/workflow/workflow.go
```diff
@@ -458,8 +487,31 @@ func (e *engine) getMaxBackoffWaitTime() int {
...
+					min = duration
+				}
+			}
+		}
+	}
+	if min == max {
+		return -1
+	}
+	if min.Seconds() < 1 {
+		return minWorkflowBackoffWaitTime
```
wonderflow, line 508: what's the value of this `minWorkflowBackoffWaitTime`

pkg/workflow/workflow.go
```diff
@@ -458,8 +487,31 @@ func (e *engine) getMaxBackoffWaitTime() int {
...
+	}
+	return int64(math.Ceil(min.Seconds()))
```
wonderflow, line 510: why need math.Ceil here?

apis/core.oam.dev/common/types.go
FogDong: If the value is seconds, it is best to specify this in the key, such as `timeoutSeconds`.

Some references:
[TimeoutSeconds in Kubernetes probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes)
[Argo Timeout](https://argoproj.github.io/argo-workflows/template-defaults/#configuring-templatedefaults-in-workflowspec)
[Argo activeDeadlineSeconds](https://argoproj.github.io/argo-workflows/walk-through/timeouts/)
[Tekton Timeout](https://tekton.dev/vault/pipelines-v0.18.1/tasks/#specifying-a-timeout)
[timeout-minutes in GitHub Action](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)

pkg/workflow/workflow.go
FogDong: 1

pkg/workflow/workflow.go
FogDong: if the `min.Seconds == 1.2`, the next requeue time should be 2 seconds later.

pkg/workflow/workflow.go
FogDong: We'll lose the `NextExecuteTime` in memory so the backoff time is re-calculated from the initial value `1`.

pkg/workflow/workflow.go
FogDong: cc @leejanee

pkg/workflow/workflow.go
FogDong: I actually set the status to the map but not copy so I think it's `set` here.

pkg/workflow/workflow.go
FogDong: In this way, the rest code can avoid indentation indeed, but we have to write `return min` twice, so I think there's no necessity to avoid indentation.

wonderflow: if so, we should make every function which has an error return like this:

```
if err==nil{
   ...
   ...
}

return err
```

This can avoid multiple `return err` block, but nobody actually did like this.

wonderflow: refer to the go code convention which is also kubevela's code convention https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

FogDong: Just in this case, there are not many differences between:
```go
if status.Phase != common.WorkflowStepPhaseRunning {
   return min
}
.... // some codes
return min
```
and 
```go
if status.Phase == common.WorkflowStepPhaseRunning {
.... // some codes
}
return min
```

And I think the last one is better. But in some cases, there's better to return the result first.

wonderflow (approved):
great job

test/e2e-multicluster-test/multicluster_standalone_test.go
```diff
@@ -130,6 +130,11 @@ var _ = Describe("Test multicluster standalone scenario", func() {
 			_app := &v1beta1.Application{}
 			g.Expect(k8sClient.Get(hubCtx, appKey, _app)).Should(Succeed())
 			_app.Status.Workflow.Suspend = false
+			for i, step := range _app.Status.Workflow.Steps {
+				if step.Type == "suspend" {
+					_app.Status.Workflow.Steps[i].Phase = oamcomm.WorkflowStepPhaseSucceeded
+				}
+			}
```
Somefive, line 137: It might be better if we have some function like `ResumeWorkflow(&app.Status.Workflow)`?

Somefive (approved):
Generally LGTM

yangsoon (approved):

PR #4141, opened 2022-06-09
Chore: fix potential bot failures

charlie0129:

KubeVela bot sometimes will fail and the log shows that it is caused by an inconsistent lockfile.

However the lockfile itself is fine, and I suspect it is caused by node version changes (as it uses the version bundled with `ubuntu-latest`, which just updated recently with npm `v8.5.5` -> `v8.11.0` and causing this issue). 

Using `latest` tag does no good, which will often cause problems when software version changes. 

It is a good practice to fix node versions (also npm versions).

So I did the following:
- fixed node version to `v14` (instead of node `v16` + npm `v8.11.0` bundled with `ubuntu-latest`; that specific `v8.11.0` version doesn't work and will fail)
- fixed ubuntu version to 20.04, just for good measure
- (btw) added cache to speed up workflow runs

Screen Shot 2022-06-09 at 6 16 02 PM

Fixes #

I have:

Since I used my repo to double check it can successfully install dependencies using GitHub Actions now, but I don't have the necessary token to run the entire job, I will continue to check how it actually works in real environments after being merged.

### DIFF
--- a/.github/workflows/issue-commands.yml
+++ b/.github/workflows/issue-commands.yml
@@ -15,7 +15,17 @@ jobs:
           repository: "oam-dev/kubevela-github-actions"
           path: ./actions
           ref: v0.4.2
-      - name: Install Actions
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: Cache Dependencies
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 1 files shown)


StevenLeiZhang (approved):
LGTM

wonderflow (approved):
great work！

PR #4167, opened 2022-06-14
Fix: fix trait customStatus error when controlPlanOnly=true

fourierr:

Fix trait customStatus error when controlPlanOnly=true

I have:

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/apply.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/apply.go
@@ -264,6 +264,10 @@ func (h *AppHandler) collectHealthStatus(ctx context.Context, wl *appfile.Worklo
 
 	var traitStatusList []common.ApplicationTraitStatus
 	for _, tr := range wl.Traits {
+		if tr.FullTemplate.TraitDefinition.Spec.ControlPlaneOnly {
+			namespace = appRev.GetNamespace()
+			wl.Ctx.SetCtx(context.WithValue(wl.Ctx.GetCtx(), multicluster.ClusterContextKey, multicluster.ClusterLocalName))
+		}
 		var traitStatus = common.ApplicationTraitStatus{
 			Type:    tr.Name,
 			Healthy: true,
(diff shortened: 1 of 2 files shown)


pkg/cue/process/handle.go
```diff
@@ -42,6 +42,7 @@ type Context interface {
 	PushData(key string, data interface{})
 	GetCtx() context.Context
 	SetCtx(context.Context)
+	SetCtxWithValue(key, value interface{})
```
wonderflow, line 45: you should not add a interface for so little convenience. 

It can be done by just one line:
```
ctx.SetCtx(context.WithValue(ctx.GetCtx(), key, value))
```

fourierr: > 
`ctx.SetCtx()` can't set a new context, If we want to set a new context, we may need a `setCtxForce` method. @wonderflow 
https://github.com/kubevela/kubevela/blob/0c190ef32948561b577ccb0ca8432aee74bcfc80/pkg/cue/process/handle.go#L330-L335

wonderflow: Let's check why it's can't be set a new one. \cc @leejanee

wonderflow: You can change that function to set a new context even there exists one

fourierr: ok, I will change it

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -264,6 +264,10 @@ func (h *AppHandler) collectHealthStatus(ctx context.Context, wl *appfile.Worklo
 
 	var traitStatusList []common.ApplicationTraitStatus
 	for _, tr := range wl.Traits {
+		if tr.FullTemplate.TraitDefinition.Spec.ControlPlaneOnly {
+			namespace = appRev.GetNamespace()
```
wonderflow, line 268: where is the namespace used?

fourierr: > where is the namespace used?

Whe `ControlPlaneOnly=true`, the trait CR will be dispatched to the `local` cluster and the namespace that is same as application

wonderflow: got

wonderflow (approved):

FogDong (approved):

PR #4193, opened 2022-06-16
Fix: clear namespace for cluster scoped resource for dispatching

Somefive:

ResourceTracker will record resources with metadata. The namespace for cluster scoped resource is incorrectly recorded. Although it does not affect the normal function, it will makes the resource view a bit weird.

Before:
image

After:
image

I have:

### DIFF
--- a/pkg/resourcekeeper/utils.go
+++ b/pkg/resourcekeeper/utils.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2022 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resourcekeeper
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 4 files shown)


pkg/resourcekeeper/utils.go
```diff
@@ -0,0 +1,35 @@
...
+
+// ClearNamespaceForClusterScopedResources clear namespace for cluster scoped resources
+func (h *resourceKeeper) ClearNamespaceForClusterScopedResources(manifests []*unstructured.Unstructured) {
+	for _, manifest := range manifests {
+		mappings, err := h.Client.RESTMapper().RESTMappings(manifest.GroupVersionKind().GroupKind(), manifest.GroupVersionKind().Version)
+		if err != nil {
+			continue
+		}
+		if len(mappings) > 0 && mappings[0].Scope.Name() == meta.RESTScopeNameRoot {
+			manifest.SetNamespace("")
```
wonderflow, line 32: the empty is usually means default, it will cause more confuse if we use so.

What about just omit the value in display? use `"***"` instead.

Somefive: Actually, it is not only a simple display problem. For example, if one application try to dispatch two ClusterRoleBinding in two namespaces, however, since ClusterRoleBinding is cluster-scoped, these two CRB are the same. If we do not clear the namespace here, the ResourceTracker will record it twice. If will also cause confusion for users while looking at the ResourceTracker.

So generally, I think it will be better to leave the underlying data clean and correct, instead of just letting the display correct.

P.S. Empty does not always mean "default" for dispatching resources. For cluster-scoped resources, the namespace field should not carry any value.

Somefive: If we use client to get a cluster-scoped resource in Kubernetes, the namespace field in that retrieved object is empty.

wonderflow (approved):

PR #4207, opened 2022-06-20
Fix: cue patch remove temp var

### LINKED ISSUE #4203: traits env/annotations conflict, invalid patch trait env into workload: field "env" not allowed in closed struct
**Describe the bug**

```
About:

  Name:      	failed-env                   
  Namespace: 	default                      
  Created at:	2022-06-20 13:13:01 +0800 CST
  Status:    	runningWorkflow              

Workflow:

  mode: StepByStep
  finished: false
  Suspend: false
  Terminated: false
  Steps
  - id:c2xqtrhdte
    name:deploy-default
    type:deploy
    phase:failed 
    message:step deploy: step deploy: run step(provider=oam,do=components-apply): Found 1 errors. [(failed to apply component local-default-0: GenerateComponentManifest: evaluate template trait=env app=failed-env: invalid patch trait env into workload: result check err: spec.template.spec.containers.0._params: field "env" not allowed in closed struct)]

Services:


```

**To Reproduce**

**Expected behavior**

**Screenshots**

```yaml
kind: Application
metadata:
  name: failed-env
spec:
  components:
  - name: failed-env
    type: webservice
    properties:
      ports:
      - name: http
        port: 8080
        expose: true
      image: busybox:1.34
      imagePullPolicy: IfNotPresent
      env:
      - name: TZ
        value: Asia/Shanghai
    traits:
    - type: annotations #################### conflict trait env
      properties:
        app.kubernetes.io/tier: bff
    - type: resource
      properties:
        requests:
          cpu: 0.1
          memory: 256Mi
        limits:
          cpu: 1
          memory: 960Mi
  policies:
  - name: topology-default
    type: topology
    properties:
      clusters: ["local"]
      namespace: default
  - name: override-default
    type: override
    properties:
      components:
      - type: webservice
        properties:
        traits:
        - type: scaler
          properties:
            replicas: 1
        - type: container-image
          properties:
            image: busybox:1.35
            imagePullPolicy: IfNotPresent
        - type: env #################### conflict trait annotations
          properties:
            env:
              PARAMS: "trait env failed"
  workflow:
    steps:
    - name: deploy-default
      type: deploy
      properties:
        policies: ["topology-default", "override-default"]

```

**KubeVela Version**

```
CLI Version: v1.3.6
Core Version: 1.3.6
GitRevision: git-54477ea
GolangVersion: go1.17.10

```

**Cluster information**

**Additional context**

Somefive:

While dealing with multiple trait patch in one component, the temp variable used in previous trait patch will remain in the following patch, which will lead to two problems:
1. Recursive explosion. For example, one trait patch use a map [1] to record all containers, then the second trait also use a map [2] to record all containers, but since the map [1] is included in the container the first trait patch to, the map[2] will include map [1] and this will cause lots of useless render.
2. The temporary variable will cause name conflict across different traits.

Fixes #4203

I have:

### DIFF
--- a/pkg/cue/model/sets/utils.go
+++ b/pkg/cue/model/sets/utils.go
@@ -358,3 +358,38 @@ func listOpen(expr ast.Node) {
 		}
 	}
 }
+
+func removeTmpVar(expr ast.Node) ast.Node {
+	switch v := expr.(type) {
+	case *ast.File:
+		for _, decl := range v.Decls {
+			removeTmpVar(decl)
+		}
+	case *ast.Field:
+		removeTmpVar(v.Value)
+	case *ast.StructLit:
+		var elts []ast.Decl
+		for _, elt := range v.Elts {
+			if field, isField := elt.(*ast.Field); isField {
+				if ident, isIdent := field.Label.(*ast.Ident); isIdent && strings.HasPrefix(ident.Name, "_") {
+					continue
+				}
+			}
+			removeTmpVar(elt)
+			elts = append(elts, elt)
+		}
+		v.Elts = elts
+	case *ast.BinaryExpr:
+		removeTmpVar(v.X)
+		removeTmpVar(v.Y)
+	case *ast.EmbedDecl:
+		removeTmpVar(v.Expr)
+	case *ast.Comprehension:
+		removeTmpVar(v.Value)
+	case *ast.ListLit:
+		for _, elt := range v.Elts {
+			removeTmpVar(elt)
+		}
+	}
+	return expr
+}
--- a/pkg/cue/model/sets/walk_test.go
+++ b/pkg/cue/model/sets/walk_test.go
@@ -23,6 +23,7 @@ import (
 	"cuelang.org/go/cue/ast"
 	"cuelang.org/go/cue/parser"
 	"github.com/bmizerany/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestWalk(t *testing.T) {
@@ -127,4 +128,34 @@ func TestWalk(t *testing.T) {
 		}).walk(f)
 	}
 
+}
+
+func TestRemoveTmpVar(t *testing.T) {
+	src := `spec: {
+    _tmp: "x"
+	list: [{
+		_tmp: "x"
+		retain: "y"
+	}, {
+		_tmp: "x"
+		retain: "z"
+	}]
+	retain: "y"
+}
+`
+	r := require.New(t)
+	var runtime cue.Runtime
+	inst, err := runtime.Compile("-", src)
+	r.NoError(err)
+	s, err := toString(inst.Value(), removeTmpVar)
+	r.NoError(err)
+	r.Equal(`spec: {
+	list: [{
+		retain: "y"
+	}, {
+		retain: "z"
+	}]
+	retain: "y"
+}
+`, s)
 }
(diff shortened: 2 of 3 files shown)


pkg/cue/model/sets/utils.go
```diff
@@ -358,3 +358,38 @@ func listOpen(expr ast.Node) {
...
+		for _, decl := range v.Decls {
+			removeTmpVar(decl)
+		}
+	case *ast.Field:
+		removeTmpVar(v.Value)
+	case *ast.StructLit:
+		var elts []ast.Decl
+		for _, elt := range v.Elts {
+			if field, isField := elt.(*ast.Field); isField {
+				if ident, isIdent := field.Label.(*ast.Ident); isIdent && strings.HasPrefix(ident.Name, "_") {
```
wonderflow, line 374: what if the temp var is from user input?

wonderflow: We may use temp var in workflowstep, will that be affected?

wonderflow: For example:

```
_data: op.Read & {
   kind: "Deployment"
   metadata: name: "my"
}
output: _data
```

Somefive: > Do we have any rules that can help identify the temp var generate from vela system?

What do you mean for "generate from vela system"? The variable conflict mentioned here is inside CUE trait definition. For issue #4203, the conflict resides in 
- https://github.com/kubevela/kubevela/blob/71a02b434b3e85c56c43d761b8db46bef94ec63e/vela-templates/definitions/internal/trait/container-image.cue#L23
- https://github.com/kubevela/kubevela/blob/71a02b434b3e85c56c43d761b8db46bef94ec63e/vela-templates/definitions/internal/trait/env.cue#L22
I think it should not have any relation with "vela-system"?

wonderflow: I got it, actually you're ommiting temp var from users.

wonderflow (approved):

pkg/cue/model/sets/utils.go
Somefive: > For example:
> 
> ```
> _data: op.Read & {
>    kind: "Deployment"
>    metadata: name: "my"
> }
> output: _data
> ```


This is okay. The output will be
```
output: {
    kind: "Deployment"
    metadata: name: "my"
    spec: {...}
}
```
Before this PR, it will be
```
output: {
    kind: "Deployment"
    metadata: name: "my"
    spec: {...}
}
_data: {
    kind: "Deployment"
    metadata: name: "my"
    spec: {...}
}
```

Somefive: > We may use temp var in workflowstep, will that be affected?

By now, this only affect trait patch.

wonderflow: This is okay. The output will be
```
output: {
    kind: "Deployment"
    metadata: name: "my"
    spec: {...}
}
```
Before this PR, it will be
```
output: {
    kind: "Deployment"
    metadata: name: "my"
    spec: {...}
}
_data: {
    kind: "Deployment"
    metadata: name: "my"
    spec: {...}
}
```

This looks good to me, we should also make this work for workflow, or we will meet the same issue soon.

FogDong: We'll not allow conflict values in workflow step since they're in one cue instance.

pkg/cue/model/sets/walk_test.go
```diff
@@ -128,3 +129,33 @@ func TestWalk(t *testing.T) {
 	}
 
 }
+
+func TestRemoveTmpVar(t *testing.T) {
+	src := `spec: {
```
FogDong, line 134: Generally LGTM, suggestions: add some test cases like below:

```
_temp: "x"
retain: _temp
```

FogDong (approved):

PR #4213, opened 2022-06-21
Feat: support shared-resource policy

Somefive:

#### Background

In KubeVela, by default, application **owns** resources.
It means that resources create by the application should only be controlled by the application that creates it.

So there are basically two requirements for application creating resources:
1. The resource must not exist before the application creating it. It exists, there will be a resource conflict error.
2. The resource is expected to be only manageable through its creator. "Others" should not be able to modify it or edit it.

While dispatching resources, the application will
1. Check if resource exists. If exists, check its labels. 
If "app.oam.dev/name" and "app.oam.dev/namespace" equals to the application's name and namespace, it means this resource is previously created by the same application and the dispatching operation now will become an update operation.
The two labels identify the owner of the resource.
2. If resource exists, but no label found, then this resource is created before this application. At this time, the application will report a resource conflict error.
3. If resource exists, and the labels point to another application, then this resource is managed by other applications. At this time, the current application will also report a resource conflict error.

With these checks, different applications cannot manage the same resource. 

#### Usage

However, there are scenarios that these two requirements are not met. One of the scenarios is sharing across different Applications.
For example, each application wants to create a ConfigMap, but their ConfigMaps are the same.

To achieve that, KubeVela application could utilize the `shared-resource` policy to make it possible.

##### create

When one resource is created as sharing resource, one special annotation `app.oam.dev/shared-by` will be added to the resource.
It will record the "sharer" of the resource in time order. The application that firstly creates the resource will set its owner labels to itself.
Then it will add itself to the sharer annotation.

##### share

When another application comes and wants to share the resource, it will check if the resource is sharable, aka there is at least one sharer in the sharer annotation.
If it is sharable, it will add itself to the sharer annotation, but not modify the content of the resource.

##### delete

With this mechanism, only the owner of the resource can modify the resource (including updating and state-keeping). Other sharer can only see that resource.
When the owner of the resource is gone (application is deleted or do not use this resource anymore), it will give the owner of the application to the next sharer. If no sharer exists, it will finally delete that resource.

See the following figures for details.

image
image

#### Example

```yaml
kind: Application
metadata:
  name: app1
spec:
  components:
    - name: ns1
      type: k8s-objects
      properties:
        objects:
          - apiVersion: v1
            kind: Namespace
            metadata:
              name: example
    - name: cm1
      type: k8s-objects
      properties:
        objects:
          - apiVersion: v1
            kind: ConfigMap
            metadata:
              name: cm1
              namespace: example
            data:
              key: value1
  policies:
    - name: shared-resource
      type: shared-resource
      properties:
        rules:
          - selector:
              resourceTypes: ["Namespace"]
```

```yaml
kind: Application
metadata:
  name: app2
spec:
  components:
    - name: ns2
      type: k8s-objects
      properties:
        objects:
          - apiVersion: v1
            kind: Namespace
            metadata:
              name: example
    - name: cm2
      type: k8s-objects
      properties:
        objects:
          - apiVersion: v1
            kind: ConfigMap
            metadata:
              name: cm2
              namespace: example
            data:
              key: value2
  policies:
    - name: shared-resource
      type: shared-resource
      properties:
        rules:
          - selector:
              resourceTypes: ["Namespace"]
```

The above two applications will dispatch the same namespace "example".
They will create two different ConfigMap inside namespace "example" respectively.

Both application use the `shared-resource` policy and declared the namespace resource as shared.
In this way, there will be no conflict for creating the same namespace.
If the `shared-resource` policy is not used, the second application will report error after it finds that the namespace "example" is managed by the first application. 

The namespace will only be recycled when both applications are removed.

I have:


apis/core.oam.dev/v1alpha1/garbagecollect_types.go
```diff
@@ -66,6 +67,29 @@ type ResourcePolicyRuleSelector struct {
 	OAMResourceTypes []string `json:"oamTypes"`
 	TraitTypes       []string `json:"traitTypes"`
 	ResourceTypes    []string `json:"resourceTypes"`
+	ResourceNames    []string `json:"resourceNames"`
+}
+
+// Match check if current rule selector match the target resource
+func (in ResourcePolicyRuleSelector) Match(manifest *unstructured.Unstructured) bool {
```
wonderflow, line 74:
```suggestion
func (in *ResourcePolicyRuleSelector) Match(manifest *unstructured.Unstructured) bool {
```

It's save for a object method to use pointer to avoid "value copy issues".

wonderflow: https://stackoverflow.com/questions/25382073/defining-golang-struct-function-using-pointer-or-not

pkg/utils/apply/apply.go
```diff
@@ -315,3 +330,43 @@ func DisableUpdateAnnotation() ApplyOption {
...
+func SharedByApp(app *v1beta1.Application) ApplyOption {
+	return func(act *applyAction, existing, desired client.Object) error {
+		// calculate the shared-by annotation
+		// if resource exists, add the current application into the resource shared-by field
+		var sharedBy string
+		if existing != nil && existing.GetAnnotations() != nil {
+			sharedBy = existing.GetAnnotations()[oam.AnnotationAppSharedBy]
+		}
+		sharedBy = AddSharer(sharedBy, app)
+		util.AddAnnotations(desired, map[string]string{oam.AnnotationAppSharedBy: sharedBy})
```
wonderflow, line 344: will the annotation updated if the apply skipped?

test/e2e-multicluster-test/multicluster_test.go
```diff
@@ -501,5 +501,59 @@ var _ = Describe("Test multicluster scenario", func() {
...
+			}, 10*time.Second).Should(Succeed())
+			Expect(k8sClient.Delete(hubCtx, app1)).Should(Succeed())
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(hubCtx, types.NamespacedName{Namespace: namespace, Name: "shared"}, &corev1.ConfigMap{})).Should(Succeed())
+				g.Expect(k8sClient.Get(hubCtx, types.NamespacedName{Namespace: namespace, Name: "non-shared-app1"}, &corev1.ConfigMap{})).Should(Satisfy(kerrors.IsNotFound))
+				g.Expect(k8sClient.Get(hubCtx, types.NamespacedName{Namespace: namespace, Name: "non-shared-app3"}, &corev1.ConfigMap{})).Should(Succeed())
+			}, 10*time.Second).Should(Succeed())
+			Expect(k8sClient.Delete(hubCtx, app3)).Should(Succeed())
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(hubCtx, types.NamespacedName{Namespace: namespace, Name: "shared"}, &corev1.ConfigMap{})).Should(Satisfy(kerrors.IsNotFound))
```
wonderflow, line 554: you should check the annotation of the shared resource, check if it matches the expected state.

pkg/utils/apply/apply.go
Somefive: Yes, it will.

apis/core.oam.dev/v1alpha1/garbagecollect_types.go
Somefive: OK.

docs/examples/app-with-policy/shared-resource-policy/shared-resource.md
```diff
@@ -0,0 +1,91 @@
+## How to share resources across applications
```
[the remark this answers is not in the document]
Somefive: OK. Another design doc with the figures is added to the design folder.

apis/core.oam.dev/v1alpha1/policy_types.go
```diff
@@ -53,3 +57,23 @@ type OverridePolicySpec struct {
 	Components []EnvComponentPatch `json:"components,omitempty"`
 	Selector   []string            `json:"selector,omitempty"`
 }
+
+// SharedResourcePolicySpec defines the spec of shared-resource policy
+type SharedResourcePolicySpec struct {
```
barnettZQG, line 62: You should create the definition for this policy.

wonderflow (approved):

pkg/utils/apply/share.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
+)
+
+const (
+	sharedBySep = ","
+)
+
+// AddSharer add sharer
+func AddSharer(sharedBy string, app *v1beta1.Application) string {
```
barnettZQG, line 32: How to guarantee non-conflict? If multiple applications are published at the same time

Somefive: Conflict will be rejected by Kubernetes APIServer. For example, if the resource has sharer A now, B & C both wants to add themselves to the sharer annotation. B&C sends the patch request at the same time.

B&C requests will be handled by kube-apiserver in some order, not concurrently. Let's say B is handled first, the sharer will become "A,B". C's request then will be rejected by kube-apiserver since it is trying to mutate based on an old resourceVersion for that resource. Then it will fail the reconcile and try again. Next time he will find sharer "A,B" at first, and then try to add C to it. The conflicts will be solved in this way.

apis/core.oam.dev/v1alpha1/policy_types.go
Somefive: shared-resource is an advanced feature. Let's reserve the creation of the policy definition in the future. (Another Issue will be launched for this.)

PR #3648, opened 2022-04-14
Feat: support kruise rollout

Somefive:
Signed-off-by: Somefive <yd219913@alibaba-inc.com>

### Description of your changes

Support Kruise Rollout, including:
1. Add trait definition for kruise rollout, allow webservice / ref-objects to attach it.
2. Add support in vela-cli.

I have:

- [x] Read and followed KubeVela's [contribution process](https://github.com/oam-dev/kubevela/blob/master/contribute/create-pull-request.md).
- [ ] [Related Docs](https://github.com/oam-dev/kubevela.io) updated properly. In a new feature or configuration option, an update to the documentation is necessary. 
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Kruise Rollout release official version.
- [ ] Add tests.

PR #4243, opened 2022-06-23
Feat: Support kruise rollout

wangyikewxgm:
base on: https://github.com/kubevela/kubevela/pull/3648

1. refactor some code, funcs can be reached by apiserver and cli
2. add tests
3. rollback logic modified
4. topology filter by owner reference, cause canary deployment will have same labels will stable deployment , the graph will have intersection, so filter by owner reference.

before:

image

after:

Fixes #

I have:


pkg/workflow/operation/operation.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/oam-dev/kubevela/pkg/oam"
+	"github.com/oam-dev/kubevela/pkg/resourcetracker"
+	"github.com/oam-dev/kubevela/pkg/rollout"
+	errors3 "github.com/oam-dev/kubevela/pkg/utils/errors"
+
+	"github.com/pkg/errors"
+)
+
+// WfOperator is opratior handler for workflow's resume/rollback/restart
+type WfOperator interface {
```
Somefive, line 41: It is cool to move the actions of Workflow into a separate package. Could it be renamed as WorkflowOperator to make it clear?

Somefive: If possible, I would prefer an alternative usage like
`NewApplicationClient(k8sClient).Workflow().Suspend(ctx, app)` but it is okay for the WorkflowOperator as well.

wangyikewxgm: > It is cool to move the actions of Workflow into a separate package. Could it be renamed as WorkflowOperator to make it clear?

Good suggestion.

wangyikewxgm: > If possible, I would prefer an alternative usage like `NewApplicationClient(k8sClient).Workflow().Suspend(ctx, app)` but it is okay for the WorkflowOperator as well.

Since the client only update workflow currently, feel free refactor a this in the future?

go.mod
```diff
@@ -314,6 +318,7 @@ require (
 replace (
 	github.com/docker/cli => github.com/docker/cli v20.10.9+incompatible
 	github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
+	github.com/openkruise/rollouts => github.com/zmberg/rollouts v0.0.0-20220412081748-c6fabf54ab85
```
wonderflow, line 321: why we need this replace?

wangyikewxgm: fixed.

pkg/workflow/operation/operation.go
Somefive: Absolutely ok.

Somefive (approved):

wonderflow (approved):
good job

PR #4270, opened 2022-06-27
Feat: store outputs for unsuccessful steps in workflow context

FogDong:

Previously, if the component was not ready, its output would not be stored in the context. This PR fixes the problem and now we can pass the status of the component even if it's failed like:

```yaml
kind: Application
metadata:
  name: err-with-always
  namespace: default
spec:
  components:
  - name: invalid
    type: webservice
    properties:
      image: invalid
      ports:
        - port: 8000
  workflow:
    steps:
      - name: comp
        type: apply-component
        timeout: 5s
        outputs:
          - name: status
            valueFrom: output.status.conditions[0].type + output.status.conditions[0].status
        properties:
          component: invalid
      - name: notification
        type: notification
        inputs:
          - from: status
            parameterKey: slack.message.text
        if: always
        properties:
          slack:
            url:
              value: <url>
```

I have:

### DIFF
--- a/pkg/workflow/hooks/data_passing.go
+++ b/pkg/workflow/hooks/data_passing.go
@@ -47,10 +47,10 @@ func Output(ctx wfContext.Context, taskValue *value.Value, step v1beta1.Workflow
 	if wfTypes.IsStepFinish(status.Phase, status.Reason) {
 		for _, output := range step.Outputs {
 			v, err := taskValue.LookupByScript(output.ValueFrom)
-			if err != nil {
+			if err != nil && !strings.Contains(err.Error(), "not found") {
 				return err
 			}
-			if v.Error() != nil {
+			if err != nil || v.Error() != nil {
 				v, err = taskValue.MakeValue("null")
 				if err != nil {
 					return err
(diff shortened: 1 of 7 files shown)


pkg/workflow/hooks/data_passing.go
```diff
@@ -47,10 +47,10 @@ func Output(ctx wfContext.Context, taskValue *value.Value, step v1beta1.Workflow
 	if wfTypes.IsStepFinish(status.Phase, status.Reason) {
 		for _, output := range step.Outputs {
 			v, err := taskValue.LookupByScript(output.ValueFrom)
-			if err != nil {
+			if err != nil && !strings.Contains(err.Error(), "not found") {
```
StevenLeiZhang, line 50: Would you extract it to be one util function, which decides the error is NotFound?

FogDong: This judgment is only for the outputs so I don't think we should extract it to a util function.

Somefive (approved):

wonderflow (approved):

PR #4291, opened 2022-06-30
Fix: context.namespace use the namespace from override policy

Somefive:

context.namespace use override namespace when rendering component and trait

I have:


wonderflow (approved):

fourierr:
I want to make sure it,  the defination of context.namespace in cue.context has changed from app's ns to pod's ns ? @wonderflow

wonderflow:
@fourierr Not exactly, the context.namespace is the namespace of component, in this PR, it will align to component which is affected by the override policy.

fourierr:
> @fourierr Not exactly, the context.namespace is the namespace of component, in this PR, it will align to component which is affected by the override policy.

you are right, but the `context.appNamespace` is useful to trait cr, trait cr has been created in app's ns for more convenient management.

wonderflow:
@fourierr By default, the trait will be created in the app namespace, so it's not necessary to use `context.appNamespace`.

PR #4295, opened 2022-07-01
Fix: goruntine leak in CUE http

wonderflow:

Fixes #

I have:


pkg/builtin/http/http.go
```diff
@@ -117,9 +117,9 @@ func (c *HTTPCmd) Run(meta *registry.Meta) (res interface{}, err error) {
 			tr.TLSClientConfig.Certificates = []tls.Certificate{cliCrt}
 		}
 
-		client.Transport = tr
+		c.client.Transport = tr
```
barnettZQG, line 120: The different meta config will use the same transport config. such as:

1. The first request with `tls_config`;
2. The second request is without `tls_config`, but it will use the last transport.

PR #4347, opened 2022-07-11
Fix: ignore empty output file

Somefive:

When addon directly has cue files that do not have output, skip it instead of reporting error.

This is useful for conditional rendering, for example, add component X if key=value.


barnettZQG (approved):
LGTM

PR #4348, opened 2022-07-11
Fix: escape resource name for label selector

Somefive:

Kubernetes resources support names containing character ':', but label does not allow it. This PR escapes the name when filling it into the value of label.


barnettZQG (approved):
LGTM

PR #4370, opened 2022-07-13
Feat: enhance addon init

charlie0129:

- Make blue color clearer on black backgrounds
- Generate sample files by default to get user started (can be skipped)
- Support `--url` to add a ref-objects component (use yaml files from URLs)
- Rename existing `readme` to `README`
- Refactor code structure

I have:

Unit-tests.

Manually tested with several possible combinations of parameters.


wonderflow:
> Rename existing readme to README

Please make sure API is compatible with this change.

charlie0129:
> Please make sure API is compatible with this change.
 
This should work just fine:
1. Before this change, the existing addons have both `readme.md` and `README.md`. They all worked, meaning changing case should breaking anything.
2. This change should only affect reading the `README.md` from disk. While most filesystems are case-insensitive, so it should read the right file.

references/cli/addon_test.go
```diff
@@ -421,16 +421,16 @@ func TestGenerateParameterString(t *testing.T) {
...
 	cmd.SetArgs([]string{})
 	err := cmd.Execute()
 	assert.ErrorContains(t, err, "required")
 
-	cmd.SetArgs([]string{"--chart", "a", "--helm-repo-url", "https://some.com", "--version", "c"})
+	cmd.SetArgs([]string{"--chart", "a", "--helm-repo", "https://some.com", "--chart-version", "c"})
 	err = cmd.Execute()
 	assert.ErrorContains(t, err, "required")
 
-	cmd.SetArgs([]string{"test-addon", "--chart", "a", "--helm-repo-url", "https://some.com", "--version", "c"})
```
wonderflow, line 433: is this a break change?

wonderflow (commented):
does any docs need to change after this PR?

references/cli/addon_test.go
charlie0129: Yes, the parameters are adjusted. But this function is for humans, there are no automation tools or other things that uses this, so it won't affect much.

The CLI examples have been updated, so the user can refer to help.

wonderflow: well, we're not release it in a formal release, so we can change it for now. But next time, we should keep the consistence

wangyikewxgm (approved):

Somefive (approved):

Issue #3924, opened 2022-05-18
[Feature] System Info & Diagnose


Somefive:
**Is your feature request related to a problem? Please describe.**

In KubeVela, there are multiple deployments running in the vela-system namespace in the control plane, including
- vela-core
- cluster-gateway
- vela-prism
- vela apiserver
- velaux

Although in vela CLI we have `vela version` command which shows the version of kubevela running in the control plane, it is still laborious to check the details of those widgets. For example, to see which feature flags are enabled or what command args is used in vela-core, user needs to get the detail YAML of the kubevela-vela-core deployment in the vela-system namespace and check the running args of the controller. Additionally, not all widgets are always using the same version number with the vela-core

So, it will be useful if we can get all detail system information through vela CLI commands, like `vela system info`.

On the other hand, we also need convenient ways to diagnose system problems like
- application handling is slow
- failed to find APIService of cluster-gateway
- failed to reach managed cluster
- requests long latency

Those system widgets sometimes are overloaded or rate-limited due to limited resources or request rate limitation. Besides, due to the Kubernetes AggregatedServer mode, the requests related to cluster-gateway/vela-prism are associated with multiple connections (vela-core -> hub apiserver -> cluster-gateway -> managed cluster apiserver). So when diagnosing these problems, system operators need to check a lot of potential causes, such as if each connection path is active and stable. This process is a bit laborious.

To help alleviate the efforts to diagnose system problems, we can have `vela system diagnose` commands to do some automatic detection and give heuristic advices to some common cases (such as allocating more resources to vela-core or checking if the control plane supports AggregatedServer).

**Describe the solution you'd like**

Developing `vela system info` and `vela system diagnose` commands.

**Describe alternatives you've considered**

Developing a diagnose container image and run a debugging pod in control plane.

**Additional context**

It would be more powerful, if the system info/diagnose is not only dealing with KubeVela widgets but also do some heuristic testing to the control plane component, like kube-apiserver.

oeular:
@Somefive Is there any solution for the above issue in detail? We want to join this issue.

Somefive:
> @Somefive Is there any solution for the above issue in detail? We want to join this issue.

It needs design and further discussion. If you have any idea for this issue, feel free to discuss it here.

oeular:
- Can output the runtime configuration information for each component of vela independently via vela CLI, including addons
- Can output the runtime environment information, including k8s, container engine, os
[...]

PR #4379, opened 2022-07-13
Feat: System Info & Diagnose

foursevenlove:
Signed-off-by: foursevenlove <foursevenlove@gmail.com>

Feat #3924

I have:

- run ` vela system info ` to check all detail system information
- run ` vela system diagnose ` to diagnose system's health

@Somefive

### DIFF
--- a/references/cli/system.go
+++ b/references/cli/system.go
@@ -1,0 +1,158 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/oam-dev/cluster-gateway/pkg/generated/clientset/versioned"
+	"github.com/oam-dev/kubevela/apis/types"
+	"github.com/oam-dev/kubevela/pkg/multicluster"
+	"github.com/oam-dev/kubevela/pkg/utils/common"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// FlagDeploymentName specifies the deployment name
+	FlagDeploymentName = "name"
+)
+
+// NewSystemCommand print system detail info
+func NewSystemCommand(c common.Args) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "system",
+		Short: "Manage system.",
+		Long:  "Manage system, incluing printing the system deployment information in vela-system namespace and diagnosing the system's health.",
+		Example: "# Check all deployments information:\n" +
+			"> vela system info\n" +
+			"# Specify a deployment name to check detail information:\n" +
+			"> vela system info --name kubevela-vela-core\n" +
+			"# Diagnose the system's health:\n" +
+			"> vela system diagnose\n",
+		Annotations: map[string]string{
+			types.TagCommandType: types.TypeSystem,
+		},
+	}
+	cmd.AddCommand(
+		NewSystemInfoCommand(c),
+		NewSystemDiagnoseCommand(c))
+	return cmd
+}
+
+// NewSystemInfoCommand prints system detail info
+func NewSystemInfoCommand(c common.Args) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "info",
+		Short: "Print the system deployment detail information in vela-system namespace.",
+		Long:  "Print the system deployment detail information in vela-system namespace.",
+		Args:  cobra.ExactValidArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// Get deploymentName from flag
+			deployName, err := cmd.Flags().GetString(FlagDeploymentName)
+			if err != nil {
+				return errors.Wrapf(err, "failed to get deployment name flag")
+			}
+			table := newUITable().AddRow("NAME", "NAMESPACE", "IMAGE", "ARGS")
+			table.MaxColWidth = 120
+			config, err := c.GetConfig()
+			if err != nil {
+				return err
+			}
+			// Get clientset
+			clientset, err := kubernetes.NewForConfig(config)
+			if err != nil {
+				panic(err)
+			}
+			// Get deploymentsClient
+			deploymentsClient := clientset.AppsV1().Deployments(types.DefaultKubeVelaNS)
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			if deployName != "" {
+				// DeployName is not empty, print the specified deployment's all args
+				deployment, err := deploymentsClient.Get(ctx, deployName, metav1.GetOptions{})
+				if err != nil {
+					panic(err)
+				}
+				table.AddRow(deployment.Name, deployment.Namespace, deployment.Spec.Template.Spec.Containers[0].Image, strings.Join(deployment.Spec.Template.Spec.Containers[0].Args, " "))
+			} else {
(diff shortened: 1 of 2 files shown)


references/cli/system.go
```diff
@@ -0,0 +1,175 @@
...
+				deployment, err := deploymentsClient.Get(ctx, deployName, metav1.GetOptions{})
+				if err != nil {
+					panic(err)
+				}
+				table.AddRow(deployment.Name, deployment.Namespace, deployment.Spec.Template.Spec.Containers[0].Image, strings.Join(deployment.Spec.Template.Spec.Containers[0].Args, " "))
+			} else {
+				// DeployName is empty, print all deployment's partial args
+				deployments, err := deploymentsClient.List(ctx, metav1.ListOptions{})
+				if err != nil {
+					panic(err)
```
charlie0129, line 101: Same as above, prefer returning an error.

references/cli/system.go
```diff
@@ -0,0 +1,175 @@
...
+			} else {
+				cmd.Println(table.String())
+			}
+			return nil
+		},
+		Annotations: map[string]string{
+			types.TagCommandType: types.TypeSystem,
+		},
+	}
+	cmd.Flags().StringP(FlagDeploymentName, "n", "", "Specify the deployment name to check detail information. If empty, it will print all deployments information. Default to be empty.")
```
charlie0129, line 119: We typically think `-n` as `namespace`. Consider choosing a different letter or use non-flag arguments.

references/cli/system.go
```diff
@@ -0,0 +1,175 @@
...
+			defer cancel()
+			if deployName != "" {
+				// DeployName is not empty, print the specified deployment's all args
+				deployment, err := deploymentsClient.Get(ctx, deployName, metav1.GetOptions{})
+				if err != nil {
+					panic(err)
+				}
+				table.AddRow(deployment.Name, deployment.Namespace, deployment.Spec.Template.Spec.Containers[0].Image, strings.Join(deployment.Spec.Template.Spec.Containers[0].Args, " "))
+			} else {
+				// DeployName is empty, print all deployment's partial args
```
charlie0129, line 98: There are user-enabled addons in `vela-system` namespace as well. Is this an intended behavior?

references/cli/system.go
```diff
@@ -0,0 +1,175 @@
...
+			deploymentsClient := clientset.AppsV1().Deployments(types.DefaultKubeVelaNS)
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			if deployName != "" {
+				// DeployName is not empty, print the specified deployment's all args
+				deployment, err := deploymentsClient.Get(ctx, deployName, metav1.GetOptions{})
+				if err != nil {
+					panic(err)
+				}
+				table.AddRow(deployment.Name, deployment.Namespace, deployment.Spec.Template.Spec.Containers[0].Image, strings.Join(deployment.Spec.Template.Spec.Containers[0].Args, " "))
```
charlie0129, line 96: If we are getting a single deployment, which means the user want some detailed info, is fitting that deployment in a single row a good choice?

references/cli/system.go
```diff
@@ -0,0 +1,175 @@
...
+		NewSystemInfoCommand(c),
+		NewSystemDiagnoseCommand(c))
+	return cmd
+}
+
+// NewSystemInfoCommand prints system detail info
+func NewSystemInfoCommand(c common.Args) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "info",
+		Short: "Print the system deployment detail information in vela-system namespace.",
```
Somefive, line 66: Notice that while dealing with SystemInfo, the install namespace of KubeVela could be something other than "vela-system", so it would be better if we can have namespace-agnostic detection logics. This can be reserved as a future enhancement, not necessary for this PR.

Somefive (commented):
Except for the comments by @charlie0129 (which are great advices), this PR, as an initial bootstrap for the `vela system` command generally LGTM. There are more details to dig for these two commands, which can be addressed in future PRs. Thanks for the contribution!

Details:
1. The logic for `vela system info` is looking for the deployments in `vela-system`. But there are potential problems for that. Users can install KubeVela controller in other namespaces instead of `vela-system` and it is also possible to install vela-irrelevant deployments in the `vela-system`. Furthermore, even deployments can have zero replicas, which indicates they are not working. Deployments could be upgrading as well, which means the current working pods might not use the latest configuration for deployments while upgrading. So it might be better to add checks for the pods and look for KubeVela controllers by labels.
2. We could add simple resource check for KubeVela controller & ClusterGateway, for example, showing the current CPU/Memory usage.
3. Adding information for environment variables is also helpful, not only args.
4. Args can be grouped, according to their usage. You can try to make some groupings according to your understanding and let's make discussion in the future.
5. The discussion of diagnose command can be reserved in the future as well.

Somefive:
Would you like to make a small video / gif / snapshots for the command and attach it in the introduction of the PR? It would be nice and straightforward for other reviewers to understand it. :)

foursevenlove:
Thank you @Somefive @charlie0129  for reviewing my code and making valuable suggestions. Based on the original, I made the following modifications:
Fix:
- 1.Returning error instead of panic.
- 2.Getting deployment by label instead of by namespace .
- 3.When getting a single deployment, the result is displayed in multi rows. 
- 4.Using -s to specify a deployment instead of -n.

Feat: 
- 1.The system info command displays the cpu and memory metrics.
- 2.The system info command displays the numbers of ready pods and desired pods.
- 3.The system info command displays the environment variables.

Finally I made a gif to show how to use the system command：

Somefive:
> Thank you @Somefive @charlie0129 for reviewing my code and making valuable suggestions. Based on the original, I made the following modifications: Fix:
> 
> * 1.Returning error instead of panic.
> * 2.Getting deployment by label instead of by namespace .
> * 3.When getting a single deployment, the result is displayed in multi rows.
> * 4.Using -s to specify a deployment instead of -n.
> 
> Feat:
> 
> * 1.The system info command displays the cpu and memory metrics.
> * 2.The system info command displays the numbers of ready pods and desired pods.
> * 3.The system info command displays the environment variables.
> 
> Finally I made a gif to show how to use the system command：  [  ](https://user-images.githubusercontent.com/90825536/181453068-bcd8fffc-954c-46b5-ba42-55dfd3093ef8.gif) [ ](https://user-images.githubusercontent.com/90825536/181453068-bcd8fffc-954c-46b5-ba42-55dfd3093ef8.gif)

Looks great! I notice that the DCO CI has some problem. You could follow the guide to fix your commit signature.

Somefive (approved):
As a first initial PR, looks good to me. Following possible enhancements in later PR:
1. For system info, the list of arguments in the demo gif looks to be truncated due to the width of the terminal. Multi-line display might be more helpful.
2. For demo gif, it is clear that when everything is right, the command works pretty well. Let's make some fake broken scenarios, for example, let controller down, and show how the system-info will show.
3. For demo gif, another interesting thing is the upgrade process. What if the kubevela-controller is upgrading and there are two different pods running at the same time. Could the system info command show their differences?
4. For system info, showing CPU/Memory usage is cool, let's add the percentage usage if the resource limit is set. For example, memory: 456Mi (45.6%).
5. For system info, there are some logs could be potentially useful. But I haven't come up with the idea of how to handle the log information. Feel free to make discussions if you have any ideas on it.

As for the diagnose, similar to [2] above, let's make some bad situation and see how this command performs.

charlie0129 (approved):
Good job!

PR #4398, opened 2022-07-15
Feat: Upgrade golangci-lint version and fix detected lints

### LINKED ISSUE #4018: [Feature] Upgrade golangci-lint to a higher version

The golangci-lint in KubeVela repo is `v1.38.0` which is kinda outdated, it's nice to upgrade it to a higher version.

At the same time, some codes in the repo don't meet the lint requirement in the higher version, so please fix them in the upgrading.

You can use the command `make lint` to check all the results.

**Additional context**

MimicTester1307:

Fixes #4018 

I have:

- Upgrade `golangci-lint` version and fix some lint checks
- linter errors ignored: `errchkjson`, `contextcheck`, and `gocritic` (`offBy1` report)
- `gosec` error ignored too because  the report referenced 'possible file inclusion through variable', but on checking, the variables used to create the files were pre-defined constants, and would not hold user-provided values.

### DIFF
--- a/references/cli/delete.go
+++ b/references/cli/delete.go
@@ -31,7 +31,7 @@ import (
 // NewDeleteCommand Delete App
 func NewDeleteCommand(c common2.Args, order string, ioStreams cmdutil.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:                   "delete APP_NAME",
+		Use:                   "delete APP_NAME1 [APP_NAME2 APP_NAME3...]",
 		DisableFlagsInUseLine: true,
 		Short:                 "Delete an application",
 		Long:                  "Delete an application.",
@@ -61,7 +61,6 @@ func NewDeleteCommand(c common2.Args, order string, ioStreams cmdutil.IOStreams)
 		if len(args) < 1 {
 			return errors.New("must specify name for the app")
 		}
-		o.AppName = args[0]
 		wait, err := cmd.Flags().GetBool("wait")
 		if err != nil {
 			return err
@@ -72,17 +71,20 @@ func NewDeleteCommand(c common2.Args, order string, ioStreams cmdutil.IOStreams)
 			return err
 		}
 		o.ForceDelete = force
-		userInput := NewUserInput()
-		if !assumeYes {
-			userConfirmation := userInput.AskBool(fmt.Sprintf("Do you want to delete the application %s from namespace %s", o.AppName, o.Namespace), &UserInputOptions{assumeYes})
-			if !userConfirmation {
-				return fmt.Errorf("stopping Deleting")
+		for _, app := range args {
+			o.AppName = app
+			userInput := NewUserInput()
+			if !assumeYes {
+				userConfirmation := userInput.AskBool(fmt.Sprintf("Do you want to delete the application %s from namespace %s", o.AppName, o.Namespace), &UserInputOptions{assumeYes})
+				if !userConfirmation {
+					return fmt.Errorf("stopping Deleting")
+				}
 			}
+			if err = o.DeleteApp(ioStreams); err != nil {
+				return err
+			}
+			ioStreams.Info(green.Sprintf("app \"%s\" deleted from namespace \"%s\"", o.AppName, o.Namespace))
 		}
-		if err = o.DeleteApp(ioStreams); err != nil {
-			return err
-		}
-		ioStreams.Info(green.Sprintf("app \"%s\" deleted from namespace \"%s\"", o.AppName, o.Namespace))
 		return nil
 	}
 
--- a/references/common/application.go
+++ b/references/common/application.go
@@ -186,7 +186,7 @@ func (o *DeleteOptions) DeleteAppWithoutDoubleCheck(io cmdutil.IOStreams) error 
 	err := o.Client.Get(ctx, client.ObjectKey{Name: o.AppName, Namespace: o.Namespace}, app)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			return fmt.Errorf("app %s already deleted or not exist", o.AppName)
+			return nil
 		}
 		return fmt.Errorf("delete application err: %w", err)
 	}


pkg/addon/create.go
```diff
@@ -247,7 +247,7 @@ func writeRequiredFiles(addonPath string, tmplContent, readmeContent, metadataCo
 	templateFilePath := path.Join(addonPath, TemplateFileName)
 	err := os.WriteFile(templateFilePath,
 		tmplContent,
-		0644)
+		0600)
```
charlie0129, line 250: This is a template for the user. We want other user to be able to access it, so please keep `0644`. You can ignore this by `// nolint:gosec`

charlie0129:
> `gosec` error ignored too because the report referenced 'possible file inclusion through variable', but on checking, the variables used to create the files were pre-defined constants, and would not hold user-provided values.

Why not fix it instead of ignoring it? Is there any difficulty?

charlie0129:
> linter errors ignored: `errchkjson`, `contextcheck`, and `gocritic`

Is there any specific reason to ignore these checks? (I agree with you for the `contextcheck` part, as it will required a large effort to fix and can be saved for future improvements.)

pkg/addon/create.go
MimicTester1307: Alright, thank you.

MimicTester1307:
> > `gosec` error ignored too because the report referenced 'possible file inclusion through variable'
> 
> Why not fix it instead of ignoring it globally?
> 
> > but on checking, the variables used to create the files were pre-defined constants, and would not hold user-provided values.
> 
> If you know a line that is fine and won't be fixed, you can use a `nolint:xxx` comment to ignore that specific line.
> 
> BTW, `possible file inclusion through variable` can sometimes be fixed using `filepath.Clean()`

I know about `filepath.Clean()`, but I figured that since the paths being referenced were pre-defined constants and not user-supplied values, it wasn't necessary to add that, but no worries, I will update the code.

Thank you for the `nolint:xxx` suggestion. I didn't know that.

charlie0129:
> I know about `filepath.Clean()`, but I figured that since the paths being referenced were pre-defined constants and not user-supplied values, it wasn't necessary to add that, but no worries, I will update the code.

Great. If you have a good reason to ignore that, feel free to just ignore the linter for that line.

MimicTester1307:
> > linter errors ignored: `errchkjson`, `contextcheck`, and `gocritic`
> 
> Is there any specific reason to ignore these checks? (I agree with you for the `contextcheck` part, as it will require a large effort to fix and can be saved for future improvements.)

For `gocritic`, it was a possible `offBy1` error, so I didn't fix it because it had worked just fine in the code base. Has there been an issue with this in the past? I'm a bit new to the code base.

For 'errchkjson`, it was that the output of the `json.Marshal` function wasn't checked. But in the outer function, there was only one return type (`[]byte`), so I figured it was intentional? Checking the output would have meant adding an extra return type of `error`, which could cause breaking changes in other places that the function is used. If there is a way to resolve that, please let me know.

pkg/addon/create.go
MimicTester1307: I've fixed this.

MimicTester1307:
> > I know about `filepath.Clean()`, but I figured that since the paths being referenced were pre-defined constants and not user-supplied values, it wasn't necessary to add that, but no worries, I will update the code.
> 
> Great! If you have a good reason to ignore that, feel free to just ignore the linter for that line.

Sure, thanks. Now that I know about `nolint:xxx`, I'll add the comment rather than just ignoring it.  What do you think about the reasons for the `errchkjson` and 'gocritic`? I can use `nolint:xxx` on those specific functions for now?

charlie0129:
> For `gocritic`, it was a possible `offBy1` error, so I didn't fix it because it had worked just fine in the code base.

Yes, you can leave it as-is. If you come with a better solution, you can fix it.

I just took a look at the errors. For example you can fix this one:
https://github.com/kubevela/kubevela/blob/01737d62b09e37f0df0c9107c96c5f5d5b5d7ca2/references/cli/uischema.go#L133
with: 
https://github.com/kubevela/kubevela/blob/01737d62b09e37f0df0c9107c96c5f5d5b5d7ca2/pkg/utils/file.go#L116-L118

> But in the outer function, there was only one return type ([]byte), so I figured it was intentional? Checking the output would have meant adding an extra return type of error, which could cause breaking changes in other places that the function is used.

Yes, you have a good reason for this. 

But for this part, it will be better that the original author gives his opinion. So I will just leave my opinion for the original author.

charlie0129:
Also, if you take a look at the CI, you will need to sign your commits (click the DCO check for details) and run `make reviewable` (make sure it passes).

charlie0129:
In case you missed it, there are golangci-lints in GitHub Actions workflows as well. You probably want to update these as well.

For example, one of them is like:

https://github.com/kubevela/kubevela/blob/01737d62b09e37f0df0c9107c96c5f5d5b5d7ca2/.github/workflows/go.yml#L17

pkg/apiserver/event/sync/store.go
```diff
@@ -18,8 +18,9 @@
~
 import (
 	"context"
```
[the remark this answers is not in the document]
FogDong: Many IDEs can lint your code with goimports when saving the code, you can set it and lint your code.

FogDong:
If some of the linters are still in discussion, we can fix part of the lint first and keep the old lint version.

BTW, you need to rebase from the master branch and re-push the code.

pkg/apiserver/event/sync/store.go
```diff
@@ -42,9 +43,10 @@ func StoreProject(ctx context.Context, name string, ds datastore.DataStore, proj
 		return err
 	}
 	if projectService != nil {
+		caser := cases.Title(language.English)
```
Somefive, line 46: Interesting fix. Related to this issue, https://github.com/golang/go/issues/48367.

pkg/addon/create.go
```diff
@@ -242,6 +242,7 @@ func createSampleFiles(addonName, addonPath string) error {
 
 // writeRequiredFiles creates required files for an addon,
 // including template.yaml, readme.md, and metadata.yaml
+//nolint:gosec
```
Somefive, line 245: // nolint:gosec

Somefive (commented):
Generally LGTM after the go import nit. Great PR. Thanks for the contribution!

MimicTester1307:
> If some of the linters are still in discussion, we can fix part of the lint first and keep the old lint version.
> 
> BTW, you need to rebase from the master branch and re-push the code.

Sure, I'll fix some of the discussed issues and do that.

pkg/addon/create.go
MimicTester1307: > Use //nolint instead of // nolint because machine-readable comments should have no space by Go convention.

From [here](https://golangci-lint.run/usage/false-positives/#Use%20//nolint%20instead%20of%20//%20nolint%20because%20machine-readable%20comments%20should%20have%20no%20space%20by%20Go%20convention.:~:text=tests%20for%20it.-,Use,because%20machine%2Dreadable%20comments%20should%20have%20no%20space%20by%20Go%20convention.,-Edit%20this%20page).

That was why I wrote it without the space.

MimicTester1307:
> Generally LGTM after the go import nit. Great PR. Thanks for the contribution!

Thank you!

pkg/addon/create.go
Somefive: It's reasonable. You are right. We can make it without the extra space.

Issue #4320, opened 2022-07-04
ResourceKeeper delete resource not belong to the current application


Somefive:
**Describe the bug**

**To Reproduce**

When deploying resources to locations where resource already exists, although the APPLY will failed, the resource will still be recorded in the ResourceTracker, which will let the resource to be recycled when the application is deleted.

**Expected behavior**

**Screenshots**

**KubeVela Version**

**Cluster information**

**Additional context**

PR #4399, opened 2022-07-18
Fix: prevent apply existing orphan resource

Somefive:

When resource exists but not owned by any application, if one new application wants to dispatch a resource at the same place, it will "TAKE OVER" the original resource instead of being prevented. This PR disallows this behaviour.

Issue https://github.com/kubevela/kubevela/issues/4320, although not directly refer to this problem, is caused by this misbehaviours for the most time.

I have:


pkg/utils/apply/apply.go
```diff
@@ -290,8 +290,11 @@ func MustBeControlledByApp(app *v1beta1.Application) ApplyOption {
 			return nil
 		}
 		appKey, controlledBy := GetAppKey(app), GetControlledBy(existing)
+		if controlledBy == "" {
+			return fmt.Errorf("%s %s/%s exists but not managed by any application now", existing.GetObjectKind().GroupVersionKind().Kind, existing.GetNamespace(), existing.GetName())
+		}
```
wonderflow, line 295: will it cause some compatibility issues? not all resources has this label

Somefive: All resources created by KubeVela application should have this label. If not, they will be bugs.

Without this condition, it is pretty dangerous because KubeVela application will be able to take over arbitrary resources (if not managed by other KubeVela Application).

For example, you want application A to create ConfigMap X but due to the configuration error, A tries to create ConfigMap Y. But if Y already exists there, A will take over Y and delete it if you fix the application A.

wonderflow: > All resources created by KubeVela application should have this label. If not, they will be bugs.

Can you describe some more details? You mean all resources dispatched by vela **already MUST** have these labels?

Somefive: Right. 

#### kube.#Apply

https://github.com/kubevela/kubevela/blob/a519a6c89d53eea3a5cff7383648eae22829c2f4/pkg/workflow/providers/kube/handle.go#L97

#### Normal Dispatch

https://github.com/kubevela/kubevela/blob/a519a6c89d53eea3a5cff7383648eae22829c2f4/pkg/controller/core.oam.dev/v1alpha2/application/generator.go#L340 prepareWorkloadAndManifests
https://github.com/kubevela/kubevela/blob/a519a6c89d53eea3a5cff7383648eae22829c2f4/pkg/appfile/appfile.go#L302 SetOAMContract
https://github.com/kubevela/kubevela/blob/a519a6c89d53eea3a5cff7383648eae22829c2f4/pkg/appfile/appfile.go#L327 generateAndFilterCommonLabels

Somefive: But still, there will be compatibility issues because some resources dispatched by kube.#Apply does not have this label in early KubeVela versions.

wonderflow: I'm not sure if the namespace will affect, not all of them have namespace

Somefive: https://github.com/kubevela/kubevela/blob/a519a6c89d53eea3a5cff7383648eae22829c2f4/pkg/appfile/appfile.go#L328 generateAndFilterCommonLabels

For common cases, the namespace should be set. 

It is true that there are legacy applications dispatch resources without namespace set. One of the alternative way is to add a controller option to allow legacy behaviour. What do you think?

wonderflow: yes, add a featureflag is better. We need to add it on release not. and it's better to keep in v1.5 only

wonderflow (approved):

PR #4401, opened 2022-07-18
Feat: Addon support app template written by cuelang.

### LINKED ISSUE #4283: [Feature]  Addon Install resources in needNamespace

Proposal about addon installing resources into specified namespace：
Because of the mechanism of application applying resources, resource with one specified namespace defined in `raw` or `k8s-objects` type component can be deployed into it (such as [fluxcd addon](https://github.com/kubevela/catalog/blob/master/addons/fluxcd/resources/deployment/helm-controller.yaml) ). But webservice or worker component which use cue template rendering resource don't have this ability yet.

So we should provide a way to let addon maker install resources into any namespace.

1. Addon maker can specify the namespace he want to deploy, since we already have a filed of `needNamespace` and all addons which using this field want to deploy resources to this namespace such as: flink-kubernetes-operator, mysql-operator and so on. So I propose install resources into the needNamespace directly.

2. Render the the application with `needNamespace` topology policy, for example if addon's meta is :

```yaml
needNamespace:
   - my-namespace
```

the Application is :

```yaml
kind: Application
metadata:
   name: some-addon
   namespace: vela-system
components:
   type: k8s-objects
   name: namespace
   properties:
      apiVersion: v1
      kind: Namespace
      metadata:
          name: my-namespace
policies:
   - type: topology
     properties:
        namespace: my-namespace
        cluster: [] // Please be aware of that target clusters will fill here while enabling.
  - type: share-resource
     name: shared-namespace
     properties:
         rules:
            - selector:
                   resourceType: ["Namespace"] // Namespace also can be created by another application
```

3. The needNamespace will be filled in topology namespace, and clusters specified by user when enable an addon will set in clusters field.

4. Since this namespace can be created by other addons, so add a shared-resource policy for namespace.

**Additional context**

wangyikewxgm:
This pr support writing basic addon application template framework by cue, so addon-developer can define the final app with parameter.Will allow user define a `parameter.cue` file  outside `resource` dir. All parameters defined in this file will affect resources in `resources` dir and `template.cue`.

So  a new addon dir is like this:
```dir
definitions/
resources/
——component1.cue
——component2.yaml
template.cue
parameter.cue
metadata.yaml
```

Besides, this pr also dose such things:

1. Keep compatibility for old addon, new version CLI can still enable old version addon.
2. If an addon defines `deployTo.runtimeCluster=true` want to deploy to runtime clusters but still don't have a topology polocy, we assume it's a legacy addon, will attach a topology policy to this app,  not two workflowSteps(deploy-local, deploy-all-runtime) as before.
3. Delete all hack code for observability addon, because all these codes only for selecting cluster and override policy for different cluster, this will handled by writing a `template.cue`.
4. Refactor some codes, put render related code to a single file.

Fixes #4283 

I have:

### DIFF
--- a/pkg/addon/render.go
+++ b/pkg/addon/render.go
@@ -23,11 +23,9 @@ import (
 	"path"
 	"strings"
 
-	cueyaml "cuelang.org/go/encoding/yaml"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/yaml"
 
 	common2 "github.com/oam-dev/kubevela/apis/core.oam.dev/common"
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1alpha1"
@@ -40,6 +38,12 @@ import (
 	addonutil "github.com/oam-dev/kubevela/pkg/utils/addon"
 )
 
+const (
+	specifyAddonClustersTopologyPolicy = "deploy-addon-to-specified-clusters"
+	addonAllClusterPolicy              = "deploy-addon-to-all-clusters"
+	renderOutputCuePath                = "output"
+)
+
 type addonCueTemplateRender struct {
 	addon     *InstallPackage
 	inputArgs map[string]interface{}
@@ -48,15 +52,17 @@ type addonCueTemplateRender struct {
 // This func can be used for addon render, supporting render app template and component.
 // Please notice the result will be stored in object parameter, so object must be a pointer type
 func (a addonCueTemplateRender) toObject(cueTemplate string, object interface{}) error {
-	bt, err := json.Marshal(a.inputArgs)
+	args := a.inputArgs
+	if args == nil {
+		args = map[string]interface{}{}
+	}
+	bt, err := json.Marshal(args)
 	if err != nil {
 		return err
 	}
+	paramFile := fmt.Sprintf("%s: %s", cuemodel.ParameterFieldName, string(bt))
+
 	var contextFile = strings.Builder{}
-	var paramFile = cuemodel.ParameterFieldName + ": {}"
-	if string(bt) != "null" {
-		paramFile = fmt.Sprintf("%s: %s", cuemodel.ParameterFieldName, string(bt))
-	}
 	// addon metadata context
 	metadataJSON, err := json.Marshal(a.addon.Meta)
 	if err != nil {
@@ -76,19 +82,19 @@ func (a addonCueTemplateRender) toObject(cueTemplate string, object interface{})
 	if err != nil {
 		return err
 	}
-	outputContent, err := out.LookupValue("output")
+	outputContent, err := out.LookupValue(renderOutputCuePath)
 	if err != nil {
 		return err
 	}
-	b, err := cueyaml.Encode(outputContent.CueValue())
-	if err != nil {
-		return err
-	}
-	return yaml.Unmarshal(b, object)
+	return outputContent.UnmarshalTo(object)
 }
 
 // generateAppFramework generate application from yaml defined by template.yaml or cue file from template.cue
 func generateAppFramework(addon *InstallPackage, parameters map[string]interface{}) (*v1beta1.Application, error) {
+	if len(addon.AppCueTemplate.Data) != 0 && addon.AppTemplate != nil {
+		return nil, ErrBothCueAndYamlTmpl
+	}
+
 	var app *v1beta1.Application
 	var err error
 	if len(addon.AppCueTemplate.Data) != 0 {
@@ -100,10 +106,7 @@ func generateAppFramework(addon *InstallPackage, parameters map[string]interface
 		app = addon.AppTemplate
 		if app == nil {
 			app = &v1beta1.Application{
-				TypeMeta: metav1.TypeMeta{APIVersion: "core.oam.dev/v1beta1", Kind: "Application"},
-				Spec: v1beta1.ApplicationSpec{
-					Components: []common2.ApplicationComponent{},
-				},
+				TypeMeta: metav1.TypeMeta{APIVersion: v1beta1.SchemeGroupVersion.String(), Kind: v1beta1.ApplicationKind},
 			}
 		}
 		if app.Spec.Components == nil {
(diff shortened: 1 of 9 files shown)


pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
+/*
+Copyright 2021 The KubeVela Authors.
```
wonderflow, line 2:
```suggestion
Copyright 2022 The KubeVela Authors.
```

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+			}
+		}
+		if app.Spec.Components == nil {
+			app.Spec.Components = []common2.ApplicationComponent{}
+		}
+	}
+
+	app.Name = addonutil.Addon2AppName(addon.Name)
+	// force override the namespace defined vela with DefaultVelaNS,this value can be modified by Env
+	app.SetNamespace(types.DefaultKubeVelaNS)
```
charlie0129, line 116: I was wondering how can this value be modified with env?

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return err
+	}
+	return yaml.Unmarshal(b, object)
+}
+
+// generateAppFramework generate application from yaml defined by template.yaml or cue file from template.cue
+func generateAppFramework(addon *InstallPackage, parameters map[string]interface{}) (*v1beta1.Application, error) {
+	var app *v1beta1.Application
+	var err error
```
wonderflow, line 93: if both appCueTemplate and appTemplate exist, you should report an error

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+)
+
+type addonCueTemplateRender struct {
+	addon     *InstallPackage
+	inputArgs map[string]interface{}
+}
+
+// This func can be used for addon render, supporting render app template and component.
+// Please notice the result will be stored in object parameter, so object must be a pointer type
+func (a addonCueTemplateRender) toObject(cueTemplate string, object interface{}) error {
```
wonderflow, line 50:
```suggestion
func (a addonCueTemplateRender) render(cueTemplate string) (*Application,error) {
```

This function can only output Application cause you have specific logic inside.

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+	bt, err := json.Marshal(a.inputArgs)
+	if err != nil {
+		return err
+	}
+	var contextFile = strings.Builder{}
+	var paramFile = cuemodel.ParameterFieldName + ": {}"
+	if string(bt) != "null" {
```
wonderflow, line 57: what if the `bt` is empty string?

pkg/addon/render.go
wonderflow: do we support `vela addon enable --namespace x` now ? It's a TODO?

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+	if len(addon.AppCueTemplate.Data) != 0 {
+		app, err = renderAppAccordingToCueTemplate(addon, parameters)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		app = addon.AppTemplate
+		if app == nil {
+			app = &v1beta1.Application{
+				TypeMeta: metav1.TypeMeta{APIVersion: "core.oam.dev/v1beta1", Kind: "Application"},
```
wonderflow, line 103: we shouldn't use hardcode string, import our standard lib.

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+
+	v, err := value.NewValue(contextFile.String(), nil, "")
+	if err != nil {
+		return err
+	}
+	out, err := v.LookupByScript(cueTemplate)
+	if err != nil {
+		return err
+	}
+	outputContent, err := out.LookupValue("output")
```
wonderflow, line 79: specify a const for `output`

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return err
+	}
+	b, err := cueyaml.Encode(outputContent.CueValue())
+	if err != nil {
+		return err
+	}
+	return yaml.Unmarshal(b, object)
```
wonderflow, line 87:
```suggestion
        return outputContent.UnmarshalTo(object)
```

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+	var paramFile = cuemodel.ParameterFieldName + ": {}"
+	if string(bt) != "null" {
+		paramFile = fmt.Sprintf("%s: %s", cuemodel.ParameterFieldName, string(bt))
+	}
+	// addon metadata context
+	metadataJSON, err := json.Marshal(a.addon.Meta)
+	if err != nil {
+		return err
+	}
+	contextFile.WriteString(fmt.Sprintf("context: metadata: %s\n", string(metadataJSON)))
```
wonderflow, line 65: will the metadataJSON be empty?

pkg/addon/render.go
wonderflow: why don't initilize the args, so it will at least become `{}`

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+			return nil, err
+		}
+	} else {
+		app = addon.AppTemplate
+		if app == nil {
+			app = &v1beta1.Application{
+				TypeMeta: metav1.TypeMeta{APIVersion: "core.oam.dev/v1beta1", Kind: "Application"},
+				Spec: v1beta1.ApplicationSpec{
+					Components: []common2.ApplicationComponent{},
+				},
```
wonderflow, line 106:
```suggestion
```

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	return app, nil
+}
+
+func attachPolicyForLegacyAddon(ctx context.Context, app *v1beta1.Application, addon *InstallPackage, args map[string]interface{}, k8sClient client.Client) error {
+	deployClusters, err := checkDeployClusters(ctx, k8sClient, args)
+	if err != nil {
+		return err
+	}
+	if isDeployToRuntime(addon) {
```
wonderflow, line 188:
```suggestion
	if !isDeployToRuntime(addon) {
	  return nil
	}
```

wonderflow: The rest code don't need to be nested after this

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+			clusterSelector := map[string]interface{}{
+				// empty labelSelector means deploy resources to all clusters
+				ClusterLabelSelector: map[string]string{},
+			}
+			properties, err := json.Marshal(clusterSelector)
+			if err != nil {
+				return err
+			}
+			policy := v1beta1.AppPolicy{
+				Name:       "deploy-addon-to-clusters",
```
wonderflow, line 200: we should clean up all these magic names, make them align.

pkg/addon/render.go
```diff
@@ -0,0 +1,278 @@
...
+			if err != nil {
+				return err
+			}
+			app.Spec.Policies = append(app.Spec.Policies, v1beta1.AppPolicy{
+				Name:       "specified-addon-clusters",
+				Type:       v1alpha1.TopologyPolicyType,
+				Properties: &runtime.RawExtension{Raw: body},
+			})
+			// addon should not contain workflow, this also update legacy addon with deploy2runtime steps
+			app.Spec.Workflow = nil
```
wonderflow, line 230: how do the users know this omit here? Will it become a black box bug?

pkg/addon/render.go
wangyikewxgm: This is by design. All addon associated application in same name is more convenient to manage. But user still can change this by set `DEFAULT_VELA_NS` to change this.

pkg/addon/render.go
wangyikewxgm: This func can be called by both render app and appCompnent, that's determined by in put object.

pkg/addon/render.go
wangyikewxgm: This is legacy logic, I just refactor this func to a single file, but I think it's fine to remove this line.

pkg/addon/render.go
wonderflow: if so ,you should make the cuepath `output` as an input, it can become much more common function

wangyikewxgm: Why we need change this the path? The appComponent and app are all in `output` block.

pkg/addon/render.go
wangyikewxgm: No,  empty meta will get `{"name":"","version":"","description":"","icon":"","invisible":false}`

pkg/addon/render.go
wonderflow: how would you deal with the case for "outputs"?

pkg/addon/render.go
wangyikewxgm: fixed, thanks.

pkg/addon/render.go
wangyikewxgm: Since the input parameter is cueTemplate file will contain one output and several outputs, and the outputs is unstructed object, I plan to add a input parameter value as `aliaOutputs []unstructed.unstructure` which fetch all outpus in this cueTemplate

wonderflow (approved):

Somefive (approved):

Issue #4326, opened 2022-07-04
[Feature] support edge application delivery


huiwq1990:
**Is your feature request related to a problem? Please describe.**

[openyurt application delivery proposal](https://github.com/huiwq1990/openyurt/blob/feat-app-delivery/docs/proposals/20220611-openyurt-application-delivery.md)

[openyurt application delivery proposal chinese version](https://github.com/huiwq1990/openyurt/blob/feat-app-delivery/docs/proposals/20220611-openyurt-application-delivery-cn.md)

**Describe the solution you'd like**

[use oam deploy edge applicaton chinese version](https://github.com/huiwq1990/openyurt/blob/feat-app-delivery/docs/proposals/20220611-openyurt-application-delivery-oam.md)

**Describe alternatives you've considered**

**Additional context**

- all information could be found in https://github.com/openyurtio/openyurt/pull/879
- the proposal has disscus with openyurt community members

FogDong:
It would be nice if you can comment on the community meeting agenda issue here kubevela/community#34 so that you can share your proposal and we can discuss this in the community meeting.

huiwq1990:
> It would be nice if you can comment on the community meeting agenda issue here [kubevela/community#34](https://github.com/kubevela/community/issues/34) so that you can share your proposal and we can discuss this in the community meeting.

We are happly to attend the meeting.

wonderflow:
Typo: subnet -> subset

huiwq1990:
@wonderflow  I think it works fine for deploy helm to a specific nodepool. If we want to deploy helm to all nodepools, did we should write override policy for all nodepools?

chivalryq:
Here is another solution with trait. There is one trait for webservice (will introduce trait for helm later) if you want to deploy them to two node pool.  
```cue
annotations: {
	type: "trait"
	annotations: {}
}
template: {
	// +patchStrategy=jsonMergePatch
	patch: {
		metadata: {
			annotations: {
				nodeSelectorLabel: "apps.openyurt.io/nodepool"
				nodeSelectorLabel: [context.replicaKey]
			}
		}
		if context.output.spec != _|_ && context.output.spec.template != _|_ {
			spec: template: metadata: annotations: {
				nodeSelectorLabel: "apps.openyurt.io/nodepool"
				nodeSelectorLabel: [context.replicaKey]
			}
		}
	}
	parameter:  null
}
```

Note that It needed a replication policy in Application. Replication policy do two things:
[...]

PR #4449, opened 2022-07-25
Feat: component replication

chivalryq:
## Description of your change

Fixes https://github.com/kubevela/kubevela/issues/4326
Added a replication policy. When using `deploy` workflow step, it could replicate components with keys.  e.g.

```yaml
kind: Application
metadata:
  name: app-replication-policy
spec:
  components:
    - name: hello-rep
      type: replica-webservice
      properties:
        image: crccheck/hello-world
      traits:
        - type: expose
          properties:
            port: [8000]
    - name: hello-no-rep
      type: webservice
      properties:
        image: crccheck/hello-world
      traits:
        - type: expose
          properties:
            port: [8001]

  policies:
    - name: comp-to-replicate
      type: override
      properties:
        selector: [ "hello-rep" ]
    - name: comp-not-replicate
      type: override
      properties:
        selector: [ "hello-no-rep" ]
    - name: target-default
      type: topology
      properties:
        clusters: [ "local" ]
    - name: replication-default
      type: replication
      properties:
        keys: ["beijing","hangzhou"]
        selector: ["hello-rep"]

  workflow:
    steps:
      - name: deploy-with-rep
        type: deploy
        properties:
          policies: ["comp-to-replicate","target-default","replication-default"]
      - name: deploy-without-rep
        type: deploy
        properties:
          policies: ["comp-not-replicate","target-default"]
```

### Apply Order of Policies

Now, `override`, `topology` and `replication` can work with `deploy` workflow step. If they are used together, the apply order is:
```text
topology -->  override -->  replication
```

###  Replication Spec

`replication` policy can **NOT** be used individually. Like topology, it need override to select target components to deploy.
The `selector` field of replication policy is used to choose which components to replicate. For example, you can specify three components with override policy and only replicate one of them. "comp2" and "comp3" will be dispatched without any replication

```yaml
  policies:
    - name: select-three
      type: override
      properties:
        selector: [ "comp1", "comp2", "comp3" ]
    - name: replication-default
      type: replication
      properties:
        keys: [ "beijing","hangzhou" ]
        selector: [ "comp1" ]
```

### Special Definition

Given that there aren't too many replication use case now. One want to use replication policy should modify the Component or Trait definition. A new context key `replicaKey` is provided when render the components. Here is an replica-webservice example in `test/e2e-test/testdata/definition/replica-webservice.yaml`

Fixes #

I have:


pkg/oam/util/helper.go
```diff
@@ -988,3 +988,11 @@ func (accessor *applicationResourceNamespaceAccessor) Namespace() string {
 func NewApplicationResourceNamespaceAccessor(appNs, overrideNs string) NamespaceAccessor {
 	return &applicationResourceNamespaceAccessor{applicationNamespace: appNs, overrideNamespace: overrideNs}
 }
+
+// ComponentNameWithReplicaKey helps generate component name with replica key
+func ComponentNameWithReplicaKey(componentName, replicaKey string) string {
+	if replicaKey == "" {
+		return componentName
+	}
+	return fmt.Sprintf("%s-%s", componentName, replicaKey)
```
wonderflow, line 997: what if a user define workload name in cue template?

pkg/workflow/providers/oam/apply.go
```diff
@@ -91,21 +90,17 @@ func (p *provider) RenderComponent(ctx wfContext.Context, v *value.Value, act wf
...
 	if err != nil {
 		return err
 	}
-	workload, traits, healthy, err := p.apply(*comp, patcher, clusterName, overrideNamespace, env)
+	workload, traits, healthy, err := p.apply(*comp, patcher, clusterName, overrideNamespace, "", env)
 	if err != nil {
 		return err
 	}
 
-	if mu != nil {
```
Somefive, line 104: Why remove this? It is used to prevent race problem.

pkg/workflow/providers/multicluster/deploy_test.go
```diff
@@ -83,31 +83,47 @@ func TestOverrideConfiguration(t *testing.T) {
 
 func TestApplyComponents(t *testing.T) {
 	r := require.New(t)
-	const n, m = 50, 5
+	const n, m = 30, 5
```
Somefive, line 86: Why reduce it to 30? I think this test is fast?

chivalryq: Misunderstood this number. I thought it should be divided by 3 exactly.

pkg/workflow/providers/oam/apply.go
chivalryq: All callers pass a nil to this param. It is not actually used.

apis/core.oam.dev/v1alpha1/policy_types.go
```diff
@@ -77,3 +79,15 @@ func (in SharedResourcePolicySpec) FindStrategy(manifest *unstructured.Unstructu
 	}
 	return false
 }
+
+// ReplicationPolicySpec defines the spec of replication policy
+type ReplicationPolicySpec struct {
+	Keys     []string `json:"keys,omitempty"`
```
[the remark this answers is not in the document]
chivalryq: Yes, see PR description. You mean add some comment for them?

pkg/oam/util/helper.go
chivalryq: It should be verified if the old style will overwrite the workload name as well.

pkg/policy/replication.go
```diff
@@ -0,0 +1,86 @@
...
+		replicationDecisions []v1alpha1.ReplicationDecision
+		compToRep            []string
+		compToKeep           = make(map[string]bool)
+	)
+	existReplicationPolicy := false
+	for _, policy := range policies {
+		if policy.Type == v1alpha1.ReplicationPolicyType {
+			existReplicationPolicy = true
+			replicateSpec := &v1alpha1.ReplicationPolicySpec{}
+			if err := pkgutils.StrictUnmarshal(policy.Properties.Raw, replicateSpec); err != nil {
```
charlie0129, line 59: This is prone to the same nil pointer dereference crashing problem when accessing `.Raw` described in #4473

apis/core.oam.dev/v1alpha1/policy_types.go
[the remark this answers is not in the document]
Somefive: Right, add comment for the fields here.

pkg/workflow/providers/oam/apply.go
Somefive: lgtm

pkg/workflow/providers/multicluster/deploy_test.go
chivalryq: n should be divided by 3 exactly to pass the designed test.

pkg/policy/replication.go
chivalryq: Fixed, thanks!

apis/core.oam.dev/v1alpha1/policy_types.go
```diff
@@ -77,3 +79,11 @@ func (in SharedResourcePolicySpec) FindStrategy(manifest *unstructured.Unstructu
 	}
 	return false
 }
+
+// ReplicationPolicySpec defines the spec of replication policy
+// Override policy should be used together with replication policy to select the deploy target components
+// ReplicationPolicySpec.Selector is the subset of selected components which will be replicated.
+type ReplicationPolicySpec struct {
```
Somefive, line 86: Add some comment for the Selector field that it will not drop components that not in the list. FYI, it might be a bit better if the description for each field is placed on the field, instead of the parent structure, but it is also acceptable here.

pkg/oam/labels.go
```diff
@@ -27,6 +27,8 @@ const (
 	LabelAppDeployment = "app.oam.dev/appDeployment"
 	// LabelAppComponent records the name of Component
 	LabelAppComponent = "app.oam.dev/component"
+	// LabelReplicaKey records the replica key of Component
+	LabelReplicaKey = "app.oam.dev/replica-key"
```
Somefive, line 31: Let's use `app.oam.dev/replicaKey` here.

Somefive (approved):
Great Job! LGTM

apis/core.oam.dev/v1alpha1/policy_types.go
```diff
@@ -77,3 +79,11 @@ func (in SharedResourcePolicySpec) FindStrategy(manifest *unstructured.Unstructu
 	}
 	return false
 }
+
+// ReplicationPolicySpec defines the spec of replication policy
+// Override policy should be used together with replication policy to select the deployment target components
+type ReplicationPolicySpec struct {
+	Keys []string `json:"keys,omitempty"`
+	// Selector is the subset of selected components which will be replicated.
+	Selector []string `json:"selector,omitempty"`
```
wonderflow, line 88: does the selector can only specify components?

Somefive: Currently yes. It works similar to the `selector` field in override policy. We can make further extension in the future.

Somefive:
FYI, @chivalryq , a CUE file might need to be added to provide OpenAPI Schema for VelaUX and auto-generate documentations for kubevela.io. We can do it in another PR.

pkg/workflow/providers/multicluster/deploy.go
```diff
@@ -82,6 +84,10 @@ func (executor *deployWorkflowStepExecutor) Deploy(ctx context.Context, policyNa
 	if err != nil {
 		return false, "", err
 	}
+	components, err = pkgpolicy.ReplicateComponents(policies, components)
```
wonderflow, line 87: what will happen if the replicaComponent don't use the replicaKey in the definition template?

apis/core.oam.dev/v1alpha1/policy_types.go
chivalryq: Yes, what else? Trait?

chivalryq:
> FYI, @chivalryq , a CUE file might need to be added to provide OpenAPI Schema for VelaUX and auto-generate documentations for kubevela.io. We can do it in another PR.

Got it.

pkg/workflow/providers/multicluster/deploy.go
Somefive: - If the definition template relies on replicaKey field, an error will be thrown for the incorrect use of that component/trait.
- If the definition template use the replicaKey field and has a fallback logic for handling the lack of replicaKey, it will be fine to use the definition with or without replicaKey.
- If the definition template does not use the replicaKey, nothing will be happened while using the replicaKey. (The component will be repeated dispatched but there is no side effect.)

apis/core.oam.dev/v1alpha1/policy_types.go
Somefive: Maybe other component selection ways, like selecting components through regex?

test/e2e-test/application_test.go
```diff
@@ -433,4 +433,38 @@ var _ = Describe("Application Normal tests", func() {
...
+					return k8sClient.Get(ctx, client.ObjectKey{Namespace: namespaceName, Name: svcName}, &svc)
+				},
+				time.Second*120, time.Millisecond*500).Should(BeNil())
+		}
+		verifySeriveDispatched("hello-rep-beijing")
+		verifySeriveDispatched("hello-rep-hangzhou")
+
+		By("Checking the services not replicated & application status")
+		verifyWorkloadRunningExpected("hello-no-rep", 1, "crccheck/hello-world")
+
```
wonderflow, line 468: please also check component didn't generate the workload `hello-rep`

pkg/workflow/providers/multicluster/deploy.go
chivalryq: Then it's depends on definitions. In definition it can check if `context.replicaKey` is nil. Just like I did in `replica-webservice.yaml`. But it's not recommended to mix component designed for replication and common ones.

apis/core.oam.dev/v1alpha1/policy_types.go
wonderflow: But it's a string, it lacks flexible?

pkg/workflow/providers/multicluster/deploy.go
wonderflow: what if one component use the replica key in the definition, but it's used in a fixed workload name? It will apply many times with the last config? What about next time? Does it cause any infinitely loop?

wonderflow (commented):
please also add definition cue file with example markdown file (and some necessary explanation of the usage.)

It's really tricky when use it in definition.

pkg/workflow/providers/multicluster/deploy.go
chivalryq: It is a problem. We can optimize it later. Like scanning the definition used when replicate.

Somefive: Yes, it will apply many times but will not cause infinite loop. Replication does not mutate configurations. Multiple apply will be the same.

apis/core.oam.dev/v1alpha1/policy_types.go
Somefive: Yes, it lacks flexibility. The consideration here is:
1. It aligns with the design of `override` policy, although `override` policy also lacks flexibility, but it is easy to understand.
2. It satisfy the scenario we currently encountered by far.
3. We could have more extension ways in the future. For now although it would be nice to make it more extensible, I do not have too much good points due to the lack of more advanced feature requests.

chivalryq:
ping @wonderflow definition cue and example md added.

wonderflow (approved):
great job

PR #4473, opened 2022-07-26
Fix: address vela-core crash due to empty policy properties

charlie0129:

The [topology doc](https://kubevela.net/docs/next/end-user/policies/references#specification-topology) shows that every property is not required.

<details>
<summary>Click to see app sample</summary>

```yaml
kind: Application
metadata:
  name: first-vela-app
spec:
  components: []
  policies:
    - name: target-default
      type: topology
      properties:
```
</details>

But if we actually don't provide any properties in topology, after applying the sample app, vela-core will 

- crash immediately due to nil pointer dereference
- will **NOT** recover after restarting, because the reconcilation starts again

So the vela-core will keep restarting over and over, and no Application can be created afterwards.

This pr fixes this problem, and similar ones.

> I encountered this problem when writing addons. Sometime it will output a empty policy due to the usage of conditional rendering policies using parameters (we can avoid this if we upgrade CUE above 0.3).

- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary. **(no permission to add labels)**

Tests are added to avoid regression

### DIFF
--- a/pkg/policy/topology.go
+++ b/pkg/policy/topology.go
@@ -18,6 +18,7 @@ package policy
 
 import (
 	"context"
+	"fmt"
 
 	prismclusterv1alpha1 "github.com/kubevela/prism/pkg/apis/cluster/v1alpha1"
 	"github.com/pkg/errors"
@@ -65,7 +66,10 @@ func GetPlacementsFromTopologyPolicies(ctx context.Context, cli client.Client, a
 	}
 	hasTopologyPolicy := false
 	for _, policy := range policies {
-		if policy.Type == v1alpha1.TopologyPolicyType && policy.Properties != nil {
+		if policy.Type == v1alpha1.TopologyPolicyType {
+			if policy.Properties == nil {
+				return nil, fmt.Errorf("topology should not have empty properties")
+			}
 			hasTopologyPolicy = true
 			topologySpec := &v1alpha1.TopologyPolicySpec{}
 			if err := utils.StrictUnmarshal(policy.Properties.Raw, topologySpec); err != nil {
--- a/pkg/policy/topology.go
+++ b/pkg/policy/topology.go
@@ -94,7 +94,6 @@ func GetPlacementsFromTopologyPolicies(ctx context.Context, cli client.Client, a
 				}
 			}
 		}
-
 	}
 	if !hasTopologyPolicy {
 		placements = []v1alpha1.PlacementDecision{{Cluster: multicluster.ClusterLocalName}}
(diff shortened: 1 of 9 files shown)


Somefive:
I agree that the controller must not panic even if there is no properties in topology policy.

But there are two things to notice:
1. Although all fields in topology policy are not required, it does not mean empty properties should be valid. Field clusters, field clusterLabelSelector are field namespace should have at least one to be set. Otherwise, the topology currently should be meaningless. BUT, it *might* be also reasonable to let the empty topology represent the default placement. (though it could be ambiguous)
2. I also wonder that is there any simlar bugs with other policy properties?

Somefive (approved):
Great catch.

Somefive:
> 2. I also wonder that is there any simlar bugs with other policy properties?

Seems `override` policy has similar problems.

pkg/policy/topology.go
```diff
@@ -94,6 +94,7 @@ func GetPlacementsFromTopologyPolicies(ctx context.Context, cli client.Client, a
 				}
 			}
 		}
+
```
Somefive, line 97: This line could be removed?

pkg/policy/topology.go
```diff
@@ -65,7 +65,7 @@ func GetPlacementsFromTopologyPolicies(ctx context.Context, cli client.Client, a
 	}
 	hasTopologyPolicy := false
 	for _, policy := range policies {
-		if policy.Type == v1alpha1.TopologyPolicyType {
+		if policy.Type == v1alpha1.TopologyPolicyType && policy.Properties != nil {
```
Somefive, line 68: I would recommend directly raise errors here to notify user that he/she misuses the topology policy.

pkg/policy/topology.go
charlie0129: yes

Somefive:
FYI, a desired and more user-friendly way to handle invalid policy properties is to let the webhook reject the requests. (It is also important for the controller to recognize the error, which is done in this PR).

charlie0129:
> FYI, a desired and more user-friendly way to handle invalid policy properties is to let the webhook reject the requests. (It is also important for the controller to recognize the error, which is done in this PR).

Good to know. I will take a look at the webhook in another PR.

pkg/policy/topology.go
charlie0129: that's better. accepted

Somefive: FYi, override policy also fit this change. :)

charlie0129: override policy also have this change now. 

since env-binding is deprecated, it is not changed.

PR #4493, opened 2022-07-28
Fix: check definition of addon whether is conflict 

wangyikewxgm:
1. fix bug check an  addon whether been used when disable.
2. add check definitions of an addon whether exist.
3. fix check addon system requirement special case bug: `actual: v1.5.0-beta.1` require:  `>=1.5.0` res: `ture`

Fixes #

I have:

### DIFF
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -1071,12 +1071,12 @@ func (h *Installer) dispatchAddonResource(addon *InstallPackage) error {
 	}
 
 	if !h.overrideDefs {
-		existDefs, err := checkConflictDefs(h.ctx, h.cli, defs, *app)
+		existDefs, err := checkConflictDefs(h.ctx, h.cli, defs, app.Name)
 		if err != nil {
 			return err
 		}
 		if len(existDefs) != 0 {
-			return fmt.Errorf(`definitions: %s in this addon already exist, if you want override them, Please use add "--override-defs" to re-enable or re-upgrade`, existDefs)
+			return produceDefConflictError(existDefs)
 		}
 	}
 
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -889,6 +889,7 @@ type Installer struct {
 	cache               *Cache
 	dc                  *discovery.DiscoveryClient
 	skipVersionValidate bool
+	overrideDefs        bool
 }
 
 // NewAddonInstaller will create an installer for addon
@@ -1067,6 +1068,16 @@ func (h *Installer) dispatchAddonResource(addon *InstallPackage) error {
 	defs, err := RenderDefinitions(addon, h.config)
 	if err != nil {
 		return errors.Wrap(err, "render addon definitions fail")
+	}
+
+	if !h.overrideDefs {
+		conflict, err := checkDefAlreadyExist(h.ctx, h.cli, defs)
+		if err != nil {
+			return err
+		}
+		if conflict {
+			return fmt.Errorf(`definitions in this addon already exist, if you want override them, Please use "vela addon enable <addon-name> --override-defs" to enable`)
+		}
 	}
 
 	schemas, err := RenderDefinitionSchema(addon)
@@ -1286,26 +1297,30 @@ func checkSemVer(actual string, require string) (bool, error) {
 	if len(require) == 0 {
 		return true, nil
 	}
-	smeVer := strings.TrimPrefix(actual, "v")
+	semVer := strings.TrimPrefix(actual, "v")
 	l := strings.ReplaceAll(require, "v", " ")
 	constraint, err := semver.NewConstraint(l)
 	if err != nil {
 		log.Logger.Errorf("fail to new constraint: %s", err.Error())
 		return false, err
 	}
-	v, err := semver.NewVersion(smeVer)
+	v, err := semver.NewVersion(semVer)
 	if err != nil {
-		log.Logger.Errorf("fail to new version %s: %s", smeVer, err.Error())
+		log.Logger.Errorf("fail to new version %s: %s", semVer, err.Error())
 		return false, err
 	}
 	if constraint.Check(v) {
 		return true, nil
 	}
 	if strings.Contains(actual, "-") && !strings.Contains(require, "-") {
-		smeVer := strings.TrimPrefix(actual[:strings.Index(actual, "-")], "v")
-		v, err := semver.NewVersion(smeVer)
+		semVer := strings.TrimPrefix(actual[:strings.Index(actual, "-")], "v")
+		if strings.Contains(require, ">=") && require[strings.Index(require, "=")+1:] == semVer {
+			// for case: `actual` is 1.5.0-beta.1 require is >=`1.5.0`
+			return false, nil
+		}
+		v, err := semver.NewVersion(semVer)
 		if err != nil {
-			log.Logger.Errorf("fail to new version %s: %s", smeVer, err.Error())
+			log.Logger.Errorf("fail to new version %s: %s", semVer, err.Error())
 			return false, err
 		}
 		if constraint.Check(v) {
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -1071,12 +1071,12 @@ func (h *Installer) dispatchAddonResource(addon *InstallPackage) error {
 	}
 
 	if !h.overrideDefs {
-		conflict, err := checkDefAlreadyExist(h.ctx, h.cli, defs)
+		existDefs, err := checkConflictDefs(h.ctx, h.cli, defs, *app)
 		if err != nil {
 			return err
 		}
-		if conflict {
-			return fmt.Errorf(`definitions in this addon already exist, if you want override them, Please use "vela addon enable <addon-name> --override-defs" to enable`)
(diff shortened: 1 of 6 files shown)


references/cli/addon.go
```diff
@@ -323,6 +326,7 @@ non-empty new arg
 	}
 	cmd.Flags().StringVarP(&addonVersion, "version", "v", "", "specify the addon version to upgrade")
 	cmd.Flags().BoolVarP(&skipValidate, "skip-version-validating", "s", false, "skip validating system version requirement")
+	cmd.Flags().BoolVarP(&overrideDefs, "override-defs", "o", false, "override already exist definitions within addon")
```
wonderflow, line 329:
```suggestion
	cmd.Flags().BoolVarP(&overrideDefs, "override", "o", false, "override existing definitions if conflict with those contained in this addon")
```

references/cli/addon.go
```diff
@@ -203,6 +205,7 @@ func NewAddonEnableCommand(c common.Args, ioStream cmdutil.IOStreams) *cobra.Com
 	cmd.Flags().StringVarP(&addonVersion, "version", "v", "", "specify the addon version to enable")
 	cmd.Flags().StringVarP(&addonClusters, types.ClustersArg, "c", "", "specify the runtime-clusters to enable")
 	cmd.Flags().BoolVarP(&skipValidate, "skip-version-validating", "s", false, "skip validating system version requirement")
+	cmd.Flags().BoolVarP(&overrideDefs, "override-defs", "o", false, "override already exist definitions within addon")
```
wonderflow, line 208:
```suggestion
	cmd.Flags().BoolVarP(&overrideDefs, "override", "o", false, "override existing definitions if conflict with those contained in this addon")
```

wonderflow: if you're desigining other override options in the future, I would suggest the option name to be "override-definitions" and with no short args("o").

pkg/addon/addon.go
```diff
@@ -1069,6 +1070,16 @@ func (h *Installer) dispatchAddonResource(addon *InstallPackage) error {
 		return errors.Wrap(err, "render addon definitions fail")
 	}
 
+	if !h.overrideDefs {
+		existDefs, err := checkConflictDefs(h.ctx, h.cli, defs, *app)
+		if err != nil {
+			return err
+		}
+		if len(existDefs) != 0 {
+			return fmt.Errorf(`definitions: %s in this addon already exist, if you want override them, Please use add "--override-defs" to re-enable or re-upgrade`, existDefs)
```
wonderflow, line 1079:
```suggestion
			return fmt.Errorf(`definitions: %s in this addon already exist, if you want override them, Please use arguments "--override" to enable`, existDefs)
```

I wonder if the case really exist when users are upgrading addons?

wonderflow: Shall we update an addon with new definitions add?

pkg/addon/addon.go
```diff
@@ -1286,26 +1297,30 @@ func checkSemVer(actual string, require string) (bool, error) {
...
 		return true, nil
 	}
 	if strings.Contains(actual, "-") && !strings.Contains(require, "-") {
-		smeVer := strings.TrimPrefix(actual[:strings.Index(actual, "-")], "v")
-		v, err := semver.NewVersion(smeVer)
+		semVer := strings.TrimPrefix(actual[:strings.Index(actual, "-")], "v")
+		if strings.Contains(require, ">=") && require[strings.Index(require, "=")+1:] == semVer {
+			// for case: `actual` is 1.5.0-beta.1 require is >=`1.5.0`
+			return false, nil
+		}
```
wonderflow, line 1320: are there any new test cases to cover it?

wonderflow: sorry ,found it

pkg/addon/addon_test.go
```diff
@@ -945,6 +945,11 @@ func TestCheckSemVer(t *testing.T) {
 			require: ">=v1.2.4-beta.3",
 			res:     false,
 		},
+		{
+			actual:  "1.5.0-beta.2",
+			require: ">=1.5.0",
+			res:     false,
+		},
```
wonderflow, line 952:
```suggestion
		},
		{
			actual:  "1.5.0-alpha.2",
			require: ">=1.5.0",
			res:     false,
		},		
```

pkg/addon/addon.go
wangyikewxgm: I think addon developer cannot avoid this situation in the future.

wonderflow: can we notify what exsiting addon contain this definition? It's user friendly:

···
definition: %s in this addon already exist in %s ...
···

pkg/addon/utils.go
```diff
@@ -409,3 +439,21 @@ func generateAnnotation(meta *Meta) map[string]string {
 func isErrorCueRenderPathNotFound(err error, path string) bool {
 	return err.Error() == fmt.Sprintf("var(path=%s) not exist", path)
 }
+
+func checkConflictDefs(ctx context.Context, k8sClient client.Client, defs []*unstructured.Unstructured, app v1beta1.Application) ([]string, error) {
```
wonderflow, line 443:
```suggestion
func checkConflictDefs(ctx context.Context, k8sClient client.Client, defs []*unstructured.Unstructured, appName string) ([]string, error) {
```

pkg/addon/utils.go
```diff
@@ -409,3 +439,21 @@ func generateAnnotation(meta *Meta) map[string]string {
...
+	var res []string
+	for _, def := range defs {
+		err := k8sClient.Get(ctx, client.ObjectKeyFromObject(def), def)
+		if err == nil {
+			owner := metav1.GetControllerOf(def)
+			if owner == nil || owner.Kind != v1beta1.ApplicationKind || owner.Name != app.Name {
+				// if addon not belong to an addon or addon name is another one, we should put them in result
```
wonderflow, line 450: the condition judgement is really clever! But the explanation here is confusing.

```suggestion
			if owner == nil || owner.Kind != v1beta1.ApplicationKind || owner.Name != app.Name {
				// owner is nil means the definition don't belong to any addons, it must conflict with us.
				// owner.Kind != v1beta1.ApplicationKind or name not equal, it also means the owner is not this addon.
```

BTW, it is possible that Definition has more than 1 owners , what if  you're checking the wrong one?

wangyikewxgm: Have consider this situation, because addon-app always is the controller owner, which can only be one.

wonderflow: I think it's better to check if this application is one of the addon

pkg/addon/utils.go
wangyikewxgm: Maybe we need use  other information（annotations or labels）about this app in the future？

pkg/addon/addon.go
wangyikewxgm: Yes, when conflict, will produce an error to tell what defs has exist.

pkg/addon/utils.go
wonderflow: please  follow the KISS (keep it simple and stupid) principle

pkg/addon/utils.go
wangyikewxgm: I think if the addon-app is controller owner of this def, that's mean the addon will modify at any time. Is it  fine to update it directly?

pkg/addon/utils.go
```diff
@@ -409,3 +440,37 @@ func generateAnnotation(meta *Meta) map[string]string {
...
+
+func produceDefConflictError(conflictDefs map[string]string) error {
+	if len(conflictDefs) == 0 {
+		return nil
+	}
+	var errorInfo string
+	for _, s := range conflictDefs {
+		errorInfo += s
+	}
+	errorInfo += "if you want override them, Please use arguments '--override' to enable \n"
```
wonderflow, line 474:
```suggestion
	errorInfo += "if you want override them, please use argument '--override' to enable \n"
```

wonderflow (approved):
LGTM after nits

pkg/addon/addon.go
wangyikewxgm: The print result is 
```
definition: helm already exist and not belong to any addon 
definition: kustomize in this addon already exist in fluxcd 
if you want override them, Please use arguments '--override' to enable

File: references/cli/addon.go [hunk 9]
```diff
@@ -203,6 +205,7 @@ func NewAddonEnableCommand(c common.Args, ioStream cmdutil.IOStreams) *cobra.Com
 	cmd.Flags().StringVarP(&addonVersion, "version", "v", "", "specify the addon version to enable")
 	cmd.Flags().StringVarP(&addonClusters, types.ClustersArg, "c", "", "specify the runtime-clusters to enable")
 	cmd.Flags().BoolVarP(&skipValidate, "skip-version-validating", "s", false, "skip validating system version requirement")
+	cmd.Flags().BoolVarP(&overrideDefs, "override-definitions", "", false, "override existing definitions if conflict with those contained in this addon")
 	return cmd
```
Somefive, line 209: What is this args for? Is there any cases for it?

Somefive (approved):

PR #4496, opened 2022-07-28
Fix: empty health policy do not check object existence

### LINKED ISSUE #4470: Component HealthCheck do not check the existence of the resource if there is no healthPolicy in ComponentDefinition
**Describe the bug**

The health check will not check if the resource exists if the component definition has no healthPolicy. TraitDefinition also has this problem.

**To Reproduce**

**Expected behavior**

**Screenshots**

**KubeVela Version**

v1.5.0-beta.4

**Cluster information**

**Additional context**

Somefive:

When dispatching resources for application component, 
- `apply-component`: directly apply component, then check health
- `deploy`: first check if component is healthy, if not healthy, apply the component

Therefore, it apply component fails, the `apply-component` will always block the workflow step (which is correct). But for `deploy` step, if apply component fails but health check passes (this is weird but could happen if we do not have healthPolicy in component definition), the workflow step will fail for the first time (since RT has no records) but then succeeded directly for the second time.

This PR fixes it by enforcing object existence check even if healthPolicy is empty.

Besides, there is another bug fixed in this PR. When trait resources uses different namespace than the component and the application, the health check will fail due to incorrect namespace settings.

Fixes #4470

I have:

### DIFF
--- a/pkg/appfile/appfile.go
+++ b/pkg/appfile/appfile.go
@@ -109,7 +109,7 @@ func (wl *Workload) EvalStatus(ctx process.Context, cli client.Client, accessor 
 // EvalHealth eval workload health check
 func (wl *Workload) EvalHealth(ctx process.Context, client client.Client, accessor util.NamespaceAccessor) (bool, error) {
 	// if health of template is not set or standard workload is managed by trait always return true
-	if wl.FullTemplate.Health == "" || wl.SkipApplyWorkload {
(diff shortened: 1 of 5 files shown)


wonderflow (commented):
any tests to cover this case?

Somefive:
> any tests to cover this case?

Will add it later

wonderflow (approved):

FogDong (approved):

Issue #3300, opened 2022-02-18
trait env and storage cannot be used at the same time, using one of them ok


yuchang01:
**Describe the bug**

```
E0218 17:14:41.513828   22564 instance.go:80] "failed to have the workload/trait unstructured" err="cue: marshal error: spec.template.spec.containers.0.env: from source" Definition="apiVersion: \"apps/v1\"\nkind:       \"Deployment\"\nspec: {\n\tselector: {\n\t\tmatchLabels: {\n\t\t\t\"app.oam.dev/component\": \"platuml-server\"\n\t\t}\n\t}\n\ttemplate: {\n\t\tspec: {\n\t\t\t// +patchKey=name\n\t\t\tcontainers: [{\n\t\t\t\tname: \"platuml-server\"\n\t\t\t\t// +patchKey=name\n\t\t\t\tenv:   _|_ // conflicting list lengths: invalid value 0 (out of bound int & >=1)\n\t\t\t\timage: \"plantuml/plantuml-server:jetty\"\n\t\t\t\tports: [{\n\t\t\t\t\tname:          \"http\"\n\t\t\t\t\tcontainerPort: 8080\n\t\t\t\t\tprotocol:      *\"TCP\" | \"UDP\" | \"SCTP\"\n\t\t\t\t}, ...]\n\t\t\t\t// +patchKey=name\n\t\t\t\tvolumeDevices: []\n\t\t\t\t// +patchKey=name\n\t\t\t\tvolumeMounts: [{\n\t\t\t\t\tname:      \"emptydir-jvm\"\n\t\t\t\t\tmountPath: \"/jvm\"\n\t\t\t\t}]\n\t\t\t}, ...{\n\t\t\t\t// +patchKey=name\n\t\t\t\tenv: []\n\t\t\t\t// +patchKey=name\n\t\t\t\tvolumeDevices: []\n\t\t\t\t// +patchKey=name\n\t\t\t\tvolumeMounts: [{\n\t\t\t\t\tname:      \"emptydir-jvm\"\n\t\t\t\t\tmountPath: 
[...]

PR #4523, opened 2022-08-02
Fix: storage patch env

Somefive:

Fixes storage patch env bug in #3300

P.S. This PR does not ensure all the possible cases in storage works. It only fixes the bug in #3300.

I have:

### DIFF
--- a/pkg/cue/model/sets/operation.go
+++ b/pkg/cue/model/sets/operation.go
@@ -115,6 +115,9 @@ func listMergeProcess(field *ast.Field, key string, baseList, patchList *ast.Lis
 			kmaps[fmt.Sprintf(key, blit.Value)] = patchList.Elts[i]
 		}
 		if !foundPatch {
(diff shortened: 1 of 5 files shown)


barnettZQG (approved):
LGTM

pkg/cue/model/sets/operation.go
```diff
@@ -115,7 +115,7 @@ func listMergeProcess(field *ast.Field, key string, baseList, patchList *ast.Lis
 			kmaps[fmt.Sprintf(key, blit.Value)] = patchList.Elts[i]
 		}
 		if !foundPatch {
-			return
+			continue
```
FogDong, line 118: Can you add a test in application_controller?

Somefive: An e2e test is added.

Somefive: UT is also added.

FogDong (approved):

PR #4570, opened 2022-08-07
Feat: provide Stage field for TraitDefinition

### LINKED ISSUE #3888: [Feature] Configmap not corrected mounted when deployment and configmap are updated together

Though we have dependencies across different components, there are still resource problems along within one component.

For example, if we have one deployment as the main workload of a component, then we use ConfigMap as a trait to mount. If we update the deployment and the configmap at the same time, the configmap might be updated after deployment pod is created. In this case, the deployment pod will use the old configmap data.

**Additional context**

### LINKED ISSUE #4143: [Feature] Dispatch resource within a component sequentially
In our scenario, we define several traits to generate `Custom Resource` for completing specific tasks. 
- In some cases, trait-A need to dispatch before trait-B and workload.
- In k8s's case, when we modify pod parameter and configmap parameter, the cm should be dispatched before deployment.
So, the ability that dispatch resource within a component sequentially is necessary.

We want to add a field `inOrder` align with `dependsOn`, and serially dispatch resource between steps and dispatch resource within steps in parallel.
There is an example for `inOrder` as follows
```yaml
kind: Application
metadata:
  name: fourierapp03
  namespace: fourier
spec:
  components:
    - name: fourierapp03-comp-01
      type: webservice
      inOrder:
        - step:
            - config
        - step:
            - workload
            - cpuscaler
      properties:
        image: nginx:latest
        ports:
          - expose: true
            port: 80
            protocol: TCP
        imagePullPolicy: IfNotPresent
      traits:
        - type: cpuscaler
          properties:
            min: 1
            max: 10
        - type: config
          properties:
            refConfigs:
              - name: fourierapp03-comp-01
                encrypt: false
                mountPath: /mount/test/cm/tomcat.yaml
 
  policies:
    - name: fourierapp03-comp-01
      type: topology
      properties:
        clusters: [ "local" ]
        namespace: fourier01
 
  workflow:
    steps:
      - name: step-group
        type: step-group
        subSteps:
          - name: fourierapp03-deploy-01
            type: deploy
            properties:
              policies: [ "fourierapp03-comp-01" ]
```

ZhongsJie96:
Signed-off-by: ZhongsJie <zhongsjie@cmbchina.com>

Fixes #4143 
Fixes #3888 

I have:

### DIFF
--- a/apis/core.oam.dev/v1beta1/core_types.go
+++ b/apis/core.oam.dev/v1beta1/core_types.go
@@ -158,7 +158,28 @@ type TraitDefinitionSpec struct {
 	// ControlPlaneOnly defines which cluster is dispatched to
 	// +optional
 	ControlPlaneOnly bool `json:"controlPlaneOnly,omitempty"`
+
+	// Stage defines the stage information to which this trait resource processing belongs.
+	// Currently, PreDispatch and PostDispatch are provided, which are used to control resource
+	// pre-process and post-process respectively.
+	// +optional
+	Stage StageType `json:"stage,omitempty"`
 }
+
+// StageType describes how the manifests should be dispatched.
+// Only one of the following stage types may be specified.
+// If none of the following types is specified, the default one
+// is DefaultDispatch.
+type StageType string
+
+const (
+	// PreDispatch means that pre dispatch for manifests
+	PreDispatch StageType = "PreDispatch"
+	// DefaultDispatch means that default dispatch for manifests
+	DefaultDispatch StageType = "DefaultDispatch"
+	// PostDispatch means that post dispatch for manifests
+	PostDispatch StageType = "PostDispatch"
+)
 
 // TraitDefinitionStatus is the status of TraitDefinition
 type TraitDefinitionStatus struct {
--- a/pkg/controller/core.oam.dev/v1alpha2/application/apply.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/apply.go
@@ -188,6 +188,17 @@ func removeResources(elements []common.ClusterObjectReference, index int) []comm
 	return elements[:len(elements)-1]
 }
 
+// getServiceStatus get specified component status
+func (h *AppHandler) getServiceStatus(svc common.ApplicationComponentStatus) common.ApplicationComponentStatus {
+	for i := range h.services {
+		current := h.services[i]
+		if current.Equal(svc) {
+			return current
+		}
+	}
+	return svc
+}
+
 // addServiceStatus recorde the whole component status.
 // reconcile run at single threaded. So there is no need to consider to use locker.
 func (h *AppHandler) addServiceStatus(cover bool, svcs ...common.ApplicationComponentStatus) {
@@ -197,7 +208,7 @@ func (h *AppHandler) addServiceStatus(cover bool, svcs ...common.ApplicationComp
 		found := false
 		for i := range h.services {
 			current := h.services[i]
-			if current.Name == svc.Name && current.Env == svc.Env && current.Namespace == svc.Namespace && current.Cluster == svc.Cluster {
+			if current.Equal(svc) {
 				if cover {
 					h.services[i] = svc
 				}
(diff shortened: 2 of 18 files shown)


apis/core.oam.dev/v1beta1/core_types.go
```diff
@@ -160,6 +160,9 @@ type TraitDefinitionSpec struct {
 	// ControlPlaneOnly defines which cluster is dispatched to
 	// +optional
 	ControlPlaneOnly bool `json:"controlPlaneOnly,omitempty"`
+	// PreApply represents dispatch manifests in advance
+	// +optional
+	PreApply bool `json:"preApply,omitempty"`
```
wonderflow, line 165: we need a better name

wonderflow (changes requested):
I think the scenario is worth of fix while the code and API is not good enough

apis/core.oam.dev/v1beta1/core_types.go
Somefive: How about `ApplyPriority` and let it be a number. Default to be 0. Higher number indicates earlier apply.

wonderflow: I doesn't work. A number is a fixed value only work on fixed comparison, while definition itself is a dynamic module can be used in any composition.

ZhongsJie96: I think it should just control which needs to be dispatch in advance and make sure it's healthy. If the concept of priority is introduced, the complexity will increase greatly. New features require knowledge of all other priorities.

Somefive: > A number is a fixed value only work on fixed comparison, while definition itself is a dynamic module can be used in any composition.

But how could you describe a dynamic comparison? At least `ApplyPriority` is a superset for `preApply`. Although I agree it has limitations.

> If the concept of priority is introduced, the complexity will increase greatly.

Yes. I agree with it. The implementation of priority will be more complex. But then we need to make it clear what this PR wants to solve and the boundary here.

For example, in current KubeVela application, even if without the PR here, user can still use Workflow / DependsOn to make resource dispatch in order. Therefore, AFAIS, this PR is not a necessity for features (The ConfigMap case could be split into `creating ConfigMap` and `mounting ConfigMap`), as discussed in previous PR. One of the big progress here is that we are moving the responsibility for resource dispatch order controlling from end-user to system-operators (who design traits and components). Then I believe it will be okay for asking system-operators to gain `the knowledge of all other priorities`.

When you want to `make sure it's healthy`, this can be difficult in KubeVela model since you need `healthPolicy` in definitions. Therefore, you have lines of codes for the implementation for this feature in this PR, which I think although useful but might be better if it can be generalized into more scenarios.

Anyway, `Priority` is just a recommendation implementation from my perspective. The code here generally LGTM. Thanks for the contribution.

oeular:
Here I suggest some comparison about the performance between the original parallel mode and the one @ZhongsJie96  provides.

apis/core.oam.dev/v1beta1/core_types.go
```diff
@@ -160,8 +160,29 @@ type TraitDefinitionSpec struct {
...
+// Only one of the following stage types may be specified.
+// If none of the following types is specified, the default one
+// is DefaultDispatch.
+type StageType string
+
+const (
+	// PreDispatch means that pre dispatch for manifests
+	PreDispatch StageType = "PreDispatch"
+	// DefaultDispatch means that default dispatch for manifests
+	DefaultDispatch StageType = "DefaultDispatch"
```
wonderflow, line 181:
```suggestion
	DefaultDispatch StageType = ""
```

Mark it empty or `Default`?

pkg/oam/labels.go
```diff
@@ -233,6 +233,9 @@ const (
 
 	// AnnotationAddonDefinitionBondCompKey indicates the definition in addon bond component.
 	AnnotationAddonDefinitionBondCompKey = "addon.oam.dev/bind-component"
+
+	// AnnotationTraitStageType is used to specify stage type
+	AnnotationTraitStageType = "trait.oam.dev/stage-type"
```
wonderflow, line 238:
```suggestion
	AnnotationTraitStageType = "trait.oam.dev/stage"
```

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -186,6 +186,17 @@ func removeResources(elements []common.ClusterObjectReference, index int) []comm
 	return elements[:len(elements)-1]
 }
 
+// getServiceStatus get specified component status
+func (h *AppHandler) getServiceStatus(svc common.ApplicationComponentStatus) common.ApplicationComponentStatus {
```
Somefive, line 190: What is this function for?

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -214,10 +225,101 @@ func (h *AppHandler) ProduceArtifacts(ctx context.Context, comps []*types.Compon
 	return h.createResourcesConfigMap(ctx, h.currentAppRev, comps, policies)
 }
 
+// collectPartialHealthStatus will collect the health status for partial manifests
+func (h *AppHandler) collectPartialHealthStatus(ctx context.Context, wl *appfile.Workload, appRev *v1beta1.ApplicationRevision, options DispatchOptions) (bool, error) {
```
Somefive, line 229: So you copied the collectHealthStatus function here?

Somefive: suggest make some common logic into separated functions for re-use

apis/core.oam.dev/v1beta1/core_types.go
ZhongsJie96: According to implement logic, the relevant information needs to be used in the future, and it can be considered to be changed to `default`.

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
ZhongsJie96: When executing according to the stage, a part of the state information needs to be kept. If the information is completely overwritten, the status will be lost during the execution.

pkg/oam/labels.go
ZhongsJie96: This is a good suggestion, I will modify it.

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
ZhongsJie96: I will refactor this part of the logic.

pkg/controller/core.oam.dev/v1alpha2/application/generator.go
```diff
@@ -358,15 +508,24 @@ func (h *AppHandler) applyComponentFunc(appParser *appfile.Parser, appRev *v1bet
...
-			if trait.Spec.ControlPlaneOnly && trait.Name == readyTrait.GetLabels()[oam.TraitTypeLabel] {
+			if trait.Name != readyTrait.GetLabels()[oam.TraitTypeLabel] {
+				continue
+			}
+
+			stageType := trait.Spec.Stage
+			if len(stageType) == 0 {
+				stageType = v1beta1.DefaultDispatch
+			}
+			util.AddAnnotations(traits[index], map[string]string{oam.AnnotationTraitStageType: string(stageType)})
```
wonderflow, line 524: why do we need to add the deploy stage inside the annotation, what's used for

pkg/controller/core.oam.dev/v1alpha2/application/generator.go
```diff
@@ -358,15 +508,24 @@ func (h *AppHandler) applyComponentFunc(appParser *appfile.Parser, appRev *v1bet
...
+
+			if trait.Spec.ControlPlaneOnly {
 				oam.SetCluster(traits[index], "local")
 				traits[index].SetNamespace(appRev.GetNamespace())
-				break
```
wonderflow, line 369: why omit the break ? \cc @wangyikewxgm

pkg/controller/core.oam.dev/v1alpha2/application/generator.go
ZhongsJie96: After rendering, the stage information needs to be obtained again to support the logic.

pkg/controller/core.oam.dev/v1alpha2/application/generator.go
ZhongsJie96: Type judgment has been added before this.

wonderflow: I  think `break` and `continue` are different logic, the `break` will execute only once in the loop while the continue means we can execute more than once if condition matches. Maybe we still need the `break` here.

Somefive (approved):
Great Job!

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
```diff
@@ -216,80 +227,124 @@ func (h *AppHandler) ProduceArtifacts(ctx context.Context, comps []*types.Compon
...
-	status.Traits = traitStatusList
+		idx := 0
+		for _, _trait := range status.Traits {
+			if _trait.Type != tr.Name {
+				status.Traits[idx] = _trait
+				idx++
+			}
+		}
+		status.Traits = status.Traits[:idx]
+	}
```
wonderflow, line 346:
```suggestion
		var oldStatus []..
		for _, _trait := range status.Traits {
			if _trait.Type != tr.Name {
				oldStatus = append(oldStatus, _trait)
			}
		}
		status.Traits = oldStatus
	}
```

Make it more clear

pkg/controller/core.oam.dev/v1alpha2/application/dispatcher.go
```diff
@@ -0,0 +1,215 @@
...
+	for _, readyTrait := range readyTraits {
+		var (
+			traitType = readyTrait.GetLabels()[oam.TraitTypeLabel]
+			stageType = DefaultDispatch
+			err       error
+		)
+		switch {
+		case traitType == definition.AuxiliaryWorkload:
+		case traitType != "":
+			if strings.Contains(traitType, "-") {
```
wonderflow, line 174: why do we need this check logic?

wonderflow: can you give some concrete examples?

pkg/controller/core.oam.dev/v1alpha2/application/dispatcher.go
```diff
@@ -0,0 +1,215 @@
...
+				}
+				status, isHealth, err := h.collectHealthStatus(ctx, wl, appRev, options.OverrideNamespace, skipWorkload,
+					ByTraitType(readyTraits, options.Traits))
+				if err != nil {
+					return false, errors.WithMessage(err, "CollectHealthStatus")
+				}
+				if options.Stage < DefaultDispatch {
+					status.Healthy = false
+					if status.Message == "" {
+						status.Message = "waiting for previous stage healthy"
```
wonderflow, line 151: why you don't check the `isHealth` and set the message directly?

pkg/controller/core.oam.dev/v1alpha2/application/apply.go
ZhongsJie96: Yes, this is a way to improve readability

pkg/controller/core.oam.dev/v1alpha2/application/dispatcher.go
ZhongsJie96: The `isHealth` here is only for the current resource. I don't think it is reasonable to directly set the  status that should belong to the workload. For example, the health status of the workload should be maintained during the post phase.

pkg/controller/core.oam.dev/v1alpha2/application/dispatcher.go
wonderflow: I got it, but why it's not handled outside? AFAIK, it should be trimed when getting the traitdefition spec detial before the application rendered.

pkg/controller/core.oam.dev/v1alpha2/application/dispatcher.go
wonderflow: OK, that's fine

wonderflow (approved):

PR #4592, opened 2022-08-10
Feat: support use update instead of patch for apply

### LINKED ISSUE #4469: [Feature] Support updating resources through replace

KubeVela controller now uses three-way-merge-patch to calculate patch for resource updates, and use MergePatch/StrategyMergePatch to update resources.

However, it will let the StateKeep allow merges of unintended data, like malicious sidecar being injected into pods. This is due to the patchStrategy defined by [Kubernetes API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#podspec-v1-core).

If we support using `replace` requests to update resources through specifying some policies in applications, it can help users to enforce strict consistency between KubeVela application and managed resources.

Add a `resource-update-strategy` policy (or other name) and let the Applicator of resources to use replace requests when find this strategy. It can be configured in application level or resource level, like using `rules` in `garbage-collect` policy.

**Additional context**

Somefive:

Fixes #4469

I have:

Notice:
1. Resources like Service have separate controller to modify its spec. Therefore, for Service resource, this configuration will not take effect.
2. KubeVela legacy rollout will not work when this flag is enabled.

### DIFF
--- a/pkg/utils/apply/apply.go
+++ b/pkg/utils/apply/apply.go
@@ -170,15 +170,26 @@ func (a *APIApplicator) Apply(ctx context.Context, desired client.Object, ao ...
 		return nil
 	}
 
-	loggingApply("patching object", desired)
-	patch, err := a.patcher.patch(existing, desired, applyAct)
-	if err != nil {
-		return errors.Wrap(err, "cannot calculate patch by computing a three way diff")
+	switch {
+	case utilfeature.DefaultMutableFeatureGate.Enabled(features.ApplyResourceByUpdate) && isUpdatableResource(desired):
+		loggingApply("updating object", desired)
+		desired.SetResourceVersion(existing.GetResourceVersion())
+		var options []client.UpdateOption
+		if applyAct.dryRun {
+			options = append(options, client.DryRunAll)
+		}
+		return errors.Wrapf(a.c.Update(ctx, desired, options...), "cannot update object")
+	default:
+		loggingApply("patching object", desired)
+		patch, err := a.patcher.patch(existing, desired, applyAct)
+		if err != nil {
+			return errors.Wrap(err, "cannot calculate patch by computing a three way diff")
+		}
+		if applyAct.dryRun {
+			return errors.Wrapf(a.c.Patch(ctx, desired, patch, client.DryRunAll), "cannot patch object")
+		}
+		return errors.Wrapf(a.c.Patch(ctx, desired, patch), "cannot patch object")
 	}
-	if applyAct.dryRun {
-		return errors.Wrapf(a.c.Patch(ctx, desired, patch, client.DryRunAll), "cannot patch object")
-	}
-	return errors.Wrapf(a.c.Patch(ctx, desired, patch), "cannot patch object")
 }
 
 func generateRenderHash(desired client.Object) (string, error) {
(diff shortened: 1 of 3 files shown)


pkg/utils/apply/apply.go
```diff
@@ -397,3 +407,14 @@ func DryRunAll() ApplyOption {
 		return nil
 	}
 }
+
+// isUpdatableResource check whether the resource is updatable
+// Resource like v1.Service cannot unset the spec field (the ip spec is filled by service controller)
+func isUpdatableResource(desired client.Object) bool {
```
wonderflow, line 413: I think I've written such kinds of code before, but I forget where.

pkg/utils/apply/apply.go
```diff
@@ -170,6 +170,16 @@ func (a *APIApplicator) Apply(ctx context.Context, desired client.Object, ao ...
 		return nil
 	}
 
+	if utilfeature.DefaultMutableFeatureGate.Enabled(features.ApplyResourceByUpdate) && isUpdatableResource(desired) {
```
wonderflow, line 173: I will use switch in this code.

switch{
case utilfeature.DefaultMutableFeatureGate.Enabled(features.ApplyResourceByUpdate) && isUpdatableResource(desired):
default:
}

Somefive: But I think there will be no other cases here, even in the future. Is there any plan for more strategies other than Update&Patch?

Somefive: A switch case is used in isUpdatableResource.

pkg/utils/apply/apply.go
Somefive: I think that part of code is removed. Let me find it in history archives.

Somefive: Are you referring to https://github.com/kubevela/kubevela/blob/8de36ff2950b3c757f912e81d2fc4be3ea4b60af/pkg/controller/core.oam.dev/v1alpha2/application/dispatch/dispatch.go#L285?

pkg/utils/apply/apply.go
wonderflow: A switch will be more clear even there's only two cases

pkg/utils/apply/apply.go
wonderflow: Yes

wonderflow: Just make sure no duplicate code

pkg/utils/apply/apply.go
Somefive: OK, agree.

pkg/utils/apply/apply.go
Somefive: No, that code only exists in release 1.1, not on master. Besides, they have different working mechanism. (Although use similar names)

fourierr:
We will use update instead of three-way-patch later?

Somefive:
> We will use update instead of three-way-patch later?

Currently not. Update has side-effects. It also increase the network traffic. But update can be more straightforward especially for state-keep. This feature provides an alternative usage for users who wants to use update instead of patch.

If this feature is needed in many cases, we might move it from a controller feature to an application feature or even resource feature, just like `apply-once` and `garbage-collect`. But for now, it is still only reserved for use in limited scenario.

pkg/utils/apply/apply.go
Somefive: Done

wonderflow (approved):

PR #4595, opened 2022-08-10
Feat: align resource selector across policies

### LINKED ISSUE #4594: ApplyOnce ResourceSelector ambiguous and not aligned with other ResourceSelector
**Describe the bug**

The implementation for FindStrategy in apply-once policy is confusing. 

https://github.com/kubevela/kubevela/blob/436ff916b65fcf3e57e7043904343e9db6bbcec1/apis/core.oam.dev/v1alpha1/applyonce_types.go#L62

The major problem is that it misuses `componentNames` and `resourceNames`.

The minor problem is that this selector is not aligned with selector in other policies (garbage-collect & shared-resource).

https://github.com/kubevela/kubevela/blob/436ff916b65fcf3e57e7043904343e9db6bbcec1/apis/core.oam.dev/v1alpha1/garbagecollect_types.go#L74

I plan to refactor this part of code. Is there any special case that might break existing usage outside test cases?
/cc @fourierr 

**To Reproduce**

**Expected behavior**

**Screenshots**

**KubeVela Version**

**Cluster information**

**Additional context**

Somefive:

Fixes #4594

I have:

/cc @fourierr

### DIFF
--- a/apis/core.oam.dev/v1alpha1/applyonce_types.go
+++ b/apis/core.oam.dev/v1alpha1/applyonce_types.go
@@ -53,15 +53,7 @@ func (in ApplyOncePolicySpec) FindStrategy(manifest *unstructured.Unstructured) 
 		return nil
 	}
 	for _, rule := range in.Rules {
-		match := func(src []string, val string) (found bool) {
-			for _, _val := range src {
-				found = found || _val == val
-			}
-			return val != "" && found
-		}
-		if (match(rule.Selector.CompNames, manifest.GetName()) && match(rule.Selector.ResourceTypes, manifest.GetKind())) ||
-			(rule.Selector.CompNames == nil && match(rule.Selector.ResourceTypes, manifest.GetKind()) ||
-				(rule.Selector.ResourceTypes == nil && match(rule.Selector.CompNames, manifest.GetName()))) {
+		if rule.Selector.Match(manifest) {
 			return rule.Strategy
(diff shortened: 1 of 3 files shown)


fourierr:
It covers my usage completely. How about exposing an operator to provide the ability of chossing "and" , "or" for user.

Somefive:
> It covers my usage completely. How about exposing an operator to provide the ability of chossing "and" , "or" for user.

I actually think for a while, but this exposure might be a bit complex for user. The AND case is more reasonable I think, as you proposed in apply-once previously. The OR case on the other hand, can be implemented through configuring multiple rules, which is a bit more complex but available. By far, I haven't seen cases use OR actually. We can leave this for future implementation if we encountered lots of such usage.

fourierr:
> > It covers my usage completely. How about exposing an operator to provide the ability of chossing "and" , "or" for user.
> 
> I actually think for a while, but this exposure might be a bit complex for user. The AND case is more reasonable I think, as you proposed in apply-once previously. The OR case on the other hand, can be implemented through configuring multiple rules, which is a bit more complex but available. By far, I haven't seen cases use OR actually. We can leave this for future implementation if we encountered lots of such usage.

👌

fourierr (approved):
lgtm

barnettZQG (approved):
LGTM

FogDong (approved):

PR #4613, opened 2022-08-12
Feat: support compression in ResourceTracker

### LINKED ISSUE #4392: [Feature] Support resourcetracker can be stored after compressing
For saving the capacity of the etcd cluster in the hub cluster, some resources are expected to less size.

For the resourcetrackers with more versions and  within all resources specifications, compressing them before saving will save much capacity, and improving usability of the hub cluster.

For implematation, we can set a parameter --feature-gate when core-controller work

Somefive:

1. Add FeatureFlag `CompressResourceTracker` for enabling this feature. When enabled, newly created ResourceTracker will use compression, legacy ones will not be changed (even updated).
2. Add `compression` field for ResourceTracker. If `.compression.type` is set to `gzip`, the ManagedResources will be compressed with gzip and stored. While retrieved, it will be decompressed.
3. A mock test is set up in unit test. 2000 similar resources are added to the ResourceTracker, the compression rate turns out to be 12% while the marshal time cost for json goes up to 4~8x.

Another example is that enabling the latest fluxcd addon to 3 clusters, before compression the size of ResourceTracker is 600kb(root)+98.9kb(versioned) while after compression the size of ResourceTracker is 70.2kb(root)+5.15kb(versioned). Compression rate ~ 10.7%.

Fixes #4392

I have:


FogDong (approved):

apis/core.oam.dev/v1beta1/resourcetracker_types.go
```diff
@@ -69,9 +70,61 @@ const (
...
+		cpy.ManagedResources = nil
+		cpy.Compression.Data = data
+		tmp.Alias = (*Alias)(cpy)
+	default:
+		return nil, compression.NewUnsupportedCompressionTypeError(string(in.Compression.Type))
+	}
+	return json.Marshal(tmp.Alias)
+}
+
+// UnmarshalJSON .
```
wonderflow, line 107:
```suggestion
// UnmarshalJSON wil unmarshal resource tracker spec according to the compression type, if feature enabled, it will unmarshal data from compression data. Note: this is not actually json unmarhsal but re-use the framework function.
```

apis/core.oam.dev/v1beta1/resourcetracker_types.go
```diff
@@ -69,9 +70,61 @@ const (
...
+	Compression           ResourceTrackerCompression `json:"compression,omitempty"`
+}
+
+// ResourceTrackerCompression the compression for ResourceTracker ManagedResources
+type ResourceTrackerCompression struct {
+	Type compression.Type `json:"type,omitempty"`
+	Data string           `json:"data,omitempty"`
+}
+
+// MarshalJSON .
```
wonderflow, line 85:
```suggestion
// MarshalJSON wil encode resource tracker spec according to the compression type, if feature enabled, it will encode data to compression data. Note: this is not actually standard json marshal process but re-use the framework function.
```

pkg/features/controller_features.go
```diff
@@ -58,6 +58,10 @@ const (
 
 	// AuthenticateApplication enable the authentication for application
 	AuthenticateApplication featuregate.Feature = "AuthenticateApplication"
+	// CompressResourceTracker enable the compression for ResourceTracker. It can be useful if you have large application
+	// that needs to dispatch lots of resources or large resources (like CRD or huge ConfigMap), which at the cost of
+	// slower processing speed due to the extra overhead for compression and decompression
+	CompressResourceTracker featuregate.Feature = "CompressResourceTracker"
```
wonderflow, line 64:
```suggestion
	CompressResourceTracker featuregate.Feature = "GzipResourceTracker"
```

Somefive: Done

apis/core.oam.dev/v1beta1/resourcetracker_types.go
Somefive: Done

apis/core.oam.dev/v1beta1/resourcetracker_types.go
Somefive: Done

barnettZQG (approved):
LGTM

charlie0129:
I have a question: why not consider some more efficient and faster compression algorithms other than `gzip`, e.g. `zstd`? 

Compression will be much faster and decompression too, with no additional CPU cycles.

Most of the time, the barrier of going from `gzip` to `zstd` is backwards compatibility, but we don't have such problems here.

Somefive:
> I have a question: why not consider some more efficient and faster compression algorithms other than `gzip`, e.g. `zstd`?
> 
> Compression will be much faster and decompression too, with no additional CPU cycles.
> 
> Most of the time, the barrier of going from `gzip` to `zstd` is backwards compatibility, but we don't have such problems here.

It is okay to make extension for that. Good idea.

Issue #4596, opened 2022-08-10
[Feature]  Enhanced horizontal scaling of app controller


fourierr:
**Is your feature request related to a problem? Please describe.**
Due to the leader-election of the controller, when there are thousands of application CRs in the hubcluster,
the new application will be delayed for updating longer and longer time.

**Describe the solution you'd like**
For this issue, i have made some experimental solution.
We want to dispatch several deployment of app controller in hubcluster.
Firstly, enabling the leader election and set `leader-election-resource-name` with different name e.g sharding-1, sharding-2.
Secondly, we add a label tag for every application CR, e.g `app.oam.dev/sharding=sharding-1`
Thirdly, add a filter to controller's eventFilter. Only the value of label tag is equal to `leader-election-resource-name`, the controller will handle the application CR.

The final effect is: application CR with label value of `sharding-1` will only be processed by app controller with `leader-selection-resource-name` value of `sharding-1`

Somefive:
Great! This solution looks good and is similar to what I have been thinking for sharding. Would you like to make a PR for that?

fourierr:
> 



> Great! This solution looks good and is similar to what I have been thinking for sharding. Would you like to make a PR for that?

Of course, i will make a PR.

wonderflow:
lgtm

wonderflow:
Just FYI, there's another solution for this issue, which don't need change the controller code: https://github.com/openkruise/controllermesh

wonderflow:
Sharding is supported now, please check https://github.com/kubevela/kubevela/pull/5360 

It will be released in 1.8.0 and you can already enjoy this feature in https://github.com/kubevela/kubevela/releases/tag/v1.8.0-alpha.1

Closing

PR #4618, opened 2022-08-13
Feat: enhanced horizontal scaling of app controller

fourierr:

Fixes https://github.com/kubevela/kubevela/issues/4596

Add sharding for app controller, and legacy app cr will be processed by sharding-1. 
I have:


cmd/core/main.go
```diff
@@ -98,6 +98,8 @@ func main() {
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "",
 		"Determines the namespace in which the leader election configmap will be created.")
+	flag.StringVar(&leaderElectionName, "leader-election-name", "sharding-1",
```
wonderflow, line 101: 1. by default it should be empty.
2. `leader-election-name` -> `app-sharding-key` \c @Somefive  any good suggestion?

cmd/core/main.go
```diff
@@ -230,7 +232,7 @@ func main() {
 	}
 	ctrl.SetLogger(klogr.New())
 
-	leaderElectionID := util.GenerateLeaderElectionID(types.KubeVelaName, controllerArgs.IgnoreAppWithoutControllerRequirement)
+	leaderElectionID := util.GenerateLeaderElectionID(types.KubeVelaName+"-"+leaderElectionName, controllerArgs.IgnoreAppWithoutControllerRequirement)
```
wonderflow, line 235: I wonder if it will cause incompatible?

wonderflow (changes requested):
1. how this PR be tested?
2. please also add some test cases
3. please add docs for this feature

cmd/core/main.go
fourierr: I think it's a leader-election configmap name and be compatible.

cmd/core/main.go
fourierr: > 1. by default it should be empty.

I think for a while, if default is empty, and we want to have three sharding, the sharding will contains "", "sharding-1", "sharding-2", it's a little strange. I'm not sure whether I'm proper.

fourierr:
> 1. how this PR be tested?
> 2. please also add some test cases
> 3. please add docs for this feature

I plan to deploy app controller with `sharding-1`, and create three app cr, 
- the first cr was default and no mark, 
- the second cr was marked with `sharding-1`. 
- the third cr was marked with `sharding-2`. 
The expected result is:
- the first cr was processed successful,
- the second cr was also processed successful,
- the third cr has not been processed.
How about this testing case? @wonderflow

cmd/core/main.go
fourierr: After thinking again, most users will only use one sharding. So It is better to default to empty.

fourierr:
At first, I wanted to set multiple shardings through `helm -- set`, but in most cases, only one app controller was deployed at the beginning. Later, it was found that it could not be satisfied, and then the app controller was gradually added sharding. In this case, helm cannot implement it, so I plan to increase the number of shardings gradually through `vela sharding add`. How about `vela sharding add`?

wonderflow:
The sharding controller only consumes those apps with sharding annotation. What's blocked on helm installation?

Somefive:
I have some design questions for this PR. I understand that sharding could help make controller horizontally scalable. But technically, I think `controller-requirement` can already satisfy the need of sharding. The only problem is the shared leader election lock cause multiple controllers cannot work concurrently.

Therefore, I think it might not be necessary to add `app.oam.dev/sharding` for applications. It should be able to reuse `app.oam.dev/controller-version-require`.

wonderflow:
> I have some design questions for this PR. I understand that sharding could help make controller horizontally scalable. But technically, I think `controller-requirement` can already satisfy the need of sharding. The only problem is the shared leader election lock cause multiple controllers cannot work concurrently.
> 
> Therefore, I think it might not be necessary to add `app.oam.dev/sharding` for applications. It should be able to reuse `app.oam.dev/controller-version-require`.

I don't think we should re-sue `app.oam.dev/controller-version-require` in this case, reasons:

1. Users can't understand the annotation, this is mainly for inner apiserver to use.
2. In some cases, it could be possible that we both these two Annotations, for example, a sharding key for some BU while there're two versions of controller to rollout.

Somefive:
> > I have some design questions for this PR. I understand that sharding could help make controller horizontally scalable. But technically, I think `controller-requirement` can already satisfy the need of sharding. The only problem is the shared leader election lock cause multiple controllers cannot work concurrently.
> > Therefore, I think it might not be necessary to add `app.oam.dev/sharding` for applications. It should be able to reuse `app.oam.dev/controller-version-require`.
> 
> I don't think we should re-sue `app.oam.dev/controller-version-require` in this case, reasons:
> 
> 1. Users can't understand the annotation, this is mainly for inner apiserver to use.
> 2. In some cases, it could be possible that we both these two Annotations, for example, a sharding key for some BU while there're two versions of controller to rollout.

I see. Then the implementation is relatively reasonable for me.

fourierr:
> The sharding controller only consumes those apps with sharding annotation. What's blocked on helm installation?
> 
> The default behavior is the default controller don't consume apps with sharding annotation.

I didn't think there was any blocks to helm.

cmd/core/main.go
wonderflow: if leaderElectionName is empty, the "-" should be omitted

pkg/controller/core.oam.dev/v1alpha2/application/application_controller.go
```diff
@@ -622,3 +627,7 @@ func (r *Reconciler) matchControllerRequirement(app *v1beta1.Application) bool {
 	}
 	return true
 }
+
+func (r *Reconciler) matchAppSharding(object client.Object) bool {
+	return object.GetLabels()[oam.LabelAppSharding] == pflag.Lookup("leader-election-name").Value.String()
```
wonderflow, line 632: don't look up every time, the performance could be better

fourierr: > don't look up every time, the performance could be better

ok, i will fix it in recently

Somefive:
Duplicated with #4705

PR #4625, opened 2022-08-16
Fix: open basic lit fails lookup path to KubeVela 1.5

Somefive:

When env exists in webservice component, the use of labels + storage traits will eliminate the env field in webservice.

For example:

```yaml
kind: Application
metadata:
  name: test
spec:
  components:
    - name: test
      properties:
        cmd:
          - sleep
          - -c
          - "86400"
        env:
          - name: testKey
            value: testValue
        image: busybox
        imagePullPolicy: IfNotPresent
      traits:
        - properties:
            key: val
          type: labels
        - properties:
            pvc:
              - accessModes:
                  - ReadWriteOnce
                mountOnly: false
                mountPath: /data
                name: lvhyca
                resources:
                  requests:
                    storage: 4096Mi
                storageClassName: default
                volumeMode: Filesystem
          type: storage
      type: webservice
```

The reason is that when `labels` trait patches, it uses `json-merge-patch`, which convert CUE into JSON and patch, then convert JSON to CUE. However, while converting JSON to CUE, the previous OpenBasicLit function will do the following conversion.
```cue
// before
"y": {
  "x": ["a", "b"]
}
// after
{
  "y": *{
    "x": *["a", "b"] | [...]
  } | _
}
``` 

This will make the `lookup` function unable to find `.y.x`. This PR fixes it by doing the following conversion.

```cue
// before
"y": {
  "x": ["a", "b"]
}
// after
"y": {
  "x": ["a", "b", ...]
  ...
}
```

Notice this problem also exists for CUE 0.4, therefore this fix need to be cherry-picked to master branch. But since we have made CUE upgrade from KubeVela v1.5 -> v1.6, there will be conflicts. So this fix cannot be automatically backported, need to be manually cherry-picked.

I have:

### DIFF
--- a/pkg/cue/model/sets/operation.go
+++ b/pkg/cue/model/sets/operation.go
@@ -22,6 +22,7 @@ import (
 
 	"cuelang.org/go/cue"
 	"cuelang.org/go/cue/ast"
+	"cuelang.org/go/cue/format"
 	"cuelang.org/go/cue/parser"
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/pkg/errors"
@@ -361,7 +362,7 @@ func jsonMergePatch(base cue.Value, patch cue.Value) (string, error) {
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to merge base value and patch value by JsonMergePatch")
 	}
-	output, err := OpenBaiscLit(string(merged))
+	output, err := openJSON(string(merged))
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to parse open basic lit for merged result")
 	}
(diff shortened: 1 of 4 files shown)


pkg/cue/model/sets/operation.go
```diff
@@ -386,9 +387,38 @@ func jsonPatch(base cue.Value, patch cue.Value) (string, error) {
...
+	f, err := parser.ParseFile("-", data, parser.ParseComments)
+	if err != nil {
+		return "", err
+	}
+	ast.Walk(f, func(node ast.Node) bool {
+		field, ok := node.(*ast.Field)
+		if ok {
+			v := field.Value
+			switch lit := v.(type) {
+			case *ast.StructLit:
```
FogDong, line 407: Better to add a pre-check to avoid adding an ellipsis if the last item is ellipisis.

FogDong (approved):

PR #4630, opened 2022-08-17
Feat: support zstd compression in resourcetracker

### LINKED ISSUE #4392: [Feature] Support resourcetracker can be stored after compressing

charlie0129:

This pr continues #4613 
Fixes #4392

[`zstd`](https://github.com/facebook/zstd) is a modern real-time compression algorithm developed by Meta, which is fast and efficient. In terms of speed, it is typically 2x as fast as the stdlib deflate/gzip in its fastest mode. The compression ratio compared to stdlib is around level 3, but usually 3x as fast.

It is not only faster, but also compresses data better. We can get size-reductions while not giving up much on performance by adopting `zstd`.

### Benchmarks

There are benchmarks which use real CRDs to simulate a real use-case in `resourcetracker_types_test.go`. The user must have large ResourceTrackers if they want to use compression, so we load real CRDs to test, instead of generating fake data.

To generalize:
- Almost no performance penalty when dealing with large RTs (>1MB)
- Small penalties (10% - 50% slower) when dealing with small RTs (20KB - 1MB)

We have some very promising results (tested on `linux/amd64`, AMD Ryzen7 4750U @1.7Ghz-forced):

---

#### 2MB ResourceTracker

It is almost as fast as no compression. This result is consistent, not a coincidence.

image

```diff
Compressed Size:
  uncompressed: 2131455 bytes   100.00%
  gzip:         273058 bytes    12.81%
  zstd:         171582 bytes    8.05%
Marshal Time:
  no compression: 38635130 ns   1.00x
  gzip:           100722238 ns  2.61x
  zstd:           39643965 ns   1.03x
```

---

#### 500KB ResourceTracker

image

```diff
Compressed Size:
  uncompressed: 508291 bytes    100.00%
  gzip:         61230 bytes     12.05%
  zstd:         34666 bytes     6.82%
Marshal Time:
  no compression: 9424367 ns    1.00x
  gzip:           26402197 ns   2.80x
  zstd:           10853595 ns   1.15x
```

---

#### 37KB ResourceTracker

image

```diff
Compressed Size:
  uncompressed: 37275 bytes     100.00%
  gzip:         6162 bytes      16.53%
  zstd:         6878 bytes      18.45%
Marshal Time:
  no compression: 878362 ns     1.00x
  gzip:           2202297 ns    2.51x
  zstd:           1342592 ns    1.53x

```

---

I have:

### DIFF
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -111,7 +111,7 @@ optimize:
 
 ##@param featureGates.enableLegacyComponentRevision if disabled, only component with rollout trait will create component revisions
 ##@param featureGates.gzipResourceTracker if enabled, resourceTracker will be compressed using gzip before being stored
-##@param featureGates.zstdResourceTracker if enabled, resourceTracker will be compressed using zstd before being stored. It is both faster and more efficient than gzip. If both gzip and zstd are enabled, zstd will be used.
+##@param featureGates.zstdResourceTracker if enabled, resourceTracker will be compressed using zstd before being stored. It is much faster and more efficient than gzip. If both gzip and zstd are enabled, zstd will be used.
 featureGates:
   enableLegacyComponentRevision: false
   gzipResourceTracker: false
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -111,7 +111,7 @@ optimize:
 
 ##@param featureGates.enableLegacyComponentRevision if disabled, only component with rollout trait will create component revisions
 ##@param featureGates.gzipResourceTracker if enabled, resourceTracker will be compressed using gzip before being stored
-##@param featureGates.zstdResourceTracker if enabled, resourceTracker will be compressed using zstd before being stored
+##@param featureGates.zstdResourceTracker if enabled, resourceTracker will be compressed using zstd before being stored. It is both faster and more efficient than gzip. If both gzip and zstd are enabled, zstd will be used.
 featureGates:
   enableLegacyComponentRevision: false
   gzipResourceTracker: false
--- a/charts/vela-core/values.yaml
+++ b/charts/vela-core/values.yaml
@@ -110,10 +110,12 @@ optimize:
   enableResourceTrackerDeleteOnlyTrigger: true
 
 ##@param featureGates.enableLegacyComponentRevision if disabled, only component with rollout trait will create component revisions
-##@param featureGates.gzipResourceTracker if enabled, resourceTracker will be compressed before stored
+##@param featureGates.gzipResourceTracker if enabled, resourceTracker will be compressed using gzip before being stored
+##@param featureGates.zstdResourceTracker if enabled, resourceTracker will be compressed using zstd before being stored
 featureGates:
   enableLegacyComponentRevision: false
   gzipResourceTracker: false
+  zstdResourceTracker: false
 
 ## @section MultiCluster parameters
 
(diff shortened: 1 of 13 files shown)


charts/vela-core/values.yaml
```diff
@@ -110,10 +110,12 @@ optimize:
   enableResourceTrackerDeleteOnlyTrigger: true
 
 ##@param featureGates.enableLegacyComponentRevision if disabled, only component with rollout trait will create component revisions
-##@param featureGates.gzipResourceTracker if enabled, resourceTracker will be compressed before stored
+##@param featureGates.gzipResourceTracker if enabled, resourceTracker will be compressed using gzip before being stored
+##@param featureGates.zstdResourceTracker if enabled, resourceTracker will be compressed using zstd before being stored
 featureGates:
   enableLegacyComponentRevision: false
   gzipResourceTracker: false
+  zstdResourceTracker: false
```
barnettZQG, line 118: If users enable both the gzip and zsyd not carefully, it is strange. How are priorities determined?

charlie0129: `zstd` have higher priority.

barnettZQG: Add some comments about this in the values file.

charlie0129: Good idea. Sure

makefiles/e2e.mk
```diff
@@ -17,7 +17,20 @@ e2e-setup-core-wo-auth:
 
 .PHONY: e2e-setup-core-w-auth
 e2e-setup-core-w-auth:
-	helm upgrade --install --create-namespace --namespace vela-system --set image.pullPolicy=IfNotPresent --set image.repository=vela-core-test --set applicationRevisionLimit=5 --set dependCheckWait=10s --set image.tag=$(GIT_COMMIT) --wait kubevela ./charts/vela-core --set authentication.enabled=true --set authentication.withUser=true --set authentication.groupPattern=* --set featureGates.gzipResourceTracker=true
```
wonderflow, line 20: We're missing the gzip test now

charlie0129: Since we only have three possible compressions, I added the gzip test to another e2e target, for now. If we have other compression methods in the future, we may use other methods.

Somefive: I suggest not doing so. Leaving the e2e-setup-core-wo-auth without any compression method. It is okay to omit one of the compression tests in the e2e test.

Somefive:
Great Job! Generally LGTM. I suggest removing the compression option in `e2e-setup-core-wo-auth` test.

makefiles/e2e.mk
charlie0129: I will remove it. But I wonder is there any specific reason for having no compression in `e2e-setup-core-wo-auth`?

Somefive: By default, we do not switch on the ResourceTracker compression (neither zstd nor gzip). So in general, for most existing KubeVela users, they are using ResourceTracker without compression, and this might persist for some time (until we switch on the compression in the future, but there are downstream codes affected by this, like vela-prism). Although logically it is okay to set up the tests all with compression, I think it is also necessary to keep tests without compression to align with the real usage. The `e2e-setup-core-w-auth` (the name is actually confusing now), enforces some advanced features in multi-cluster e2e test. So it is okay to put new features and mechanism in this test.

Somefive (approved):
Good Job!

wonderflow (approved):
great job

PR #4689, opened 2022-09-02
Feat: highlight the Yaml text of `vela top`

### LINKED ISSUE #4685: [Feature] Highlight the Yaml text of `vela top`

I  want to add the feature that the Yaml text will be highlighted like this:

image

About the highlight color I don't have some good idea, if someone give me  some suggestions, I will be very appreciative.

**Additional context**

Hanmengnan:

image

Fixes #4685

I have:

### DIFF
--- a/references/cli/top/component/table.go
+++ b/references/cli/top/component/table.go
@@ -56,6 +56,7 @@ func (t *Table) Start() {
 
 // Stop table component
 func (t *Table) Stop() {
+	t.Clear()
 }
 
 // Hint return key action menu hints of the component
--- a/references/cli/top/component/table.go
+++ b/references/cli/top/component/table.go
@@ -56,7 +56,6 @@ func (t *Table) Start() {
 
 // Stop table component
 func (t *Table) Stop() {
-	t.Clear()
 }
 
 // Hint return key action menu hints of the component
--- a/references/cli/top/config/color.go
+++ b/references/cli/top/config/color.go
@@ -63,4 +63,10 @@ const (
 	PodSucceededPhase = "[purple::]"
 	// PodFailedPhase is pod failed phase text color
 	PodFailedPhase = "[red::]"
+	// YamlKeyColor is the color of key in highlighted yaml
+	YamlKeyColor = "mediumturquoise"
+	// YamlColonColor is the color of colon in highlighted yaml
+	YamlColonColor = "white"
+	// YamlValueColor is the color of value in highlighted yaml
+	YamlValueColor = "orangered"
 )
--- a/references/cli/top/model/managed_resource.go
+++ b/references/cli/top/model/managed_resource.go
@@ -18,6 +18,7 @@ package model
 
 import (
 	"context"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/oam-dev/kubevela/pkg/velaql/providers/query"
--- a/references/cli/top/model/managed_resource.go
+++ b/references/cli/top/model/managed_resource.go
@@ -18,7 +18,6 @@ package model
 
 import (
 	"context"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/oam-dev/kubevela/pkg/velaql/providers/query"
--- a/references/cli/top/view/application_view.go
+++ b/references/cli/top/view/application_view.go
@@ -43,7 +43,6 @@ func (v *ApplicationView) Name() string {
 func (v *ApplicationView) Init() {
 	v.CommonResourceView.Init()
 	v.SetTitle(fmt.Sprintf("[ %s ]", v.Title()))
-	v.BuildHeader()
 	v.bindKeys()
 }
 
@@ -74,6 +73,7 @@ func (v *ApplicationView) InitView(ctx context.Context, app *App) {
 
 // Update refresh the content of body of view
 func (v *ApplicationView) Update() {
+	v.BuildHeader()
 	v.BuildBody()
 }
 
--- a/references/cli/top/view/application_view_test.go
+++ b/references/cli/top/view/application_view_test.go
@@ -66,7 +66,7 @@ func TestApplicationView(t *testing.T) {
 
 	t.Run("stop", func(t *testing.T) {
 		appView.Stop()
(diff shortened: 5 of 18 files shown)


references/cli/top/component/table.go
```diff
@@ -56,7 +56,6 @@ func (t *Table) Start() {
 
 // Stop table component
 func (t *Table) Stop() {
-	t.Clear()
```
wonderflow, line 59: why removed

wonderflow (commented):
can you please show the effect?

references/cli/top/component/table.go
Hanmengnan: First, the logic of rendering table is that the rendering of table header and table body are separate, because the table header is same of the same kind of resource table, so header will be rendered only once.

I think i can clear the body when table is stopped, so i clear it. But i find table header is cleared in the same time.  In the next time rendering, the header will disappear. So I decide to remove the `Clear`.


But just now, I find i also need `Clear`, because if i don't clear the table body, in the case  that the older table is longer  and the latter table is short, the older table's data will be left in new table.  

Thanks for helping me fix a bug! 😆

wonderflow: can we add a test to prevent it happen next time especially when other maintainer contribute this part?

Hanmengnan: A test about logic here? 

For example, the content of table is empty when it is stopped?

Hanmengnan: Maybe your idea is add more test case to prevent bugs happened? Ok, i will  add more test cases as i can.

chivalryq:
Can this project helps? It seems to be more mature. https://github.com/andreazorzetto/yh

Hanmengnan:
> Can this project help? It seems to be more mature. https://github.com/andreazorzetto/yh

Thanks for your advice,  principle of this package is print colorized text by using linux ANSI color control code. But vela top use the tview package, which has build a higher level package above linux ANSI, and if we want to make text to be colorized on tview component, we should use the tview color principle. 

In fact, i have try the package you mentioned, this result is that it doesn't work. 😭

chivalryq:
> > Can this project help? It seems to be more mature. [andreazorzetto/yh](https://github.com/andreazorzetto/yh)
> 
> Thanks for your advice, principle of this package is print colorized text by using linux ANSI color control code. But vela top use the tview package, which has build a higher level package above linux ANSI, and if we want to make text to be colorized on tview component, we should use the tview color principle.
> 
> In fact, i have try the package you mentioned, this result is that it doesn't work. 😭

Got it. Highlight YAML by regex match is also acceptable. It's a pity we can not reuse an existing project.

charlie0129:
regex-matching can do the job. But in case you want a better way to manipulate yaml than regex-matching, I have an idea.

You can build the yaml AST tree, then do the highlighting job based on the syntax components. This is most accurate and can stick to the yaml spec.

There is a project to help: https://github.com/goccy/go-yaml

Hanmengnan:
> regex-matching can do the job. But in case you want a better way to manipulate yaml than regex-matching, I have an idea.
> 
> You can build the yaml abstract syntax tree, then do the highlighting job based on the syntax components. This is most accurate and can stick to the yaml spec.
> 
> There is a project to build the abstract syntax tree for you: https://github.com/goccy/go-yaml

Thanks for your advice, if i don't misunderstand the usage of the package, it need to unmarshal the yaml to a struct or by yaml path to get the value.  By this way, we can get the value and manipulate each key value pair more finely. 

But i have a puzzle, it seems that because all kinds of resource may have some differences, so i need define a struct for each kind of resource, or define a common struct to get the common key value pair such as "apiVersion".

For the first choice, the cost maybe a little lot. And for the latter, it may not get the better effect.

Hanmengnan:
By the way, about the highlight method, I refer to the implementation of K9s, which also uses regex-matching.

yangsoon:
> By the way, about the highlight method, I refer to the implementation of K9s, which also uses regex-matching.

Great Job, I think the regex-matching way is enough.

charlie0129:
> regex-matching can do the job

**I also think the regex one is good enough for this purpose.** Because I see your pursuit of good code, the AST method is just a FYI one just in case you want to do it, which will cost much time.

> it need to unmarshal the yaml to a struct or by yaml path to get the value

To be honest, I haven't used that package myself, but I don't think so. Generally, it is similar to how a language compiler frontend works. For example, we have code to analyze the CUE components using CUE AST in KubeVela before v1.6 https://github.com/kubevela/kubevela/blob/17872f97050bb0d039b6f7f0f7245fe7f714fe9f/pkg/cue/model/sets/walk.go#L47-L111 The process is like: you build the AST, then you traverse the tree to know what and where each syntax component is. Based on the syntax component type, you do the highlighting. AST is at the core of every language parser. Most IDEs without a language server do accurate highlighting in this way. It is better than token-based syntax highlighting, worse than semantic highlighting (which requires a language server).

Hanmengnan:
> > regex-matching can do the job
> 
> **I also think the regex one is good enough for this purpose.** Because I see your pursuit of good code and you said the pkg @chivalryq mentioned will not work, the AST method is just a FYI one just in case you want to do it, which will cost much time.
> 
> > it need to unmarshal the yaml to a struct or by yaml path to get the value. But i have a puzzle, it seems that because all kinds of resource may have some differences, so i need define a struct for each kind of resource, or define a common struct to get the common key value pair such as "apiVersion".
> 
> To be honest, I haven't used that package myself, but I don't think this is how highlighting works. Generally, it is similar to how a language compiler frontend works. For example, we have code to analyze the CUE components using CUE AST in KubeVela before v1.6
> 
> https://github.com/kubevela/kubevela/blob/17872f97050bb0d039b6f7f0f7245fe7f714fe9f/pkg/cue/model/sets/walk.go#L47-L111
> 
> The process is like: you build the AST, then you traverse the tree to know what and where each syntax component is. Based on the syntax component type, you do the highlighting. AST is at the core of every language parser. Most IDEs without a language server do accurate highlighting in this way. It is better than token-based syntax highlighting, worse than semantic highlighting (which requires a language server).

Yes thanks, I got it.

yangsoon (approved):

Somefive (approved):

PR #4693, opened 2022-09-05
Chore: use functions from kubevela/pkg

Somefive:

Reuse code from `kubevela/pkg` repo. Including
- multi-cluster roundtripper
- controller client
- monitorcontext

I have:


wonderflow (approved):

FogDong (approved):

pkg/multicluster/utils.go
```diff
@@ -35,22 +34,19 @@ import (
...
 )
 
-type contextKey string
-
 const (
-	// ClusterContextKey is the name of cluster using in client http context
-	ClusterContextKey = contextKey("ClusterName")
 	// ClusterLocalName specifies the local cluster
-	ClusterLocalName = velatypes.ClusterLocalName
+	ClusterLocalName = multiclusterpkg.Local
```
chivalryq, line 49: It seems sometimes we import the package as "pkgXXX" and sometimes "xxxpkg" and sometimes "velaxxx". Can you explain the rule?

chivalryq (approved):

pkg/multicluster/utils.go
chivalryq: I want to comment and hit "approve" unexpectedly

Somefive: velaxxx is okay. I think we do not need xxxvela.

Somefive: The package alias here is not a recommended way. I prefer to use multicluster.Local here but since we have a internal multicluster package and a external multicluster package, here I use multiclusterpkg to workaround, but pkgmulticluster is also okay.

Somefive: refactored.

Issue #4702, opened 2022-09-07
[Feature] Support when the template content is changed, the update of a single component only affects its own k8s resources


zxbyoyoyo:
**Describe the bug**

Scenario description:

now we will put multiple microservice components in an Application, each release is a single microservice component release
YAML example:
````
app1
  component:
    - type: micro-service
      name: s1
      props: s1 parameter
    - type: micro-service
      name: s2
      props: s2 parameter
````
When the content of the micro-service component template is changed, the parameters of s2 are modified at this time, and when the Application update is triggered, the k8s resources associated with the s1 and s2 components will be updated. I want that s1 will not be updated; 
To put it simply, i want the release of s2 service does not affect s1 service

**To Reproduce**

1. The YAML files of ComponentI used. container only one env
vela def apply microservice.yaml
```
microservice: {
	type: "component"
	annotations: {}
	labels: {}
	description: "Describes long-running, scalable, containerized services that have a stable network endpoint to receive external network traffic from customers."
	attributes: {
		workload: {
			definition: {
				apiVersion: "apps/v1"
				kind:       "Deployment"
			}
			type: "deployments.apps"
		}
	}
}
template: {

	output: {
		apiVersion: "apps/v1"
		kind:       "Deployment"
		spec: {
			selector: matchLabels: {
				"app": context.name
			}
			template: {
				metadata: {
					labels: {
						"app": context.name
					}
					annotations: {
						"version": parameter.version
					}
				}

				spec: {
					containers: [{
						name:  context.name
						image: parameter.image
						ports: [{
							containerPort: parameter.port
						}]
						env: [{
							name:  "default_env01"
							value: "default_env01"
						}]

					}]

				}
			}
		}
	}

	parameter: {
		image:    string
		port?:    int
		version?: string
	}

}

```
2. The YAML file of Application I applied.
``` shell
cat <<EOF | kubectl apply -f -
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: microservice-app
spec:
  components:
    - name: s1
      type: microservice
      properties:
        image: crccheck/hello-world
        port: 8080
        version: "1"
    - name: s2
      type: microservice
      properties:
        image: crccheck/hello-world
        port: 8080
        version: "1"
EOF
```
3. See Output
```
 kubectl get pods                                                              
NAME                                   READY   STATUS    RESTARTS   AGE
s1-7cbdd6bdf7-bhpz2                    1/1     Running   0          6m
[...]

PR #4715, opened 2022-09-13
Feat: support to add the deploy version to the pod annotation

### LINKED ISSUE #4597: The pod list is duplicated when the application is upgrading.
**Describe the bug**

deploy

As shown in GIF.

**Expected behavior**

1. Old and new instances are displayed normally.
2. The version of the instance is correct.

**KubeVela Version**

1.5.0

barnettZQG:

Fixes #4597

The component revision does not take effect as default. We could insert the deploy version to the pod template.

image

I have:


charts/vela-core/templates/defwithtemplate/webservice.yaml
```diff
@@ -369,6 +379,9 @@ spec:
         	// +usage=If addRevisionLabel is true, the revision label will be added to the underlying pods
         	addRevisionLabel: *false | bool
 
+        	// +usage=If addDeployVersionAnnotation is true, the deploy version annotation will be added to the underlying pods
+        	addDeployVersionAnnotation: *false | bool
```
wonderflow, line 383: why do we need DeployVersion when we already have a revision?

wonderflow (changes requested):
we must be caution as it will make the pod to restart everytime we upgrade the application.

wonderflow:
In some issues like https://github.com/kubevela/kubevela/issues/4702 , we don't want to affect the component when no properties are changed.

charts/vela-core/templates/defwithtemplate/webservice.yaml
barnettZQG: The component revision does not take effect as default.

barnettZQG:
> we must be caution as it will make the pod to restart everytime we upgrade the application.

This is an optional parameter. If users want to insert the deploy version to the pod, it is very useful.

charts/vela-core/templates/defwithtemplate/webservice.yaml
barnettZQG: The DeployVersion is generated by the business platform. The WebService definition provides an example use case to help users know how to pass the version number to the pod.

wonderflow: the business Platform should generate the DeployVersion inside the `annotations.annotations`

charts/vela-core/templates/defwithtemplate/daemon.yaml
```diff
@@ -467,6 +472,9 @@ spec:
         		ip: string
         		hostnames: [...string]
         	}]
+
+        	// +usage=If addDeployVersionAnnotation is true, the deploy version annotation will be added to the underlying pods
+        	addDeployVersionAnnotation: *false | bool
```
wonderflow, line 477: addPublishVersion

wonderflow: sideaffect: addPublishVersion will affect all components even the component has no change.

wonderflow: 1. addRevisionLabel is `ignore` now, is it really deprecated now?

- vela rollout depends on it when we need traffic rolling, it don't affect vela rollout with only replica rollout.
- addRevisionLabel will automatically inject revision on component workload change.
- addRevisionLabel only restart when component workload changed while traits changes don't affect the revision.

2. the usage of addPublishVersion

- publishVersion is currently the application revision specified by user, it works through velaux to application with whole lifecycle.
- publishVersion can help restart pods when new version published, adding this lable can implement this feature. Now we don't restart pods on new release, we can't tell the pod version.
- sideaffect: no automation, need to specify unique publishVersion manually on every release.


3. Users can have work around now

 - Add labels manually.

Somefive: I would suggest making reuse of appRevision. Although PublishVersion is clear and manually controllable, there are times that user wants to make automatic changes. I wonder if appRevision could handle both the automatic case and the manual case.

PR #4773, opened 2022-09-21
Fix: memory leak of the apiserver

barnettZQG:

There are two points:

1. The `json.Marshal` and `json.Unmarshal` consume much memory.
2. Not call the `Done` function of the queue. The app object will pile up in a queue.

Now:

image

image

|  Item  | Before | Now |
|  ----  | ----  | ----  |
| Application Number  | 700+ |  1.4K |
| Application Create/Update  | 3 request/second |   10 request/second |
| KubeVela Core Memory  | 600Mi+ | 1057Mi |
| APIServer Memory(Master) | 900Mi+ | 300Mi | 

I have:

### DIFF
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -162,7 +162,7 @@ func main() {
 
 	if pprofAddr != "" {
 		// Start pprof server if enabled
-		go pkgutils.EnablePprof(context.Background(), pprofAddr)
+		go pkgutils.EnablePprof(pprofAddr, nil)
 	}
 
 	if logFilePath != "" {
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -22,8 +22,6 @@ import (
 	goflag "flag"
 	"fmt"
 	"io"
-	"net/http"
-	"net/http/pprof"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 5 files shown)


pkg/utils/pprof.go
```diff
@@ -0,0 +1,57 @@
...
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+	pprofServer := http.Server{
+		Addr:    pprofAddr,
+		Handler: mux,
+	}
+	defer func() {
+		ctx, cancelFunc := context.WithTimeout(ctx, 60*time.Second)
+		defer cancelFunc()
```
wonderflow, line 44:
```suggestion
	go func() {
		<- ctx.Done()
```

wonderflow: the outside context need to have a cancel func

cmd/core/main.go
```diff
@@ -163,36 +162,7 @@ func main() {
...
-					klog.Error(err, "Failed to shutdown debug HTTP server")
-				}
-			}()
-
-			if err := pprofServer.ListenAndServe(); !errors.Is(http.ErrServerClosed, err) {
-				klog.Error(err, "Failed to start debug HTTP server")
-				panic(err)
-			}
-		}()
+		go pkgutils.EnablePprof(context.Background(), pprofAddr)
```
wonderflow, line 165: you must wrapper context.Background to have a cancel function

barnettZQG: I think there is no need to cancel the context. There are many `os.Exit(1)` function calls. There is no graceful exit control.

wonderflow: But you're actually writing dead code if you don't pass a useful context here.

image

barnettZQG: For the apiserver, There is a graceful exit control. So the common function provides this ability.

wonderflow (approved):

FogDong (approved):

PR #4850, opened 2022-10-11
Fix: change LabelResourceRuleFormat from Annotations to Labels

chengleqi:

Fixes #

I have:

### DIFF
--- a/pkg/velaql/providers/query/tree.go
+++ b/pkg/velaql/providers/query/tree.go
@@ -880,8 +880,8 @@ func mergeCustomRules(ctx context.Context, k8sClient client.Client) error {
 			format      string
 			err         error
 		)
-		if item.Annotations != nil {
-			format = item.Annotations[oam.LabelResourceRuleFormat]
+		if item.Labels != nil {
+			format = item.Labels[oam.LabelResourceRuleFormat]
 		}
 		switch format {
 		case oam.ResourceTopologyFormatJSON:
--- a/pkg/velaql/providers/query/tree_test.go
+++ b/pkg/velaql/providers/query/tree_test.go
@@ -1483,26 +1483,38 @@ childrenResourceType:
 	It("test merge rules", func() {
 		Expect(k8sClient.Create(ctx, &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "vela-system"}})).Should(SatisfyAny(BeNil(), util.AlreadyExistMatcher{}))
 		cloneSetConfigMap := v1.ConfigMap{TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: "ConfigMap"},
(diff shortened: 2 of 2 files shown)


pkg/velaql/providers/query/tree.go
```diff
@@ -880,8 +880,8 @@ func mergeCustomRules(ctx context.Context, k8sClient client.Client) error {
 			format      string
 			err         error
 		)
-		if item.Annotations != nil {
-			format = item.Annotations[oam.LabelResourceRuleFormat]
+		if item.Labels != nil {
+			format = item.Labels[oam.LabelResourceRuleFormat]
```
wonderflow, line 884: please add a test for this case.

chengleqi: OK. will do.

chengleqi:
> BTW, we must keep the consistency, so the annotation should also be checked.

I think it maybe a mistake. See [kruise-rollout plugin](https://github.com/kubevela/catalog/blob/d799c26bfbef247e8a019d715b330164f1255882/addons/kruise-rollout/template.cue#L30)

wonderflow:
> > BTW, we must keep the consistency, so the annotation should also be checked.
> 
> I think it maybe a mistake. See [kruise-rollout plugin](https://github.com/kubevela/catalog/blob/d799c26bfbef247e8a019d715b330164f1255882/addons/kruise-rollout/template.cue#L30)

But why it works? \cc @wangyikewxgm

chengleqi:
> > > BTW, we must keep the consistency, so the annotation should also be checked.
> > 
> > 
> > I think it maybe a mistake. See [kruise-rollout plugin](https://github.com/kubevela/catalog/blob/d799c26bfbef247e8a019d715b330164f1255882/addons/kruise-rollout/template.cue#L30)
> 
> But why it works? \cc @wangyikewxgm

It cannot get `oam.LabelResourceRuleFormat` key in Annotations so the [format](https://github.com/kubevela/kubevela/blob/fb45a94bb8ec04691f2fdcb79a3a5c46eac832c1/pkg/velaql/providers/query/tree.go#L884) is "" , so it has the same behavior like `oam.ResourceTopologyFormatYAML` in [label](https://github.com/kubevela/kubevela/blob/fb45a94bb8ec04691f2fdcb79a3a5c46eac832c1/pkg/velaql/providers/query/tree.go#L889).

wonderflow (commented):
OK, I've checked that all our addons use label. https://github.com/kubevela/catalog/search?q=rules.oam.dev%2Fresource-format

We don't need to keep the compatible. But let's add one test case to cover the json format.

chengleqi:
> OK, I've checked that all our addons use label. https://github.com/kubevela/catalog/search?q=rules.oam.dev%2Fresource-format
> 
> We don't need to keep the compatible. But let's add one test case to cover the json format.

OK. I'll try.

PR #4852, opened 2022-10-11
Fix: prevent rerun app while upgrade due to old apprev lack app workflow

Somefive:

In version before v1.6.0-alpha.4, applications with revision under v1.5.7 (not released yet) and revision under v1.6.0-alpha.4 (released), application revision (without PublishVersion) does not record workflow in Application Spec inside, which will lead to inequality with revisions created by latest KubeVela.

This PR adds additional rules for bypassing the comparing application workflow when
1. PublishVersion is not used
2. Application is created under KubeVela version v1.5.7. 

I have:


wonderflow (approved):
LGTM

PR #4926, opened 2022-10-26
Feat: add mode in workflow template

FogDong:

add mode in workflow template

I have:

cc @barnettZQG


charts/vela-core/crds/core.oam.dev_workflows.yaml
```diff
@@ -34,150 +34,16 @@ spec:
             type: string
           metadata:
             type: object
-          steps:
```
wonderflow, line 37: why the steps field changed?

wonderflow (commented):
why do we need to add mode to workflow CRD?

charts/vela-core/crds/core.oam.dev_workflows.yaml
FogDong: This field is not changed, it's just the diff displayed by github.

FogDong:
> why do we need to add mode to workflow CRD?

When we want to run a `WorkflowRun` From `Workflow`, we want to set the mode in the `Workflow` instead of setting the mode every time when running it.

cc @barnettZQG

barnettZQG (approved):

Somefive (approved):

PR #4954, opened 2022-10-29
Fix: do not change the workflow after the env changed

barnettZQG:

Enable users to edit the workflow to determine targets for the application to be published to the environment.

I have:

### DIFF
--- a/pkg/apiserver/domain/service/env.go
+++ b/pkg/apiserver/domain/service/env.go
@@ -184,23 +184,6 @@ func checkEqual(old, new []string) bool {
 	return reflect.DeepEqual(old, new)
 }
 
-func (p *envServiceImpl) updateAppWithNewEnv(ctx context.Context, envName string, env *model.Env) error {
-
-	// List all apps inside the env
-	apps, err := listApp(ctx, p.Store, apisv1.ListApplicationOptions{Env: envName})
-	if err != nil {
-		return err
-	}
-	for _, app := range apps {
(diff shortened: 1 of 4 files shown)


pkg/apiserver/domain/service/env.go
```diff
@@ -222,9 +205,7 @@ func (p *envServiceImpl) UpdateEnv(ctx context.Context, name string, req apisv1.
 		return nil, bcode.ErrEnvTargetConflict
 	}
 
-	var targetChanged bool
 	if len(req.Targets) > 0 && !checkEqual(env.Targets, req.Targets) {
```
StevenLeiZhang, line 208: Shall we check if any APP exists(deployed) in this ENV? If this ENV already has APP, reject the update request.
That means user need delete APP before updating ENV's target.

barnettZQG: I think we do not need to check the app when adding a new target to the environment. For deleting the targets, maybe we should check. What do you think?

StevenLeiZhang: > I think we do not need to check the app when adding a new target to the environment. For deleting the targets, maybe we should check. What do you think?

Yes, I think
1. if user only update 'description' or 'alias', no need pre-check.
2. if user want to remove some targets from ENV, need pre-check.
3. if user want to add more targets, no need pre-check.

barnettZQG: @StevenLeiZhang Could review again.

FogDong (approved):

PR #4985, opened 2022-11-02
Fix: app contains app override the child app with parent app label

Somefive:

When parent application creates child application, the resource of child application will carry the parent application label, due to the incorrect label inheritence.

I have:

### DIFF
--- a/pkg/appfile/appfile.go
+++ b/pkg/appfile/appfile.go
@@ -329,7 +329,7 @@ func (af *Appfile) generateAndFilterCommonLabels(compName string) map[string]str
 		oam.LabelAppComponent: compName,
 	}
 	// merge application's all labels
-	finalLabels := util.MergeMapOverrideWithDst(Labels, af.AppLabels)
+	finalLabels := util.MergeMapOverrideWithDst(af.AppLabels, Labels)
 	filterLabels, ok := af.AppAnnotations[oam.AnnotationFilterLabelKeys]
 	if ok {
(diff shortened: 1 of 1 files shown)


pkg/appfile/appfile.go
```diff
@@ -329,7 +329,7 @@ func (af *Appfile) generateAndFilterCommonLabels(compName string) map[string]str
 		oam.LabelAppComponent: compName,
 	}
 	// merge application's all labels
-	finalLabels := util.MergeMapOverrideWithDst(Labels, af.AppLabels)
+	finalLabels := util.MergeMapOverrideWithDst(af.AppLabels, Labels)
```
wonderflow, line 332: why we should exchange the order?

wonderflow (approved):

pkg/appfile/appfile.go
Somefive: The resources of one child application (identified as B) should carry `app.oam.dev/name: B` label.

However since the child application (B) is created and managed by the parent application (identified as A), B itself carries `app.oam.dev/name: A` label.

The KubeVela controller passes the labels of the child application B to its resources, but should not override `app.oam.dev/name: B`.

The logic here previously let `app.oam.dev/name: A` (the label passed from the child application B) overrides the label `app.oam.dev/name: B` (the label that indicates the owner of the resource). By changing the order here, we let the latter one to have higher priority.

PR #5002, opened 2022-11-04
Fix: fix the stuck problem and use LRU cache to promote the speed of loading topology

### LINKED ISSUE #5001: [Bug] In `vela top`, the UI interaction will get stuck when interacting with remote clusters with poor network
**Describe the bug**

In `vela top`, the UI interaction will get stuck when interacting with remote clusters with poor network.

**To Reproduce**

When we try to show the resource topology of applications deployed in overseas clusters, it takes a long time to show the topology, which causes subsequent interactions to get stuck.

**Expected behavior**

**Screenshots**

**KubeVela Version**

**Cluster information**

**Additional context**

Hanmengnan:
This problem mainly comes from two problems:

The underlying logic of the `tview` framework is that it places the operations of drawing UI on the main thread for execution to avoid conflicts. And it recommends that all operations that cause conflicts should do so. In addition, it provides a function `QueueUpdateDraw` to enable the drawing UI operations in other threads to be called in order in the main thread.

However, in `vela top`, we have a system performance component to show the performance of the vela core and gateway. This component will refresh regularly. When the cluster is overseas or the network is poor, network IO will take too long, causing the UI refresh of this component to be too slow and blocking all subsequent UI drawing operations. In the user's opinion, the UI is stuck.

So here we use a **lazy loading** mechanism. This component updates data regularly, but its UI drawing operation does not execute immediately, so it does not block the UI drawing of the main thread, but the UI will redraw with the redraw triggered by user's UI operation by the way.

---

In addition, topology loading takes a long time, and the application topology will not change frequently. Therefore, the number of network requests for application topology queries should be reduced. Therefore, we have introduced an LRU cache to cache the latest five application resource topologies, which can greatly improve the topology loading speed in the future.

Signed-off-by: HanMengnan <1448189829@qq.com>

Fixes #5001

I have:

@wonderflow


wonderflow (approved):

Somefive (commented):
Generally LGTM, several additional questions.
1. The implementation of LRU cache I think is not a unique use for vela top. Have you tried to reuse some existing golang library or refer to some Kubernetes implementation? If there is no suitable ones, I suggest moving the implementation of LRU to some common package, like `pkg/utils/cache` or even the `kubevela/pkg` library.
2. The default number of resource topology cache is set to constant value 5. Is there any reason for the number of 5? For example, will the number of cache 50 greatly affect the memory use?
3. It is good that we can introduce cache and other optimization method to `vela top`, like the async update between UI and retrieving data. However, the initial design of `vela top` aims at a lightweight CLI tool for displaying data. The cache and async update and future optimization tools could help improve the use experiences of `vela top` indeed, but it makes the architecture and logic of `vela top` bigger and more complex. Although it is okay to keep codes in the main repo as long as the codes are clean and tested, I would suggest in a long term, some commonly used techniques or interfaces could be moved out of the `top` package. Just try not to become another huge apiserver module in `kubevela/kubevela`. :)

FYI, the above questions are not required to be solved to merge this PR. Just some recommends and discussion.

Hanmengnan:
> Generally LGTM, several additional questions.
> 
> 1. The implementation of LRU cache I think is not a unique use for vela top. Have you tried to reuse some existing golang library or refer to some Kubernetes implementation? If there is no suitable ones, I suggest moving the implementation of LRU to some common package, like `pkg/utils/cache` or even the `kubevela/pkg` library.
> 2. The default number of resource topology cache is set to constant value 5. Is there any reason for the number of 5? For example, will the number of cache 50 greatly affect the memory use?
> 3. It is good that we can introduce cache and other optimization method to `vela top`, like the async update between UI and retrieving data. However, the initial design of `vela top` aims at a lightweight CLI tool for displaying data. The cache and async update and future optimization tools could help improve the use experiences of `vela top` indeed, but it makes the architecture and logic of `vela top` bigger and more complex. Although it is okay to keep codes in the main repo as long as the codes are clean and tested, I would suggest in a long term, some commonly used techniques or interfaces could be moved out of the `top` package. Just try not to become another huge apiserver module in `kubevela/kubevela`. :)
> 
> FYI, the above questions are not required to be solved to merge this PR. Just some recommends and discussion.

Thanks, I got it.

Hanmengnan:
> Generally LGTM, several additional questions.
> 
> 1. The implementation of LRU cache I think is not a unique use for vela top. Have you tried to reuse some existing golang library or refer to some Kubernetes implementation? If there is no suitable ones, I suggest moving the implementation of LRU to some common package, like `pkg/utils/cache` or even the `kubevela/pkg` library.
> 2. The default number of resource topology cache is set to constant value 5. Is there any reason for the number of 5? For example, will the number of cache 50 greatly affect the memory use?
> 3. It is good that we can introduce cache and other optimization method to `vela top`, like the async update between UI and retrieving data. However, the initial design of `vela top` aims at a lightweight CLI tool for displaying data. The cache and async update and future optimization tools could help improve the use experiences of `vela top` indeed, but it makes the architecture and logic of `vela top` bigger and more complex. Although it is okay to keep codes in the main repo as long as the codes are clean and tested, I would suggest in a long term, some commonly used techniques or interfaces could be moved out of the `top` package. Just try not to become another huge apiserver module in `kubevela/kubevela`. :)
> 
> FYI, the above questions are not required to be solved to merge this PR. Just some recommends and discussion.

Hanmengnan:
> Generally LGTM, several additional questions.
> 
> 1. The implementation of LRU cache I think is not a unique use for vela top. Have you tried to reuse some existing golang library or refer to some Kubernetes implementation? If there is no suitable ones, I suggest moving the implementation of LRU to some common package, like `pkg/utils/cache` or even the `kubevela/pkg` library.
> 2. The default number of resource topology cache is set to constant value 5. Is there any reason for the number of 5? For example, will the number of cache 50 greatly affect the memory use?
> 3. It is good that we can introduce cache and other optimization method to `vela top`, like the async update between UI and retrieving data. However, the initial design of `vela top` aims at a lightweight CLI tool for displaying data. The cache and async update and future optimization tools could help improve the use experiences of `vela top` indeed, but it makes the architecture and logic of `vela top` bigger and more complex. Although it is okay to keep codes in the main repo as long as the codes are clean and tested, I would suggest in a long term, some commonly used techniques or interfaces could be moved out of the `top` package. Just try not to become another huge apiserver module in `kubevela/kubevela`. :)
> 
> FYI, the above questions are not required to be solved to merge this PR. Just some recommends and discussion.

1. I reused an existing golang LRU library.
2. I think your proposal to expand the cache is correct, but I think that too much cache will affect the garbage collection of cached objects. In addition, cached view objects are not small objects, and users usually do not view too many topologies at a time. Therefore, I appropriately increase the cache size to 10.

Somefive (approved):
Great!

PR #5014, opened 2022-11-07
Fix: vela status will always get external ip first

chivalryq:

`vela status` sometimes print internal IP even if there are external IP for nodes in cluster.
This also have an impact when deploy add-on VelaUX and influence the first-time installation experience.

Fix this by always printing the node's external IP if there are any.

image

Fixes #

I have:


wonderflow:
What it works like in ACK？

barnettZQG (approved):
Great!

chivalryq:
> What it works like in ACK？

@wonderflow  

Now: If first node returned from APIServer have external IP, then return it. Else return this node's internal IP.

After fix:
If there are node with external IP, return the external IP.
If there aren't any, return the first node's internal IP.

In short, this PR make the result more predictable instead of random.

PR #5090, opened 2022-11-18
Feat: add support for compressing apprev using gzip and zstd

### LINKED ISSUE #5048: [Feature] support compress apprevision using gzip or zstd

AppRevision could be large as the default number is 10 for every application. It could cause list-watch time out when the number is too large. 

We need to compress it just like the resourceTracker (https://github.com/kubevela/kubevela/issues/4392).

**Additional context**

charlie0129:

Similar to #4630 , this pr adds compression support (`gzip` and `zstd`) for ApplicationRevisions.

Performance is also similar to #4630 so I didn't benchmark this change.

Some test results using `fluxcd` addon:

Uncompressed: 420,968 bytes

```console
$ kubectl describe apprev addon-fluxcd-v4 -n vela-system | wc -c # uncompressed
420968
```

zstd: 42,837 bytes

```console
$ k describe apprev addon-fluxcd-v1 -n vela-system | wc -c # zstd
42837
```

Fixes #5048

I have:


Somefive (commented):
Great Job! Amazing compression and very clean code! Great appreciation for the work.

Some discussions:
1. I think the `pkg/utils/compression` now is pretty clean, functional and extensible. It could be moved into `kubevela/pkg` repo and reused by more controllers like `WorkflowRun`.
2. I see that there is a bunch of field copy code in apis/core.oam.dev/v1beta1/applicationrevision.go  . These copy code means any edition to the ApplicationRevision field in the future would need to modify in every place they are copied, which is a bit hard to track. I'm not sure if there is any better solutions for the code. But if there is some great ideas that can reduce these duplicated codes, it would be great help.

charlie0129:
> 1. I think the `pkg/utils/compression` now is pretty clean, functional and extensible. It could be moved into `kubevela/pkg` repo and reused by more controllers like `WorkflowRun`.

Sure, I will open a pull request there to move `pkg/utils/compression` to `kubevela/pkg`, and later refactor this pr to use `kubevela/pkg/util/compression` instead.


> 2. I see that there is a bunch of field copy code in apis/core.oam.dev/v1beta1/applicationrevision.go

The copying is to only copy the fields that need compression, and later compress them. Since not all fields need compression, e.g., `ApplicationRevisionSpec.Compression`, so I selectively copied them.

Some possible solution that came to my mind:
- Use Go reflect to loop over the fields and selectively copy them. But this can be a bit hacky.
- Embed all the fields that need compression in another struct, e.g., `CompressedFields` , then embed `CompressedFields` in `ApplicationRevisionSpec` and use `inline` JSON tag to keep the JSON structure of `ApplicationRevisionSpec` the same. Now we only need to compress `CompressedFields`. No need for selectively copying.

Somefive:
> > 1. I think the `pkg/utils/compression` now is pretty clean, functional and extensible. It could be moved into `kubevela/pkg` repo and reused by more controllers like `WorkflowRun`.
> 
> Sure, I will open a pull request there to move `pkg/utils/compression` to `kubevela/pkg`, and later refactor this pr to use `kubevela/pkg/util/compression` instead.
> 
> > 2. I see that there is a bunch of field copy code in apis/core.oam.dev/v1beta1/applicationrevision.go
> 
> The copying is to only copy the fields that need compression, and later compress them. Since not all fields need compression, e.g., `ApplicationRevisionSpec.Compression`, so I selectively copied them.
> 
> Some possible solutions that came to my mind:
> 
> * Use Go reflect to loop over the fields and selectively copy them. But this can be a bit hacky.
> * Embed all the fields that need compression in another struct, e.g., `CompressedFields` , then embed `CompressedFields` in `ApplicationRevisionSpec` and use `inline` JSON tag to keep the JSON structure of `ApplicationRevisionSpec` the same. Now we only need to compress `CompressedFields`. No need for selectively copying. **But after some experimentation, this will break crds, so it is not practically feasible.**
> 
> I don't have any good solution currently.

I see. I'm a bit curious how the use of `CompressedFields` will break CRDs, even with the use of `inline` tag? Could you give some details?

Somefive:
> > 1. I think the `pkg/utils/compression` now is pretty clean, functional and extensible. It could be moved into `kubevela/pkg` repo and reused by more controllers like `WorkflowRun`.
> 
> Sure, I will open a pull request there to move `pkg/utils/compression` to `kubevela/pkg`, and later refactor this pr to use `kubevela/pkg/util/compression` instead.
> 
> > 2. I see that there is a bunch of field copy code in apis/core.oam.dev/v1beta1/applicationrevision.go
> 
> The copying is to only copy the fields that need compression, and later compress them. Since not all fields need compression, e.g., `ApplicationRevisionSpec.Compression`, so I selectively copied them.
> 
> Some possible solutions that came to my mind:
> 
> * Use Go reflect to loop over the fields and selectively copy them. But this can be a bit hacky.
> * Embed all the fields that need compression in another struct, e.g., `CompressedFields` , then embed `CompressedFields` in `ApplicationRevisionSpec` and use `inline` JSON tag to keep the JSON structure of `ApplicationRevisionSpec` the same. Now we only need to compress `CompressedFields`. No need for selectively copying. **But after some experimentation, this will break crds, so it is not practically feasible.**
> 
> I don't have any good solution currently.

@charlie0129 How about making some interface like `Compressable` and use `GetCompressedFields` and `SetCompresseFields`? Will it be helpful?

charlie0129:
> I see. I'm a bit curious how the use of `CompressedFields` will break CRDs, even with the use of `inline` tag? Could you give some details?

Hmm, it can be I wrote something wrong last time I tried so it break CRDs. I retried this method again today, will wait and see the test results to see if it works

charlie0129:
> How about making some interface like `Compressable` and use `GetCompressedFields` and `SetCompresseFields`? Will it be helpful?

If we use this method, we still need to write `Get/SetCompressedFields` which need to be updated everytime a new field is added. Just like we discussed before

> These copy code means any edition to the ApplicationRevision field in the future would need to modify in every place they are copied, which is a bit hard to track.

So, I retried the JSON `inline` method, and it seems to work (: . I will later move the compression util to `kubevela/pkg` repo.

references/cli/revision_test.go
```diff
@@ -713,25 +715,22 @@ func TestPrintApprev(t *testing.T) {
...
+			assert.Contains(t, strings.ReplaceAll(tc.out.String(), " ", ""), strings.ReplaceAll(tc.out.String(), " ", ""))
 		})
 	}
 }
 
-func tableOut(name, pv, s, hash, bt, status, size string) string {
+func tableOut(name, pv, s, hash, bt, status string) string {
 	table := newUITable().AddRow("NAME", "PUBLISH_VERSION", "SUCCEEDED", "HASH", "BEGIN_TIME", "STATUS", "SIZE")
-	table.AddRow(name, pv, s, hash, bt, status, size)
+	table.AddRow(name, pv, s, hash, bt, status)
```
Somefive, line 732: Why the size here is removed?

Somefive: So the size of the AppRev here could be compressed and the size might not match the original size. Can we show the size of the AppRev as before and add extra hint for compression, such as 
```shell
STATUS    SIZE 
SUCCEED   62B (Compressed)
```

charlie0129: Because the newly added `compression` field will increase the size, so the test will fail. To fix the tests, I need to recalculate the size in the test cases. 

After some thought, I figured out to just not test the `size` field. The purpose of this test is to test the `SUCCEED` and `STATUS` field anyway. Also, since `size` is affected by apprev struct, everytime a new field is added to apprev, we always need to recalculate the size to fix this test.

charlie0129: > So the size of the AppRev here could be compressed and the size might not match the original size. Can we show the size of the AppRev as before and add extra hint for compression, such as

good idea

Somefive: > Because the newly added `compression` field will increase the size, so the test will fail. To fix the tests, I need to recalculate the size in the test cases.
> 
> After some thought, I figured out to just not test the `size` field. The purpose of this test is to test the `SUCCEED` and `STATUS` field anyway. Also, since `size` is affected by apprev struct, everytime a new field is added to apprev, we always need to recalculate the size to fix this test.

I see. Just to confirm that the size still shows since this feature can be helpful to estimate the storage used by AppRev in ETCD. The value check for the size field could be ignored, although recalculating it works as well, because the structure of ApplicationRevision will not change frequently.

charlie0129: Hmm, interesting... Just about to edit it. It turns out the code that this test tests isn't used anywhere lol. will do some refactoring

charlie0129: image

added compression ratio

Somefive (approved):

barnettZQG (approved):
lgtm

makefiles/e2e.mk
```diff
@@ -32,20 +32,21 @@ e2e-setup-core-wo-auth:
...
+	    --set applicationRevisionLimit=5                \
+	    --set dependCheckWait=10s                       \
+	    --set image.tag=$(GIT_COMMIT)                   \
+	    --wait kubevela                                 \
+	    ./charts/vela-core                              \
+	    --set authentication.enabled=true               \
+	    --set authentication.withUser=true              \
+	    --set authentication.groupPattern=*             \
+	    --set featureGates.zstdResourceTracker=true     \
+	    --set featureGates.zstdApplicationRevision=true
```
wonderflow, line 49: does it mean we don't test uncompressed apprevision any more?

charlie0129: There are targets that don't use compressions, so we can say uncompressed apprev is still tested. 

https://github.com/kubevela/kubevela/blob/13029c07767a129c545abc9652ec94b2d2b79433/makefiles/e2e.mk#L21-L31

But yes, if we want to test every combination, we may want some changes to our testing infrastructure. I'm not sure how this can be done elegantly.

wonderflow: OK, I think it's fine in this case, I'm just afraid if we haven't test any cases for uncompressed mode.

PR #5116, opened 2022-11-24
Fix: apiserver image build and add CI check for it

wonderflow:
Signed-off-by: Jianbo Sun <jianbo.sjb@alibaba-inc.com>

Fixes https://github.com/kubevela/kubevela/actions/runs/3537032744/jobs/5936644544

I have:

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -173,3 +173,63 @@ jobs:
         run: |
           move .\bin\vela .\bin\vela.exe
           .\bin\vela.exe version
+
+  check-core-image-build:
+    runs-on: ubuntu-latest
+    needs: detect-noop
+    if: needs.detect-noop.outputs.noop != 'true'
+    steps:
+      - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -173,3 +173,35 @@ jobs:
         run: |
           move .\bin\vela .\bin\vela.exe
           .\bin\vela.exe version
+
+    image-build-check:
```
barnettZQG, line 177: Is this a repeat job? There is a `Registry` workflow to check building images.

wonderflow: The Registry CI only works in release/master branch, I'm making it to be PR normal check.

barnettZQG: We could set the Registry workflow to work in all branches?

wonderflow: That is a more complicate CI and contains secrets, we don't need to do that.

barnettZQG (approved):

PR #5118, opened 2022-11-24
Fix: bug of filter registry func will modify origin data

wangyikewxgm:
filter registry func flaky which will cause flaky tests.

Signed-off-by: 楚岳 <wangyike.wyk@alibaba-inc.com>

Fixes #

I have:

### DIFF
--- a/pkg/addon/utils.go
+++ b/pkg/addon/utils.go
@@ -505,10 +505,14 @@ func checkBondComponentExist(u unstructured.Unstructured, app v1beta1.Applicatio
 
 // FilterDependencyRegistries will return all registries besides the target registry itself
 func FilterDependencyRegistries(i int, rs []Registry) []Registry {
-	registries := make([]Registry, len(rs))
-	copy(registries, rs)
-	if i < len(registries) {
-		return append(registries[:i], registries[i+1:]...)
+	if i >= len(rs) {
+		return rs
 	}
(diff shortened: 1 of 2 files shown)


pkg/addon/utils.go
```diff
@@ -504,7 +504,9 @@ func checkBondComponentExist(u unstructured.Unstructured, app v1beta1.Applicatio
 }
 
 // FilterDependencyRegistries will return all registries besides the target registry itself
-func FilterDependencyRegistries(i int, registries []Registry) []Registry {
+func FilterDependencyRegistries(i int, rs []Registry) []Registry {
+	registries := make([]Registry, len(rs))
+	copy(registries, rs)
 	if i < len(registries) {
```
charlie0129, line 510: What if `i<0`?

Something like this will do (possibly copies fewer elements)

```go
func FilterDependencyRegistries(i int, rs []Registry) []Registry {
	if i >= len(rs) {
		return rs
	}
	if i < 0 {
		return rs
	}
	ret := make([]Registry, len(rs)-1)
	copy(ret, rs[:i])
	copy(ret[i:], rs[i+1:])
	return ret
}
```

wangyikewxgm: Good suggestion.

charlie0129 (approved):

PR #5161, opened 2022-12-06
Feat: support data-passing in deploy step

chivalryq:

### Intro

This PR add component level input/output capability for `deploy` step. Now `deploy` can correctly deal with components like:

```YAML
  components:
    - name: podinfo
      outputs:
        - name: message
          valueFrom: output.status.conditions[0].message
      properties:
        image: stefanprodan/podinfo:4.0.3
      type: webservice
    - name: configmap
      properties:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: deployment-msg
      type: raw
      inputs:
        - from: message
          parameterKey: data.msg
```

### Variable matching mechanism

There are two possible variable matching ways:
1. First try to match the var strictly. Find var with key `<cluster>/<namespace>/<replication>/<var>`.

image

2. As fallback, if no matching in the first step, restriction on <replication> can be given up. Try to match var with `<cluster>/<namespace>//<var>`. Namely the <replication> is empty.

image

Fixes #

I have:

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/generator.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/generator.go
@@ -332,13 +332,12 @@ func (h *AppHandler) checkComponentHealth(appParser *appfile.Parser, appRev *v1b
 			return false, nil, nil, err
 		}
 
-		_, isHealth, err := h.collectHealthStatus(auth.ContextWithUserInfo(ctx, h.app), wl, appRev, overrideNamespace, false)
+		_, output, outputs, isHealth, err := h.collectHealthStatus(auth.ContextWithUserInfo(ctx, h.app), wl, appRev, overrideNamespace, false)
 		if err != nil {
 			return false, nil, nil, err
 		}
 
-		workload, traits, err := getComponentResources(auth.ContextWithUserInfo(ctx, h.app), manifest, wl.SkipApplyWorkload, h.r.Client)
-		return isHealth, workload, traits, err
+		return isHealth, output, outputs, err
 	}
 }
 
@@ -390,7 +389,7 @@ func (h *AppHandler) applyComponentFunc(appParser *appfile.Parser, appRev *v1bet
 			if err := h.Dispatch(ctx, clusterName, common.WorkflowResourceCreator, dispatchResources...); err != nil {
 				return nil, nil, false, errors.WithMessage(err, "Dispatch")
 			}
-			_, isHealth, err = h.collectHealthStatus(ctx, wl, appRev, overrideNamespace, false)
+			_, _, _, isHealth, err = h.collectHealthStatus(ctx, wl, appRev, overrideNamespace, false)
 			if err != nil {
 				return nil, nil, false, errors.WithMessage(err, "CollectHealthStatus")
 			}
--- a/pkg/controller/core.oam.dev/v1alpha2/application/generator.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/generator.go
@@ -333,6 +333,10 @@ func (h *AppHandler) checkComponentHealth(appParser *appfile.Parser, appRev *v1b
 		}
 
 		_, isHealth, err := h.collectHealthStatus(auth.ContextWithUserInfo(ctx, h.app), wl, appRev, overrideNamespace, false)
+		if err != nil {
+			return false, nil, nil, err
+		}
+
 		workload, traits, err := getComponentResources(auth.ContextWithUserInfo(ctx, h.app), manifest, wl.SkipApplyWorkload, h.r.Client)
 		return isHealth, workload, traits, err
 	}
--- a/pkg/controller/core.oam.dev/v1alpha2/application/generator.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/generator.go
@@ -307,20 +307,20 @@ func (h *AppHandler) renderComponentFunc(appParser *appfile.Parser, appRev *v1be
 }
 
 func (h *AppHandler) checkComponentHealth(appParser *appfile.Parser, appRev *v1beta1.ApplicationRevision, af *appfile.Appfile) oamProvider.ComponentHealthCheck {
-	return func(baseCtx context.Context, comp common.ApplicationComponent, patcher *value.Value, clusterName string, overrideNamespace string, env string) (bool, error) {
+	return func(baseCtx context.Context, comp common.ApplicationComponent, patcher *value.Value, clusterName string, overrideNamespace string, env string) (bool, *unstructured.Unstructured, []*unstructured.Unstructured, error) {
 		ctx := multicluster.ContextWithClusterName(baseCtx, clusterName)
 		ctx = contextWithComponentNamespace(ctx, overrideNamespace)
 		ctx = contextWithReplicaKey(ctx, comp.ReplicaKey)
 
 		wl, manifest, err := h.prepareWorkloadAndManifests(ctx, appParser, comp, appRev, patcher, af)
 		if err != nil {
-			return false, err
+			return false, nil, nil, err
 		}
 		wl.Ctx.SetCtx(auth.ContextWithUserInfo(ctx, h.app))
 
 		readyWorkload, readyTraits, err := renderComponentsAndTraits(h.r.Client, manifest, appRev, clusterName, overrideNamespace, env)
 		if err != nil {
-			return false, err
+			return false, nil, nil, err
 		}
 		checkSkipApplyWorkload(wl)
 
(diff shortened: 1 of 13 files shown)


pkg/workflow/providers/multicluster/deploy.go
```diff
@@ -165,13 +183,148 @@ func (t *applyTask) dependents() []string {
...
+		outputs = append(outputs, t.varKey(output.Name))
+	}
+	return outputs
+}
+
+func (t *applyTask) varKey(v string) string {
+	return fmt.Sprintf("%s/%s/%s", t.placement.Cluster, t.placement.Namespace, v)
+}
+
+func (t *applyTask) patchInput(inputs *ioCache, build valueBuilder) error {
```
Somefive, line 206: Function implementation here is a bit redundant. Way to optimize:
1. Create a empty cue.Value x
2. x = x.FillValue(cue.ParsePath(""), comp.Properties)
3. for key, val := range inputs { x = x.FillValue(cue.ParsePath(key), val) }
4. x.UnmarshalTo(comp.Properties)
So that we can avoid explicit json parsing and FillValueByScript, reduce the number of errors we need to handle in this function. BTW, this logic could be separated into some util packages.

pkg/workflow/providers/multicluster/deploy.go
```diff
@@ -186,8 +339,25 @@ func applyComponents(ctx context.Context, apply oamProvider.ComponentApply, heal
 		taskHealthyMap[tasks[i].key()] = res.healthy
 	}
 
+	reapplyTask := getReapplyTasks(tasks, taskHealthyMap)
+
+	var outputResult []*applyTaskResult
+	if len(reapplyTask) > 0 {
```
Somefive, line 345: reapplyTasks are done tasks but need to reapply and get outputs?

Somefive (commented):
Good job. Some suggestions for reducing the code complexity are commented.

pkg/workflow/providers/multicluster/deploy.go
```diff
@@ -165,13 +183,148 @@ func (t *applyTask) dependents() []string {
 	return dependents
 }
 
+func (t *applyTask) inputs() []string {
+	var inputs []string
+	for _, input := range t.component.Inputs {
+		inputs = append(inputs, t.varKey(input.From))
+	}
+	return inputs
```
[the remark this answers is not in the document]
chivalryq: Then we have to make stemInput struct exported in `kubevela/workflow/api/v1alpha1/types.go` or It can not be clarified as `varKey` parameter.

pkg/workflow/providers/multicluster/deploy.go
chivalryq: yes, or I guess we have to do changes to make `healthCheck` to return the manifests. This change
1. will change a lot places except deploy step.
2. make the function less cohesive

pkg/workflow/providers/multicluster/deploy.go
chivalryq: I remove the marshal process on the first step. However, we can not use a random cue.Value to do the patch, because the runtime is not the same with values from input.

pkg/workflow/providers/multicluster/deploy.go
```diff
@@ -165,13 +183,148 @@ func (t *applyTask) dependents() []string {
...
+}
+
+func (t *applyTask) varKey(v string) string {
+	return fmt.Sprintf("%s/%s/%s", t.placement.Cluster, t.placement.Namespace, v)
+}
+
+func (t *applyTask) patchInput(inputs *ioCache, build valueBuilder) error {
+	comp, err := json.Marshal(t.component)
+	if err != nil {
+		return errors.Wrap(err, "marshal component")
```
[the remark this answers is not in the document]
chivalryq: Why we shouldn't use errors.Wrap? to flatten the error?

deploy-out.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: core.oam.dev/v1beta1
```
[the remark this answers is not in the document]
chivalryq: just draft, sorry

pkg/workflow/providers/multicluster/deploy.go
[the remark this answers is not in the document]
Somefive: I think `varKey` is not a value but a function?

pkg/workflow/providers/multicluster/deploy.go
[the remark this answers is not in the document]
Somefive: https://stackoverflow.com/questions/61933650/whats-the-difference-between-errors-wrapf-errors-errorf-and-fmt-errorf

It is okay to use `errors.Wrap`. I would suggest we use `fmt.Errorf` and unify the error handling code style.

pkg/workflow/providers/multicluster/deploy.go
Somefive: I think the input is not necessarily a cue.Value?

chivalryq: When we get input from workflowContext, it is value.Value.

Somefive: I see. Well, let's leave the issue here. It is a non-blocking issue for this PR. I will go back and handle it in the future.

chivalryq: @Somefive Here's some correction: we are not getting input from workflowContext. We will re-apply healthy components and dynamically compute the output. Using same value.Value's runtime is critical here.

wonderflow (commented):
1. please increase the test coverage of the `deploy.go` file
2. please add description for this PR

pkg/workflow/providers/multicluster/deploy.go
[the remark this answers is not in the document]
Somefive: I see the problem.

test/e2e-multicluster-test/multicluster_test.go
```diff
@@ -784,5 +784,47 @@ var _ = Describe("Test multicluster scenario", func() {
 				g.Expect(kerrors.IsNotFound(k8sClient.Get(hubCtx, deployKey, deploy))).Should(BeTrue())
 			}, 20*time.Second).Should(Succeed())
 		})
+
+		FIt("Test application with input/output in deploy step", func() {
```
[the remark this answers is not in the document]
chivalryq: Yes, done

pkg/controller/core.oam.dev/v1alpha2/application/generator.go
```diff
@@ -329,11 +329,16 @@ func (h *AppHandler) checkComponentHealth(appParser *appfile.Parser, appRev *v1b
...
+			return false, nil, nil, err
 		}
 
 		_, isHealth, err := h.collectHealthStatus(auth.ContextWithUserInfo(ctx, h.app), wl, appRev, overrideNamespace, false)
-		return isHealth, err
+		if err != nil {
+			return false, nil, nil, err
+		}
+
+		workload, traits, err := getComponentResources(auth.ContextWithUserInfo(ctx, h.app), manifest, wl.SkipApplyWorkload, h.r.Client)
```
Somefive, line 340: I think this line of code will add extra "get" requests, right?

pkg/workflow/providers/multicluster/deploy.go
```diff
@@ -148,80 +164,230 @@ func overrideConfiguration(policies []v1beta1.AppPolicy, components []common.App
...
+			task.healthy = pointer.Bool(healthy)
+			if healthy {
+				err = task.generateOutput(output, outputs, cache, makeValue)
+			}
+			return &applyTaskResult{healthy: healthy, err: err, task: task}
+		}, slices.Parallelism(parallelism))
+
+		for _, res := range checkResults {
+			taskHealthyMap[res.task.key()] = res.healthy
+			if !res.healthy || res.err != nil {
```
Somefive, line 343: If is not healthy, the result will be put in healthCheckError? The variable name healthCheckError looks a bit confusing.

Somefive (commented):
Generally looks good, great job!

pkg/workflow/providers/multicluster/deploy.go
chivalryq: It's a little confusing and wrong. Maybe we should call it `unhealthResult`?

Somefive: unhealthyResults looks better

pkg/controller/core.oam.dev/v1alpha2/application/generator.go
chivalryq: Yes, it will get one more time for resources. Maybe I can make a optimization. Let me try.

Somefive: great!

Somefive (commented):
Good job! LGTM. Let's fix the tests.

Somefive (approved):

wonderflow (approved):

PR #5189, opened 2022-12-14
Feat: return notFound error when deleting app

suwliang3:
Signed-off-by: suwanliang_yewu <suwanliang_yewu@cmss.chinamobile.com>

When deleting an app from a namespace,if there is not the app in the namespace, it should return the error of "notFound".

I have:


wonderflow (changes requested):
In which case why we should not found error? If I want to delete an app, it's not exist, then we can say it's already deleted. There should be no error.

wonderflow (commented):
well, please sign the DCO

suwliang3:
> well, please sign the DCO

Done

PR #5233, opened 2022-12-26
Chore: update workflow version to add restart from step and fix bugs

### LINKED ISSUE #5178: [Bug] vela debug cannot work for workflow steps with invalid resource name
**Describe the bug**

`vela debug` records the step data into configmap with the name of the workflowstep as part of the configmap name.

However, in the case where workflowstep name is not valid resource name, such as `addon-grafana` which use name like "Install Kubernetes API Datasource", the debug will fail and report `configmaps "addon-grafana-Install Kubernetes API Datasource-debug-44967" not found`. We should use hash value or replace space as dash(-) in the name in avoid of the bug.

**To Reproduce**

**Expected behavior**

**Screenshots**

**KubeVela Version**

**Cluster information**

**Additional context**

FogDong:

Update workflow version to add restart from step and fix bugs.

This PR:
* Add step name-generator in mutating webhook.
* Add restart from a specified step in cli.
* Fixes #5178

Fixes

I have:

### DIFF
--- a/pkg/workflow/operation/operation.go
+++ b/pkg/workflow/operation/operation.go
@@ -22,6 +22,8 @@ import (
 	"io"
 
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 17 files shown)


pkg/workflow/operation/operation.go
```diff
@@ -269,24 +272,62 @@ func (wo appWorkflowOperator) Rollback(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to parse appfile: %w", err)
+	}
+	stepStatus, cm, err := wfUtils.CleanStatusFromStep(appFile.WorkflowSteps, status.Steps, *appFile.WorkflowMode, cm, step)
+	if err != nil {
 		return err
 	}
-
+	status.Steps = stepStatus
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
wonderflow, line 313: when will the confilct happen? I don't like to use retryOnConflict as it may cover up bugs which can be dangerous.

FogDong: If the controller updates the resources at the same time, it may cause conflict error, it's better to wrap the conflict error for the users so that they don't need to retry.

Somefive (approved):

PR #5237, opened 2022-12-28
Fix: remove cluster field from pod view in vela top and optimize some ui

### LINKED ISSUE #5229: Incorrect cluster field in vela top pod view
**Describe the bug**

When I compare managed resources view and pod view in vela top, I found the `Cluster` field is correct in managed resources view, but not in pod. view.

**To Reproduce**

**Expected behavior**

The `Cluster` field in pod view is correct.

**Screenshots**

image
type "Enter" ⬇️
image

**KubeVela Version**

CLI version v1.6.4

**Cluster information**

**Additional context**

Related PR: #5228

Hanmengnan:

Fixes #5229

I have:

### DIFF
--- a/references/cli/top/component/crumbs.go
+++ b/references/cli/top/component/crumbs.go
@@ -35,11 +35,7 @@ func NewCrumbs(config *config.ThemeConfig) *Crumbs {
 		Flex:  tview.NewFlex(),
 		style: config,
 	}
-	c.init()
 	return c
-}
-
-func (c *Crumbs) init() {
 }
 
 // StackPop change itself when accept "pop" notify from app's main view
--- a/references/cli/top/model/pod.go
+++ b/references/cli/top/model/pod.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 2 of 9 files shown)


chivalryq (commented):
Just remove "cluster" field is OK, but can we pass the cluster info from the "managed resource" panel?

Hanmengnan:
> Just remove "cluster" field is OK, but can we pass the cluster info from the "managed resource" panel?

Can we guarantee the cluster of pod is the same as the cluster of managed resource the pod belonging to?

chivalryq:
> Can we guarantee the cluster of pod is the same as the cluster of managed resource the pod belonging to?

If the pod is in a different cluster as managed resource it belongs to, then the managed resource have the capability to dispatch resources across the clusters. For now I think we have  few use case like that. So it's ok to pass the cluster info.

Hanmengnan:
> > Can we guarantee the cluster of pod is the same as the cluster of managed resource the pod belonging to?
> 
> If the pod is in a different cluster as managed resource it belongs to, then the managed resource has the capability to dispatch resources across the clusters. For now I think we have few use case like that. So it's ok to pass the cluster info.

I got it. I have updated code.

Somefive (approved):

PR #5246, opened 2022-12-31
Fix: dry run from revision application patch error

LiuFang07:
Signed-off-by: old.prince <di7zhang@gmail.com>

Fixes:Dry-run from revision application,Problems caused by resource version lower than the current version

I have:

### DIFF
--- a/pkg/apiserver/domain/service/application.go
+++ b/pkg/apiserver/domain/service/application.go
@@ -1513,6 +1513,13 @@ func (c *applicationServiceImpl) DryRunAppOrRevision(ctx context.Context, appMod
 		}
 	case "REVISION":
 		app, _, err = c.getAppModelFromRevision(ctx, appModel.Name, dryRunReq.Version)
+		originalApp := &v1beta1.Application{}
+		if err := c.KubeClient.Get(ctx, types.NamespacedName{
+			Name:      app.Name,
+			Namespace: app.Namespace,
+		}, originalApp); err == nil {
+			app.ResourceVersion = originalApp.ResourceVersion
+		}
 		if err != nil {
 			return nil, err
 		}


pkg/apiserver/domain/service/application.go
```diff
@@ -1513,6 +1513,13 @@ func (c *applicationServiceImpl) DryRunAppOrRevision(ctx context.Context, appMod
 		}
 	case "REVISION":
 		app, _, err = c.getAppModelFromRevision(ctx, appModel.Name, dryRunReq.Version)
+		originalApp := &v1beta1.Application{}
+		if err := c.KubeClient.Get(ctx, types.NamespacedName{
+			Name:      app.Name,
+			Namespace: app.Namespace,
+		}, originalApp); err == nil {
+			app.ResourceVersion = originalApp.ResourceVersion
```
barnettZQG, line 1521: If not set the resource version, what error will occur?

LiuFang07: k8s will find that the current resource version is higher than the dryrun resource version, causing the update to fail.
detailed errors:cannot patch object: Operation cannot be fulfilled on applications.core.oam.dev \"demo\": the object has been modified; please apply your changes to the latest version and try again

StevenLeiZhang: > k8s will find that the current resource version is higher than the dryrun resource version, causing the update to fail. detailed errors:cannot patch object: Operation cannot be fulfilled on applications.core.oam.dev "demo": the object has been modified; please apply your changes to the latest version and try again

Should we update the Dryrun version, when deploy Application? In order to keep Dryrun version follows the deployed Application

barnettZQG: @LiuFang07  How to reproduce? Generating the application spec from the application config or revision should get a consistent result. So, if we need to set the resource version, both ways need to do it.

LiuFang07: @barnettZQG 
reproduce:
curl --location --request POST 'http://example.doamin.com/api/v1/applications/policy/dry-run' \
--header 'Authorization: Bearer xxxxxx' \
--header 'Content-Type: application/json' \
--data-raw '{
    "dryRunType": "REVISION",
    "version": "20221230183820839" //not the currently running version
}'


Now,Generating the application spec from the application config,The renderOAMApplication method set the resource version.

barnettZQG (approved):
LGTM

PR #5253, opened 2023-01-03
Test: prevent notification step definition test from failing when re-run.

jamesdobson:

The E2E test `Test notification step definition` in `definition_test.go` fails when it is run on a cluster where it has already run.

The problem is that it creates resources in the `default` namespace instead of in the namespace created for the test case. The resources in `default` do not get cleaned up when the test finishes, so subsequent test executions fail.

I have:

- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary. (seems not necessary)

I tested by:
1) Setting up the code without the fix
2) Running `make e2e-test` once
3) Observing the `noti` and `noti-secret` Applications and the `my-secret` Secret still present in the `default` namespace after the tests have finished.
4) Running `make e2e-test` again and seeing the test fail
5) Manually removing the Applications and Secret from the `default` namespace.
6) Applying the fix in this PR.
7) Ran `make e2e-test` to see the test pass
8) Ran `make e2e-test` again to see the test continue to pass
9) Confirmed that the Applications and Secret are not present in the `default` namespace.

None.

### DIFF
--- a/test/e2e-test/definition_test.go
+++ b/test/e2e-test/definition_test.go
@@ -227,25 +227,37 @@ var _ = Describe("ComponentDefinition Normal tests", func() {
 	})
 
 	It("Test notification step definition", func() {
+		By("Install notification workflow step definition")
 		_, file, _, _ := runtime.Caller(0)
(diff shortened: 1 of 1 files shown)


wonderflow (approved):
great job, thanks!

PR #5260, opened 2023-01-04
Fix: can't find the component pod view

barnettZQG:

Managing resources at the same time for helm and addon causes the resource to be deleted when the chart is upgraded. Now, we only manage the common velaQL view in the chart.

If users encounter the ConfigMap resource conflict when upgrading the vela-core, they only need to delete the ConfigMap and retry.

I have:

### DIFF
--- a/charts/vela-core/templates/velaql/component-pod.yaml
+++ b/charts/vela-core/templates/velaql/component-pod.yaml
@@ -1,4 +1,3 @@
-{{- if not (lookup "v1" "ConfigMap" (include "systemDefinitionNamespace" .) "component-pod-view") }}
 apiVersion: v1
 data:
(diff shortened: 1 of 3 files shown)


wonderflow:
>If users encounter the ConfigMap resource conflict when upgrading the vela-core, they only need to delete the ConfigMap and retry.

It's not convenient, and will cause lot's of questions in the community. We have a `vela install` command, can we handle that in the command?

barnettZQG:
@wonderflow Good idea. I will add a start check in `vela install`.

barnettZQG:
@wonderflow  PTAL

barnettZQG:
Fixes by #5268

Issue #5191, opened 2022-12-14
多集群交付应用然后删除application PHASE一直处于deleting


wxuedong:
vela veriosn  1.6.2

手动删除应用后application一直存在 且PHASE 是deleting
 ```
$ k get application.core.oam.dev  | grep deleting
multi-cluster-addon-fluxcd                    flux-system-namespace   raw           deleting         true               3h39m
 ```

应用定义：

```
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: multi-cluster-addon-fluxcd
  namespace: control-system
spec:
  components: 
    - name: flux-system-namespace 
      type: raw
      properties:
        ----
        ----
  policies:
  - name: topology-vela-clusters
    type: topology
    properties:
    - name: topology-clusters
      type: topology
      properties:
        clusterLabelSelector:
           xxx:xxx

```

wxuedong:
事件错误信息：

```
Warning  FailedGC  5m38s               Application  failed to sweep resourcetrackers to be deleted: failed to get resource ....: dial tcp xxx.xxx.xxx: connect: connection timed out
```

Somefive:
Seems your application is unable to clean up the target resource because the remote cluster lost connection?

wxuedong:
> Seems your application is unable to clean up the target resource because the remote cluster lost connection?

也许是连接丢失了 ，我认为在丢失连接的时候应该不影响其他集群的操作，当前的现象是卡住不动了

Somefive:
For this application, you are trying to delete it. If one of the cluster (where the application has resources in) lost connection, it will block the deletion of this application. The other applications that do not use this broken cluster should not be affected.

The reason for the blocking of deletion is one of our design. Because there could be scenarios when cluster lost connection but join back later. In this case we do not want the application to skip the resource recycle in the temporarily lost cluster. But it does cause some inconvenience for the usage. We are planning to make some optimization to help improve the experiences but the blocking behaviour will remain the same.

Issue #4700, opened 2022-09-07
delete application containing forbidden object is failed due to FailedGC


TommyHwang:
**Describe the bug**

When attempt to delete an application that performs a forbidden action(e.g. create ConfigMap in `kube-system`) with vela-core that uses `authentication-with-user` option, it is failed to delete the resourcetracker and stopped indefinitely.

**To Reproduce**

1. Run vela-core with `authentication-with-user` option
2. Create user and rbac

<details>

```bash
vela auth gen-kubeconfig --user test-user
```

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: test-user
  namespace: default
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: test-user
```

</details>

3. Create an application that contains objects in forbidden namespace (e.g. vela-system)

<details>

```yaml
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: k8s-objects
  namespace: default
spec:
  components:
  - name: configmap
    type: k8s-objects
    properties:
      objects:
        - apiVersion: v1
          kind: ConfigMap
          metadata:
            name: k8s-objects
            namespace: vela-system
          data:
            valid "true"
```

</details>

4. Attempt to delete the application

```
kubectl delete application k8s-objects
```

5. Check the error

```bash
    Steps:
      First Execute Time:  2022-09-07T06:40:51Z
      Id:                  73t92xc8rc
      Last Execute Time:   2022-09-07T06:41:43Z
      Message:             run step(provider=oam,do=component-apply): Dispatch: Found 1 errors. [(cannot get object: configmaps "k8s-objects" is forbidden: User "test-user" cannot get resource "configmaps" in API group "" in the namespace "vela-system")]
      Name:                configmap
      Phase:               failed
      Reason:              FailedAfterRetries
      Type:                apply-component
    Suspend:               false
    Terminated:            true
Events:
  Type     Reason           Age                    From         Message
  ----     ------           ----                   ----         -------
  Normal   Revisioned       5m48s (x6 over 5m53s)  Application  Revisioned successfully
  Normal   PolicyGenerated  5m48s (x6 over 5m53s)  Application  Policy generated successfully
  Normal   Rendered         5m48s (x6 over 5m53s)  Application  Rendered successfully
[...]

PR #5266, opened 2023-01-04
Feat: support interactive mode to manually skip encountered errors

Somefive:

Rework `vela delete` command. Now `vela delete` supports:
1. delete multiple applications `vela delete app-1 app-2 -n example`
2. delete application with orphan option `vela delete app-1 --orphan`
3. **NEW** `vela delete` by default will wait and print out the deletion error message controller encountered. This can be helpful for user to understand why the deletion of application has not finished and know what specific resource deletion is blocking the deletion of the application.
4. `vela delete -f` inherits the function provided by old codes but would might be refactored in the future. It now works more like delete at client-side.
5. **NEW** support interactive deletion for application. This is useful when unexpected error happens to the recycle of resources, like cluster disconnection or authorization failure. In other words, it provides a solution for #4700 and #5191.

```shell
Delete applications

 Delete KubeVela applications. KubeVela application deletion is associated with the recycle of underlying resources. By
default, the resources created by the KubeVela application will be deleted once it is not in use or the application is
deleted. There is garbage-collect policy in KubeVela application that you can use to configure customized recycle rules.

 This command supports delete application in various modes. Natively, you can use it like "kubectl delete app<app-name>
". In the cases you only want to delete the application but leave the resources there, you can use the --orphan
parameter. In the cases the server-side controller is uninstalled, or you want to manually skip some errors in the
deletion process (like lack privileges or handle cluster disconnection), you can use the --force parameter.

Usage:
  vela delete

Examples:
  # Delete an application
  vela delete my-app
  
  # Delete multiple applications in a namespace
  vela delete app-1 app-2 -n example
  
  # Delete all applications in one namespace
  vela delete -n example --all
  
  # Delete application without waiting to be deleted
  vela delete my-app --wait=false
  
  # Delete application without confirmation
  vela delete my-app -y
  
  # Force delete application at client-side
  vela delete my-app -f
  
  # Delete application by orphaning resources and skip recycling them
  vela delete my-app --orphan
  
  # Delete application interactively
  vela delete my-app -i

Flags:
      --all                delete all the application under the given namespace
  -e, --env string         The environment name for the CLI request
  -f, --force              force delete the application
  -h, --help               help for delete
  -i, --interactive        delete the application interactively
  -n, --namespace string   If present, the namespace scope for this CLI request
  -o, --orphan             delete the application and orphan managed resources
  -w, --wait               wait util the application is deleted completely (default true)

Global Flags:
```

Demo:
ezgif-3-0599f41d23

I have:

### DIFF
--- a/references/cli/delete.go
+++ b/references/cli/delete.go
@@ -90,6 +90,7 @@ func NewDeleteCommand(c common2.Args, order string, ioStreams cmdutil.IOStreams)
 	cmd.PersistentFlags().BoolVarP(&o.Wait, "wait", "w", false, "wait util the application is deleted completely")
 	cmd.PersistentFlags().BoolVarP(&o.ForceDelete, "force", "f", false, "force to delete the application")
 	cmd.PersistentFlags().BoolVarP(&o.Orphan, "orphan", "o", false, "delete the application and orphan managed resources")
+	cmd.PersistentFlags().BoolVarP(&o.IgnoreErrors, "ignore-errors", "", false, "delete the application and ignore deletion errors")
 	addNamespaceAndEnvArg(cmd)
 	return cmd
 }
--- a/references/cli/delete.go
+++ b/references/cli/delete.go
@@ -17,80 +17,384 @@ limitations under the License.
 package cli
 
 import (
-	"errors"
+	"context"
 	"fmt"
+	"strings"
+	"time"
 
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/kubevela/pkg/multicluster"
+	"github.com/kubevela/pkg/util/slices"
 	"github.com/spf13/cobra"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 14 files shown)


wonderflow (approved):

references/cli/delete.go
```diff
@@ -17,79 +17,384 @@ limitations under the License.
...
+	}
+	if mr.Namespace != "" {
+		src += fmt.Sprintf(", namespace %s", mr.Namespace)
+	}
+	groups := strings.Split(mr.APIVersion, "/")
+	group := "." + groups[0]
+	if len(groups) == 0 {
+		group = ""
+	}
+	return fmt.Sprintf("%s%s %s %s", strings.ToLower(mr.Kind), group, mr.Name, src)
```
barnettZQG, line 230: The option is not clear. If there are multiple resource trackers that include the same resource, how do the users select the resource that wanted to delete?

Could we mark the error resource?

Somefive: That resource will be removed from all ResourceTrackers.

I think marking error resource itself would not solve the probelm?

barnettZQG (approved):

FogDong (approved):

chivalryq (approved):

PR #5295, opened 2023-01-10
Fix: don't return err if subresource type is not found when listing application resources.

hnd4r7:
Signed-off-by: hnd4r7 <307365651@qq.com>

Subresource list loop should continue if not found.

`ChildrenResourcesRule` of `HelmRelease` workload has 2 subresource crd type that cannot be found in my remote cluster(my `fluxcd` addon version 2.1.9 and vela-core version 1.6.5 ).  The method `iterateListSubResources` then returns err `the server could not find the requested resource (get gateways.gateway.networking.k8s.io)` which will cause other subresources (such as pods of statefulSet) of my helm workload cannot be seen as well.
```
{
	ResourceType: ResourceType{APIVersion: "gateway.networking.k8s.io/v1alpha2", Kind: "HTTPRoute"},
},
{
	ResourceType: ResourceType{APIVersion: "gateway.networking.k8s.io/v1alpha2", Kind: "Gateway"},
},
```

Fixes #

I have:

- [X] Read and followed KubeVela's [contribution process](https://github.com/kubevela/kubevela/blob/master/contribute/create-pull-request.md).
- [X] Run `make reviewable` to ensure this PR is ready for review.

### DIFF
--- a/pkg/velaql/providers/query/tree.go
+++ b/pkg/velaql/providers/query/tree.go
@@ -25,6 +25,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	v12 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


chivalryq:
Thanks for your contribution, here is some discuss about this patch since there is not any issues linked.

1. Can you give an screenshot of problem this PR focused on? And one after fix.
2. Please add some test for this fix.

wonderflow (commented):
please sign the DCO

hnd4r7:
Here is the problem. 

image

I cannot list pods of helm workloads in a remote cluster.  But It works fine in my local cluster.  After digging into the source code,  I find the problem is that the crd `gateways.gateway.networking.k8s.io` is registered in my local cluster but not in the remote cluster. 

image.

image

I do not know how the crd is used.  But it can be found in the kubevela repo. [gateway.networking.k8s.io_gateways.yaml](https://github.com/kubevela/kubevela/blob/master/pkg/velaql/providers/query/testdata/gateway/crds/gateway.networking.k8s.io_gateways.yaml)

Unit tests in [pkg/velaql/providers/query/tree_test.go](https://github.com/kubevela/kubevela/blob/master/pkg/velaql/providers/query/tree_test.go#L1557 ) have covered the `meta.IsNoMatchError` and the `runtime.IsNotRegisteredError` which will be caught in the local master cluster.

But in my case, the crd type is registered in the master cluster, but not in the remote cluster. So the list request goes through the master cluster to the remote cluster apiserver which doesn't have that crd type.

Whatever, I think the subresources listing process should continue in cases the type not found error occurred.

Before fix, 
image

After fix,
image

wonderflow (approved):

PR #5360, opened 2023-01-18
Feat: support sharding in controller

Somefive:

# Application Controller Sharding

### Background

As many adopters start to choose KubeVela to manage thousands of Applications in production, the performance issue becomes increasingly critical.

The typical Kubernetes controller usually run one controller to manage all related custom resources. The operator reconciliation is designed to be relatively lightweight. In KubeVela, to support customized delivery process and manage lifecycles of tens of resources in one application makes the reconciliation process heavier compared to typical controllers.

Although multiple replicas of KubeVela controller could provide high availability, in avoid of multiple concurrent reconciliation for the same application (which could lead to conflict problems), only one replica could work and this is achieved by acquiring the same leader-election lock.

So usually, users add more resources to KubeVela controller (more CPUs, Memory...) and set up more threads and larger rate limiter to support more applications in the system. This can lead to potential problems in different cases:
1. There is limitations for the growth of resources for a single KubeVela controller. If the Kubernetes cluster of the control plane uses lots of small machines, a single KubeVela controller can only run one of them, instead of spreading out.
2. The failure of the single KubeVela controller could block all the delivery process of applications. The failure could be caused by various reasons and some frequently seen reasons (like OOM or Crash due to some weird application input) are not recoverable by restarting controller.
3. In multi-tenancy scenario, there is no fairness that could be guaranteed if some users have a huge number of applications. They could make the controller run slow and users with a small number of applications could also be affected, due to the sharing of the single KubeVela controller.

Therefore, this KEP proposes to support dividing a single large KubeVela controller into multiple small application controllers for run.

### Proposal

image

When running with `--enable-sharding` arg in KubeVela core controller, the KubeVela core controller will run in sharding mode.
- If the `--shard-id` arg is set to `master`, this one will run in **master** mode.
- If the `--shard-id` arg is set to some other value, like `shard-0`, this one will run in **slave** mode.

#### master mode

The **master** mode will enable all the controllers, such as ComponentDefinitionController, TraitDefinitionController, ApplicationController, Webhooks, etc.

The application controller will only handle applications with the label `controller.core.oam.dev/scheduled-shard-id: master`, and only the applications, applicationrevisions and resourcetrackers that carries this label will be watched and cached.

By default, it will watch the pods within the same namespace as it runs. The pods with labels `app.kubernetes.io/name: vela-core` and carries `controller.core.oam.dev/shard-id` label key will be selected and their health status will be recorded. The selected ready pods will be registered as schedulable shards. The mutating Webhook will automatically assign shard-id for non-assigned applications when requests come in.

#### slave mode

The **slave** mode controller will only start the ApplicationController and will not enable others like Webhook or ComponentDefinitionController. It is dedicated to applications that carries the matched labels `controller.core.oam.dev/scheduled-shard-id=<shard-id>`.

### Example

First, install KubeVela with sharding enabled. This will let the KubeVela core controller deployed in **master** mode.

```shell
helm install kubevela kubevela/vela-core -n vela-system --set sharding.enabled=true
```

Second, deploy **slave mode** application controller.

There are different ways to do it.
1. Use addons for installation. `vela addon enable vela-core-shard-manager nShards=3`. Supported by https://github.com/kubevela/catalog/pull/606
2. Use kubectl to copy the deployment of **master** and modify it. `kubectl get deploy kubevela-vela-core -oyaml -n vela-system | sed 's/schedulable-shards=/shard-id=shard-0/g' | sed 's/instance: kubevela/instance: kubevela-shard/g' | sed 's/shard-id: master/shard-id: shard-0/g' | sed 's/name: kubevela/name: kubevela-shard/g' | kubectl apply -f -` This will create a copy of the **master** vela-core and run a **slave** replica with shard-id as *shard-0*.
3. Just create a new deployment and set the labels to match the above mentioned requirements.

> In the case you do not want dynamic discovery for available application controller, you can specify what shards are schedulable by add arg `--schedulable-shards=shard-0,shard-1` to the **master mode** vela-core.

### Future Work

This Webhook implemented will only schedule applications when
1. The application is going to be mutated (create or update).
2. The application has not been scheduled yet (no `controller.core.oam.dev/scheduled-shard-id` label).
3. There is available controller to assign (the **master mode** controller will not be automatically schedulable in order to divide the burden of handling application into **slave mode** controllers).

So it cannot handle re-schedule scenario and cannot make automatic scheduling for previous application.

For the next step, we can support:
1. `vela up <app-name> --shard-id <shard-id>` Use vela CLI to help user manually reschedule application. NOTE: reschedule application not only need to reset the label of that application but also need to reset the labels of related applicationrevisions and resourcetrackers.
2. Support a background scheduler or a scheduling script to make schedule for all unscheduled application.
3. Support `DisableAutoSchedule` for the case where user wants to disable the automatic schedule for the mutating webhook.
4. Observability tools (Prometheus, Loki, Grafana) should be able to collect data from all vela-core controllers in sharding mode.

### Extend Usage
1. The sharding package could potentially be moved to `kubevela/pkg` if later we find it also helpful for other controller like `kubevela/workflow`.
2. It is possible to let different shards to use different vela-core version so as to handle applications in multiple versions by different controller.
3. It is possible to let different tenant to use different vela-core controller to help improve fairness and reduce the damage scope when one controller fails.
4. The original multi-replica KubeVela core controller for high availability is deferred to each shard's deployment, which means each shard could have multiple replicas (and only one can work for one time, but multiple shards can be parallelized).

### Tech Details
1. Since the cache of Application/ApplicationRevision/ResourceTracker are divided into multiple shards, the memory consumption of each controller is expected to be divided as well. Other resources like `ConfigMap` (used by Workflow Context) are not divided.
2. The validating webhook needs to read ApplicationRevision when `publish-version` is set in application, but since ApplicationRevision is sharded, therefore the cache of the vela-core in **master mode** will not hold the ApplicationRevision. So here we use native Kubernetes API request to read ApplicationRevision. This may lead to performance drop for validating webhook and this is disabled by default when sharding enabled.

### FAQ

Q: What will happen when the master one is down?
A: If the webhook is enabled, the webhook will be down if the master one is down. Then the mutating webhook and validating webhook will fail so no new application creation or change will be accepted. Old applications that are scheduled to master will not be processed anymore. Others that are scheduled to other shards will not be affected. If the webhook is not enabled, then only applications scheduled to master shard will be affected. Users can still create or change applications in this case.

Q: What will happen when a slave mode controller is down?
A: For applications that are not scheduled to that shard, nothing will happen. For applications that are scheduled to the broken shard, succeeded applications will not run state-keep or gc but the delivered resources will not be touched. For applications that are still running workflow, they will not be handled but can be recovered instantly when the controller is restarted.

Q: What will happen if one user delete the sharding label while the application is already in running state?
A: If the webhook is enabled, this behaviour will be prevented. The sharding label will inherit the original one. If the webhook is not enabled, the application will not be handled anymore (no state-keep, no gc and no update).

I have:


pkg/controller/sharding/scheduler.go
```diff
@@ -0,0 +1,215 @@
...
+	return available
+}
+
+func (in *dynamicDiscoveryScheduler) schedule() (string, bool) {
+	available := in.availableShards()
+	if len(available) == 0 {
+		return "", false
+	}
+	// nolint
+	return available[rand.Intn(len(available))], true
```
barnettZQG, line 201: Advice uses Round Robin as the default schedule rule.

wonderflow:
We need to discuss the failover case and best configuration:

* when the master mode controller is down, everything is down. so we still need at least 3 replicas for master mode controller.
* when the slave mode controller (naming `slave-0`) is down, what will happen to these two cases?
   1. applications in scheduling that marked as `slave-0` to handle.
   2. applications already handled by `slave-0` but with error that needs to reconcile again.
   3. how can we discover applications in the above two scenarios?

Another question is, what will happen if one user delete the sharding label while the application is already in running state?

charts/vela-core/README.md
```diff
@@ -153,6 +153,8 @@ helm install --create-namespace -n vela-system kubevela kubevela/vela-core --wai
 | `authentication.withUser`     | Application authentication will impersonate as the request User                                                            | `true`               |
 | `authentication.defaultUser`  | Application authentication will impersonate as the User if no user provided in Application                                 | `kubevela:vela-core` |
 | `authentication.groupPattern` | Application authentication will impersonate as the request Group that matches the pattern                                  | `kubevela:*`         |
+| `sharding.enabled`            | Enable sharding for core controller                                                                                        | `false`              |
```
wonderflow, line 156: should the default value to be true?

pkg/controller/sharding/scheduler.go
Somefive: Good idea.

charts/vela-core/values.yaml
```diff
@@ -268,3 +268,9 @@ authentication:
   withUser: true
   defaultUser: kubevela:vela-core
   groupPattern: kubevela:*
+
+## @param sharding.enabled Enable sharding for core controller
+## @param sharding.shards The shards available for scheduling. If empty, dynamic discovery will be used.
+sharding:
+  enabled: false
```
wonderflow, line 275: why don't make the default value to be true?

wonderflow: what should one user do when they want to upgrade to the sharding mode when they don't enable it at first installation

wonderflow: I noticed we need new role and rolebinding configuration, that might be hard for users to learn

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/oam-dev/kubevela
 go 1.19
 
 require (
-	cuelang.org/go v0.5.0-alpha.1
+	cuelang.org/go v0.5.0-beta.2
```
wonderflow, line 6: shall we upgrade the cue lib here?  \cc @FogDong

hack/e2e/end_e2e_core.sh
```diff
@@ -12,7 +13,7 @@ while [ $n -le 60 ];do
     if_exist=$(docker exec k3d-k3s-default-server-0 sh -c "test -f $file && echo 'ok'")
     echo "$if_exist"
     if [ -n  "$if_exist" ];then
-        docker exec k3d-k3s-default-server-0 cat "$file" > /tmp/e2e-profile.out
+        docker exec k3d-k3s-default-server-0 cat "$file" >> /tmp/e2e-profile.out
```
wonderflow, line 16: Why we need to change it here? \cc @chivalryq

pkg/controller/sharding/flags.go
```diff
@@ -0,0 +1,29 @@
...
+
+package sharding
+
+import (
+	"github.com/spf13/pflag"
+)
+
+// AddFlags add sharding flags
+func AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&EnableSharding, "enable-sharding", EnableSharding, "Whether to enable sharding.")
```
wonderflow, line 25: does the salve mode controller need to specify this?

```suggestion
	fs.BoolVar(&EnableSharding, "enable-sharding", EnableSharding, "When sharding enabled, the controller will run as..., refer to ... for details.")
```

pkg/controller/sharding/flags.go
```diff
@@ -0,0 +1,29 @@
...
+	fs.StringVar(&ShardID, "shard-id", ShardID, "The id for sharding. If empty, no sharding.")
+	fs.StringSliceVar(&SchedulableShards, "schedulable-shards", SchedulableShards, "The shard ids that are available for scheduling. If empty, dynamic discovery will be used.")
+	fs.DurationVar(&DynamicDiscoverySchedulerResyncPeriod, "dynamic-discovery-scheduler-resync-period", DynamicDiscoverySchedulerResyncPeriod, "The resync period for default dynamic discovery scheduler.")
```
wonderflow, line 28:
```suggestion
	fs.DurationVar(&DynamicDiscoverySchedulerResyncPeriod, "sharding-slaves-discovery-period", DynamicDiscoverySchedulerResyncPeriod, "The resync period for default dynamic discovery scheduler.")
```

pkg/controller/sharding/scheduler.go
```diff
@@ -0,0 +1,215 @@
...
+// Start .
+func (in *staticScheduler) Start(ctx context.Context) {
+	klog.Infof("staticScheduler started, shards: [%s]", strings.Join(in.shards, ", "))
+}
+
+// Schedule the target object to a random shard
+func (in *staticScheduler) Schedule(o client.Object) bool {
+	if _, scheduled := GetScheduledShardID(o); !scheduled {
+		if len(in.shards) > 0 {
+			// nolint
```
wonderflow, line 70: for what?

Somefive:
> We need to discuss the failover case and best configuration:
> 
> * when the master mode controller is down, everything is down. so we still need at least 3 replicas for master mode controller.
> * when the slave mode controller (naming `slave-0`) is down, what will happen to these two cases?
>   
>   1. applications in scheduling that marked as `slave-0` to handle.
>   2. applications already handled by `slave-0` but with error that needs to reconcile again.
>   3. how can we discover applications in the above two scenarios?
> 
> Another question is, what will happen if one user delete the sharding label while the application is already in running state?

> when the master mode controller is down, everything is down

Precisely speaking, when the master mode controller is down, only webhook will be down, so application will not be able to mutate. The old applications are still ok to run.

Besides, in the case the webhook is not enabled (which also means automatic scheduling is not enabled), users can still create application and manually schedule them into working shards (by adding label).

It is true that the master one should have multiple replicas, but I think the main reason for that is for the webhook, not for the controller.

> * when the slave mode controller (naming `slave-0`) is down, what will happen to these two cases?

For the applications that are marked as `slave-0`, they will not be handled any more (no matter running or error), in this PR. That's why we have future work to do after this PR. There are two solutions.
1. Use `vela schedule` command to manually re-schedule the app to other active shards.
2. Use a background scheduler that will watch for the inactive applications and automatically re-schedule them.

> Another question is, what will happen if one user delete the sharding label while the application is already in running state?

This PR does not prevent modifying the sharding label. The removal of sharding label will let the application not be handled by any shards any more. In running state, nothing will happen. (Application will not run state-keep as well.)

hack/e2e/end_e2e_core.sh
Somefive: Because in the multicluster-e2e-test and apiserver-e2e-test, the sharding option is enabled and we have two controller logs to collect.

go.mod
Somefive: This upgrade is brought by the upgrade of kubevela/pkg. But in kubevela/kubevela, this upgrade will be overwritten by the `replace` below.

pkg/controller/sharding/scheduler.go
Somefive: rand.Intn will be reported as an issue in golang-ci-lint because it is pseudo random number.

pkg/controller/sharding/scheduler.go
wonderflow: I don't prefer the schedule algorithm, but do we expose metrics of how many applications the controller is handling?  This is important.

wonderflow:
we should turn off the validation webhook when opening the sharding mechanism.

pkg/controller/sharding/scheduler.go
Somefive: This is exposed by the standard metrics endpoint in Prometheus format. We need to add support for the Grafana to collect them.

charts/vela-core/README.md
Somefive: Currently no. If the default value set to be true, then when no shards added, application will not be handled.

pkg/controller/sharding/flags.go
Somefive: Yes.

wonderflow:
By default, the webhook will routing requests to all replicas no matter the replica is master or slave.

1. The webhook is pointing to the service.
```
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  annotations:
    meta.helm.sh/release-name: kubevela
    meta.helm.sh/release-namespace: vela-system
...
    service:
      name: vela-core-webhook
      namespace: vela-system
      path: /mutating-core-oam-dev-v1beta1-componentdefinitions
      port: 443
```

2. The service is selecting to all pods:

```
apiVersion: v1
kind: Service
metadata:
  name: vela-core-webhook
  namespace: vela-system
...
spec:
...
  ports:
  - name: https
    port: 443
    protocol: TCP
    targetPort: 9443
  selector:
    app.kubernetes.io/instance: kubevela
    app.kubernetes.io/name: vela-core
  sessionAffinity: None
  type: ClusterIP
```

pkg/controller/sharding/flags.go
Somefive: changed to sharding-slave-discovery-resync-period

charts/vela-core/values.yaml
Somefive: > why don't make the default value to be true?

By default, in the sharding mode, the master will not work on applications.

Somefive: > what should one user do when they want to upgrade to the sharding mode when they don't enable it at first installation

Just run `helm upgrade --set sharding.enabled=true`

Somefive: > I noticed we need new role and rolebinding configuration, that might be hard for users to learn

User does not need to know the role/rolebinding configuration.

wonderflow: > By default, in the sharding mode, the master will not work on applications.

You should add that to the doc and KEP, I missed this message at first time.

Somefive:
> we should turn off the validation webhook when opening the sharding mechanism. ping @barnettZQG maybe we can add that validate on dry-run.

FeatureGate `ValidateComponentWhenSharding` is added and by default it is set to false.

Somefive:
> we should turn off the validation webhook when opening the sharding mechanism. ping @barnettZQG maybe we can add that validate on dry-run.

FeatureGate `ValidateComponentWhenSharding` is added and by default it is set to false.

pkg/controller/sharding/scheduler.go
Somefive: > Advice uses Round Robin as the default schedule rule.

Added to the dynamicDiscoveryScheduler. The staticDiscoveryScheduler will use random still.

design/vela-core/sharding.md
```diff
@@ -0,0 +1,78 @@
...
+
+When running with `--enable-sharding` arg in KubeVela core controller, the KubeVela core controller will run in sharding mode.
+- If the `--shard-id` arg is set to `master`, this one will run in **master** mode.
+- If the `--shard-id` arg is set to some other value, like `shard-0`, this one will run in **slave** mode.
+
+#### master mode
+
+The **master** mode will enable all the controllers, such as ComponentDefinitionController, TraitDefinitionController, ApplicationController, Webhooks, etc.
+
+The application controller will only handle applications with the label `controller.core.oam.dev/scheduled-shard-id: ""`, and only the applications, applicationrevisions and resourcetrackers that carries this label will be watched and cached.
```
wonderflow, line 30: master?

pkg/controller/sharding/flags.go
```diff
@@ -0,0 +1,29 @@
...
+
+package sharding
+
+import (
+	"github.com/spf13/pflag"
+)
+
+// AddFlags add sharding flags
+func AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&EnableSharding, "enable-sharding", EnableSharding, "When sharding enabled, the controller will run as master (shard-id=master) or slave mode (shard-id is not master). Refer to https://github.com/kubevela/kubevela/blob/master/design/vela-core/sharding.md for details.")
```
wonderflow, line 25: please also update the chart value annotation

wonderflow, line 25:
```suggestion
	fs.BoolVar(&EnableSharding, "enable-sharding", EnableSharding, "When sharding enabled, the controller will run as master (shard-id=master) or slave mode (shard-id is any non-empty string except master). Refer to https://github.com/kubevela/kubevela/blob/master/design/vela-core/sharding.md for details.")
```

pkg/controller/sharding/scheduler.go
```diff
@@ -0,0 +1,219 @@
...
+			in.mu.Lock()
+			in.candidates = map[string]map[string]bool{}
+			in.mu.Unlock()
+			for _, obj := range in.store.List() {
+				in._registerPod(obj)
+			}
+			available := in.availableShards()
+			klog.Infof("dynamicDiscoveryScheduler resync finished, available shards: [%s]", strings.Join(available, ", "))
+		}
+		time.Sleep(in.resyncPeriod)
```
wonderflow, line 151: time.Ticker is better?

design/vela-core/sharding.md
Somefive: ok

pkg/controller/sharding/flags.go
Somefive: ok

pkg/controller/sharding/scheduler.go
Somefive: ok

pkg/controller/sharding/flags.go
Somefive: ok

charts/vela-core/values.yaml
Somefive: Now the master shard by default also accepts applications. So slaves are not required.

charts/vela-core/README.md
Somefive: Behaviour changed. Master shard by default accept applications now.

wonderflow (approved):

charts/vela-core/README.md
```diff
@@ -132,27 +134,29 @@ helm install --create-namespace -n vela-system kubevela kubevela/vela-core --wai
...
+| `authentication.enabled`      | Enable authentication for application                                                                                                                              | `false`              |
+| `authentication.withUser`     | Application authentication will impersonate as the request User                                                                                                    | `true`               |
+| `authentication.defaultUser`  | Application authentication will impersonate as the User if no user provided in Application                                                                         | `kubevela:vela-core` |
+| `authentication.groupPattern` | Application authentication will impersonate as the request Group that matches the pattern                                                                          | `kubevela:*`         |
+| `sharding.enabled`            | When sharding enabled, the controller will run as master mode. Refer to https://github.com/kubevela/kubevela/blob/master/design/vela-core/sharding.md for details. | `false`              |
```
StevenLeiZhang, line 158: user did not enable sharding, when install vela. After that, if user need enable it in order to manage more apps, can user enable sharding?

charts/vela-core/values.yaml
```diff
@@ -268,3 +272,9 @@ authentication:
   withUser: true
   defaultUser: kubevela:vela-core
   groupPattern: kubevela:*
+
+## @param sharding.enabled When sharding enabled, the controller will run as master mode. Refer to https://github.com/kubevela/kubevela/blob/master/design/vela-core/sharding.md for details.
+## @param sharding.schedulableShards The shards available for scheduling. If empty, dynamic discovery will be used.
+sharding:
```
StevenLeiZhang, line 278: Can `vela install ` command install kubevela with sharding enabled?

references/cli/up.go
```diff
@@ -272,6 +299,7 @@ func NewUpCommand(f velacmd.Factory, order string, c utilcommon.Args, ioStream u
 	cmd.Flags().StringVarP(&o.File, "file", "f", o.File, "The file path for appfile or application. It could be a remote url.")
 	cmd.Flags().StringVarP(&o.PublishVersion, "publish-version", "v", o.PublishVersion, "The publish version for deploying application.")
 	cmd.Flags().StringVarP(&o.RevisionName, "revision", "r", o.RevisionName, "The revision to use for deploying the application, if empty, the current application configuration will be used.")
+	cmd.Flags().StringVarP(&o.ShardID, "shard-id", "s", o.ShardID, "The shard id assigned to the application. If empty, it will not be used.")
```
StevenLeiZhang, line 302: If vela is installed with sharding mode, which controller will manage the Applications without label `controller.core.oam.dev/scheduled-shard-id` ?

charts/vela-core/README.md
Somefive: Yes, user can use `helm upgrade --set sharding.enabled` to enable it. But he needs to use `vela up --shard-id` command to manually schedule the original applications to `master` shard.

references/cli/up.go
Somefive: No controller will handle those applications. Those ones need to be scheduled through `vela up --shard-id` (provided in this PR) or customized scheduler.

charts/vela-core/values.yaml
Somefive: Yes.

design/vela-core/sharding.md
```diff
@@ -0,0 +1,89 @@
...
+
+The typical Kubernetes controller usually run one controller to manage all related custom resources. The operator reconciliation is designed to be relatively lightweight. In KubeVela, to support customized delivery process and manage lifecycles of tens of resources in one application makes the reconciliation process heavier compared to typical controllers.
+
+Although multiple replicas of KubeVela controller could provide high availability, in avoid of multiple concurrent reconciliation for the same application (which could lead to conflict problems), only one replica could work and this is achieved by acquiring the same leader-election lock.
+
+So usually, users add more resources to KubeVela controller (more CPUs, Memory...) and set up more threads and larger rate limiter to support more applications in the system. This can lead to potential problems in different cases:
+1. There is limitations for the growth of resources for a single KubeVela controller. If the Kubernetes cluster of the control plane uses lots of small machines, a single KubeVela controller can only run one of them, instead of spreading out.
```
FogDong, line 12:
```suggestion
1. There are limitations for the growth of resources for a single KubeVela controller. If the Kubernetes cluster of the control plane uses lots of small machines, a single KubeVela controller can only run one of them, instead of spreading out.
```

FogDong (approved):

charlie0129 (commented):
typo here

image

design/vela-core/sharding.md
Somefive: Fixed.

.github/workflows/apiserver-test.yml
```diff
@@ -180,7 +180,9 @@ jobs:
           make e2e-apiserver-test
 
       - name: Stop kubevela, get profile
-        run: make end-e2e-core
+        run: |
+          make end-e2e-core
+          CORE_NAME=kubevela-shard sh ./hack/e2e/end_e2e_core.sh
```
chivalryq, line 185: Since `end-e2e-core` is a makefile target, It'd be better if we wrap `CORE_NAME=kubevela-shard sh ./hack/e2e/end_e2e_core.sh` into another one to keep style here consistent.

Somefive: add `make end-e2e-core-shards`

chivalryq (approved):

PR #5379, opened 2023-01-30
Fix: use the namespace specified in the resource if -n is not s…

### LINKED ISSUE #5368: vela adopt should use namespace of the resource to be the default app namespace when not specified

When I try to adopt a resource as vela application, it reports an error:

```
$ vela adopt deployment/kube-system/metrics-server --apply
error: failed to apply application /metrics-server: cannot get object: an empty namespace may not be set when a resource name is provided
```

After debugging, I found I must specify the namespace for the application, so both the following cases work:

* Adopt application in `default` namespace:
```
$ vela adopt deployment/kube-system/metrics-server --apply  -n default
I0129 18:03:44.568639   60076 apply.go:121] "creating object" name="metrics-server" resource="core.oam.dev/v1beta1, Kind=Application"
resources adopted in app default/metrics-server
```

* Adopt application in `kube-system` namespace:
```
$ vela adopt deployment/kube-system/metrics-server --apply  -n kube-system
I0129 18:03:53.569639   60091 apply.go:121] "creating object" name="metrics-server" resource="core.oam.dev/v1beta1, Kind=Application"
resources adopted in app kube-system/metrics-server
```

This is not convenient that I must specify the namespace.

Use the namespace specified in the resource:

```
$ vela adopt deployment/kube-system/metrics-server --apply
resources adopted in app kube-system/metrics-server
```

**Additional context**

basuotian:
Fix #5368 

Signed-off-by: Basuotian <basuoluomiu@gmail.com>

I have:

### DIFF
--- a/references/cli/adopt.go
+++ b/references/cli/adopt.go
@@ -160,6 +160,9 @@ func (opt *AdoptOptions) Complete(f velacmd.Factory, cmd *cobra.Command, args []
 		}) {
 			opt.AppName = opt.NativeResourceRefs[0].Name
 		}
+		if opt.AppNamespace == "" {
+			opt.AppNamespace = opt.NativeResourceRefs[0].Namespace
+		}
 	case adoptTypeHelm:
 		if len(args) > 0 {
 			opt.HelmReleaseName = args[0]
--- a/references/cli/adopt.go
+++ b/references/cli/adopt.go
@@ -52,6 +52,7 @@ import (
 	velacmd "github.com/oam-dev/kubevela/pkg/cmd"
 	cmdutil "github.com/oam-dev/kubevela/pkg/cmd/util"
 	"github.com/oam-dev/kubevela/pkg/utils/apply"
+	"github.com/oam-dev/kubevela/pkg/utils/env"
 	"github.com/oam-dev/kubevela/pkg/utils/util"
 )
 
@@ -133,6 +134,9 @@ func (opt *AdoptOptions) parseResourceRef(f velacmd.Factory, cmd *cobra.Command,
 		or.Name = parts[1]
 		if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
 			or.Namespace = velacmd.GetNamespace(f, cmd)
+			if or.Namespace == "" {
+				or.Namespace = env.DefaultEnvNamespace
+			}
 		}
 	case 3:
 		or.Namespace = parts[1]
--- a/references/cli/adopt_test.go
+++ b/references/cli/adopt_test.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2021 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cli
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	velacmd "github.com/oam-dev/kubevela/pkg/cmd"
+	"github.com/oam-dev/kubevela/pkg/utils/util"
+)
+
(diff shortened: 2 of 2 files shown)


references/cli/adopt.go
```diff
@@ -160,6 +160,9 @@ func (opt *AdoptOptions) Complete(f velacmd.Factory, cmd *cobra.Command, args []
 		}) {
 			opt.AppName = opt.NativeResourceRefs[0].Name
 		}
+		if opt.AppNamespace == "" {
+			opt.AppNamespace = opt.NativeResourceRefs[0].Namespace
```
wonderflow, line 164: 1. what if the silce is zero?
2. what if the namespace is still empty? Can we downgrade to the environment( from vela env and the kubeconfig context)?

basuotian: 1. from previous code, if opt.parseResourceRef didn't return err, the slice would not be zero, else it would not reach the code
2. i will reconsider about the downgrade solution

basuotian: velacmd.GetNamespace is already able to downgrade to env and kubeconfig context when -n is not set, but after testing, i find the env usage is: 
- vela adopt deployment/kube-system/metrics-server -e default   (default namespace is set)
- vela adopt deployment/kube-system/metrics-server -e namespace=default (default namespace is not set, but this is more human-readable)

basuotian: velacmd.GetNamespace function is already able to downgrade to env and kubeconfig context when -n is not set, but i think the env logic is too subtle, it would match the first namespace with label `namespace.oam.dev/env` except default namespace, here is the case:

- default namespace is set
```
vela adopt deployment/monitoring-system/metrics-server -e default
```
- kube-system namespace is not set
```
vela adopt deployment/monitoring-system/metrics-server -e kube-system
```
- vela-system namespace is set as it has label `namespace.oam.dev/env=syc-vela-system`
```
vela adopt deployment/monitoring-system/metrics-server -e syc-vela-system
```

wonderflow: > velacmd.GetNamespace function

Great, just use this function should be fine.

> the env logic is too subtle

What's the results for these cases? I think we're discusing the case that we didn't specify any env or namespace, but your case specify env, then just use the env logic should be fine?

basuotian: The env logic would match the first namespace with label `namespace.oam.dev/env`, so the downgrade is  --env flags, then kubeconfig context.
I am not sure about the expect "env" downgrade is --env flags or just environment variable like `KUBE_NAMESPACE=kube-system`

basuotian: As the velacmd.GetNamespace is called many times, in this issue i think the env logic shoue be fine

wonderflow: if velacmd.GetNamespace works, then the `opt.AppNamespace` won't be empty, why could your changes take effect?

As to this issue, I think we need this case work:

```
vela adopt deployment/metrics-server --apply
```

It will adopt deployment in the default namespace named `metrics-server` to be an application in the default namespace.

wonderflow: Coud you add a test case for that?

basuotian: In velacmd.GetNamespace, the namespace comes from `--namespace > --env > kubeconfig context`, so if we don't provide any args above, the namespace would be empty.

My changes use the namespace of the k8s object resourcesRef (deployment/kube-system/metrics-server) when velacmd.GetNamespace get empty namespace, it would be the last downgrade choice, so this case can work:

```
$ vela adopt deployment/kube-system/metrics-server --apply
resources adopted in app kube-system/metrics-server
```

If we need this case work, it needs to change the opt.parseResourceRef and i will make another commit
```
vela adopt deployment/metrics-server --apply
```

wonderflow:
```
vela adopt deployment/metrics-server --apply
```

I think it means using the "default" namespace for deployment metrics-server.

basuotian: OK, the last commit implement this case:

```
$vela adopt deployment/nginx-deployment --apply
resources adopted in app default/nginx-deployment
```

wonderflow: great, can you help add some test cases

basuotian: done, add two test cases without -n

```
$ vela adopt deployment/kube-system/metrics-server --apply
resources adopted in app kube-system/metrics-server

$vela adopt deployment/nginx-deployment --apply
resources adopted in app default/nginx-deployment
```

wonderflow (approved):
LGTM after CI passed

wonderflow:
please format the code as the CI reported

Somefive (approved):

PR #5416, opened 2023-02-02
Feat: add deploy inline policy

Somefive:

Add deploy inline policy and support loading definitions when lack in the revision.

I have:

### DIFF
--- a/charts/vela-core/templates/defwithtemplate/deploy.yaml
+++ b/charts/vela-core/templates/defwithtemplate/deploy.yaml
@@ -26,7 +26,7 @@ spec:
         	//+usage=If set to false, the workflow will suspend automatically before this step, default to be true.
         	auto: *true | bool
         	//+usage=Declare the policies that used for this deployment. If not specified, the components will be deployed to the hub cluster.
-        	policies?: [...string]
+        	policies: *[] | [...string]
         	//+usage=Maximum number of concurrent delivered components.
         	parallelism: *5 | int
         	//+usage=If set false, this step will apply the components with the terraform workload.
--- a/vela-templates/definitions/internal/workflowstep/deploy.cue
+++ b/vela-templates/definitions/internal/workflowstep/deploy.cue
@@ -20,7 +20,7 @@ template: {
 		//+usage=If set to false, the workflow will suspend automatically before this step, default to be true.
 		auto: *true | bool
 		//+usage=Declare the policies that used for this deployment. If not specified, the components will be deployed to the hub cluster.
-		policies?: [...string]
+		policies: *[] | [...string]
 		//+usage=Maximum number of concurrent delivered components.
 		parallelism: *5 | int
 		//+usage=If set false, this step will apply the components with the terraform workload.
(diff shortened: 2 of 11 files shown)


charts/vela-core/templates/defwithtemplate/deploy.yaml
```diff
@@ -21,15 +21,23 @@ spec:
         	policies:                 parameter.policies
         	parallelism:              parameter.parallelism
         	ignoreTerraformComponent: parameter.ignoreTerraformComponent
+        	inlinePolicies:           parameter.inlinePolicies
+        }
+        #Policy: {
+        	type:  string
+        	name?: string
+        	properties: {...}
```
FogDong, line 29: Will the properties be empty?

Somefive: > Will the properties be empty?

No, the policy type here needs to be one of "topology", "override", "replication". I have just added constraints on the policy type.

chivalryq (commented):
If I get it right, we allow to `ParseWorkloadFromRevisionAndClient` because the inline override policy can introduce new Component/Trait? And the component/trait referred in non-inline policy have been added into AppRevision.

Somefive:
> If I get it right, we allow to `ParseWorkloadFromRevisionAndClient` because the inline override policy can introduce new Component/Trait? And the component/trait referred in non-inline policy have been added into AppRevision.

Yes. Because inlinePolicies can be patched through workflow input/output so it is hard to detect definitions that needs to be used during the CreateAppRevision stage before workflow runs.

chivalryq (approved):

vela-templates/definitions/internal/workflowstep/deploy.cue
```diff
@@ -15,15 +15,23 @@ template: {
...
 		auto: *true | bool
 		//+usage=Declare the policies that used for this deployment. If not specified, the components will be deployed to the hub cluster.
-		policies?: [...string]
+		policies: *[] | [...string]
 		//+usage=Maximum number of concurrent delivered components.
 		parallelism: *5 | int
 		//+usage=If set false, this step will apply the components with the terraform workload.
 		ignoreTerraformComponent: *true | bool
+		//+usage=The policies that embeds in the `deploy` step directly
+		inlinePolicies: *[] | [...#Policy]
```
barnettZQG, line 35: Adding this field for the `deploy` step may not be a good idea. This leads the user to choose between the `policies` and `inlinePolicies` fields. The `policies` field is recommended. 

But, If we want to define an `xx-deploy` step type with some default policies, this field is very useful.

Somefive: I see. So you recommend not exposing inlinePolicies in deploy.cue but in op.#Deploy right?

barnettZQG: Yes

Somefive: Updated.

barnettZQG (approved):
LGTM

FogDong (approved):

PR #5479, opened 2023-02-10
Feat: add cli command `vela metrics` to display resource number and resource status 

### LINKED ISSUE #5374: [Feature] Add `--metrics` option to `vela status` command

The resource metrics currently used by the application is a very important status information, which can help developers understand the current resource usage of the application in detail. 

Therefore, I hope we can introduce an option `-- metrics` in the `vela status` to display the resource indicator status.

These informations can be divided into two categories:
- Number of related resources: Managed resource, Pod, Container, Node, Cluster.
- Physical resource metrics: Storage, CPU, Memory, Network Received Bytes, Network Send Bytes

**Additional context**

Hanmengnan:
Signed-off-by: HanMengnan <1448189829@qq.com>

The resource metrics currently used by the application is a very important status information, which can help developers understand the current resource usage of the application in detail.

Therefore, I want to add a cli command ` vela metrics` and introduce a new option `--metrics` to `vela status` to display the resource metrics status.

These information can be divided into two categories:

- Number of Kubernetes resources: 
  - Managed resource
  - Pod
  - Container
  - Node
  - Cluster
- Underlying Physical Resources consumed: 
  - Storage
  - CPU 
  - Memory

> Alter: To collect CPU, Memory metrics, you need install metrics-server in your Kubernetes env.

image

image

Fixes #5374

I have:


wonderflow:
Great job! 

`vela status --metrics` is fine, introducing `vela metrics` command seems a bit overkill.

Let's hold the use of `vela metrics`, I think when we have more features on observability, we can use this command.

references/cli/status.go
```diff
@@ -182,6 +195,7 @@ func NewAppStatusCommand(c common.Args, order string, ioStreams cmdutil.IOStream
 	cmd.Flags().BoolVarP(&detail, "detail", "d", false, "display more details in the application like input/output data in context. Note that if you want to show the realtime details of application resources, please use it with --tree")
 	cmd.Flags().StringP("detail-format", "", "inline", "the format for displaying details, must be used with --detail. Can be one of inline, wide, list, table, raw.")
 	cmd.Flags().StringVarP(&outputFormat, "output", "o", "", "raw Application output format. One of: (json, yaml, jsonpath)")
+	cmd.Flags().BoolP("metrics", "m", false, "show resource num nad resource metrics of the application")
```
wonderflow, line 198:
```suggestion
	cmd.Flags().BoolP("metrics", "m", false, "show resource quota and consumption metrics of the application")
```

references/cli/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+	fmt.Printf("    * Number of Pods: %d\n", metrics.Resource.PodNum)
+	fmt.Printf("    * Number of Containers: %d\n", metrics.Resource.ContainerNum)
+	fmt.Printf("    * Number of Managed Resource: %d\n", metrics.Resource.SubresourceNum)
+	fmt.Printf("    * Number of Nodes: %d\n", metrics.Resource.NodeNum)
+	fmt.Printf("    * Number of Clusters: %d\n", metrics.Resource.ClusterNum)
+	fmt.Println()
+	fmt.Printf("Underlying Physical Resoures consumed:\n")
+	fmt.Printf("    * Total CPU(cores): %d m\n", metrics.Status.CPU)
+	fmt.Printf("    * Total MEMORY(bytes): %d Mi\n", metrics.Status.Memory)
+	fmt.Printf("    * Total Storage(bytes): %d Gi\n", metrics.Status.Storage)
```
wonderflow, line 99: can we also check the resource limit? while it's fine to seprate the feature into another PR.

references/cli/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/oam-dev/kubevela/apis/types"
+	"github.com/oam-dev/kubevela/pkg/utils/common"
+	cmdutil "github.com/oam-dev/kubevela/pkg/utils/util"
+	referencesCommon "github.com/oam-dev/kubevela/references/common"
+)
+
+// NewAppMetricsCommand creates metrics command
+func NewAppMetricsCommand(c common.Args, order string, ioStreams cmdutil.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "metrics APP_NAME",
```
wonderflow, line 37: let's use `vela status --metrics` only.

wonderflow:
The warning log is annoying, let's try to remove it. Does the message important? @wangyikewxgm 

image

references/cli/metrics.go
Hanmengnan: ok

references/cli/metrics.go
Hanmengnan: ok, i will add them later.

references/cli/status.go
Hanmengnan: ok, I will fix it.

StevenLeiZhang:
since App may expose customized metrics except  CPU/MEM/storage, can this command show them?

references/cli/status.go
```diff
@@ -182,6 +198,7 @@ func NewAppStatusCommand(c common.Args, order string, ioStreams cmdutil.IOStream
 	cmd.Flags().BoolVarP(&detail, "detail", "d", false, "display more details in the application like input/output data in context. Note that if you want to show the realtime details of application resources, please use it with --tree")
 	cmd.Flags().StringP("detail-format", "", "inline", "the format for displaying details, must be used with --detail. Can be one of inline, wide, list, table, raw.")
 	cmd.Flags().StringVarP(&outputFormat, "output", "o", "", "raw Application output format. One of: (json, yaml, jsonpath)")
+	cmd.Flags().BoolP("metrics", "m", false, "show resource quota and consumption metrics of the application")
```
StevenLeiZhang, line 201: since App may expose customized metrics except CPU/MEM/storage, can this command show them?

Somefive:
> The warning log is annoying, let's try to remove it. Does the message important? @wangyikewxgm
> 
> image

This is not a single issue only related to resource tree. It is a general issue for all CLI commands to print out controller typed logs. I would suggest dealing it in a separate work, not only fix it here.

Hanmengnan:
> 

I saw your PR about shielding logs in vela top. Is the logic about shielding logs in other CLI commands similar to it? Perhaps we should deal with this problem uniformly in the upper code of the CLI?

Somefive:
Some issues from my perspective. (Not blocking this PR)
1. The `metrics` name is a bit ambiguous. I would prefer not using the name of `metrics`.
2. Is there any differences between `vela status --metrics` and `vela metrics`? If not, I would prefer not inventing new commands but instead reuse the `vela status`.
3. Will it be better to use `vela status --resource-usage(-u)`?

references/common/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+	Storage uint64
+}
+
+// ApplicationResourceStatus is the status of application resource
+type ApplicationResourceStatus struct {
+	NodeNum        int
+	ClusterNum     int
+	SubresourceNum int
+	PodNum         int
+	ContainerNum   int
```
StevenLeiZhang, line 54: The cluster may be controlled by ServiceMesh(Istio), and ServiceMesh may inject some sidecar containers automatically. Does `ContainerNum` count them?
If yes, may user feel confused? because it is not specified by user in Application.

Hanmengnan:
> Some issues from my perspective. (Not blocking this PR)
> 
> 1. The `metrics` name is a bit ambiguous. I would prefer not using the name of `metrics`.
> 2. Is there any differences between `vela status --metrics` and `vela metrics`? If not, I would prefer not inventing new commands but instead reuse the `vela status`.
> 3. Will it be better to use `vela status --resource-usage(-u)`?

Yeah, I know. I am just interest how to fix it.

Somefive:
> > 
> 
> I saw your PR about shielding logs in vela top. Is the logic about shielding logs in other CLI commands similar to it? Perhaps we should deal with this problem uniformly in the upper code of the CLI?

Absolutely correct. But we need to notice that some CLI commands relies on klog to print out necessary information so directly hiding all logs for all CLI commands might break some CLI commands. So this rework is better to be done in another PR and needs to check if other CLI commands works after the refactoring.

Somefive:
> > Some issues from my perspective. (Not blocking this PR)
> > 
> > 1. The `metrics` name is a bit ambiguous. I would prefer not using the name of `metrics`.
> > 2. Is there any differences between `vela status --metrics` and `vela metrics`? If not, I would prefer not inventing new commands but instead reuse the `vela status`.
> > 3. Will it be better to use `vela status --resource-usage(-u)`?
> 
> Yeah, I know. I am just interest how to fix it.

Great. I think reusing `vela status` is more clear but I think the flags for `vela status` needs some better reorganization. But "reorganization" is not required to be done in this PR. We can add new flags to `vela status` for this capability for now and refactor it in the future.

image

Somefive:
> Some issues from my perspective. (Not blocking this PR)
> 
> 1. The `metrics` name is a bit ambiguous. I would prefer not using the name of `metrics`.
> 2. Is there any differences between `vela status --metrics` and `vela metrics`? If not, I would prefer not inventing new commands but instead reuse the `vela status`.
> 3. Will it be better to use `vela status --resource-usage(-u)`?

/cc @wonderflow What do you think?

Hanmengnan:
> 

Yes, actually I have removed `metrics` command and only reserved the option for `vela status`. About the name of option, how do you think about it? @wonderflow

references/common/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// LoadApplicationMetrics load application resource metrics
+func LoadApplicationMetrics(c client.Client, conf *rest.Config, app *appv1beta1.Application) (*ApplicationMetrics, error) {
+	appResList, err := ListApplicationResource(c, app.Name, app.Namespace)
+	if err != nil {
+		return nil, err
+	}
+	components := make([]string, 0)
+	for _, resource := range appResList {
```
StevenLeiZhang, line 163: Does `appResList` only have Workload? Do it has Service, ConfigMap or other resources?
Non-workload resource does not have pod.

references/common/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+	pods := ListApplicationPods(c, app, components)
+
+	appMetrics := &ApplicationMetricsStatus{}
+	appMetrics.LoadApplicationMetrics(c, conf, pods)
+
+	clusters := make(map[string]struct{})
+	nodes := make(map[string]struct{})
+	containerNum := 0
+	for _, r := range appResList {
+		clusters[r.Cluster] = struct{}{}
```
StevenLeiZhang, line 175: For multi-clusters case, target clusters are specified in Policy

references/common/metrics.go
Hanmengnan: I think some questions can be answer at same time. Now we use `*v1beta1.ResourceTracker` object to track application's resource. So, if resource belong to an application, it can be tracked out.

Hanmengnan: When we statistic the number of Managed Resource, I think Service, ConfigMap or other resources should be counted. Because we just statistic the son resource of an application, although it doesn't have pod

references/common/metrics.go
Hanmengnan: I think some questions can be answer at same time. Now we use `*v1beta1.ResourceTracker` object to track application's resource. So, if resource belong to an application, it can be tracked out. 

I am not very clear about the logic of ResourceTracker, but from my perspective I think these sidecar containers you mentioned won't be trcaked.

references/common/metrics.go
Hanmengnan: ResourceTracker can track resource whose attribute will be Policy affected. So, I think we can get the cluster information.

references/cli/status.go
Hanmengnan: Now I think the answer is not.

chivalryq: Like some prometheus metrics exported in `http:<endpoint>:8084/metrics`? How do you think @StevenLeiZhang

wonderflow:
I think `vela status --metrics` is good. We can expose other metrics related in this option in the future. 

 --resource-usage(-u) is a bit long and not intuitive, we must avoid names to be hard to remember.

Somefive:
> I think `vela status --metrics` is good. We can expose other metrics related in this option in the future.
> 
> --resource-usage(-u) is a bit long and not intuitive, we must avoid names to be hard to remember.

It is true `resource-usage` is a bit long. `metrics` could lead to misunderstanding as above, like @StevenLeiZhang mentioned. Will user think that `vela status --metrics` will show the exposed Prometheus metrics inside the application?

FYI, `kubectl` uses `kubectl top nodes` or `kubectl top pods` to show the resources.

Anyway, it is just a bit of my concerns. Great Job, @Hanmengnan !

Somefive (approved):

wonderflow:
>Will user think that `vela status --metrics` will show the exposed Prometheus metrics inside the application

Actually this is a good direction that `vela metrics` can provide, I don't think it will cause misleading, it's just because this feature is not fully supported.

We can expose app metrics in the command and refer to a grafana URL if o11y addon enabled, that could be really good experience.

PR #5498, opened 2023-02-14
Fix: read-only definition in cue spec

chivalryq:

read-only parameter defined in CUE have some issue, related field in go:
```go
// ReadOnlyPolicyRule defines the rule for read-only resources
type ReadOnlyPolicyRule struct {
	Selector ResourcePolicyRuleSelector `json:"selector"`
}
```

Fixes #

I have:


Somefive (changes requested):
Using apiVersion & kind in ref-objects is a legacy behaviour. Use `resource` instead.

Somefive (approved):

wonderflow (commented):
we need some test case to cover

PR #5505, opened 2023-02-14
Feat: support resume a specific suspend step in workflow

FogDong:

Support resume a specific suspend step in workflow like:

```
vela workflow resume --step=step1
```

I have:

### DIFF
--- a/pkg/apiserver/domain/service/workflow.go
+++ b/pkg/apiserver/domain/service/workflow.go
@@ -25,7 +25,6 @@ import (
 	"strconv"
 	"strings"
 
-	workflowv1alpha1 "github.com/kubevela/workflow/api/v1alpha1"
 	"helm.sh/helm/v3/pkg/time"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -35,6 +34,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
+	workflowv1alpha1 "github.com/kubevela/workflow/api/v1alpha1"
 	wfContext "github.com/kubevela/workflow/pkg/context"
 	"github.com/kubevela/workflow/pkg/cue/model/value"
 	wfTypes "github.com/kubevela/workflow/pkg/types"
@@ -52,6 +52,7 @@ import (
 	"github.com/oam-dev/kubevela/pkg/oam"
 	pkgUtils "github.com/oam-dev/kubevela/pkg/utils"
 	"github.com/oam-dev/kubevela/pkg/utils/apply"
+	"github.com/oam-dev/kubevela/pkg/workflow/operation"
 )
 
 // LogSourceResource Read the step logs from the pod stdout.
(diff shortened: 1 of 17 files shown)


pkg/apiserver/domain/service/workflow.go
```diff
@@ -789,19 +789,36 @@ func (w *workflowServiceImpl) TerminateRecord(ctx context.Context, appModel *mod
...
 
 // ResumeWorkflow resume workflow
-func ResumeWorkflow(ctx context.Context, kubecli client.Client, app *v1beta1.Application) error {
+func ResumeWorkflow(ctx context.Context, kubecli client.Client, app *v1beta1.Application, stepName string) error {
 	app.Status.Workflow.Suspend = false
 	steps := app.Status.Workflow.Steps
+	found := stepName == ""
+
 	for i, step := range steps {
 		if step.Type == wfTypes.WorkflowStepTypeSuspend && step.Phase == workflowv1alpha1.WorkflowStepPhaseRunning {
```
barnettZQG, line 798: If we add a provider function to suspend the current step, the step type that needs to resume is not only `suspend`. Maybe we should remove the condition `step.Type == wfTypes.WorkflowStepTypeSuspend` when the step name is not empty. What do you think?

FogDong: Currently, we don't have this `pause/suspend` provider and only `suspend` step can work. I think we can change the condition when we have such capabilities. Besides, we need to change a lot if we add such providers, eg, maybe add some phase like `suspending`.

pkg/apiserver/domain/service/workflow.go
```diff
@@ -789,19 +789,36 @@ func (w *workflowServiceImpl) TerminateRecord(ctx context.Context, appModel *mod
...
+				} else if stepName == sub.Name {
+					steps[i].SubStepsStatus[j].Phase = workflowv1alpha1.WorkflowStepPhaseSucceeded
+					found = true
+					break
+				}
 			}
 		}
 	}
+	if !found {
+		return fmt.Errorf("can not find step %s", stepName)
```
barnettZQG, line 820: Here should return a `bcode` error, the HTTP status code should be 400.

pkg/apiserver/interfaces/api/application.go
```diff
@@ -530,6 +530,7 @@ func (c *application) GetWebServiceRoute() *restful.WebService {
 		Param(ws.PathParameter("appName", "identifier of the application.").DataType("string").Required(true)).
 		Param(ws.PathParameter("workflowName", "identifier of the workflow").DataType("string")).
 		Param(ws.PathParameter("record", "identifier of the  workflow record").DataType("string")).
+		Param(ws.QueryParameter("step", "resume the workflow with specific step").DataType("string")).
```
barnettZQG, line 533: We should also change the pipeline API.

pkg/apiserver/domain/service/workflow.go
FogDong: This function is used by api and cli, we should not wrap bcode here, maybe in the upper layer.

Somefive (approved):

barnettZQG (approved):
lgtm

chivalryq (approved):
LGTM

StevenLeiZhang (approved):
LGTM

Issue #5512, opened 2023-02-15
Support customized .terraformrc , credentials.tfrc.json and credentials-helper for private terraform registry


ramekris3163:
Sometimes users want to use customized .terraformrc , credentials.tfrc.json and credentials helper files. Here's one use case:

Use self-hosted JFrog Artifactory to proxy and cache AWS provider resource. These files are used to point to the self-hosted services and provide credentials. This is related to https://github.com/kubevela/terraform-controller/issues/350 

The schematic terraform cue must support passing the required configuration for mounting the above files by passing it on to the terraform controller if present.

chivalryq:
Fixed in #5550

PR #5513, opened 2023-02-15
Feat: add support for private terraform registry.

ramekris3163:

Fixes # https://github.com/kubevela/kubevela/issues/5512

I have:

- [] Run `make reviewable` to ensure this PR is ready for review.
- [] Added `backport release-x.y` labels to auto-backport this PR if necessary.

Have tested in a local cluster with terraform-configuration which connects to a private terraform registry. Passing the terraformRC and terraformcredentials to the configuration from the component.


chivalryq:
Before changing the CRD, please consider this. Will terraform credential info will be stored in ComponentDefinition? If so, it can be a fallback if  the component properties has no related field. But there are several reasons I don't recommend to do this:

1. If we bump up the version of imported terraform-controller, user can fill the field like `terraformCredentialsSecretReference` in component instance properties. 
2. Once the the credential info is stored in ComponentDefinition, it can only refer to one secret. This is unsafe in multi-tenancy scenario. On the other hand if you want to point to different secret, platform team have to create many ComponentDefinitions according to Secrets.
3. The `schematic.Terraform` is now blocking some unified process in Controller. It can be fully refactored using `schematic.CUE`. The same to `schematic.Helm` and `schematic.Kube`. This is another topic and I'm not gonna expand it here too much.

ramekris3163:
> Before changing the CRD, please consider this. Will terraform credential info will be stored in ComponentDefinition? If so, it can be a fallback if the component properties has no related field. But there are several reasons I don't recommend to do this:
> 
> 1. If we bump up the version of imported terraform-controller, user can fill the field like `terraformCredentialsSecretReference` in component instance properties.
> 2. Once the the credential info is stored in ComponentDefinition, it can only refer to one secret. This is unsafe in multi-tenancy scenario. On the other hand if you want to point to different secret, platform team have to create many ComponentDefinitions according to Secrets.
> 3. The `schematic.Terraform` is now blocking some unified process in Controller. It can be fully refactored using `schematic.CUE`. The same to `schematic.Helm` and `schematic.Kube`. This is another topic and I'm not gonna expand it here too much.

@chivalryq do you have a sample of how to set the component instance properties. Is everything set in component instance properties passed onto the terraform-controller configuration?

ramekris3163:
> Before changing the CRD, please consider this. Will terraform credential info will be stored in ComponentDefinition? If so, it can be a fallback if the component properties has no related field. But there are several reasons I don't recommend to do this:
> 
> 1. If we bump up the version of imported terraform-controller, user can fill the field like `terraformCredentialsSecretReference` in component instance properties.
> 2. Once the the credential info is stored in ComponentDefinition, it can only refer to one secret. This is unsafe in multi-tenancy scenario. On the other hand if you want to point to different secret, platform team have to create many ComponentDefinitions according to Secrets.
> 3. The `schematic.Terraform` is now blocking some unified process in Controller. It can be fully refactored using `schematic.CUE`. The same to `schematic.Helm` and `schematic.Kube`. This is another topic and I'm not gonna expand it here too much.

@chivalryq do you have a sample of how to set the component instance properties. Is everything set in component instance properties passed onto the terraform-controller configuration?

chivalryq:
Please check `generateTerraformConfigurationWorkload` in `pkg/appfile/appfile.go`. It will unmarshal all field to Configuration Spec. 
Example:
```yaml
kind: Application
metadata:
  name: my-vswitch
spec:
  components:
    - name: my-vswitch
      type: alibaba-vswitch
      properties:
        create_vpc: false
        vpc_id: "vpc-bp1mp89jxm3wjsggw9ybi"
        vswitch_cidr: "172.16.8.0/24"
        vswitch_name: "my-tf-vswitch"
        zone_id: "cn-hangzhou-b"
        writeConnectionSecretToRef:
          name: "my-vswitch-info"
          namespace: prod
        terraformCredentialsSecretReference:
          name: my-creds-sec
          namespace: prod
```
In the above application, I fill out the `terraformCredentialsSecretReference`. And it'll be unmarshalled to `Configuration.Spec`. So, if you don't have requirements to put creds secret in ComponentDefinition. It's OK to not to change the CRD.

PR #5530, opened 2023-02-17
Feat: default enable multi-stage apply

Somefive:

1. default enable multi-stage apply
2. expose trait deploy service after deployment

I have:


charts/vela-core/templates/defwithtemplate/expose.yaml
```diff
@@ -44,6 +44,7 @@ spec:
         	// +usage=Specify what kind of Service you want. options: "ClusterIP","NodePort","LoadBalancer","ExternalName"
         	type: *"ClusterIP" | "NodePort" | "LoadBalancer" | "ExternalName"
         }
+  stage: PostDispatch
```
StevenLeiZhang, line 47: Do we need add this field for more Traits?such as Gateway, HPA?

Somefive: We can add that for other traits if needed. The reason here to set expose trait to use the stage of PostDispatch is that on some platform we found that the service creation will fail when the deployment is not created. And the parallelized dryrun requests will be rejected so the deployment delivery will be blocked as well. To handle that deadlock I add PostDispatch to expose trait. But depending on the use case, parallelized dispatch would be more effective.

wonderflow (approved):

PR #5568, opened 2023-02-24
Fix: The resource topology fails to display the pods under the job #5566

### LINKED ISSUE #5566: The resource topology fails to display the pods under the job
**Describe the bug**
When the component is a task type, the resource topology map in the running state cannot display the pod associated with the job。

**To Reproduce**

**Expected behavior**
Under the resource topology, the job can display its associated pod

**Screenshots**

**KubeVela Version**
1.6.4

**Cluster information**
v1.19.9

**Additional context**

msun1996:

Fixes #5566

I have:

I have carried out local debugging to verify this function, and it can normally solve the display problem of the job-associated pod under the resource topology

### DIFF
--- a/pkg/velaql/providers/query/tree.go
+++ b/pkg/velaql/providers/query/tree.go
@@ -109,6 +109,15 @@ func init() {
 			GroupResourceType: GroupResourceType{Group: "apps", Kind: "DaemonSet"},
 		},
 		ChildrenResourcesRule{
+			GroupResourceType: GroupResourceType{Group: "batch", Kind: "Job"},
+			SubResources: buildSubResources([]*SubResourceSelector{
+				{
+					ResourceType: ResourceType{APIVersion: "v1", Kind: "Pod"},
+					listOptions:  defaultWorkloadLabelListOption,
+				},
+			}),
+		},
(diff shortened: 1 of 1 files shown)


pkg/velaql/providers/query/tree.go
```diff
@@ -215,6 +215,15 @@ func init() {
 			}),
 			GroupResourceType: GroupResourceType{Group: "batch", Kind: "CronJob"},
 		},
+		ChildrenResourcesRule{
+			GroupResourceType: GroupResourceType{Group: "batch", Kind: "Job"},
+			SubResources: buildSubResources([]*SubResourceSelector{
+				{
+					ResourceType: ResourceType{APIVersion: "v1", Kind: "Pod"},
+					listOptions:  jobLabelListOption,
```
barnettZQG, line 223: I think the Job resource could use the `defaultWorkloadLabelListOption`.

msun1996: Yes, my reference is wrong. Check the k8s code and find that getPodsForJob uses "spec", "selector"

barnettZQG (approved):
LGTM

Issue #5364, opened 2023-01-22
[Feature] [LFX] Support auto generation of CUE schema and docs from Go struct


FogDong:
## Project description

In our provider system, we can use our defined go functions in cue schema.

The go provider usually has a `parameter` and `return`, for example, in a `http` provider, we have struct like:

```
// input: parameters

type RequestVars struct {
	Method  string `json:"method"`
	URL     string `json:"url"`
	Request struct {
		Body string `json:"body"`
	} `json:"request"`
	Header  http.Header `json:"header"`
	Trailer http.Header `json:"trailer"`
}

// output: returns
// ResponseVars is the vars for http response
type ResponseVars struct {
	Body       string      `json:"body"`
	Header     http.Header `json:"header"`
	Trailer    http.Header `json:"trailer"`
	StatusCode int         `json:"statusCode"`
}

```

And the CUE schema is like:

```
package http

#Do: {
	#do:       "do"
	#provider: "http"

	// +usage=The params of this action
	$params: {
		// +usage=The method of HTTP request
		method: *"GET" | "POST" | "PUT" | "DELETE" | "PATCH" | "CONNECT" | "OPTIONS" | "TRACE"
		// +usage=The url to request
		url: string
		// +usage=The request config
		request?: {
			// +usage=The request body
			body?: string
			// +usage=The header of the request
			header?: [string]: string
			// +usage=The trailer of the request
			trailer?: [string]: string
			...
		}
	}
	// +usage=The response of the request will be filled in this field after the action is executed
	$returns: {
		// +usage=The body of the response
		body?: string
		// +usage=The header of the response
		header?: [string]: [...string]
		// +usage=The trailer of the response
		trailer?: [string]: [...string]
		// +usage=The status code of the response
		statusCode: int
		...
	}
	...
}
```

The field is the same as these in the struct, so it's possible to support an auto generator of CUE schema from Go struct.
After is schema is generated, we can also support generating the docs from CUE schema with its comments automatically.

## Expected outcomes

1. Support the auto generator of CUE schema from Go struct.
2. Support the auto generator of docs from CUE schema.
3. The generators will be the command of vela cli.

## Mentors

@FogDong 

## Expected size of the project

200 hours

## Difficult level

Medium.

## Reference

Go struct example: https://github.com/kubevela/pkg/blob/main/cue/cuex/providers/http/http.go
CUE schema example: https://github.com/kubevela/pkg/blob/main/cue/cuex/providers/http/http.cue
[...]

PR #5585, opened 2023-02-28
Feat: CUE generator with type conversion

iyear:

Part of #5364 

This PR completes type conversion, file generation, comment copying, tag parsing, and AST conversion. But it's not yet production ready.

I have:

The PR contains many new files and code. So append tests to this PR or file a new PR?

- The conversion rules and description are in the README, which helps understand the code.
- Should I write TODOs in the original issue for tracking tasks?


iyear:
Please start the review after I fix the CI failing checks. Thanks!

FogDong (commented):
Please add some tests.

references/cuegen/comment.go
```diff
@@ -0,0 +1,136 @@
...
+func makeComment(cg *goast.CommentGroup) *cueast.CommentGroup {
+	if cg == nil {
+		return nil
+	}
+
+	var comments []*cueast.Comment
+
+	for _, comment := range cg.List {
+		c := comment.Text
+
```
FogDong, line 84: Add some length check like:

```
if len(c) < 1 {
  continue
}
```

references/cuegen/types.go
```diff
@@ -0,0 +1,92 @@
...
+type typeInfo map[gotypes.Type]*goast.StructType
+
+func getTypeInfo(p *packages.Package) typeInfo {
+	m := make(typeInfo)
+
+	for _, f := range p.Syntax {
+		goast.Inspect(f, func(n goast.Node) bool {
+			// record all struct types
+			if t, ok := n.(*goast.StructType); !ok {
+				m[p.TypesInfo.TypeOf(t)] = t
```
FogDong, line 36: What will happen if `t` is not *goast.StructType, it will not be stored in `m`? Maybe some tests will make this easier to understand.

references/cuegen/generator.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return g.write(w, f)
+}
+
+func (g *Generator) write(w io.Writer, f *cueast.File) error {
+	if err := astutil.Sanitize(f); err != nil {
+		return err
+	}
+
+	b, err := cueformat.Node(f, cueformat.Simplify())
```
FogDong, line 94: Maybe we should add `cue.Docs(true)` in `format.Node` so that all the comments will be shown.

iyear:
Thanks for the review! I will append tests later.

references/cuegen/types.go
iyear: Fixed

references/cuegen/generator.go
iyear: `cue.Docs()` can't be used in `cueformat.Node()`

references/cuegen/comment.go
iyear: Only text starting with `/*` or `//` will be added to `cg.List`, so len(c) must be >=2

iyear:
Tests have been added, and test coverage has reached 90.4%. But I don't know why CI still failed.

references/cuegen/comment.go
FogDong: Is there any panic risks?

references/cuegen/testdata/valid/enum.cue
```diff
@@ -0,0 +1,14 @@
...
+Enum: {
+	a: "abc" | "def" | "ghi"
+	b: 1 | 2 | 3
+	c: true | false
+	d: 1.1 | 2.2 | 3.3
+	e: "abc" | "def" | *"ghi"
+	f: 1 | *2 | 3
+	g: true | *false
+	// if default value is first enum, '*' will not be added
+	h: 1.1 | 2.2 | 3.3
```
FogDong, line 12: If we have multiple enums(>=3) in one field, actually we need a default value, otherwise the cue will failed to generate openapi. If the users do not specify the default value, maybe we need to set the first enum as the default value.

references/cuegen/testdata/valid/exported.go
```diff
@@ -0,0 +1,38 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+type A struct {
+	Field1 string `json:"field1"`
+	field2 string
```
FogDong, line 21: The field without json tag won't be exported to cue makes sense to me, so for some specific type like `unstructered`, we can register the type as `any`, great job!

references/cuegen/testdata/valid/comment.cue
```diff
@@ -0,0 +1,68 @@
+package test
```
FogDong, line 1: Can we merge all the test files?

references/cuegen/testdata/invalid/default.go
```diff
@@ -0,0 +1,22 @@
+/*
```
FogDong, line 1: Perhaps we can merge the test cases into one file and use the struct name to distinguish.

references/cuegen/util_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIdent(t *testing.T) {
+	tests := []struct {
+		name  string
+		isDef bool
+		want  string
+	}{
+		{name: "test", isDef: true, want: "_#test"}, // invisible (first letter is lowercase)
```
FogDong, line 36: Why convert the field with lowercase in the first letter as an invisible def?

references/cuegen/util.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func basicType(x *gotypes.Basic) cueast.Expr {
+	// byte is an alias for uint8 in go/types
+
+	switch t := x.String(); t {
+	case "uintptr":
+		return ident("uint64", false)
+	case "byte":
+		return ident("uint8", false)
```
FogDong, line 51: Why not `bytes`?

references/cuegen/util_test.go
```diff
@@ -0,0 +1,128 @@
...
+		wantErr bool
+		want    *cueast.BasicLit
+	}{
+		{name: "int", typ: gotypes.Typ[gotypes.Int], v: "123", wantErr: false, want: &cueast.BasicLit{Kind: cuetoken.INT, Value: "123"}},
+		{name: "int8", typ: gotypes.Typ[gotypes.Int8], v: "123", wantErr: false, want: &cueast.BasicLit{Kind: cuetoken.INT, Value: "123"}},
+		{name: "int16", typ: gotypes.Typ[gotypes.Int16], v: "123", wantErr: false, want: &cueast.BasicLit{Kind: cuetoken.INT, Value: "123"}},
+		{name: "int32", typ: gotypes.Typ[gotypes.Int32], v: "123", wantErr: false, want: &cueast.BasicLit{Kind: cuetoken.INT, Value: "123"}},
+		{name: "int64", typ: gotypes.Typ[gotypes.Int64], v: "123", wantErr: false, want: &cueast.BasicLit{Kind: cuetoken.INT, Value: "123"}},
+		{name: "uint", typ: gotypes.Typ[gotypes.Uint], v: "123", wantErr: false, want: &cueast.BasicLit{Kind: cuetoken.INT, Value: "123"}},
+		{name: "uint8", typ: gotypes.Typ[gotypes.Uint8], v: "123", wantErr: false, want: &cueast.BasicLit{Kind: cuetoken.INT, Value: "123"}},
```
FogDong, line 98: Is this the byte case?

references/cuegen/testdata/valid/exported.go
iyear: The field here is ignored because it is not exported. If an exported `FieldName` does not have a json tag, it will still be written to the cue with `FieldName`

references/cuegen/comment.go
iyear: Although I think go ast will guarantee that length>=2, I'll add the judgment later..

references/cuegen/testdata/valid/enum.cue
iyear: - Even if Generator marks the first value as the default, `cueformat.Simplify()` will remove `*`.

- It seems that enum without a default value is legal for openapi generation.

Test code:
```go
s := `
	#Def: {
		h: 1.1 | 2.2 | 3.3
		i: 1 | *2 | 3
	}
`
r := cuecontext.New()
bytes, err := openapi.Gen(r.CompileString(s), nil)
if err != nil {
	panic(err)
}
fmt.Println(string(bytes))
```

Result:
```json
{
  "openapi": "3.0.0",
  "info": {
    "title": "Generated by cue.",
    "version": "no version"
  },
  "paths": {},
  "components": {
    "schemas": {
      "Def": {
        "type": "object",
        "required": [
          "h",
          "i"
        ],
        "properties": {
          "h": {
            "type": "number",
            "enum": [
              1.1,
              2.2,
              3.3
            ]
          },
          "i": {
            "type": "integer",
            "enum": [
              2,
              1,
              3
            ],
            "default": 2
          }
        }
      }
    }
  }
}

```
image

references/cuegen/testdata/valid/comment.cue
iyear: Let me update it later.

references/cuegen/util_test.go
iyear: `ident` comes from the official cue implementation, which believes that def starting with lowercase should only be visible within the CUE package.

references/cuegen/util.go
iyear: This is the base type conversion, so it just convert byte in Golang to uint8 in CUE(CUE don't have `byte` type, only `bytes`). 

And `[]byte` and `[...]byte` are processed at https://github.com/kubevela/kubevela/pull/5585/files#diff-f3ec0bb030f3add13296ed258157b227f061b722669287e2ad2cc069d76e2723R101 and https://github.com/kubevela/kubevela/pull/5585/files#diff-f3ec0bb030f3add13296ed258157b227f061b722669287e2ad2cc069d76e2723R110

references/cuegen/util_test.go
iyear: In the unit test I used `gotypes.Typ` to quickly generate `*gotypes.Basic`. However, in gotypes, `Byte = Uint8` https://github.com/golang/go/blob/master/src/go/types/basic.go#L45. 

So `gotypes.Typ[gotypes.Byte].String() == "uint8"`, it doesn't go through `case "byte` at all. Only when goast parses the file, it appears `Type.String() == "byte"` .

iyear: So in this unit test, I did not test `byte` case.

references/cuegen/testdata/invalid/default.go
iyear: This may can't be merged, because Generator will return directly when it encounters an error. If merged, Generator ends after encountering the first struct.

references/cuegen/util_test.go
iyear: After some thought, I decided to remove this code. It's useless for us.

iyear: Removed

references/cuegen/testdata/valid/enum.cue
iyear: Hi, any suggestion or review?

FogDong (approved):
LGTM! A great start to the generator project, well done!

references/cuegen/convert.go
```diff
@@ -0,0 +1,480 @@
...
+					comments = append(comments, &cueast.Comment{Text: "//"})
+					continue
+				}
+				comments = append(comments, &cueast.Comment{Text: "// " + l[i:]})
+			}
+		}
+	}
+
+	return &cueast.CommentGroup{List: comments}
+}
```
wonderflow, line 480: I think this is a basic but important package, the scenarios are very accurate, what's the test coverage?

iyear: It's 89.1%. Some duplicate checks resulted in some branches not being covered.

references/cuegen/tag_test.go
```diff
@@ -0,0 +1,193 @@
...
+	}{
+		{"empty", "", map[string]string{}},
+		{"only_key", "key", map[string]string{"key": ""}},
+		{"one_kv", "key:value", map[string]string{"key": "value"}},
+		{"multi_kv", "key1:value1;key2:value2", map[string]string{"key1": "value1", "key2": "value2"}},
+		{"bool", "key1;key2", map[string]string{"key1": "", "key2": ""}},
+		{"bool_and_kv", "key1;key2:value2", map[string]string{"key1": "", "key2": "value2"}},
+		{"kv_and_bool", "key1:value1;key2", map[string]string{"key1": "value1", "key2": ""}},
+		{"multi_kv_and_bool", "key1:value1;key2:value2;key3", map[string]string{"key1": "value1", "key2": "value2", "key3": ""}},
+		{"escape", "key1:\"value\"", map[string]string{"key1": `"value"`}},
```
chivalryq, line 135: Sorry but what's meaning here by "supporting escape" because I think 
```go
"key1:\"value\""
```
means `key1:"value"` literally. It's transformed by Golang and it's not your parsing logic to support the escaping. If there's any end to end test case about this feature can you point it?

Somefive (approved):

references/cuegen/util.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"fmt"
+	gotypes "go/types"
+	"strconv"
+
+	cueast "cuelang.org/go/cue/ast"
+	cuetoken "cuelang.org/go/cue/token"
+)
+
+var _ = ident("", true) // nolint. will be used in the future. TODO(iyear): remove
```
chivalryq, line 28: Can be removed now?

references/cuegen/tag_test.go
iyear: Sorry, it's mistake, I will remove this case. The `escape` in tag is used to escape the separator(`;`).

And does this feature require e2e testing? I'm not familiar with this, can you help me?

references/cuegen/util.go
iyear: Related code will be committed in another PR.

iyear: Updated

references/cuegen/tag_test.go
iyear: Updated

chivalryq: There's some similar test case marked relative to escaping. What do you mean by escape the separator? Can you add some suitable test case?

iyear: If you want to use a separator(`;`) in the key or value, you need to use the escape, otherwise it will be treated as a new field.

iyear: More tests added

chivalryq: Got that.

references/cuegen/tag_test.go
```diff
@@ -132,8 +132,8 @@ func TestParseExtTag(t *testing.T) {
 		{"bool_and_kv", "key1;key2:value2", map[string]string{"key1": "", "key2": "value2"}},
 		{"kv_and_bool", "key1:value1;key2", map[string]string{"key1": "value1", "key2": ""}},
 		{"multi_kv_and_bool", "key1:value1;key2:value2;key3", map[string]string{"key1": "value1", "key2": "value2", "key3": ""}},
-		{"escape", "key1:\"value\"", map[string]string{"key1": `"value"`}},
-		{"escape_2", "key1:value1\\;vv", map[string]string{"key1": "value1;vv"}},
+		{"escape_value", "key1:value\\;1", map[string]string{"key1": "value;1"}},
```
chivalryq, line 135: 1. Suggestion here, and this is not the most required but made it easier to understand.
```suggestion
		{"escape_value", `key1:value\;1`, map[string]string{"key1": "value;1"}},
```
2. Does the escape only support the separator `;`? How about `,`. Will it detect `\\` means `\`? 
3. Let's add more test about edge case. For example, 
	1. `\` is the last character in string.
	2. `\` followed by a character which can't be escaped.

iyear: Thanks for review, bugs were found in edge cases and I have fixed it.

Currently only the escape for `;` is supported, other escapes may not be necessary in real world scenarios, so we can add it if someone really needs it...🤔

chivalryq: I see. Now if `\` is not followed by a `;` then it means `\` literally. I think other special characters like `,` in this case will appears in the future sooner or later. Why don't we support it instead of leaving it to be fixed.

chivalryq (approved):
Good job! Generally LGTM.

references/cuegen/tag_test.go
iyear: It's supported now: 1d43b31bb19a3c8d5f7344ae30b194bec6aba306

There are many new tests, which need to be review.

PR #5598, opened 2023-03-03
Feat: remove loop reduction & filter unnecessary apprev update

Somefive:

Add two optimization:
1. Remove `loop-reduction` optimization and let it be enabled by default. This is because the workflow state transition has changed its mechanism and if we do not enable loop reduction by default, the application will run twice gc every reconcile.
2. The application revision updates always runs no matter whether it is needed. However, usually, the application revision do not change when state-keep-oriented reconcile happens. This PR reduct it to optimize performance.

Result:
image
Optimization 1 can reduce the reconcile time from 125ms to 96ms, increase 23% performance for state-keep-oriented reconcile under simple application delivery scenario.
image
Optimization 2 can reduce the reconcile time from 96ms to 70ms, increase 27% performance for state-keep oriented reconcile under simple application delivery scenario.

I have:


pkg/controller/core.oam.dev/v1alpha2/application/revision.go
```diff
@@ -856,6 +856,9 @@ func (h *AppHandler) FinalizeAndApplyAppRevision(ctx context.Context) error {
 		}
 		return err
 	}
+	if apiequality.Semantic.DeepEqual(gotAppRev.Spec, appRev.Spec) {
```
wonderflow, line 859: please also compare the annotations and labels from metadata

Somefive: ok

wonderflow (approved):

FogDong (approved):

PR #5611, opened 2023-03-05
Fix: return the error when get the resource

Feizekai:

Fixes #

I have:

### DIFF
--- a/pkg/config/factory.go
+++ b/pkg/config/factory.go
@@ -624,7 +624,7 @@ func (k *kubeConfigFactory) IsExist(ctx context.Context, namespace, name string)
 	var secret v1.Secret
 	if err := k.cli.Get(ctx, pkgtypes.NamespacedName{Namespace: namespace, Name: name}, &secret); err != nil {
 		if apierrors.IsNotFound(err) {
-			return false, nil
+			return false, fmt.Errorf("fail to find the resource %s", name)
 		}
 		return false, err
 	}


Somefive:
Could you give more details for why we need this PR? Like some usage example or scenario?

Somefive:
BTW, please fixes the DCO.

wonderflow (changes requested):
we don't need to return the error as the function is checking if exists

Reetika2509:
Hi @Feizekai — thanks for the contribution.

Taking a closer look: IsExist intentionally returns (false, nil) on IsNotFound — that's the expected "doesn't exist, no error" path, whilegenuine failures already return via return false, err. Returning an error on NotFound would change the contract and make callers treat alegitimate "not found" as a failure, so this would actually introduce a regression rather than fix one. That's along the lines of what @Somefive was asking about in the earlier comments.

Since the change isn't needed, I'm going to close it. If you hit a concrete scenario where IsExist swallows an error it shouldn't, please open an issue with the details and we'll be happy to look. Thanks again!

Feizekai:
> Hi @Feizekai — thanks for the contribution.
> 
> Taking a closer look: IsExist intentionally returns (false, nil) on IsNotFound — that's the expected "doesn't exist, no error" path, whilegenuine failures already return via return false, err. Returning an error on NotFound would change the contract and make callers treat alegitimate "not found" as a failure, so this would actually introduce a regression rather than fix one. That's along the lines of what @Somefive was asking about in the earlier comments.
> 
> Since the change isn't needed, I'm going to close it. If you hit a concrete scenario where IsExist swallows an error it shouldn't, please open an issue with the details and we'll be happy to look. Thanks again!

I got u.Its magically that u review the pr, even it has been 3 yrs😂

PR #5644, opened 2023-03-09
Fix:  loading definitions if lack in the revision when calculate dispatch stage of trait.

wangyikewxgm:

When we add a trait to a component with an inline policy, the trait's definition isn't stored in the application revision. As a result, to determine if the traitDefinition's stage is predispatch or postdispatch, we must retrieve it from the cluster.

This pr is a fix for https://github.com/kubevela/kubevela/pull/5416

Fixes #

I have:

### DIFF
--- a/pkg/controller/core.oam.dev/v1alpha2/application/dispatch_test.go
+++ b/pkg/controller/core.oam.dev/v1alpha2/application/dispatch_test.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2022 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


wonderflow (commented):
can you describe why do we need this PR?

wangyikewxgm:
> can you describe why do we need this PR?

When we add a trait to a component with an inline policy, the trait's definition isn't stored in the application revision. As a result, to determine if the traitDefinition's stage is predispatch or postdispatch, we must retrieve it from the cluster

Somefive (approved):
LGTM

FogDong:
> /cc @FogDong I think this PR is a temporary fix as it possibly means when introducing new trait via inlinePolicy we would get the definition multiple times. It would be better if the design of our new TemplateLoader could use cache to optimize this. :)

Actually I think we shouldn't get trait from the cluster for inline policy. Instead, the trait that used in the policy should also be stored in the revision.

wonderflow (changes requested):
Is there a way to achieve it in one step? Let's not use temporary solutions. @Somefive @FogDong @wangyikewxgm

Somefive:
No, it is associated the mechanism of loading definitions so it is out of the scope of workflowstep itself.

FogDong (approved):

Somefive:
> Is there a way to achieve it in one step? Let's not use temporary solutions. @Somefive @FogDong @wangyikewxgm

It is a temporary solution but by far we do not have better solutions.

wonderflow (approved):

PR #5683, opened 2023-03-15
Feat: memory optimization via enhance informer cache

Somefive:

Apply 3 methods to optimize the memory usage.
1. Drop `managedFields` and `kubectl.kubernetes.io/last-applied-configuration` when storing Application, ApplicationRevision, ResourceTracker in informer. (This only affect the cache, etcd storage not change)
2. Reduce the memory usage of ApplicationRevision by sharing the storage of ComponentDefinition, TraitDefinition, WorkflowStepDefinition. (Let them use the same pointer if the hash equals)
3. Disable the informer cache for ConfigMap caused by Workflow Context. (At the cost of slower Workflow Context updates)

Before Optimization
image

After Optimization
image

3k app, 9k apprev, RSS 1.02G(Alloc: 401M, Sys: 897M) -> 739M(Alloc: 203M, Sys: 707M), enhance 25%~30%.

Refactor:
1. Remove unused flags for optimization.

I have:


charts/vela-core/values.yaml
```diff
@@ -18,7 +18,7 @@ concurrentReconciles: 4
 
 ## @param controllerArgs.reSyncPeriod The period for resync the applications
 controllerArgs:
-  reSyncPeriod: 5m
+  reSyncPeriod: 3m
```
wonderflow, line 21: let's seprate the state keep change in a new PR along with the belowing kubeClient QPS

charts/vela-core/values.yaml
```diff
@@ -88,19 +88,23 @@ healthCheck:
 ##@param optimize.markWithProb Optimize ResourceTracker GC by only run mark with probability. Side effect: outdated ResourceTracker might not be able to be removed immediately.
 ##@param optimize.disableComponentRevision Optimize componentRevision by disabling the creation and gc
 ##@param optimize.disableApplicationRevision Optimize ApplicationRevision by disabling the creation and gc.
-##@param optimize.disableWorkflowRecorder Optimize workflow recorder by disabling the creation and gc.
 ##@param optimize.enableInMemoryWorkflowContext Optimize workflow by use in-memory context.
 ##@param optimize.disableResourceApplyDoubleCheck Optimize workflow by ignoring resource double check after apply.
 ##@param optimize.enableResourceTrackerDeleteOnlyTrigger Optimize resourcetracker by only trigger reconcile when resourcetracker is deleted.
+##@param optimize.informerCache if enabled, informer cache will be optimized by removing unnecessary fields will be filtered
```
wonderflow, line 94: can you explain what's the unnecessary fields? we need to answer if these fields are unnecessary, why don't make it default to true?

charts/vela-core/values.yaml
```diff
@@ -88,19 +88,23 @@ healthCheck:
...
 optimize:
   cachedGvks: ""
   markWithProb: 0.1
   disableComponentRevision: true
   disableApplicationRevision: false
-  disableWorkflowRecorder: false
   enableInMemoryWorkflowContext: false
   disableResourceApplyDoubleCheck: false
   enableResourceTrackerDeleteOnlyTrigger: true
+  informerCache: false
```
barnettZQG, line 105: Why default value is false?

charts/vela-core/values.yaml
Somefive: Okay

charts/vela-core/values.yaml
Somefive: We can make it default to be true. The unnecessary fields are the fields that are recorded in etcd storage but vela-core does not need to use them for logic computation, like `managedFields` or `last-applied-configuration`.

charts/vela-core/values.yaml
Somefive: I will change it to be true.

pkg/cache/informer.go
```diff
@@ -0,0 +1,265 @@
...
+				DefaultDefinitionCache.Get().RemapRevision(rev)
+			}
+		})
+		opts.TransformByObject[&v1beta1.Application{}] = wrapTransformFunc(func(app *v1beta1.Application) {})
+		opts.TransformByObject[&v1beta1.ResourceTracker{}] = wrapTransformFunc(func(rt *v1beta1.ResourceTracker) {})
+	}
+}
+
+// NewClientDisableCacheFor get ClientDisableCacheFor for building controller
+func NewClientDisableCacheFor() []client.Object {
```
wonderflow, line 259:
```suggestion
func Resources2DisableCache() []client.Object {
```

Somefive: This is aligned with the naming of controller-runtime manager. See https://github.com/kubevela/kubevela/blob/afc5c8c2db4b5a322ad785fbd10ad84bcd9a54e6/cmd/core/app/server.go#L165

charts/vela-core/values.yaml
```diff
@@ -88,19 +88,23 @@ healthCheck:
...
   markWithProb: 0.1
   disableComponentRevision: true
   disableApplicationRevision: false
-  disableWorkflowRecorder: false
   enableInMemoryWorkflowContext: false
   disableResourceApplyDoubleCheck: false
   enableResourceTrackerDeleteOnlyTrigger: true
+  informerCache: true
+  applicationRevisionDefinitionStorage: true
+  disableWorkflowContextConfigMapCache: true
```
wonderflow, line 107: Shall we provide a new optimize section for these fine-grained cache items?


```suggestion
  disableCacheItems:
    disableWorkflowContextConfigMapCache: false
```

BTW, the parameter name is confusion, it's default to disable the cache or not?

charts/vela-core/values.yaml
wonderflow: if so, please don't expose a chart parameter to add complexity

pkg/cache/informer.go
wonderflow: they're different, the naming is trying to become a sentence:

Client disable cache for "some resources", so the field should be "some resources", not copy-pase the name.

pkg/cache/optimize.go
```diff
@@ -60,6 +62,9 @@ func BuildCache(ctx context.Context, scheme *runtime.Scheme, shardingObjects ...
 				return nil, err
 			}
 		}
+		if OptimizeInformerCache && OptimizeApplicationRevisionDefinitionStorage {
+			go DefaultDefinitionCache.Get().Start(ctx, c, ApplicationRevisionDefinitionCachePruneDuration)
```
wonderflow, line 66: does all cache using a go routine? can we align with them?

pkg/controller/flags.go
```diff
@@ -28,20 +28,22 @@ import (
...
 	fs.BoolVar(&application.DisableAllComponentRevision, "optimize-disable-component-revision", false, "Optimize ComponentRevision by disabling the creation and gc. Side effect: rollout cannot be used. If you don't use rollout trait, you can switch it on to reduce the storage and improve performance.")
 	fs.BoolVar(&application.DisableAllApplicationRevision, "optimize-disable-application-revision", false, "Optimize ApplicationRevision by disabling the creation and gc. Side effect: application cannot rollback. If you don't need to rollback applications, you can switch it on to reduce the storage and improve performance.")
-	fs.BoolVar(&workflow.DisableRecorder, "optimize-disable-workflow-recorder", false, "Optimize workflow recorder by disabling the creation and gc. Side effect: workflow will not record application after finished running. If you do not use VelaUX, you can switch it on to improve performance.")
```
wonderflow, line 44: does it a break change?

Somefive: No break change. It is an empty flag now.

pkg/cache/optimize.go
Somefive: yes. all definition cache use one single goroutine.

charts/vela-core/values.yaml
chivalryq: I think it should be 
```suggestion
    disableCacheItems:
	    workflowContextConfigMapCache: false
```

Somefive: Changed to default enabled featuregate

pkg/cache/informer.go
Somefive: ok

wonderflow (approved):

FogDong (approved):

barnettZQG (approved):
LGTM

PR #5690, opened 2023-03-16
Feat: add adopt all command

FogDong:

Add `vela adopt --all` to automatically adopt resources with builtin rules(https://github.com/kubevela/kubevela/blob/4abb2c9f0813c4bbb15de8cd174f25947026e468/references/cli/resource-topology/builtin-rule.cue) to an Application.

Example:
render1678966987405

I have:


references/cli/adopt.go
```diff
@@ -404,11 +558,21 @@ var (
 		the command and make your own assemble rules for the adoption application. You can
 		refer to https://github.com/kubevela/kubevela/blob/master/references/cli/adopt-templates/default.cue
 		to see the default implementation of adoption rules.
+
+		If you want to adopt all resources with resource topology rule to Applications,
+		you can use: 'vela adopt --all'. The resource topology rule can be customized by
+		'--resource-topology-rule' flag.
```
wonderflow, line 564: does the resource topology use the same rule file? if not, maybe we can call this one to be `relation-rule`

wonderflow (approved):
great work, generally LGTM

references/cli/adopt.go
FogDong: Currently the rule is not the same, but we're gonna replace it with the same rule in the future.

PR #5839, opened 2023-04-12
Feat: initial provider generator

iyear:

Part of #5364 

This is initial impl of provider generator. The impl is a bit hacky currently and I will refactor it in the future. So it's helpful to just look into testings instead of impl code.

How it works: 
1. Provider generator extracts providers from `map[string]github.com/kubevela/pkg/cue/cuex/runtime.ProviderFn` in target go file.
2. Call cuegen to generate struct decls.
3. Modify and compose decls, and call cuegen `Format`.

I have:

### DIFF
--- a/references/cuegen/generators/provider/provider.go
+++ b/references/cuegen/generators/provider/provider.go
@@ -49,10 +49,10 @@ type provider struct {
 
 // Options is options of generation
 type Options struct {
-	File     string    // Go file path
-	Writer   io.Writer // target writer
-	AnyTypes []string  // option cuegen.WithAnyTypes
-	Nullable bool      // option cuegen.WithNullable
+	File     string                 // Go file path
+	Writer   io.Writer              // target writer
+	Types    map[string]cuegen.Type // option cuegen.WithTypes
+	Nullable bool                   // option cuegen.WithNullable
 }
 
 // Generate generates cue provider from Go struct
@@ -65,7 +65,7 @@ func Generate(opts Options) (rerr error) {
 	// make options
 	genOpts := make([]cuegen.Option, 0)
 	// any types
-	genOpts = append(genOpts, cuegen.WithAnyTypes(opts.AnyTypes...))
+	genOpts = append(genOpts, cuegen.WithTypes(opts.Types))
 	// nullable
 	if opts.Nullable {
 		genOpts = append(genOpts, cuegen.WithNullable())
--- a/references/cuegen/generators/provider/provider.go
+++ b/references/cuegen/generators/provider/provider.go
@@ -1,0 +1,198 @@
+/*
+Copyright 2023 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


references/cuegen/generators/provider/provider.go
```diff
@@ -0,0 +1,198 @@
...
+
+	cueast "cuelang.org/go/cue/ast"
+	cuetoken "cuelang.org/go/cue/token"
+	"golang.org/x/tools/go/packages"
+
+	"github.com/oam-dev/kubevela/references/cuegen"
+)
+
+const (
+	typeProviderFnMap          = "map[string]github.com/kubevela/pkg/cue/cuex/runtime.ProviderFn"
```
FogDong, line 33: We can make these const config as an options struct with some default values.

references/cuegen/generators/provider/provider.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Generate generates cue provider from Go struct
+func Generate(opts Options) (rerr error) {
+	g, err := cuegen.NewGenerator(opts.File)
+	if err != nil {
+		return err
+	}
+
+	// make options
+	genOpts := make([]cuegen.Option, 0)
```
FogDong, line 66: Maybe it's better to have a unify way to convert `Option` to `cuegen.Option`.

references/cuegen/generators/provider/testdata/valid.cue
```diff
@@ -0,0 +1,100 @@
...
+	#provider: "test"
+	$params: {
+		// +usage=The cluster to use
+		cluster: string
+		// +usage=The resource to patch
+		resource: {
+			...
+		}
+		// +usage=The patch to be applied to the resource with kubernetes patch
+		patch: {
```
FogDong, line 89: The patch data in pkg is like:

```
		// +usage=The patch to be applied to the resource with kubernetes patch
		patch: *{
			// +usage=The type of patch being provided
			type: "merge"
			data: {...}
		} | {
			// +usage=The type of patch being provided
			type: "json"
			data: [...{...}]
		} | {
			// +usage=The type of patch being provided
			type: "strategic"
			data: {...}
		}
```

This is because list and struct are different in CUE. But in GO, we can cover that with any.
Can we use tag to distinguish the differences in Go in generator?

FogDong: Or we can change the way we write in Go code.

references/cuegen/generators/provider/provider.go
iyear: I can't get the point. provider-gen only exposes some options, so it can prevent options such as type filter from being modified by the caller.

references/cuegen/generators/provider/testdata/valid.cue
iyear: Perhaps a little difficult, tag's expressiveness is poor. And for go any, what is the difference between `[]{...}`  and `{...}` ?

references/cuegen/generators/provider/provider.go
iyear: Get it.

references/cuegen/generators/provider/testdata/valid.cue
FogDong: For go, there are no differences. But in CUE, it's different. Checkout  https://cuelang.org/play/?id=PhxG9btJpjZ#cue@export@cue

iyear:
Ready for review.

/cc @FogDong @Somefive

FogDong (approved):

Somefive (approved):

PR #5865, opened 2023-04-17
Fix: install dependency is invalid for runtime addon when it's cluster arg is nil

zhaohuiweixiao:

install dependency is invalid for runtime addon when it's cluster arg is nil. Because the runtime addon will be render all clusters when there is no clusters arg.

I have:

`vela cluster join config --name devops`
`vela addon enable fluxcd --clusters={local}`
`vela addon enable rollout`
then you will find the fluxcd is not automatically installed on the devops cluster after enable rollout.

@wangyikewxgm

### DIFF
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -1015,18 +1015,11 @@ func (h *Installer) installDependency(addon *InstallPackage) error {
 			continue
 		}
 		depHandler := *h
-		// get dependency addon original parameters
-		depArgs, depArgsErr := GetAddonLegacyParameters(h.ctx, h.cli, dep.Name)
+		// reset dependency addon clusters parameter
+		depArgs, depArgsErr := getDependencyArgs(h.ctx, h.cli, dep.Name, depClusters)
 		if depArgsErr != nil {
-			if !apierrors.IsNotFound(depArgsErr) {
-				return depArgsErr
-			}
+			return depArgsErr
 		}
-		if depArgs == nil {
-			depArgs = map[string]interface{}{}
-		}
-		// reset the cluster arg
-		depArgs[types.ClustersArg] = depClusters
 
 		depHandler.args = depArgs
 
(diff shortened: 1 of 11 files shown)


pkg/addon/addon.go
```diff
@@ -1110,6 +1120,27 @@ func checkDependencyNeedInstall(ctx context.Context, k8sClient client.Client, de
 	return needInstallAddonDep, depClusters, nil
 }
 
+// getDependencyArgs resets the dependency clusters arg according needed install depClusters
+func getDependencyArgs(ctx context.Context, k8sClient client.Client, depName string, depClusters []string) (map[string]interface{}, error) {
```
wangyikewxgm, line 1124: If the func name is `getXXX`, it should not change any thing inside.

pkg/addon/addon.go
```diff
@@ -1092,6 +1085,23 @@ func checkDependencyNeedInstall(ctx context.Context, k8sClient client.Client, de
...
+		// if addon clusters is nil, meaning to deploy to all clusters
+		if addonClusters == nil {
+			hasTopologyPolicy := false
+			topologyPolicyValue := map[string]interface{}{}
+			for _, policy := range depApp.Spec.Policies {
+				if policy.Type == "topology" {
+					hasTopologyPolicy = true
+					unmarshalErr := json.Unmarshal(policy.Properties.Raw, &topologyPolicyValue)
+					if unmarshalErr != nil {
+						return true, nil, unmarshalErr
```
wangyikewxgm, line 1097: If an error is generated, first return value is useless, please keep it as default, which is `false`.

pkg/addon/addon.go
```diff
@@ -1092,6 +1085,23 @@ func checkDependencyNeedInstall(ctx context.Context, k8sClient client.Client, de
...
+					}
+					break
+				}
+			}
+			need := hasTopologyPolicy && topologyPolicyValue["clusterLabelSelector"] == nil
+			return need, nil, nil
+		}
 		// get the runtime clusters of current dependency addon
 		for _, r := range depApp.Status.AppliedResources {
 			if r.Cluster != "" && !stringslices.Contains(depClusters, r.Cluster) {
```
wangyikewxgm, line 1107: why we still need to judge the status of application?

pkg/addon/addon.go
```diff
@@ -1092,6 +1085,23 @@ func checkDependencyNeedInstall(ctx context.Context, k8sClient client.Client, de
...
+				if policy.Type == "topology" {
+					hasTopologyPolicy = true
+					unmarshalErr := json.Unmarshal(policy.Properties.Raw, &topologyPolicyValue)
+					if unmarshalErr != nil {
+						return true, nil, unmarshalErr
+					}
+					break
+				}
+			}
+			need := hasTopologyPolicy && topologyPolicyValue["clusterLabelSelector"] == nil
```
wangyikewxgm, line 1102: What is the topology policy already is `clusterLabelSelector: {}`?

pkg/addon/addon.go
zhaohuiweixiao: This is because the number of clusters may not be the same when between the fluxcd addon is installed and the rollout addon is installed.
When rollout clusters is nil, the clusters parameter of the fluxcd plugin is removed, so that the fluxcd addon is reinstalled on all clusters at the current time; However, if rollout clusters is not nil, the clusters parameter of the fluxcd addon will not be removed, and if the cluster is added between these two addon installations, the fluxcd plugin will not be re-installed. 
For example, when installing fluxcd for the first time without specifying clusters (`vela addon enable fluxcd`), there are c1 and c2 clusters. Then add c3 and install rollout(`vela addon enable rollout --clusters={c3}`). If use clusterLabelSelector to determine whether to install , then the fluxcd don't need to install when installing rollout, and we will miss the new joined c3 clusters for fluxcd.

wangyikewxgm (approved):

PR #5870, opened 2023-04-17
fix: when upgrade addon without version, the addon's version will be update by mistake

zhaohuiweixiao:

Fixes # when upgrade addon without version, the addon's version will be update by mistake

I have:

1. `vela addon enable fluxcd --version 2.1.9`
2. `vela upgrade enable fluxcd onlyHelmComponents=true`
after upgrade fluxcd, you will find the fluxcd version is the latest version on the addon registry.

@wangyikewxgm

### DIFF
--- a/references/cli/addon.go
+++ b/references/cli/addon.go
@@ -324,9 +324,14 @@ non-empty new arg
 					return fmt.Errorf("addon directory %s not found in local", addonOrDir)
 				}
 				name = addonOrDir
-				_, err = pkgaddon.FetchAddonRelatedApp(context.Background(), k8sClient, addonOrDir)
+				app, err := pkgaddon.FetchAddonRelatedApp(context.Background(), k8sClient, addonOrDir)
 				if err != nil {
 					return errors.Wrapf(err, "cannot fetch addon related addon %s", addonOrDir)
+				}
+				labels := app.GetLabels()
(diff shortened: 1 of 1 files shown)


wangyikewxgm:
I think it makes sense to upgrade the addon to the latest version if the user doesn't specify a version during the upgrade process.

zhaohuiweixiao:
> I think it makes sense to upgrade the addon to the latest version if the user doesn't specify a version during the upgrade process.

for fluxcd, I just wanted to update the onlyHelmComponents parameter, so when upgrade, i will not assign the version, but after the upgrade, the version was secretly updated. 
Update, on the other hand, means updating only the parameters specified by the user. for other addon args, it was not updated when no assigned. why only version was updated?

wangyikewxgm:
> > I think it makes sense to upgrade the addon to the latest version if the user doesn't specify a version during the upgrade process.
> 
> for fluxcd, I just wanted to update the onlyHelmComponents parameter, so when upgrade, i will not assign the version, but after the upgrade, the version was secretly updated. Update, on the other hand, means updating only the parameters specified by the user. for other addon args, it was not updated when no assigned. why only version was updated?

Upgrade" itself means upgrading to a new version, so I think it's reasonable to update to the latest version if no specific version is specified. If you want to keep the current version, you should add the "--version" parameter to fix the version.

PR #5947, opened 2023-04-30
Fix: adding topology strategy to addons to be valid only for the yaml type.

### LINKED ISSUE #5932: [Feature] A proposed change to automatically add topology policy rules to addon
The current implementation will automatically add the topology policy for the addon which could be deployed to the runtime clusters but do not have the topology policy, which will modify the user's addon and cause the template set in the addon to be inconsistent with the actual rendered template content (the topology policy is added). 
Then the rules for defining addon are not completely white-box, and user-defined addon are different from the ones actually rendered.

Considering that there is no way to define different topology policies based on clusters parameter in the yaml template of the addon, it is necessary to keep the logic of automatically adding the topology policy to the addon, but it is recommended to only apply this change to yaml addon to minimize the black-box behavior.

This way we can achieve a completely white-box definition of cue type addons, where what you see is what you get for cue type addons.

Therefore, it is recommended to restrict the rules that automatically add topology strategy to addons to be valid only for the yaml type.

**Additional context**

the model-serving template in addon catalog:
image

the rendered template after rendered:
image

Hanmengnan:

To achieve a completely white-box definition of cue type addons, adding topology strategy to addons to be valid only for the yaml type.

So when determining whether to attach the policy, to consider whether the addon was generated from a cue file.

Fixes #5932

I have:

### DIFF
--- a/pkg/addon/render.go
+++ b/pkg/addon/render.go
@@ -424,6 +424,10 @@ func checkNeedAttachTopologyPolicy(app *v1beta1.Application, addon *InstallPacka
 	if !isDeployToRuntime(addon) {
 		return false
 	}
+	// To achieve a completely white-box definition of cue type addons, adding topology strategy to addons to be valid only for the yaml type.
+	if len(addon.AppCueTemplate.Data) != 0 {
+		return false
+	}
 	for _, policy := range app.Spec.Policies {
 		if policy.Type == v1alpha1.TopologyPolicyType {
 			klog.Warningf("deployTo in metadata will NOT have any effect. It conflicts with %s policy named %s. Consider removing deployTo field in addon metadata.", v1alpha1.TopologyPolicyType, policy.Name)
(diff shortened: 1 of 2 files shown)


zhaohuiweixiao:
I think this check logic shoud put before the runtime check because the template type's priority is higher than the runtime addon.
also, there's no need to actually construct a template cue in the unit test, because we're just checking to see if it's data is empty and don't care about the contents of the file.
And I'm working on this issue(https://github.com/kubevela/kubevela/issues/5932) myself, so you don't need to follow up it.

Hanmengnan:
> I think this check logic shoud put before the runtime check because the template type's priority is higher than the runtime addon. also, there's no need to actually construct a template cue in the unit test, because we're just checking to see if it's data is empty and don't care about the contents of the file. And I'm working on this issue(#5932) myself, so you don't need to follow up it.

OK, I get it.

PR #6002, opened 2023-05-16
Feat(#5861): Support addon dependencies version ranges

### LINKED ISSUE #5861: Addon dependencies version range
As a KubeVela addon developer, when defining dependencies for an addon in metadata.yaml, I would like to set a version range for these dependencies, so that my addon will be restricted from being enabled in a cluster where the dependency version doesn't match the specified range.

I see in the [docs](https://kubevela.io/docs/platform-engineers/addon/intro) that there is a dependencies[].version property, but KubeVela doesn't seem to use this value to enforce any version requirements. Ideally this property would work similarly to the existing `system` property.

For example, in "metadata.yaml":
```yaml
name: my-addon
version: 0.1.0

dependencies:
  - name: fluxcd
    version: ">= 2.3.3, < 3.0.0"
  - name: foo
    version: ">= 0.1.0, < 1.0.0"

system:
  vela: ">= v1.7.0"
  kubernetes: ">= 1.24.0"
```

Specific behavior:
* When enabling addon, check whether dependencies are enabled within existing cluster.
* For any dependency that is already enabled, verify that version matches provided version range.
  * If version is outside range, fail. Do not update dependency version.
* For any dependency that is not enabled, attempt to enable while considering version range.
  * Fetch all versions of the addon, and enable the latest version that matches the version range.
  * If no addon version is available that matches the version range, fail.

merusso:

This change enables addon maintainers to define version ranges for dependencies in an addon's metadata.yaml file.

This behavior is similar to the version range allowed in the `system` section of the metadata file. The version range expression for `dependencies` follows the same format as for `system`.

Example:

```yaml
dependencies:
  - name: addon1
    version: ">= 2.3.3, < 3.0.0"
  - name: addon2
    version: ">= 0.1.0, < 1.0.0"
```

When installing an addon, the behavior varies depending on whether the dependency is already installed.

If a dependency is already installed, the installed version will be validated against the version range, and installation will fail with an error if there's a mismatch.
If a dependency is not installed, the version range will be used to select the addon version to be installed. If no addon version matching the range exists, the installation will fail with an error.

Fixes #5861

I have:

Added unit tests for new functions.
Manually tested on an existing cluster.


pkg/addon/addon.go
```diff
@@ -1072,6 +1092,210 @@ func (h *Installer) installDependency(addon *InstallPackage) error {
...
+	// sort versions in descending order
+	sort.Sort(sort.Reverse(semver.Collection(sortedVersions)))
+	for _, v := range sortedVersions {
+		sortedVersionStrings = append(sortedVersionStrings, v.String())
+	}
+	return sortedVersionStrings
+}
+
+// AddonInfo contains summary information about an addon
+type AddonInfo struct {
```
anoop2811, line 1217: I think the github review is valid. Should we call it addon.Info with the package prefix explicitly showing this is for an addon?

merusso:
FYI, I'm having issues running `make reviewable` successfully, looking into it.

pkg/addon/addon.go
```diff
@@ -1072,6 +1092,210 @@ func (h *Installer) installDependency(addon *InstallPackage) error {
...
+		// get installed version
+		installedAddon := installedAddons[dep.Name]
+		// versions length must be 1
+		if len(installedAddon.AvailableVersions) != 1 {
+			err = multierr.Append(err, errors.New("installedAddon.Versions length must be 1"))
+			continue
+		}
+		installedVersion := installedAddon.AvailableVersions[0]
+
+		match, _ := checkSemVer(installedVersion, dep.Version)
```
StevenLeiZhang, line 1125: For compatibility, some existed addons may not specify `dep.Version`. What  `match` value will be returned? 
For those existed addons without 'dep.Version`, do we need reinstall their dependencies?

merusso: The expected behavior is for `dep.Version == nil` to accept any version. The function currently works this way, and there are test cases to cover this.

Currently this relies on `checkSemVer` returning `true` when 2nd param (`required`) is nil. However, I can make this more explicit by adding this check before calling `checkSemVer`.

pkg/addon/addon.go
```diff
@@ -1072,6 +1092,210 @@ func (h *Installer) installDependency(addon *InstallPackage) error {
...
+	Name              string
+	Description       string
+	AvailableVersions []string
+}
+
+type addonInfoMap map[string]addonInfo
+
+// listAvailableAddons fetches a collection of addons available in the given
+// registry. Returns a map of AddonInfo grouped by addon name.
+func listAvailableAddons(registry Registry) (addonInfoMap, error) {
```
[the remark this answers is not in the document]
merusso: I would prefer if we could use existing code for `listAvailableAddons` and `listInstalledAddons` but I couldn't find existing functions to use.

Unfortunately the `listAddons` func in [references/cli/addon.go](https://github.com/kubevela/kubevela/blob/aa36edbc3a1a561d626321503b85a03b4f153424/references/cli/addon.go#L962) has a lot of internal logic instead of calling a reusable func in pkg/. I think `listAddons` could use some refactoring to address this.

In [my comment above](https://github.com/kubevela/kubevela/pull/6002/files#r1196782209), I suggested extending the Source interface to replace these new funcs.

pkg/addon/addon.go
```diff
@@ -1072,6 +1092,210 @@ func (h *Installer) installDependency(addon *InstallPackage) error {
...
+				AvailableVersions: a.AvailableVersions,
+			}
+		}
+	}
+	return availableAddons, nil
+}
+
+// listInstalledAddons fetches a collection of addons installed in the cluster.
+// Returns a map of AddonInfo grouped by addon name.
+func listInstalledAddons(ctx context.Context, k8sClient client.Client) (addonInfoMap, error) {
```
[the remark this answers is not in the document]
merusso: Good question. Let's continue discussion in one of the threads for `listAvailableAddons`.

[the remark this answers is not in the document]
merusso: This function adds one k8s API call to list Applications in the KubeVela namespace. I'm not aware of any performance issues here or any more performant way to get this data.

pkg/addon/addon.go
merusso: I updated this function to continue (skip version check) when `dep.Version == nil`

pkg/addon/addon.go
merusso: I'll resolve this conversation for now since the lint check is resolved.

pkg/addon/addon.go
```diff
@@ -1072,6 +1092,221 @@ func (h *Installer) installDependency(addon *InstallPackage) error {
...
+		if !ok {
+			err = multierr.Append(err, fmt.Errorf(
+				"addon %s has unresolvable dependency %s, required version '%s', no available addon with name",
+				addon.Name, dep.Name, dep.Version))
+		} else {
+			if dep.Version == "" {
+				continue
+			}
+
+			_, err2 := calculateDependencyVersionToInstall(dep, nil, availableAddons)
```
zhaohuiweixiao, line 1152: Maybe we could return a variable that records the version of each dependency addon that needs to be installed, so we don't have to run this method again to get the version that the addon later.
Thus, The logic that installedAddons are not nil  in calculateDependencyVersionToInstall function is also not needed.

pkg/addon/addon.go
```diff
@@ -1072,6 +1092,221 @@ func (h *Installer) installDependency(addon *InstallPackage) error {
...
+			err = multierr.Append(err, errors.New("installedAddon.Versions length must be 1"))
+			continue
+		}
+		installedVersion := installedAddon.AvailableVersions[0]
+
+		match, _ := checkSemVer(installedVersion, dep.Version)
+		if !match {
+			err = multierr.Append(err, fmt.Errorf(
+				"addon %s has unresolvable dependency %s, required version '%s', installed version '%s'",
+				addon.Name, dep.Name, dep.Version, installedVersion))
```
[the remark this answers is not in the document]
merusso: I was imagining this would work like traditional validation workflows that capture and return multiple validation errors. This has the following benefits:

1. Users can identify and fix multiple errors to avoid a trial and error loop.
2. It's cheap to evaluate all errors, because the expensive part (API calls) is performed once, not for every error.

For an example, consider how Kubernetes validation errors are returned when applying an invalid object:

```
$ kubectl apply -f - << EOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: invalid
a: a1
b: b1
c: c1
EOF

error: error validating "STDIN": error validating data: [ValidationError(ConfigMap): unknown field "a" in io.k8s.api.core.v1.ConfigMap, ValidationError(ConfigMap): unknown field "b" in io.k8s.api.core.v1.ConfigMap, ValidationError(ConfigMap): unknown field "c" in io.k8s.api.core.v1.ConfigMap]; if you choose to ignore these errors, turn validation off with --validate=false
```

pkg/addon/addon.go
merusso: That's true, we could do this. I worried that this might make the validation function less clean since it would have 2 separate responsibilities:
1. Validate dependencies
2. Calculate dependency versions to install

zhaohuiweixiao: This concern is warranted. If we can use to replace validateAddonDependencies calculateDependencyVersionToInstall directly, ValidateAddonDependencies logic is also have to look for the appropriate version can be installed. we can use calculateDependencyVersionToInstall directly, if no suitable version directly return error, if found will continue.

pkg/addon/addon.go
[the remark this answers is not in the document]
zhaohuiweixiao: OK. This is fine from the perspective of following Kubernetes validation rules. Now that addon don't rely on much, it might not be too expensive to evaluate all errors.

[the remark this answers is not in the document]
zhaohuiweixiao: OK. This is fine from the perspective of following Kubernetes validation rules. Now that addon don't rely on much, it might not be too expensive to evaluate all errors.

pkg/addon/addon_test.go
```diff
@@ -1242,3 +1242,349 @@ func TestGenerateConflictError(t *testing.T) {
...
+	}
+	for _, tc := range testCases {
+		res, err := calculateDependencyVersionToInstall(tc.dep, tc.installedAddons, tc.availableAddons)
+		assert.Equal(t, tc.res, res, tc.caseName)
+		assert.Equal(t, tc.err, err, tc.caseName)
+	}
+}
+
+func TestListAvailableAddons(t *testing.T) {
+	t.Skip("TODO: implement this test")
```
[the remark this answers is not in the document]
merusso: Since Registry is currently a struct, using a mock may take a lot of refactoring. It might be more practical to write integration tests instead.

pkg/addon/addon.go
merusso: If I understand you correctly, you're suggesting that I use calculateDependencyVersionToInstall instead of validateAddonDependencies? If so, I don't believe I can do this currently because calculateDependencyVersionToInstall doesn't "validate" installed addons -- it returns the installed addon version regardless of whether the installed addon version matches the `dependency.Version`.

I will investigate some further refactoring to see if I can simplify validateAddonDependencies further.

merusso: I refactored validateAddonDependencies to be a minimal wrapper over calculateDependencyVersionToInstall. Please review again :) .

pkg/addon/addon.go
```diff
@@ -1072,6 +1092,210 @@ func (h *Installer) installDependency(addon *InstallPackage) error {
...
+	Name              string
+	Description       string
+	AvailableVersions []string
+}
+
+type addonInfoMap map[string]AddonInfo
+
+// listAvailableAddons fetches a collection of addons available in the given
+// registry. Returns a map of AddonInfo grouped by addon name.
+func listAvailableAddons(registry Registry) (addonInfoMap, error) {
```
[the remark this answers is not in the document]
merusso: I added `ListAddonInfo()` to `Registry` where it likely belongs. This simplifies `listAvailableAddons`.

pkg/addon/addon_test.go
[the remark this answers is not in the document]
merusso: I created an interface `AddonInfoLister` so I can mock the registry in unit tests.

pkg/addon/addon.go
```diff
@@ -999,6 +1001,20 @@ func (h *Installer) getAddonMeta() (map[string]SourceMeta, error) {
 
 // installDependency checks if addon's dependency and install it
 func (h *Installer) installDependency(addon *InstallPackage) error {
+	installedAddons, err := listInstalledAddons(h.ctx, h.cli)
+	if err != nil {
+		return err
+	}
+	availableAddons, err := listAvailableAddons(*h.r)
```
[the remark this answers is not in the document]
merusso: This has been added. Please review.

merusso:
All checks pass now. I still need to do some testing, but once that completes, I think this is "ready for review" (no longer Draft)

chivalryq (approved):
LGTM, great job!

Somefive (approved):

FogDong (approved):

pkg/addon/addon.go
merusso: I eventually decided that this type should be exported, since I added a method to Registry that returns this, so I renamed this to ItemInfo.

wangyikewxgm (approved):

merusso:
I'll work on updating the documentation [here](https://kubevela.io/docs/platform-engineers/addon/intro#basic-information-file) :)

wonderflow:
@merusso great job!

merusso:
Thank you all!

PR #6003, opened 2023-05-16
Feat: support resource update policy

### LINKED ISSUE #5594: [Feature] ImmutableAfterDeploy
We have a spec field in our CRD definition it corresponds to the real resource's name. It is a unique field. We don't expect to be another duplicate with the same resource name. To ensure that we added a validation webhook. This field is immutable also. There is no way to change this field with CRD edit. Our webhook provides immutability and uniqueness checks. This field is also corresponds to our CRD's metada.name field. That means if this field changes then whole CRD recreated in deployments. 

When it comes to Kubevela UI side we couldn't give immutability to this field. If we add the setting `immutable: true` from UI definitions, then there is no way to create a new one because it is completely not editable. I think we need a setting like `ImmutableAfterDeploy` this way we can enter a new record and it will be immutable after the deployment.

Somefive:

1. Support "resource-update" policy to allow users customize the update behavior for selected resources.
2. Rename FeatureGate "ApplyResourceByUpdate" to "ApplyResourceByReplace" to avoid confusion.

`resource-update` policy can allow users to customize the update behavior for selected resources.

```yaml
kind: Application
metadata:
  name: recreate
spec:
  components:
    - type: k8s-objects
      name: recreate
      properties:
        objects:
          - apiVersion: v1
            kind: Secret
            metadata:
              name: recreate
            data:
              key: dgo=
            immutable: true
  policies:
    - type: resource-update
      name: resource-update
      properties:
        rules:
          - selector:
              resourceTypes: ["Secret"]
            strategy:
              recreateFields: ["data.key"]
```
By specifying `recreateFields`, the application will recreate the target resource (**Secret** here) when the field changes (`data.key` here). If the field is not changed, the application will use the normal update (**patch** here).

```yaml
kind: Application
metadata:
  name: recreate
spec:
  components:
    - type: k8s-objects
      name: recreate
      properties:
        objects:
          - apiVersion: v1
            kind: ConfigMap
            metadata:
              name: recreate
            data:
              key: val
  policies:
    - type: resource-update
      name: resource-update
      properties:
        rules:
          - selector:
              resourceTypes: ["ConfigMap"]
            strategy:
              op: replace
```
By specifying `op` to `replace`, the application will update the given resource (ConfigMap here) by replace. Compared to **patch**, which leverages three-way merge patch to only modify the fields managed by KubeVela application, "replace" will update the object as a whole and wipe out other fields even if it is not managed by the KubeVela application. It can be seen as an "application-level" *ApplyResourceByReplace*.

Fixes #5594

I have:


test/e2e-multicluster-test/multicluster_test.go
```diff
@@ -1166,5 +1166,55 @@ var _ = Describe("Test multicluster scenario", func() {
...
+			By("update configmap")
+			Eventually(func(g Gomega) {
+				cm := &corev1.ConfigMap{}
+				g.Expect(k8sClient.Get(ctx, appKey, cm)).Should(Succeed())
+				cm.Data["extra"] = "extra-val"
+				g.Expect(k8sClient.Update(ctx, cm)).Should(Succeed())
+			}).WithPolling(2 * time.Second).WithTimeout(20 * time.Second).Should(Succeed())
+
+			By("update application")
+			Expect(yaml.Unmarshal([]byte(strings.ReplaceAll(strings.ReplaceAll(string(bs), "key: dgo=", "key: dnZ2Cg=="), "key: val", "key: val2")), app)).Should(Succeed())
```
chivalryq, line 1196: Any differences if I patch or update the same "key" from "val" to "val2"?

PR #6016, opened 2023-05-22
Chore: set CLI verbose output to be hidden by default

### LINKED ISSUE #5561: Hide the output of the warning log while the CLI tool is working
When using the CLI tool, we often encounter the output of hidden warning logs. The output of these logs is unnecessary and boring. So can we hide the output of these logs while the CLI tool is working.

> 
> image
> This is not a single issue only related to resource tree. It is a general issue for all CLI commands to print out controller typed logs. I would suggest dealing it in a separate work, not only fix it here.
> _Originally posted by @Somefive in https://github.com/kubevela/kubevela/issues/5479#issuecomment-1426644723_

In some CLI tools, we have hidden the log output, but in fact it is cumbersome to configure each tool. We should provide a unified entry so that we can configure the log output level.

Ref: #5448

Somefive:

Add "--verbosity"/"-V" global flag for vela cli. By default 0.

When 0, all "klog.Info" will be omitted. "klog.Error" will be written to stderr.

image

When greater than 0, "klog.Info" will be written to stdout.

image

When greater than 3, the call file and time will be printed.

image

When greater than 6, all details will be printed.

image

Fixes #5561

I have:

### DIFF
--- a/references/cli/cli.go
+++ b/references/cli/cli.go
@@ -25,6 +25,7 @@ import (
 	gov "github.com/hashicorp/go-version"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
@@ -36,6 +37,7 @@ import (
 	"github.com/oam-dev/kubevela/pkg/utils/helm"
 	"github.com/oam-dev/kubevela/pkg/utils/system"
 	"github.com/oam-dev/kubevela/pkg/utils/util"
+	velalog "github.com/oam-dev/kubevela/references/cli/log"
 	"github.com/oam-dev/kubevela/version"
 )
 
@@ -138,9 +140,16 @@ func NewCommandWithIOStreams(ioStream util.IOStreams) *cobra.Command {
 
 	fset := flag.NewFlagSet("logs", flag.ContinueOnError)
 	klog.InitFlags(fset)
+	pfset := pflag.NewFlagSet("logs", pflag.ContinueOnError)
+	pfset.AddGoFlagSet(fset)
+	pflg := pfset.Lookup("v")
+	pflg.Name = "verbosity"
+	pflg.Shorthand = "V"
 
+	klog.SetLogger(velalog.NewLogger("vela-cli"))
 	// init global flags
 	cmds.PersistentFlags().BoolVarP(&assumeYes, "yes", "y", false, "Assume yes for all user prompts")
+	cmds.PersistentFlags().AddFlag(pflg)
 	return cmds
 }
 
--- a/references/cli/log/logger.go
+++ b/references/cli/log/logger.go
@@ -1,0 +1,132 @@
+/*
+Copyright 2023 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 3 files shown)


chivalryq:
Consider add a short version. Say `-V`? WDYT

Somefive:
> Consider add a short version. Say `-V`? WDYT

The "-v" is used by some commands, like vela install. Upper case is okay for me, but not aligned with the `kubectl -v`. /cc @wonderflow @wangyikewxgm WDYT

Somefive:
> `vela --verbose` is better? `-V` is a bit wield

This is how "KinD" use the flag.
```
kind creates and manages local Kubernetes clusters using Docker container 'nodes'

Usage:
  kind [command]

Available Commands:
  build       Build one of [node-image]
  completion  Output shell completion code for the specified shell (bash, zsh or fish)
  create      Creates one of [cluster]
  delete      Deletes one of [cluster]
  export      Exports one of [kubeconfig, logs]
  get         Gets one of [clusters, nodes, kubeconfig]
  help        Help about any command
  load        Loads images into nodes
  version     Prints the kind CLI version

Flags:
  -h, --help              help for kind
      --loglevel string   DEPRECATED: see -v instead
  -q, --quiet             silence all stderr output
  -v, --verbosity int32   info log verbosity, higher value produces more output
      --version           version for kind
```

This is k3d
```
https://k3d.io/
k3d is a wrapper CLI that helps you to easily create k3s clusters inside docker.
Nodes of a k3d cluster are docker containers running a k3s image.
All Nodes of a k3d cluster are part of the same docker network.

Usage:
  k3d [flags]
  k3d [command]

Available Commands:
  cluster      Manage cluster(s)
  completion   Generate completion scripts for [bash, zsh, fish, powershell | psh]
  config       Work with config file(s)
  help         Help about any command
  image        Handle container images.
  kubeconfig   Manage kubeconfig(s)
  node         Manage node(s)
  registry     Manage registry/registries
  version      Show k3d and default k3s version

Flags:
  -h, --help         help for k3d
      --timestamps   Enable Log timestamps
      --trace        Enable super verbose output (trace logging)
      --verbose      Enable verbose output (debug logging)
      --version      Show k3d and default k3s version
```

This is kubectl.
```
    --tls-server-name='':
	Server name to use for server certificate validation. If it is not provided, the hostname used to contact the
	server is used

    --token='':
	Bearer token for authentication to the API server

    --user='':
	The name of the kubeconfig user to use

    --username='':
	Username for basic authentication to the API server

    -v, --v=0:
	number for the log level verbosity

    --vmodule=:
	comma-separated list of pattern=N settings for file-filtered logging (only works for the default text log
	format)

    --warnings-as-errors=false:
	Treat warnings received from the server as errors and exit with a non-zero exit code
```

So, if we would like to use the verbose level, I think `verbosity` is the best choice. It would be good if we use `-v` for short name but that could cause a conflict with other commands using `-v` for version.

It you prefer `--verbose`, then it should be a boolean flag and we might also need `--trace` for it. Less configuration capability and less confusion. The advantage of using `--verbosity` is that it is aligned with all other Kubernetes project's internal logging behaviour. The advantage of using `--verbose` is that it is easier for user to understand and use.

@wonderflow @FogDong  WDYT

wonderflow:
Great, verbosity +1, suppurt `--verbosity` and `-V` is fine.

Somefive:
> Great, verbosity +1, suppurt `--verbosity` and `-V` is fine.

Done

wonderflow (approved):

chivalryq (approved):

PR #6033, opened 2023-05-24
Feat: support force resource location with dispatch

### LINKED ISSUE #6032: [Feature] Indicate the resources to be dispatched to which cluster

We are developing custom components and traits to provide the capability to deploy cloud resources with the application. Sometimes we need to control where CRDs are generated. Trait definition has a "controlPlaneOnly" attribute which helps us to achieve what we want. But we can't do it for component definitions. 

If the desired resources are optional we provide the resource with a custom trait with the "controlPlaneOnly" attribute. But sometimes resources are mandatory for the component and we can't provide it as a trait.

For example, we developed a web service component. The component deploys the application to the remote clusters but also it generates auxiliary resources to bootstrap the observability configuration for the application. These CRDs are don't need on the remote clusters, we want to dedicate these to the hub cluster. 

@Somefive : 
> One of the way that could be possible is to add some specific annotation to the rendered resource, like `app.oam.dev/cluster: local` and let the controller to be able to recognize that and put it on the hub cluster, as we cannot design only one field to describe multiple resources rendered by the componentdefinition.

Somefive:

In this PR, we support setting the cluster to dispatch with the rendered resource directly, through setting the "app.oam.dev/cluster" label manually.

For example,
```yaml
kind: Application
metadata:
  name: app
spec:
  components:
    - type: k8s-objects
      name: app
      properties:
        objects:
          - apiVersion: v1
            kind: ConfigMap
            metadata:
              name: "x"
              labels:
                app.oam.dev/cluster: local
            data:
              key: "x"
          - apiVersion: v1
            kind: ConfigMap
            metadata:
              name: "y"
            data:
              key: "y"
  policies:
    - type: topology
      name: topology
      properties:
        clusters: ["cluster-worker"]
```
The above application will create the ConfigMap "x" in the local cluster and the ConfigMap "y" in the "cluster-worker" cluster.

With this capability, it is possible for users to design ComponentDefinition/TraitDefinitions that only deliver resources to the hub control plane no matter what topology is used.

NOTE: This label is originally attached to the rendered resources before they are applied with the topology policy. So this PR reuse that label and does not affect existing system.

Fixes #6032

I have:

### DIFF
--- a/test/e2e-multicluster-test/testdata/app/app-with-fixed-location.yaml
+++ b/test/e2e-multicluster-test/testdata/app/app-with-fixed-location.yaml
@@ -1,0 +1,29 @@
+apiVersion: core.oam.dev/v1beta1
+kind: Application
+metadata:
+  name: app
+spec:
+  components:
+    - type: k8s-objects
+      name: app
+      properties:
+        objects:
+          - apiVersion: v1
+            kind: ConfigMap
+            metadata:
+              name: "x"
+              labels:
+                app.oam.dev/cluster: local
+            data:
+              key: "x"
+          - apiVersion: v1
+            kind: ConfigMap
+            metadata:
+              name: "y"
+            data:
+              key: "y"
+  policies:
+    - type: topology
+      name: topology
+      properties:
+        clusters: ["cluster-worker"]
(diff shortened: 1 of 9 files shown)


test/e2e-multicluster-test/testdata/app/app-with-fixed-location.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - type: k8s-objects
+      name: app
+      properties:
+        objects:
+          - apiVersion: v1
+            kind: ConfigMap
+            metadata:
+              name: "x"
+              labels:
+                app.oam.dev/cluster: local
```
FogDong, line 16: So this label will override the cluster in topology, right?
If we write a workflow step like:
```
resource: op.#Apply & {
  cluster: "test"
  value: {
            apiVersion: v1
            kind: ConfigMap
            metadata:
              name: "x"
              labels:
                app.oam.dev/cluster: local
  }
}
```
The config map will be applied to the local cluster, right?

Somefive: Right.

PR #6059, opened 2023-05-31
Feat: support custom cascading gc policy

Somefive:

The default behaviour of deleting Job in Kubernetes will orphan the underlying pods. The initial design is to ensure the logs of pods can be reserved. But there are cases that KubeVela users do not want to orphan pods and want to delete them when Job or upper layer application is recycled.

This PR adds a "cascading" fields in the "garbage-collect" policy, which allows user to customize the deletion behaviour for selected resources. If not set, KubeVela controller will use the native Kubernetes behaviour for deletion.

Example:
```yaml
kind: Application
metadata:
  name: gc-propagation
spec:
  components:
    - type: k8s-objects
      name: cascading-gc
      properties:
        objects:
          - apiVersion: batch/v1
            kind: Job
            spec:
              template:
                spec:
                  containers:
                    - name: pi
                      image: perl:5.34.0
                      command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
                  restartPolicy: Never
              backoffLimit: 4
  policies:
    - type: garbage-collect
      properties:
        rules:
          - selector:
              componentNames: ["cascading-gc"]
            propagation: cascading
```
In the above example, the job will recycle the underlying pods when it is deleted. If the policy is not configured, the pods will be left even if the job and the app is deleted.

I have:


.github/workflows/e2e-test.yml
```diff
@@ -80,12 +80,6 @@ jobs:
           go install github.com/onsi/ginkgo/v2/ginkgo
           go get github.com/onsi/gomega/...
 
-      - name: Setup KinD
```
chivalryq, line 83: Why are we deleting this?

apis/core.oam.dev/v1alpha1/garbagecollect_policy_types.go
```diff
@@ -73,6 +76,16 @@ const (
...
 )
 
+// GarbageCollectPropagation the deletion propagation setting similar to metav1.DeletionPropagation
+type GarbageCollectPropagation string
+
+const (
+	// GarbageCollectPropagationOrphan orphan child resources while deleting target resources
+	GarbageCollectPropagationOrphan = "orphan"
+	// GarbageCollectPropagationBackground delete child resources in background while deleting target resources
+	GarbageCollectPropagationBackground = "background"
```
chivalryq, line 86: How about `cascading`? I think backgroud is kinda confusing since we don't allow `foreground` GC option.

.github/workflows/e2e-test.yml
Somefive: Because we have two "Setup KinD" in e2e-test.

apis/core.oam.dev/v1alpha1/garbagecollect_policy_types.go
Somefive: Actually, "foreground" is okay for client-go but we should not allow it here. It will block the controller and should only be a CLI case.

Somefive: I also think `cascading` as a good choice as well.

Somefive: Renamed to `cascading`.

Somefive: The Type still named as Propagation to aligned with the controller-runtime setting.

Somefive: I think about "cascading" as a good choice because that's how kubectl use for flags. But I later confirmed from the bare API parameter and see the PropagationPolicy used for DELETE api.
image

Somefive: Great, I renamed the value of "background" to "cascading".

vela-templates/definitions/internal/policy/garbage-collect.cue
```diff
@@ -12,6 +12,8 @@ template: {
 		selector: #ResourcePolicyRuleSelector
 		// +usage=Specify the strategy for target resource to recycle
 		strategy: *"onAppUpdate" | "onAppDelete" | "never"
+		// +usage=Specify the deletion propagation strategy for target resource to delete
+		propagation?: "orphan" | "cascading"
```
FogDong, line 16: Which propagation will be used by default?

Somefive: The default case is having no propagation settings. For Kubernetes, when the propagation setting is absent, it will determine the propagation by resources. For example, for deployments, it will be "cascading" (or in other word "background"), but for jobs, it will be "orphan". It works like "auto".

FogDong (approved):

chivalryq (approved):

PR #6063, opened 2023-05-31
Fix: vela def vet can be used without Kubernetes

Somefive:

`vela def vet` should not need Kubernetes environment to run. This is useful because CUE community can therefore run CI without KinD or KubeBuilder in their docker container to verify if the version upgrade of CUE won't break the usage of KubeVela.

I have:


FogDong (approved):

references/cli/def.go
```diff
@@ -772,7 +773,7 @@ func NewDefinitionRenderCommand(c common.Args) *cobra.Command {
 				}
 				config, err := c.GetConfig()
 				if err != nil {
-					return err
+					klog.Infof("ignore kubernetes cluster, unable to get kubeconfig: %s", err.Error())
```
wonderflow, line 776: which parts of feature in this command rely on this config?

Somefive: So it is due to the usage of legacy PackageDiscover. But it is actually not working now so the config is not required. We will remove it entirely in https://github.com/kubevela/kubevela/pull/6031

chivalryq (approved):

PR #6065, opened 2023-05-31
Fix: addon args is not changed when reinstall it without args

### LINKED ISSUE #6064: addon args value is not changed when reinstall it without args
**Describe the bug**
addon args value is not changed when reinstall it without args.

**To Reproduce**
1、vela addon enable fluxcd --clusters={local}
2、vela addon enable fluxcd

**Expected behavior**

fluxcd clusters value is nil

**Screenshots**

image

**KubeVela Version**
1.6.5

zhaohuiweixiao:
### Summary
🗑️🧩🔧

Add feature to uninstall addons by deleting secret file. Modify `pkg/addon/addon.go` to delete the secret file that stores the addon's arguments when uninstalling an addon.

> _`deleteArgsSecret`_
> _Uninstalling addons now_
> _Autumn leaves fall fast_

### Walkthrough
*  Add a function to delete the secret file for addon arguments (`[link](https://github.com/kubevela/kubevela/pull/6065/files?diff=unified&w=0#diff-dfc3f6e3b7231b7b41cf82b8e561f3cbf8191d214886e21d798f287f693806dbR847-R858)` in `pkg/addon/addon.go`)

Fixes #6064

I have:

1、vela addon enable fluxcd --clusters={local}
2、vela addon enable fluxcd

@wangyikewxgm

### DIFF
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -844,6 +844,18 @@ func RenderArgsSecret(addon *InstallPackage, args map[string]interface{}) *unstr
 	return u
 }
 
+// deleteArgsSecret delete the addon's args secret file
+func deleteArgsSecret(ctx context.Context, k8sClient client.Client, addonName string) {
+	var sec v1.Secret
+	err := k8sClient.Get(ctx, client.ObjectKey{Namespace: types.DefaultKubeVelaNS, Name: addonutil.Addon2SecName(addonName)}, &sec)
+	if err == nil {
+		err = k8sClient.Delete(ctx, &sec)
+		if err != nil {
+			fmt.Printf("fail to delete %v addon parameter file", addonName)
+		}
+	}
+}
+
 // FetchArgsFromSecret fetch addon args from secrets
 func FetchArgsFromSecret(sec *v1.Secret) (map[string]interface{}, error) {
 	res := map[string]interface{}{}
@@ -1290,6 +1302,9 @@ func (h *Installer) dispatchAddonResource(addon *InstallPackage) error {
 		if err != nil {
 			return err
 		}
+	} else {
+		// delete addon args secret file
+		deleteArgsSecret(h.ctx, h.cli, addon.Name)
 	}
 	return nil
 }
--- a/pkg/addon/addon.go
+++ b/pkg/addon/addon.go
@@ -845,15 +845,18 @@ func RenderArgsSecret(addon *InstallPackage, args map[string]interface{}) *unstr
 }
 
 // deleteArgsSecret delete the addon's args secret file
-func deleteArgsSecret(ctx context.Context, k8sClient client.Client, addonName string) {
+func deleteArgsSecret(ctx context.Context, k8sClient client.Client, addonName string) error {
 	var sec v1.Secret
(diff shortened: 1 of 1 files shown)


Somefive:
Can we leave out an arg like "vela addon enable --reset-values" to make this issue work?

For example, in helm, when we run "helm upgrade", it will keep the parameters for the previous installation by default, which is "vela addon" currently follows. You can checkout the parameters in "helm upgrade -h".

```shell
      --reset-values                               when upgrading, reset the values to the ones built into the chart
      --reuse-values                               when upgrading, reuse the last release's values and merge in any overrides from the command line via --set and -f. If '--reset-values' is specified, this is ignored
```

zhaohuiweixiao:
> Can we leave out an arg like "vela addon enable --reset-values" to make this issue work?
> 
> For example, in helm, when we run "helm upgrade", it will keep the parameters for the previous installation by default, which is "vela addon" currently follows. You can checkout the parameters in "helm upgrade -h".
> 
> ```shell
>       --reset-values                               when upgrading, reset the values to the ones built into the chart
>       --reuse-values                               when upgrading, reuse the last release's values and merge in any overrides from the command line via --set and -f. If '--reset-values' is specified, this is ignored
> ```

@Somefive ，But now this is not only a problem when upgrading, but also when reinstalling. when run `vela addon enable fluxcd --clusters={local}` first, then run `vela addon enable fluxcd`, the parameter file previous installation is also be retained.

Somefive:
> > Can we leave out an arg like "vela addon enable --reset-values" to make this issue work?
> > For example, in helm, when we run "helm upgrade", it will keep the parameters for the previous installation by default, which is "vela addon" currently follows. You can checkout the parameters in "helm upgrade -h".
> > ```shell
> >       --reset-values                               when upgrading, reset the values to the ones built into the chart
> >       --reuse-values                               when upgrading, reuse the last release's values and merge in any overrides from the command line via --set and -f. If '--reset-values' is specified, this is ignored
> > ```
> 
> @Somefive ，But now this is not only a problem when upgrading, but also when reinstalling. when run `vela addon enable fluxcd --clusters={local}` first, then run `vela addon enable fluxcd`, the parameter file previous installation is also be retained.

You mean `vela addon enable fluxcd --clusters={local}`  -> `vela addon disable fluxcd` -> `vela addon enable fluxcd`? Is there a disable operation?

Somefive (approved):
Sounds reasonable for me now. I'm okay with this change.

pkg/addon/addon.go
```diff
@@ -844,6 +844,18 @@ func RenderArgsSecret(addon *InstallPackage, args map[string]interface{}) *unstr
...
+	var sec v1.Secret
+	err := k8sClient.Get(ctx, client.ObjectKey{Namespace: types.DefaultKubeVelaNS, Name: addonutil.Addon2SecName(addonName)}, &sec)
+	if err == nil {
+		err = k8sClient.Delete(ctx, &sec)
+		if err != nil {
+			fmt.Printf("fail to delete %v addon parameter file", addonName)
+		}
+	}
+}
+
```
wangyikewxgm, line 858: If the error returned from the `get` function is not `errorNotFound`, please return the error.

It's highly recommended to handle this logic using the following pattern:
```
if err := k8sClient.Get(xxx); err == nil {
    // Handle successful get operation
} else {
    // Handle error
    if err != errorNotFound {
        return err
    }
}
```

wangyikewxgm (approved):

PR #6066, opened 2023-05-31
Fix: can not get the APISchema when registry is not version registry

zhaohuiweixiao:
### Summary
🐛🌐🚀

Fixed a bug in displaying UI data of addon packages in the web console. Changed `FindAddonPackagesDetailFromRegistry` to use `UIMetaOptions` instead of `CLIMetaOptions` in `pkg/addon/helper.go`.

> _`UIMetaOptions`_
> _Fix addon UI data bug_
> _Winter console glows_

### Walkthrough
* Fix UI data bug for addon packages by using `UIMetaOptions` instead of `CLIMetaOptions` in `FindAddonPackagesDetailFromRegistry` ([link](https://github.com/kubevela/kubevela/pull/6066/files?diff=unified&w=0#diff-c2cf26c57a54e0094a439c4c9f01e066d75f01eefc0264bfbcdc9d1524161d7cL257-R257))

FindAddonPackagesDetailFromRegistry can not get the APISchema data when registry is not version registry

I have:

1. add git type addon registry
2. enable fluxcd from this registry with local clusters
3. enable fluxcd from this registry without args

@wangyikewxgm

### DIFF
--- a/pkg/addon/helper.go
+++ b/pkg/addon/helper.go
@@ -254,7 +254,7 @@ func FindAddonPackagesDetailFromRegistry(ctx context.Context, k8sClient client.C
 				if !ok {
 					continue
 				}
-				uiData, err := r.GetUIData(&sourceMeta, CLIMetaOptions)
+				uiData, err := r.GetUIData(&sourceMeta, UIMetaOptions)
 				if err != nil {
 					continue
 				}


Somefive:
Could you explain a bit more about why this change can fix the bug?

/cc @wangyikewxgm

zhaohuiweixiao:
> Could you explain a bit more about why this change can fix the bug?
> 
> /cc @wangyikewxgm

the APISchema is only rendered if the GetParameter is true in the GetUIDataFromReader func.

image

Somefive:
@zhaohuiweixiao Yes. But the function you are editing I think is used both in UI and CLI, is it right?

zhaohuiweixiao:
> > Could you explain a bit more about why this change can fix the bug?
> > /cc @wangyikewxgm
> 
> the APISchema is only rendered if the GetParameter is true in the GetUIDataFromReader func.
> 
> image



> @zhaohuiweixiao Yes. But the function you are editing I think is used both in UI and CLI, is it right?

@Somefive Now, the GetUIDataFromReader only used in cli mode(vela addon status command). But it has bug in the cli mode.
when the registry is not a version registry(such as a git registry), the vela addon status -v command can't display the parameter of the addon.
image
Infact, it should dispaly all the fluxcd parameters:
image

wangyikewxgm (approved):
Good catch！

PR #5739, opened 2023-03-24
Feat: application_controller should be responsible for deleting application revision

dhanusaputra:
### Description of your changes

Fixes #5710 

I have:

- [x] Read and followed KubeVela's [contribution process](https://github.com/kubevela/kubevela/blob/master/contribute/create-pull-request.md).
- [ ] [Related Docs](https://github.com/kubevela/kubevela.io) updated properly. In a new feature or configuration option, an update to the documentation is necessary. 
- [ ] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

PR #6074, opened 2023-06-02
Fix: application gc not blocked by apprev gc

### LINKED ISSUE #5710: [Feature] application_controller should be responsible for deleting application revision

When application is deleted, the application revision is deleted by Kubernetes following the OwnerReference mechanism. This could be slow when the system has high load. We should let application controller to handle it as well as a double check.

Move pkg/controller/core.oam.dev/v1alpha2/application/apply.go:garbageCollection to pkg/resourcekeeper/gc.go:GarbageCollect and support cleaning up application revision when application is fully removed.

**Additional context**

Somefive:

Issue #5710 mention that we should let the application controller to handle the garbage collection of application revision. Previously we rely on the ownerReference mechanism to recycle application revision for application. However, in the case that Kubernetes does not recycle those revisions immediately, it could lead to problems for recreated same-named applications.

PR #5739 refactors the code, moving the gc part into resourceKeeper but does not solve this problem. This PR is to work continue on that and let the application controller to delete application revisions before removing the finalizer of application.

Fixes #5710

I have:

### DIFF
--- a/pkg/resourcekeeper/gc_rev.go
+++ b/pkg/resourcekeeper/gc_rev.go
@@ -41,27 +41,19 @@ import (
 	"github.com/oam-dev/kubevela/pkg/utils"
 )
 
-// execute garbage collection functions, including:
-// - clean up legacy app revisions
+// GarbageCollectApplicationRevision execute garbage collection functions, including:
+// - clean up app revisions
 // - clean up legacy component revisions
 func (h *gcHandler) GarbageCollectApplicationRevision(ctx context.Context) error {
 	t := time.Now()
 	defer func() {
 		metrics.AppReconcileStageDurationHistogram.WithLabelValues("gc-app-rev").Observe(time.Since(t).Seconds())
 	}()
-	collectFuncs := []garbageCollectFunc{
-		garbageCollectFunc(cleanUpApplicationRevision),
-		garbageCollectFunc(cleanUpWorkflowComponentRevision),
+	if err := cleanUpComponentRevision(ctx, h); err != nil {
+		return err
 	}
-	for _, collectFunc := range collectFuncs {
-		if err := collectFunc(ctx, h); err != nil {
-			return err
-		}
-	}
-	return nil
+	return cleanUpApplicationRevision(ctx, h)
 }
-
-type garbageCollectFunc func(ctx context.Context, h *gcHandler) error
 
 // cleanUpApplicationRevision check all appRevisions of the application, remove them if the number of them exceed the limit
 func cleanUpApplicationRevision(ctx context.Context, h *gcHandler) error {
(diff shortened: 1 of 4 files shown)


test/e2e-multicluster-test/multicluster_test.go
```diff
@@ -1336,5 +1336,58 @@ var _ = Describe("Test multicluster scenario", func() {
...
+			By("Deleting")
+			_app := &v1beta1.Application{}
+			Expect(k8sClient.Get(ctx, appKey, _app)).Should(Succeed())
+			Expect(k8sClient.Delete(ctx, _app)).Should(Succeed())
+
+			By("Check application existing after rt recycled")
+			Eventually(func(g Gomega) {
+				rts := &v1beta1.ResourceTrackerList{}
+				g.Expect(k8sClient.List(ctx, rts, client.MatchingLabels{oam.LabelAppName: _app.Name, oam.LabelAppNamespace: _app.Namespace})).Should(Succeed())
+				g.Expect(len(rts.Items) == 0).Should(BeTrue())
```
chivalryq, line 1374:
```suggestion
				g.Expect(len(rts.Items)).Should(Equal(0))
```

test/e2e-multicluster-test/multicluster_test.go
```diff
@@ -1336,5 +1336,58 @@ var _ = Describe("Test multicluster scenario", func() {
...
+				g.Expect(k8sClient.Update(ctx, _rev)).Should(Succeed())
+			}).WithPolling(2 * time.Second).WithTimeout(10 * time.Second).Should(Succeed())
+
+			By("Deleting")
+			_app := &v1beta1.Application{}
+			Expect(k8sClient.Get(ctx, appKey, _app)).Should(Succeed())
+			Expect(k8sClient.Delete(ctx, _app)).Should(Succeed())
+
+			By("Check application existing after rt recycled")
+			Eventually(func(g Gomega) {
```
chivalryq, line 1371: I think here we want to check application recycling is blocked for its apprev is not deleted. So here I think the async assertion func should be `Consistently`. It means the assertion is true for a period of time. 
If `Eventually` is used, the first time the check passed will pass the whole assertion. Image that app is not deleted because it hasn't been reconciled which can pass the case.   The PR purpose -- application is deleted only after its apprev is deleted. -- is not actually tested.

pkg/resourcekeeper/gc_rev.go
```diff
@@ -41,28 +41,20 @@ import (
...
 	}()
-	collectFuncs := []garbageCollectFunc{
-		garbageCollectFunc(cleanUpApplicationRevision),
-		garbageCollectFunc(cleanUpWorkflowComponentRevision),
-	}
-	for _, collectFunc := range collectFuncs {
-		if err := collectFunc(ctx, h); err != nil {
-			return err
-		}
+	if err := cleanUpComponentRevision(ctx, h); err != nil {
```
chivalryq, line 52: Does the order of recycling component revision and application revision matter?

Somefive: No. The order does not matter.

test/e2e-multicluster-test/multicluster_test.go
Somefive: OK

test/e2e-multicluster-test/multicluster_test.go
Somefive: Simply replacing `Eventually` with `Consistently` could lead to first time failure. Splitted in to one `Eventually` and one `Consistently`.

FogDong (approved):

PR #6184, opened 2023-07-04
Feat: add configmap trait

nuclearwu:
### Summary
🆕🛠🗑

Improved the `configmap` trait to support more options and scenarios. Refactored the template logic to a separate cue file for better modularity.

> _`Configmap` trait is reborn from the ashes_
> _A new template with more power and features_
> _Patch and output the secrets of the pods_
> _Don't look back at the old and deprecated codes_

### Walkthrough
*  Update `configmap` trait definition to use new cue template ([link](https://github.com/kubevela/kubevela/pull/6184/files?diff=unified&w=0#diff-58af627805e7dd0bf440ed57903f12ef93792a1a1cdb4ba65dd71137aae1f714L2-R100), [link](https://github.com/kubevela/kubevela/pull/6184/files?diff=unified&w=0#diff-58af627805e7dd0bf440ed57903f12ef93792a1a1cdb4ba65dd71137aae1f714L54-R127))
* Add new cue template for `configmap` trait to `vela-templates/definitions/internal/configmap.cue` ([link](https://github.com/kubevela/kubevela/pull/6184/files?diff=unified&w=0#diff-70b633cd83779d0be13f90cfdbf5e3e3a12677b9584aa5e69a5f747990bc6ce1R1-R123))

Fixes #

I have:


wonderflow:
can you describe a bit more about why we should make it as a built-in trait?

nuclearwu:
> can you describe a bit more about why we should make it as a built-in trait?
@wonderflow When using webservice component, our users hope to configure confimap directly and create corresponding confimap resources simultaneously. Considering that adding configmap resource creation function to webservice will violate the principle of separation of concerns, So consider introducing configmap as a trait approach.

nuclearwu:
> can you describe a bit more about why we should make it as a built-in trait?

When using webservice component, our users hope to configure confimap directly and create corresponding confimap resources simultaneously. Considering that adding configmap resource creation function to webservice will violate the principle of separation of concerns, So consider introducing configmap as a trait approach. @wonderflow

nuclearwu:
> can you describe a bit more about why we should make it as a built-in trait?

When using webservice component, our users hope to configure confimap directly and create corresponding confimap resources simultaneously. Considering that adding configmap resource creation function to webservice will violate the principle of separation of concerns, So consider introducing configmap as a trait approach.

wonderflow:
@nuclearwu the `storage` trait can create configmap in your case.

nuclearwu:
> @nuclearwu the `storage` trait can create configmap in your case.



> `storage` trait can create configmap in your case.

@wonderflow Yes,I know, however, our users have reported that the storage trait is complicated to use and some parameter configurations are not used, so consider a more simplified confimap trait. So far, I've just extracted confimap from storage. Next, I'm going to remove some of the configuration to make the confimap trait more suitable for typical usage scenarios.

nuclearwu:
ping @Somefive

references/docgen/def-doc/trait/configmap.eg.md
```diff
@@ -0,0 +1,28 @@
...
+spec:
+  components:
+    - name: express-server
+      type: webservice
+      properties:
+        image: oamdev/hello-world
+        ports:
+          - port: 8000
+      traits:
+        - type: storage
```
wonderflow, line 16: the example is still using storage

vela-templates/definitions/internal/trait/configmap.cue
```diff
@@ -0,0 +1,123 @@
...
+			defaultMode: *420 | int
+			readOnly:    *false | bool
+			data?: {...}
+			items?: [...{
+				key:  string
+				path: string
+				mode: *511 | int
+			}]
+		}]
+	}
```
wonderflow, line 122: I don't see the parameter is more simple than storage. If you say storage is hard to use, we must show enough difference here for a new configmap trait.

wonderflow (changes requested):
-

wonderflow (changes requested):
I'm against the change. I suggest you provide the trait as your own platform capability or provide an addon instead of putting as built-in one.

FogDong (commented):
Why don't we use `storage` instead?

PR #6232, opened 2023-07-24
Revert "Fix: add deploy as built-in step to avoid parse in offline mode"

wonderflow:
Reverts kubevela/kubevela#6201

test e2e tests that fluxcd-addon fail to install


wonderflow:
this PR reverted while the issue can be solved in #6234

PR #6228, opened 2023-07-24
Feat: apply-component supports namespace

nuclearwu:
Signed-off-by: wuzhongjian wuzhongjian_yewu@cmss.chinamobile.com

### Summary
🆕📝🚀

Add support for specifying namespace in workflow steps. This change adds a `namespace` parameter to the `apply-component` workflow step definition and logic in various files, and updates the documentation accordingly.

> _`apply-component`_
> _now accepts `namespace`_
> _a spring cleaning cue_

### Walkthrough
*  Add `namespace` parameter to `apply-component` workflow step definition and handler ([link](https://github.com/kubevela/kubevela/pull/6228/files?diff=unified&w=0#diff-d04ff4062f6d641448d51084a8340da502252bad348799cd4f8d89f5fdbbdfa7R22-R23), [link](https://github.com/kubevela/kubevela/pull/6228/files?diff=unified&w=0#diff-4170e9b730a11fdd789c7b33a0871e4a67c4e37910815cc262286ac8ec076212R255), [link](https://github.com/kubevela/kubevela/pull/6228/files?diff=unified&w=0#diff-4170e9b730a11fdd789c7b33a0871e4a67c4e37910815cc262286ac8ec076212L291-R294), [link](https://github.com/kubevela/kubevela/pull/6228/files?diff=unified&w=0#diff-b294236ceeaba10f0beef13117ca02e3063a2c33a6f7c67ae0f2195e865fa9e2L8-R10), [link](https://github.com/kubevela/kubevela/pull/6228/files?diff=unified&w=0#diff-b294236ceeaba10f0beef13117ca02e3063a2c33a6f7c67ae0f2195e865fa9e2L21-R23), [link](https://github.com/kubevela/kubevela/pull/6228/files?diff=unified&w=0#diff-17fef7796b683f2995e4bd6650af842e7d1c0c116471d2f206ec405b5c192256R17-R18))
* Update example and documentation of using the `apply-component` workflow step ([link](https://github.com/kubevela/kubevela/pull/6228/files?diff=unified&w=0#diff-9e546a8bf33fb0f29a44924f3621c3016273cc56d76682fd38be99e39f1bca1aL27-R28))

Fixes https://github.com/kubevela/kubevela/issues/6223

I have:

- [x] Read and followed KubeVela's [contribution process](https://github.com/kubevela/kubevela/blob/master/contribute/create-pull-request.md).
[...]

PR #6247, opened 2023-08-07
Chore: revert "Feat: apply-component supports namespace"

chivalryq:
Reverts kubevela/kubevela#6228 due to the lack of test to `k8s-objects` type component. 

## details

In a application with `k8s-objects` component, a default workflow will be generated with `apply-component` step with no `namespace` specified. It means the objects in this componets will be dispatched to application's namespace.

Now it is the reason the CI is blocking.

### DIFF
--- a/charts/vela-core/templates/defwithtemplate/apply-component.yaml
+++ b/charts/vela-core/templates/defwithtemplate/apply-component.yaml
@@ -19,7 +19,5 @@ spec:
         	component: string
         	// +usage=Specify the cluster
         	cluster: *"" | string
-        	// +usage=Specify the namespace
-        	namespace: *"" | string
         }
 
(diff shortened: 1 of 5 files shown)


FogDong (approved):
So this feature breaks `op.#Apply`?

wonderflow (approved):

chivalryq:
> So this feature breaks `op.#Apply`?

If you only looks `op.#Apply`, it's OK if users are specifying `apply-componnet` manually and they can choose to specify `namespace` or leave it empty. 

But the key problem is `apply-component` is also generated by application with only components. In the generating step, `namespace` won't be set. With new `apply-component` logic, it means to override the namespaces of component with app's namespace. That, as a result, breaks several cases in e2e-test

PR #6355, opened 2023-09-19
Fix: validate cue template in webhook handler and add policyvaladatin…

caiqi1111:
…gwebhook

validate cue template in webhook handler and add policydefinition valadatingwebhook

Fixes #

I have:


pkg/webhook/core.oam.dev/v1beta1/componentdefinition/validating_handler_test.go
```diff
@@ -169,5 +174,52 @@ var _ = Describe("Test ComponentDefinition validating handler", func() {
...
+			Expect(resp.Allowed).Should(BeTrue())
+		})
+		It("Test cue template validation failed", func() {
+			cd.Spec = v1beta1.ComponentDefinitionSpec{
+				Schematic: &common.Schematic{
+					CUE: &common.CUE{
+						Template: inValidCueTemplate,
+					},
+				},
+			}
```
chivalryq, line 211: This component is denied because its `Workload` field isn't set. You should check the reason.
```go
Expect(string(resp.Result.Reason)).Should(ContainSubstring("hello: reference \"world\" not found"))
```

pkg/webhook/core.oam.dev/v1beta1/policydefinition/validator_test.go
chivalryq, line 1: The file name should be validating_handler_test.go

pkg/webhook/core.oam.dev/v1beta1/componentdefinition/validating_handler_test.go
caiqi1111: Nice idea! I add `workload` type field and check the error reason.

chivalryq:
Now there's some issues simply validate the original CUE file because the `context` block is added in vela-core. So you may also need to add `context` data pass these definitions.

Errror from CI:
```
Error: cannot create object: admission webhook "validating.core.oam-dev.v1beta1.componentdefinitions" denied the request: output.metadata.name: reference "context" not found (and 3 more errors)
make: *** [makefiles/e2e.mk:11: e2e-setup-core-post-hook] Error 1
```

caiqi1111:
I have ignore "context not found" error, but it seams not work.
`
// ValidateCueTemplate validate cueTemplate
func ValidateCueTemplate(cueTemplate string) error {

	val := cuecontext.New().CompileString(cueTemplate)
	re := regexp.MustCompile(ContextRegex)
	// ignore context not found error
	if val.Err() != nil && !re.MatchString(val.Err().Error()) {
		return val.Err()
	}

	err := val.Validate()
	if err != nil && !re.MatchString(err.Error()) {
		return err
	}
	return nil
}
`

pkg/webhook/utils/utils.go
```diff
@@ -53,3 +58,20 @@ func ValidateDefinitionRevision(ctx context.Context, cli client.Client, def runt
 	}
 	return nil
 }
+
+// ValidateCueTemplate validate cueTemplate
+func ValidateCueTemplate(cueTemplate string) error {
```
chivalryq, line 63: Unit test should be added to this function.

caiqi1111: ok, I added unit test and fix CI error.

chivalryq (approved):

Issue #6467, opened 2024-03-01
[Bug] Parameter 'log-file-path' does not take effect


msun1996:
**Describe the bug**
Parameter `log-file-path` does not take effect

**To Reproduce**
`manager --log-file-path=/export/logs/vela.log ...`  
You will find that the log has been output to the console, and the log has not been created and output to the file `/export/logs/vela.log`

**Expected behavior**
Log parameters such as `log-file-path`, `log-file-max-size`, `log-debug` can take effect normally

**Screenshots**

**KubeVela Version**
1.9.2

**Cluster information**
1.19

**Additional context**

**Troubleshooting reasons:**
```
func (s *CoreOptions) Flags() cliflag.NamedFlagSets {
        fss := cliflag.NamedFlagSets{}
	gfs := fss.FlagSet("generic")
	gfs.StringVar(&s.LogFilePath, "log-file-path", s.LogFilePath, "The file to write logs to.")
	gfs.Uint64Var(&s.LogFileMaxSize, "log-file-max-size", s.LogFileMaxSize, "Defines the maximum size a log file can grow to, Unit is megabytes.")
	gfs.BoolVar(&s.LogDebug, "log-debug", s.LogDebug, "Enable debug logs for development purpose")
	kfs := fss.FlagSet("klog")
	utillog.AddFlags(kfs)

	if s.LogDebug {
		_ = kfs.Set("v", strconv.Itoa(int(commonconfig.LogDebug)))
	}

	if s.LogFilePath != "" {
		_ = kfs.Set("logtostderr", "false")
		_ = kfs.Set("log_file", s.LogFilePath)
		_ = kfs.Set("log_file_max_size", strconv.FormatUint(s.LogFileMaxSize, 10))
	}
}
```
As implemented above, in the Flags configuration phase, `s.LogFilePath` must be the default empty string. Therefore, the log configuration will not take effect in `klog`.

**Solution:**
Log configuration and flush moved to `run` function
![image](https://github.com/kubevela/kubevela/assets/27460507/aa378cbd-c4b7-4d14-b807-eb27c9e37bf8)

EraKin575:
Hi! I would like to work on this issue

PR #6523, opened 2024-05-27
Fix: parameter 'log-file-path' does not take effect (#6467)

msun1996:

copilot:all

Fixes #

I have:

### DIFF
--- a/cmd/core/app/options/options.go
+++ b/cmd/core/app/options/options.go
@@ -17,14 +17,12 @@ limitations under the License.
 package options
 
 import (
-	"strconv"
 	"time"
 
 	pkgclient "github.com/kubevela/pkg/controller/client"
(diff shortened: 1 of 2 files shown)


Reetika2509:
Hi @msun1996 — thank you for this, and for correctly pinpointing the root cause (klog flags set in Flags() before parsing).

 Since you opened this, the team has dug further and found that reconfiguring the log file after klog initializes isn't reliable — klog initializes too early to be updated, so moving the flag.Set calls to run() doesn't fully resolve it. The agreed direction for #6467 is now in #6875, which (rather than silently trying to redirect) warns users on --log-file-path and points them to the supported klog flags(--logtostderr=false --log_file=<path>).

Given that, and that this PR conflicts with the refactored cmd/core/app/ layout, I'm going to close it in favor of #6875. Really appreciate the contribution and the accurate diagnosis — please keep them coming!

Issue #6435, opened 2024-01-04
[Feature] Support for explicit versioning of definitions


Kolossi:
## Description

This feature request is to add the facility for component/trait etc definitions to explicitly specify a version as part of the definition.

The Application will then have the ability to refer to the definition version when using a definition.

## Exisiting situation

Definitions currently have a version field but this is automatically assigned:

The [Version Control for Definitions](https://kubevela.io/docs/platform-engineers/x-def-version/) docs have:
> When an [OAM definition](https://kubevela.io/docs/getting-started/definition/) is updated, KubeVela will _**automatically**_ generate a new revision of this definition.
> [my emphasis]

This is a simple version number, being an integer.  Being auto-assigned causes problems in not being able to predict what version will need to be used in the Application file, and the version number may be different in different environments such as dev,staging,prod)

## Possible Solutions

My preference as the originator would be to use [semantic versioning](https://semver.org/) for the definition version.

This could either be by repurposing the existing version field, or by adding a new version field.

Referencing the defintion within the Application could be allowed as follows:

```yaml
    # latest (no version specified)
    type: wb-service
```

```yaml
    # specific
    type: wb-service@v3.0.1
```
```yaml
    # non-specific
    type: >wb-service@v1.2
```

## Alternatives

- a current workaround is to call the definition e.g. "wbserver-v1.0.1" then "wbserver-v1.0.2" etc. This would meet requirements but has knock-on effects e.g. bloat of `vela def list`
- During Kubevela Community call 4th-Jan-2024, @linaking-guidewire mentioned that the versioning could use the CRD alpha/beta/v1 format.

## Additional considerations

- is any form of old version cleanup wanted/possible? (could it automatically discover used and unused versions in the cluster)
- what actions if any are to be taken regarding checking whether a given version is available?
- can anything be done to easy deployment of a given defintion version to multiple clusters

## References

This proposal originated from a [slack thread](https://cloud-native.slack.com/archives/C01BLQ3HTJA/p1703583587366029) created by Frederick Boismenu.

It was also discussed on the [Kubevela Community call 4th-Jan-2024](https://www.youtube.com/watch?v=6YCtH5RHb3g).

Kolossi:
I'll admit I've not a great understanding of the CRD alpha/beta/v1 versioning - by which I mean how does the consumer identify breaking/non-breaking changes, patches etc?

[...]

PR #6574, opened 2024-07-10
Feat(design): Enhancement proposal first draft for Definition versioning

shreyasHpandya:
Refs #6435

Feature proposal

### DIFF
--- a/design/vela-core/definition-versioning.md
+++ b/design/vela-core/definition-versioning.md
@@ -1,0 +1,180 @@
+# Versioning Support for KubeVela Definitions
+
+<!-- toc -->
+- [Versioning Support for KubeVela Definitions](#versioning-support-for-kubevela-definitions)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+    - [How to handle existing components](#how-to-handle-existing-components)
+  - [Acceptance Criteria](#acceptance-criteria)
+  - [Design Details](#design-details)
+<!-- /toc -->
+
+## Summary
+
+Support explicit versioning for KubeVela Definitions and a way to specify which
+version of the Definition is to be used in an Application spec.
+
+
+## Motivation
+
+OAM/KubeVela Definitions (referred to as Definitions or Components for the rest
+of the document) are the basic building blocks of the KubeVela platform. They
+expose a contract similar to an API contract, which evolves from minor to major
+versions. Definitions currently have a version field, but it is automatically
+assigned and is just an integer.
+
+Such a versioning scheme does not provide any hint on the type of change
+(patch/bug, minor or major)
+
+Applications are composed of definitions that the KubeVela engine stitches
+together. KubeVela automatically upgrades/reconciles the application to the
+latest when no Definition version is specified. While we don't want ideally
+application developers to bother with such details, there are use cases where
+upgrading to the latest Definition version will need to be controlled.
+
+- It is an issue faced by many Kubevela community members; see the reference
+  below.
+
+- This can be mitigated by specifying a version number, but as the current
+  versioning scheme does not provide hints on the type of change, it can hardly
+  be automated.
+
+### Goals
+
+- Support Definitions versioning using SemVer
+- Allow pinning, specific and non-specific versions of a Definition in the
+  application
+
+### Non-Goals
+
+- A complicated versioning feature
+- Support for version range in Application. For eg.`type: my-component@>1.2.0`
+
+## Proposal
+
+Definitions to explicitly specify a version as part of the definition as
+described below:
+
+    apiVersion: core.oam.dev/v1beta1
+    kind: ComponentDefinition
+    metadata:
+    name: <ComponentDefinition name>
+    annotations:
+        definition.oam.dev/description: <Function description>
+    spec:
+        version: <semver>
+        workload: # Workload Capability Indicator
+            definition:
+                apiVersion: <Kubernetes Workload resource group>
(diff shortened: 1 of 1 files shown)


FogDong (approved):
LGTM! Any insights from @anoop2811 @Kolossi

Kolossi:
Firstly, great work @shreyasHpandya in putting this design doc together and getting some movement on the proposal 👏🏻 

Some comments only in the spirit of making the proposal the best it can be 🙂 ...

> KubeVela will initiate the reconciliation of the application as soon as a new version of a component is available

&

> This is also consistent with existing behaviour where we always use the latest version of the Definition when the application reconciles. 

Unless I'm wrong, it is not existing behaviour to initiate the reconcilliation of the application as soon as a new version of a component is available?  Sure it picks up the latest version "when the application reconciles" as per the second comment, but that's not the same thing. Either way these two statements are conflicting, or rather are not referring to the same thing.

I would not support changing this behaviour.  If someone wants auto-upgrade to latest version, then have them specify 'magic values `@*` or `@latest` (the latter my personal preference).

The problem with it auto-upgrading is that it is possible to have a deploy work but then be broken way down the line if a new component version is introduced, with this being pretty hard to track down.  I would not use auto-upgrade functionality in a live system.   Worse, with this proposed design it is easy to forget to add the version number and engage this auto-upgrade without intentionally signing up for that.  I genuinely think that spells reputational risk to kubevela with things breaking "on their own" at a much later date.  I get that others feel differently and this is my 2c.  They can always use "@latest" to intentionally get auto-upgrade.  Those who don't get what I believe to be existing functionality.

If I'm correct and we don't currently auto-upgrade applications when a new definition is available, then we are saying we would markedly change the functionality of kubevela if someone upgrades kubevela to include this proposal, and potentially all of their applications will be rereconciled and upgraded as well even if they have no knowledge of or wish to engage in the versioning? This doubles the reputational risk imho.

As such I disagree with the "Proposal Notes" in the section "How to handle existing components"

> Updates to Definitions should never force the application spec to change. If a new component is changing something significant it should be a Definition with a new name and not the new version of the existing Definition.

I believe changing "application spec" simply means that the change indicates a major version update (and not a minor or patch update).  A major version update in semver can be expected to be a breaking change.  Unless the definition has changed to no longer represent its name or previously described high-level functionality, it should not be necessary to make one of a new name e.g "mything-v2" indeed avoiding the need for this should be a goal of this feature.

>  For instance avoid changing anything except `version` in the spec when a component is enrolled on new versioning.

I'm not clear what this is saying, can it be clarified?

Would be very happy to hear thoughts on my comments and continue the conversation 🙂 

---

FYI @FogDong  @anoop2811

dhiguero:
Thanks @shreyasHpandya for consolidating the proposal into a doc. I agree with @Kolossi in that we should clarify the expected behaviour when matching versions depending on how the application specifies the component version, related to:

> When selecting a Definition version in the application spec, it is possible to select a non-specific version, like v1.0 or v1, the KubeVela will auto-upgrade the application to use the latest version in the specified version series. For instance, if the application specifies type: my-component-type@1.0 and v1.0.1 of the Definition is available, KubeVela will re-render the application using this version.

In particular, I think we should support scenarios where newer component versions are deployed in cluster for final testing, and we should provide a method to protect the existing applications from automatically being updated to use those since they may not ready (Just a variation of the scenario we introduce a new component version). If I understood correctly the pinned version proposal, the following cases would reflect the expected behavior with an application that makes use of the versioning.

| Application version| New component version | Behavior |
|---|---|---|
|my-component-type@1 | v1.0.y | Expect deployed application to be updated from using component v1 to v1.0.y |
|my-component-type@1 | v1.x.0 | Expect deployed application to be updated from using component v1 to v1.x.0 |
|my-component-type@1 | v1.x.y | Expect deployed application to be updated from using component  v1 to v1.x.y |
|my-component-type@1 | v2.0.0 | Expect deployed application to remain using v1 as this a major change |
|my-component-type@1.2.3 | v1.0.y | Expect deployed application to not change as version is pinned |
|my-component-type@1.2.3 | v1.x.0 | Expect deployed application to not change as version is pinned |
|my-component-type@1.2.3 | v1.x.y | Expect deployed application to not change as version is pinned |
|my-component-type@1.2.3 | v2.0.0 | Expect deployed application to not change as version is pinned |

Now, let's consider an update scenario, where we deploy the new kubevela version in an existing cluster with applications that use component `my-component-type` with no specific version since that was not supported before. If I understood correctly, the proposed expected behavior is that KubeVela will treat the existing `my-component-type` as a legacy component specification, and if an application does not sets any component version as it will be the case with existing definition, the same component will be used, and no change in the deployed applications should happen. Is this correct?

| Application version| New component version | Behavior |
|---|---|---|
|my-component-type | v1.0.y | Expect deployed application to not change as version is pinned to the legacy component |
|my-component-type | v1.x.0 | Expect deployed application to not change as version is pinned to the legacy component |
|my-component-type | v1.x.y | Expect deployed application to not change as version is pinned to the legacy component |
|my-component-type | v2.0.0 | Expect deployed application to not change as version is pinned to the legacy component |

Hopefully the tables help define the different scenarios. Are these cases matching the proposal?

anoop2811:
I think we can clarify explicit behaviors as mentioned by Daniel and Paul. Since Shreyas is also part of Guidewire (as me), our focus was based on how we were thinking. I think we should advertise this proposal more in the kubevela channel for more inputs from the community.

anoop2811:
@shreyasHpandya : Is this PR still required? Should we close this in lieu of the changes that have already been merged?

shreyasHpandya:
I will close this @anoop2811; this is superseded by other merged PRs now

PR #6645, opened 2024-10-28
Feat: Added configurable env specific arbitrary render context

hprotzek:
This is just a POC, to get feedback if this is a feasible idea to continue. 

We have a Kubevela setup with a lot of environments for different tenants. Every tenant has own settings, that we would like to use in our own custom components and traits when we render them. 

The idea is to have a configmap per vela env, that is added to the cue context and then can be used to render components and traits.

Any opinion if we can continue with that idea and make a proper PR our of it?

### DIFF
--- a/pkg/controller/core.oam.dev/v1beta1/application/generator.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/generator.go
@@ -18,6 +18,7 @@ package application
 import (
 	"context"
 	"encoding/json"
+	v1 "k8s.io/api/core/v1"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


vaibhav0096:
Hey @hprotzek 
KubeVela already has a built-in read-object WorkflowStepDefinition that covers this use case without a controller change.

keep your per-env ConfigMap, read it with a read-object workflow step, expose the values as step outputs, and feed them into your component via the apply-component step's inputs (mapped with parameterKey).
 So the env-specific config flows into rendering through the normal workflow data mechanism

Each environment just gets its own ConfigMap (same name, different values), and the same Application renders environment-specific output. Full example here:
 https://kubevela.io/docs/end-user/workflow/built-in-workflow-defs/#read-object

Issue #6534, opened 2024-06-13
The Webhook validating.core.oam.dev.v1alpha2.traitdefinitions is cluster scoped which is incorrect


shreyasHpandya:
**Describe the bug**

The webhook scope at https://github.com/kubevela/kubevela/blob/master/charts/vela-core/templates/admission-webhooks/validatingWebhookConfiguration.yaml#L38 is incorrect. The Traitdefinition resource is a namespace scoped resource so this webhook is not working as intended for the Traitdefinition.

**KubeVela Version**
v1.9.11

PR #6600, opened 2024-08-15
KEP to support Semantic versioning of Components 

kanchan-dhamane:
### Description of your changes

copilot:all

## Summary

Support explicit versioning for KubeVela Components and a way to specify which
version of the Component is to be used in an Application spec.

## Motivation

OAM/KubeVela Definitions (referred to as Definitions or Components for the rest
of the document) are the basic building blocks of the KubeVela platform. They
expose a contract similar to an API contract, which evolves from minor to major
versions. Definitions currently do have a `Revision` field in the Status, but it is an auto-incrementing integer.

Such a versioning scheme does not denote the type of the change (patch/bug, minor or major).

Applications are composed of Components that the KubeVela engine stitches
together. KubeVela automatically upgrades/reconciles the Application to the
latest when no Component Revision is specified. While we don't ideally want
Application developers to bother with such details, there are use cases where an automatic upgrade to the latest Component version is not desired.

- This can be resolved by specifying a Revision number when referring to the Component in an Application, but as the current versioning scheme does not provide hints on the type of change, it cannot be automated well.

Fixes https://github.com/kubevela/kubevela/issues/6435

I have:

- [ ] Read and followed KubeVela's [contribution process](https://github.com/kubevela/kubevela/blob/master/contribute/create-pull-request.md).
- [ ] [Related Docs](https://github.com/kubevela/kubevela.io) updated properly. In a new feature or configuration option, an update to the documentation is necessary. 
- [ ] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

PR #6648, opened 2024-11-04
Feat: Semantic versioning support for Definitions

kanchan-dhamane:

Fixes https://github.com/kubevela/kubevela/issues/6435
Fixes https://github.com/kubevela/kubevela/issues/6534

Implements KEP - https://github.com/kubevela/kubevela/pull/6648/files#diff-c6da3da65cafd2d214c7ee0801d9db6ffa1105cb277e45f63bc3ba985209a308 
 
Documentation Update: https://github.com/kubevela/kubevela.github.io/pull/1343

Changes:
- Adds an optional "Version" field for all Definition Specs.
- Adds the following new validations to Webhooks for Definitions:
        - Validate the "Version" field follows Semantic versioning.
        - Dis-allow conflicting versioning fields ( Name annotation, Spec.Version)
- Adds the following new validations to Webhooks for Application:
        - Dis-allow the use of both the "publishVersion" & "autoUpdate" annotations.
- Enahnce "multiStageComponentApply" feature to support auto updates.

Boy Scout Changes:
- Fixes Plugin e2e tests broken by the fix for https://github.com/kubevela/kubevela/issues/6534.
- Fixes the dryRun and livediff commands to respect the "-n" namespace flag.
- Fixes the Application ValidationWebhook to respect the "-n" namespace flag.

I have:


apis/core.oam.dev/v1beta1/componentdefinition_types.go
```diff
@@ -27,6 +27,9 @@ import (
 
 // ComponentDefinitionSpec defines the desired state of ComponentDefinition
 type ComponentDefinitionSpec struct {
+	// +optional
+	Version string `json:"version"`
```
FogDong, line 31:
```suggestion
	Version string `json:"version,omitempty"`
```

apis/core.oam.dev/v1beta1/core_types.go
```diff
@@ -164,6 +164,9 @@ type TraitDefinitionSpec struct {
 	// pre-process and post-process respectively.
 	// +optional
 	Stage StageType `json:"stage,omitempty"`
+
+	// +optional
+	Version string `json:"version"`
```
FogDong, line 169:
```suggestion
	Version string `json:"version,omitempty"`
```

pkg/oam/util/helper.go
```diff
@@ -235,7 +250,24 @@ func GetCapabilityDefinition(ctx context.Context, cli client.Reader, definition
...
+		definitionType = common.PolicyType
+	case *v1beta1.WorkflowStepDefinition:
+		definitionType = common.WorkflowStepType
+	default:
+		return definitionType, fmt.Errorf("invalid definition")
+	}
+	return definitionType, nil
+}
+
+func fetchDefinitionRev(ctx context.Context, cli client.Reader, definitionName string, definitionType common.DefinitionType, fctx map[string]string) (bool, *v1beta1.DefinitionRevision, error) {
```
FogDong, line 270: I don't think we need a map here, we can just pass the parameter `autoUpdate bool`.

pkg/controller/core.oam.dev/v1beta1/application/dispatcher.go
```diff
@@ -210,11 +210,11 @@ func (h *AppHandler) generateDispatcher(appRev *v1beta1.ApplicationRevision, rea
 	return manifestDispatchers, nil
 }
 
-func getTraitDispatchStage(client client.Client, traitType string, appRev *v1beta1.ApplicationRevision) (StageType, error) {
+func getTraitDispatchStage(client client.Client, traitType string, appRev *v1beta1.ApplicationRevision, fctx map[string]string) (StageType, error) {
 	trait, ok := appRev.Spec.TraitDefinitions[traitType]
 	if !ok {
 		trait = &v1beta1.TraitDefinition{}
-		err := oamutil.GetCapabilityDefinition(context.Background(), client, trait, traitType)
+		err := oamutil.GetCapabilityDefinition(context.Background(), client, trait, traitType, fctx)
```
FogDong, line 217: You can get app annotation from `appRev.metadata.annotations[autoupdate]`

pkg/appfile/parser.go
```diff
@@ -253,7 +254,8 @@ func (p *Parser) parseWorkflowStepsForLegacyRevision(ctx context.Context, af *Ap
 			continue
 		}
 		def := &v1beta1.WorkflowStepDefinition{}
-		if err := util.GetCapabilityDefinition(ctx, p.client, def, workflowStep.Type); err != nil {
+
+		if err := util.GetCapabilityDefinition(ctx, p.client, def, workflowStep.Type, fctx); err != nil {
```
FogDong, line 258: You can get app annotation from `af.AppAnnotations[autoupdate]`

FogDong (commented):
We do not need to pass `fctx` from beginning to end, we can always get app annotation from appfile or application revision. Also, please use `bool autoUpdate` instead of the whole annotations, thanks!

kanchan-dhamane:
> We do not need to pass `fctx` from beginning to end, we can always get app annotation from appfile or application revision. Also, please use `bool autoUpdate` instead of the whole annotations, thanks!

Thanks @FogDong for reviewing PR. 

We have refactored the code to reduce the noise created by fctx map. Please take a look. 

Since the `appFile` is not available in all the functions in the Parser, we would need to pass the `autoUpdate` variable to each as a separate parameter. Initially we too thought of passing it as a boolean, but considering the number of layers and functions involved, settled on making it more generic so that we can reuse it to pass additional data in future. Hope that makes sense.

pkg/controller/core.oam.dev/v1beta1/application/dispatcher.go
FogDong: We don't need to pass fctx from all the way. For example, in this function, you can just pass `appRev.metadata.annotations`

pkg/appfile/parser.go
```diff
@@ -253,7 +257,8 @@ func (p *Parser) parseWorkflowStepsForLegacyRevision(ctx context.Context, af *Ap
 			continue
 		}
 		def := &v1beta1.WorkflowStepDefinition{}
-		if err := util.GetCapabilityDefinition(ctx, p.client, def, workflowStep.Type); err != nil {
+
+		if err := util.GetCapabilityDefinition(ctx, p.client, def, workflowStep.Type, p.FunctionalCtx); err != nil {
```
FogDong, line 261: pass `af.AppAnnotation`

pkg/appfile/parser.go
```diff
@@ -471,15 +476,15 @@ func (p *Parser) fetchAndSetWorkflowStepDefinition(ctx context.Context, af *Appf
 		return nil
 	}
 	def := &v1beta1.WorkflowStepDefinition{}
-	if err := util.GetCapabilityDefinition(ctx, p.client, def, workflowStepType); err != nil {
+	if err := util.GetCapabilityDefinition(ctx, p.client, def, workflowStepType, p.FunctionalCtx); err != nil {
```
FogDong, line 479: pass af.AppAnnotation

FogDong:
Have you checked my comments? For a lot of functions, we don't need to pass anything.
For example, in `parser.go`,
```
	if err := util.GetCapabilityDefinition(ctx, p.client, def, workflowStepType, p.FunctionalCtx); err != nil {
```
Instead of this p.FunctionalCtx, just pass af.AppAnnotation.

FogDong:
Yeah I think we just need the annotation. The app context here seems a little bit misleading.

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	cuelang.org/go v0.9.2
 	github.com/AlecAivazis/survey/v2 v2.1.1
 	github.com/FogDong/uitable v0.0.5
+	github.com/Masterminds/semver v1.5.0
```
merusso, line 9: Shouldn't we use `Masterminds/semver/v3` instead, which is already included?

```suggestion
	github.com/Masterminds/semver v1.5.0
```

merusso: We may want to upgrade this library to the latest version, currently v3.3.1

See: https://github.com/Masterminds/semver/releases

pkg/controller/core.oam.dev/v1beta1/core/revison.go
```diff
@@ -22,6 +22,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/Masterminds/semver"
```
merusso, line 25:
```suggestion
	"github.com/Masterminds/semver/v3"
```

pkg/oam/util/helper.go
```diff
@@ -39,6 +40,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/Masterminds/semver"
+
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/common"
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/condition"
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
```
merusso, line 47:
```suggestion
	"github.com/oam-dev/kubevela/apis/core.oam.dev/common"
	"github.com/oam-dev/kubevela/apis/core.oam.dev/condition"
	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
	"github.com/Masterminds/semver/v3"
```

pkg/oam/util/helper.go
```diff
@@ -39,6 +40,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/Masterminds/semver"
+
```
merusso, line 44:
```suggestion
	"github.com/Masterminds/semver/v3"
```

merusso:
Could you include a YAML example for an Application using this new versioning scheme? Maybe add something to [docs/examples](https://github.com/kubevela/kubevela/tree/master/docs/examples)?

I'm trying to understand how this change impacts consumers.

apis/core.oam.dev/v1beta1/componentdefinition_types.go
```diff
@@ -27,6 +27,9 @@ import (
 
 // ComponentDefinitionSpec defines the desired state of ComponentDefinition
 type ComponentDefinitionSpec struct {
+	// +optional
+	Version string `json:"version,omitempty"`
```
merusso, line 31: Does this mean that a cluster can only have one version of a component at a time? What happens when consumers refer to a version that's different from the ComponentDefinition version?

kanchan-dhamane: We are implementing 1.1 proposal. Please refer to the Acceptance Criteria mentioned in [this KEP](https://github.com/kubevela/kubevela/pull/6600/files?short_path=c6da3da#diff-c6da3da65cafd2d214c7ee0801d9db6ffa1105cb277e45f63bc3ba985209a308). 
Will update the PR to add the KEP and docs as part of the same PR.

kanchan-dhamane: PR description updated.

design/vela-core/definition-versioning.md
```diff
@@ -0,0 +1,321 @@
...
+But, this versioning scheme has the following issues: 
+
+- The `DefinitionRevision` does not denote the type of the change (patch/bug, minor or major). This hinders automation of automatic upgrades.
+- The current scheme also doesn't allow much control over automatic upgrades to new Component Revisions. KubeVela automatically upgrades/reconciles the Application to the
+latest when no Component Revision is specified. 
+  > While we don't ideally want Application developers to bother with such details, there are use cases where
+  > an automatic upgrade to the latest Component version is not desired.
+
+
+### Goals
```
merusso, line 48: Does this design allow a definition author to support 2 separate major versions of a definition simultaneously? In other words, are the following use cases supported?

>**AS A** Component author\
>**I SHOULD** be able to publish a new major version of my Component (v1.x -> v2.x)\
>**SO THAT** I can introduce breaking changes to my definition (in v2.x) without breaking existing consumers (using v1.x).

>**AS A** Component author\
>**I SHOULD** be able to publish changes for 2 different major versions of my Component simultaneously (v1.x, v2.x)\
>**SO THAT** I can continue updating the old version (v1.x) until consumers have migrated to the new one (v2.x).

If yes, how?

Let's say a definition has revision v1.0.0 and v2.0.0, and the author wants to publish v1.0.1 and v2.0.1.

How would this be done? Would the author sequentially apply these changes to the same ComponentDefinition, once for v1.0.1 and once for v2.0.1? Would the author directly apply a DefinitionRevision each for v1.0.1 and v2.0.1?

kanchan-dhamane: Both scenarios are supported. Please refer the [documentation PR](https://github.com/kubevela/kubevela.github.io/pull/1343) for more information.

go.mod
kanchan-dhamane: We'll create a separate issue to address this.

pkg/oam/util/helper.go
semmet95: Thanks for the suggestions @merusso 
We will run some tests to make sure the dependency update doesn't break anything and then push the changes

FogDong (commented):
Why all the tests are skipped in github action?

kanchan-dhamane:
> Why all the tests are skipped in github action?

Hi @FogDong, since all test cases passed in the previous run and there were no changes to the code, the tests were skipped in this run.

FogDong (approved):
Thanks for supporting this!

PR #6733, opened 2025-03-20
Fix: support task component to be one-time run to completion

### LINKED ISSUE #6732: KubeVela task component should support a one-time run to completion functionality
**Describe the bug**

In KubeVela, we have a task component that creates a Kubernetes Job internally. However, when we modify properties of the task, the KubeVela controller attempts to update the existing job, which fails since updating a Job is not allowed in Kubernetes. The purpose of the Task component is to trigger a one-time run to completion. Therefore, whenever an application containing this component is deployed, we always want a new job to be triggered, even if no changes have been made to the spec.
**To Reproduce**

1. Apply the below application containing task component
```
kind: Application
metadata:
  name: app-worker
spec:
  components:
    - name: mytask
      type: task
      properties:
        image: ubuntu:22.04
        count: 3
        cmd: [ "sleep", "30" ]
```

2. Modify the above Application by changing count from 3 to 2 in properties. This does not create a new Job and KubeVela controller throws below errors in the log:
`E0320 05:31:44.026520       1 task.go:252] "do steps" err="run step(provider=oam,do=component-apply): Dispatch: pre-dispatch dryrun failed: Found 1 errors. [(cannot patch object: Job.batch \"mytask\" is invalid: spec.completions: Invalid value: 2: field is immutable)]" application="default/app-worker" controller="application" resource_version="184063" generation=2 publish_version="" step_name="mytask" step_type="builtin-apply-component" spanID="i-p4uqyb4z.execute application workflow.ixwymmovkh"`
**Expected behavior**

We expect new Job to be created and no errors in KubeVela controller.
**Screenshots**

**KubeVela Version**

CLI Version: 1.10.1
Core Version: 1.10.2
GitRevision: git-711c9f00
GolangVersion: go1.22.11

**Cluster information**
Client Version: v1.31.2
Kustomize Version: v5.4.2
Server Version: v1.30.4+k3s1

**Additional context**

shivin:
…nt name

In KubeVela, we have a task component that creates a Kubernetes Job internally. However, when we modify properties of the task, the KubeVela controller attempts to update the existing job, which fails since updating a Job is not allowed in Kubernetes. The purpose of the Task component is to trigger a one-time run to completion. Therefore, whenever an application containing this component is deployed, we always want a new job to be triggered, even if no changes have been made to the spec.

To address this, I've implemented a solution to ensure that each time the application is applied, a unique job name is generated, and the task component creates a new Job instead of modifying the existing one. Here's a summary of the changes:

Custom Component Definition:
Modify the ComponentDefinition of the task to generate the metadata.name for the Job resource using a unique combination of the application name and component name. This guarantees that the Job name will be unique for every application deployment, based on the application and component names.

Application Definition:
Use the metadata.generateName property in the application YAML instead of metadata.name to ensure that a unique name is generated automatically for each deployment.

Fixes #6732

I have:

Followed https://kubevela.net/docs/v1.4/contributor/code-contribute and installed KubeVela. Applied below application multiple times and verified it creates new Job every time without any errors:

```
kind: Application
metadata:
  generateName: app-worker
spec:
  components:
    - name: mytask
      type: task
      properties:
        image: ubuntu:22.04
        count: 2
        cmd: [ "sleep", "30" ]
```

### DIFF
--- a/charts/vela-core/templates/defwithtemplate/task.yaml
+++ b/charts/vela-core/templates/defwithtemplate/task.yaml
@@ -14,6 +14,7 @@ spec:
         output: {
         	apiVersion: "batch/v1"
         	kind:       "Job"
+        	metadata: name: "\(context.appName)-\(context.name)"
         	spec: {
         		parallelism: parameter.count
         		completions: parameter.count
--- a/vela-templates/definitions/internal/component/task.cue
+++ b/vela-templates/definitions/internal/component/task.cue
@@ -44,6 +44,9 @@ template: {
 	output: {
 		apiVersion: "batch/v1"
 		kind:       "Job"
+		metadata: {
+			name: "\(context.appName)-\(context.name)"
+		}
 		spec: {
 			parallelism: parameter.count
 			completions: parameter.count


anoop2811 (approved):
LGTM

PR #6818, opened 2025-06-25
Fix(cli): return non-zero exit code on `vela def render` error

### LINKED ISSUE #6651: Incorrect `vela def render` exit code on render error
**Describe the bug**
When running `vela def render` command with a faulty `.cue` file input, it fails to render, prints the error to stderr, but exit code remains 0.

Not sure if intentional, `vela def vet` does print the same error, but with exit code 1

**To Reproduce**
```sh
$ echo "1234" > /tmp/test.cue

$ vela def render /tmp/test.cue
failed to render test.cue, reason: failed to parse CUE: no metadata found, invalid

$ echo "$?"
0
```

**Expected behavior**
Non-zero exit code

**KubeVela Version**
1.9.13 and 1.10.0-alpha.1

**Cluster information**
N/A

**Additional context**
https://github.com/kubevela/kubevela/blob/0f780dec75fadf4c081df44051234011b4d136c2/references/cli/def.go#L840

7h3-3mp7y-m4n:

copilot:all

Fixes #6651 

This PR ensures that when rendering a .cue file fails due to invalid CUE syntax or missing metadata, vela def render returns a non-zero exit code.
I have:

Manual testing done:
The fix was verified manually by running:
``` bash
echo "1234" > /tmp/test.cue
./vela def render /tmp/test.cue
echo $?
```

### DIFF
--- a/references/cli/def.go
+++ b/references/cli/def.go
@@ -837,12 +837,16 @@ func NewDefinitionRenderCommand(c common.Args) *cobra.Command {
 					return errors.Wrapf(err, "failed to read directory %s", args[0])
 				}
 			}
+			var renderErrs []error
 			for i, inputFilename := range inputFilenames {
-				if err = render(inputFilename, outputFilenames[i]); err != nil {
-					if _, err = fmt.Fprintf(cmd.ErrOrStderr(), "failed to render %s, reason: %v", inputFilename, err); err != nil {
-						return errors.Wrapf(err, "failed to write err")
-					}
+				err := render(inputFilename, outputFilenames[i])
+				if err != nil {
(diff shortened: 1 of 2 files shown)


Chaitanyareddy0702:
Hi @7h3-3mp7y-m4n ,
It appears that the unit tests are failing because [`cmd.Execute()`]( https://github.com/kubevela/kubevela/pull/6818/files#diff-21d4c7864884250de747fe68c3ca0c1a3fcacf64bb613ea06f242075d5909fdcR571) is expecting a error, but It appears that the [condition]( https://github.com/kubevela/kubevela/pull/6818/files#diff-739f4f64cd699090ae6f9705af491dc7e42decf8d0340ef88109bc2333f32bdeR843) is never met.  As a result, the errors are not being appended into `renderErrs` array which is causing the function return the error as nil.  Could you please fix the logic? Thanks

PR #6828, opened 2025-07-09
Feat: Enhanced Status Reporting

briankane:

copilot:all

Fixes #

I have:

  - https://github.com/kubevela/kubevela.github.io/pull/1367
- [x] Added `backport release-x.y` labels to auto-backport this PR if necessary. (N/A)

- Unit Tests
- e2e Tests
- Manual Verifications


pkg/cue/definition/health/health.go
```diff
@@ -0,0 +1,182 @@
...
+	val := cuecontext.New().CompileString(buff)
+	healthy, err := val.LookupPath(value.FieldPath(IsHealthPolicy)).Bool()
+	if err != nil {
+		return false, errors.WithMessage(err, "evaluate health status")
+	}
+	return healthy, nil
+}
+
+func GetStatus(templateContext map[string]interface{}, request *StatusRequest) (*StatusResult, error) {
+	message, _ := getStatusMessage(templateContext, request.Custom, request.Parameter)
```
linaking: Errors should be returned or at least logged

pkg/controller/core.oam.dev/v1beta1/application/apply.go
```diff
@@ -248,10 +248,14 @@ func (h *AppHandler) collectTraitHealthStatus(comp *appfile.Component, tr *appfi
...
-	traitStatus.Message, err = tr.EvalStatus(templateContext)
 	if err != nil {
 		return common.ApplicationTraitStatus{}, nil, errors.WithMessagef(err, "app=%s, comp=%s, trait=%s, evaluate status message error", appName, comp.Name, tr.Name)
 	}
+	statusResult, err := tr.EvalStatus(templateContext)
+	if err == nil {
+		traitStatus.Message = statusResult.Message
+		traitStatus.Status = statusResult.Status
+	}
 	return traitStatus, extractOutputs(templateContext), nil
```
linaking:
```suggestion
	return traitStatus, extractOutputs(templateContext), err
```

pkg/cue/definition/health/health.go
briankane: Thanks, good spot. These errors aren't really errors so much as the function is returning errors when some of the cue fields simply aren't set. Logging for now and will review the original functions a bit more in the follow ups to prevent errors when !val.Exists()

apis/core.oam.dev/common/types.go
```diff
@@ -135,6 +135,9 @@ type Status struct {
 	// HealthPolicy defines the health check policy for the abstraction
 	// +optional
 	HealthPolicy string `json:"healthPolicy,omitempty"`
+	// Status defines a detailed status map that can be displayed to the user
+	// +optional
+	Status string `json:"status,omitempty"`
```
eolatham: Couple of questions here:
1. The description of `status` says it's a map, but the type is actually string. Is the type incorrect?
2. I think having two fields named `status` and `customStatus` in the same object will be confusing to users. If `status` is meant to be a map of generic properties, should we name it something like `properties`, or `customProperties` (for parity with `customStatus`) instead?

apis/core.oam.dev/common/types.go
```diff
@@ -172,6 +175,7 @@ type ApplicationComponentStatus struct {
 	// WorkloadDefinition is the definition of a WorkloadDefinition, such as deployments/apps.v1
 	WorkloadDefinition WorkloadGVK              `json:"workloadDefinition,omitempty"`
 	Healthy            bool                     `json:"healthy"`
+	Status             map[string]string        `json:"status,omitempty"`
```
eolatham: Similar comment here regarding the field naming. The name `status` might be easily confused with `healthy`, and it doesn't indicate that the type is a map. I suggest naming it `properties` or `customProperties`. This would apply to the field in `ApplicationTraitStatus` as well.

apis/core.oam.dev/common/types.go
briankane: @eolatham Yeah can see how this is confusing. In cue it is a map, but gets converted to a string representation of the cue to be evaluated later. Will update shortly for clarity. 

Agreed on the naming also. I think `customStatus` was a bad choice but we've to live with it. 

What about `status.details`?

eolatham: Ah, I see. I'd be okay with `status.details` too👍🏻

pkg/cue/definition/health/health.go
```diff
@@ -0,0 +1,189 @@
...
+	iter, err := contextVal.Fields(cue.All())
+	if err != nil {
+		return nil, errors.WithMessage(err, "get context fields")
+	}
+	for iter.Next() {
+		contextLabels = append(contextLabels, iter.Label())
+	}
+
+	cueBuffer := runtimeContextBuff + "\n" + statusFields
+	val := cueCtx.CompileString(cueBuffer)
```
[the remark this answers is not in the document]
FogDong: Can you fix this?

pkg/controller/core.oam.dev/v1beta1/application/application_controller_test.go
```diff
@@ -3886,7 +3886,7 @@ var _ = Describe("Test Application Controller", func() {
 		curApp := &v1beta1.Application{}
 		Expect(k8sClient.Get(ctx, appKey, curApp)).Should(BeNil())
 		// should not panic anymore, refer to https://github.com/cue-lang/cue/issues/1828
-		// Expect(curApp.Status.Phase).Should(Equal(common.ApplicationWorkflowFailed))
+		// Expect(curApp.Details.Phase).Should(Equal(common.ApplicationWorkflowFailed))
```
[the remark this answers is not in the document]
FogDong: Maybe we can just delete this test case

test/e2e-test/testdata/definition/deployment-with-status.yaml
```diff
@@ -0,0 +1,53 @@
...
+                }]
+              }
+            }
+          }
+        }
+  status:
+    customStatus: |
+      message: "\(context.output.status.readyReplicas) / \(context.output.status.replicas) replicas are ready"
+    healthPolicy: |
+      isHealth: context.output.status.replicas == context.output.status.readyReplicas
```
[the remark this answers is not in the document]
briankane: Not needed as it's intended for a test

PR #6854, opened 2025-08-14
Feat: Allow gating with components `dependsOn` field

### LINKED ISSUE #6852: KubeVela dependsOn ignores failed component status and deploys dependent components anyway
KubeVela dependsOn ignores failed component status and deploys dependent components anyway

**To Reproduce**
 
Steps to reproduce the behaviour:
1. The YAML file of Application I applied.
```
kind: Application
metadata:
  name: dependencytest
  namespace: default
spec:
  components:
    - name: express-server
      type: webservice
      properties: {} # Intentionally not setting mandatory image to make this fail
    - name: worker-service
      type: webservice
      dependsOn:
        - express-server
      properties:
        image: nginx:1.20

```
2. Wait 2-3 minutes for retries to complete and the express-server component status to become failed
3. Check the status of the application  `vela status dependencytest`
Workflow:
```
 mode: DAG-DAG
  finished: true
  Suspend: false
  Terminated: true
  Steps
  - id: zbbxxk79qg
    name: express-server
    type: apply-component
    phase: failed
    message: function call error for : GenerateComponentManifest: evaluate base template component=express-server app=dependencytest: failed to have the workload/trait unstructured: output.spec.template.spec.containers.0.image: incomplete value string (value: cue/format: unsupported node type <nil>)
  - id: tv4848wbnc
    name: worker-service
    type: apply-component
    phase: succeeded  # ❌ This should NOT succeed while dependency is failed
```
 
**Expected behaviour** Components with dependsOn should wait until their dependencies are successfully running. 
 
**Actual**: Components with dependsOn deploy regardless of dependency failure status
**Impact**: Breaks dependency orchestration and can lead to runtime failures in dependent services.

**KubeVela Version** 1.10.x

Chaitanyareddy0702:

This PR introduces a feature that **automatically generates a dependency-aware workflow** for applications using `dependsOn`, simplifying the user experience and making dependency management more intuitive.  Previously, to enforce that dependent components would not deploy if a dependency failed, users were required to explicitly define workflow steps. For applications with simple `dependsOn` relationships, this added boilerplate configuration and was not always intuitive. This behavior was raised in issue #6852.

**Enhancement:** This PR enhances the workflow generation logic. Now, if an application contains components with `dependsOn` relationships but **lacks an explicit workflow**, KubeVela will dynamically generate a Directed Acyclic Graph (DAG) workflow that respects these dependencies. This ensures that the intended dependency orchestration works out-of-the-box, preventing dependent components from deploying when their dependencies fail.

Closes #6852

I have:
- [x] Read and followed KubeVela's [[contribution process](https://github.com/kubevela/kubevela/blob/master/contribute/create-pull-request.md)]
- [ ] [[Related Docs](https://github.com/kubevela/kubevela.io)](https://github.com/kubevela/kubevela.io) updated properly. As a new feature, documentation has been updated to reflect this new default behavior.

Tested and verified the changes locally by deploying an application with `dependsOn` relationships (without an explicit workflow) and confirming that:
1. A dependency-aware workflow is automatically created.
2. Dependent components are not deployed when a parent component fails.
3. Added the relavent test cases

### DIFF
--- a/test/e2e-test/testdata/app/app_component_depends_on_fail.yaml
+++ b/test/e2e-test/testdata/app/app_component_depends_on_fail.yaml
@@ -7,7 +7,7 @@ spec:
     - name: failing-database
       type: webservice
       properties:
-        image: "nginx:invalid-tag"  # This will fail to pull
+        image: ""  # This will fail to pull
         ports:
           - port: 3306
             expose: false


oanasc:
Thanks for the fix. Would be really helpful if we had this in the next release

FogDong (commented):
Can we add a test for this case?

PR #6877, opened 2025-09-01
Chore: Upgrade cuelang version to v0.14.0

Chaitanyareddy0702:

copilot:all

Fixes #

I have:


references/docgen/parser_test.go
```diff
@@ -749,7 +749,7 @@ func TestExtractParameterFromFiles(t *testing.T) {
  containerName | Specify the name of the target container, if not set, use the component name. | string | false | empty 
  replace | Specify if replacing the whole environment settings for the container. | bool | false | false 
  env | Specify the  environment variables to merge, if key already existing, override its value. | map[string]string | true |  
- unset | Specify which existing environment variables to unset. | []string | true |  
+ unset | Specify which existing environment variables to unset. | list | true |
```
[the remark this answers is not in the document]
roguepikachu: "list" accurately represents what the field accepts - it could be an empty list or a list of strings. If we said "[]string", that would be misleading because an empty list [] is also valid.

charts/vela-core/templates/defwithtemplate/apply-deployment.yaml
```diff
@@ -13,12 +13,8 @@ spec:
...
     cue:
       template: |
-        import (
-        	"strconv"
-        	"strings"
-        	"vela/kube"
-        	"vela/builtin"
-        )
+        import "vela/kube"
+        import "vela/builtin"
```
anoop2811: Why are we missing the strconv and strings package import? Same comment across the board

pkg/cue/convert.go
```diff
@@ -47,8 +47,16 @@ func GetParameters(templateStr string) ([]types.Parameter, error) {
 		if iter.Selector().IsDefinition() {
 			continue
 		}
+		// Use String() instead of Unquoted() to avoid panic on pattern parameter selectors
+		// Unquoted() panics unless the selector is a StringLabel with a concrete name
+		selector := iter.Selector()
```
briankane: Can we wrap this label retrieval in a utility function to avoid the repetition?
Also: Is .Label() deprecated/removed???

references/docgen/parser_test.go
briankane, line 752: Is `[]string` no longer available?

pkg/definition/definition.go
```diff
@@ -225,26 +225,53 @@ func (def *Definition) ToCUEString() (string, error) {
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to parse template cue string")
 	}
-	f = fix.File(f)
-	var importDecls, templateDecls []ast.Decl
+
+	// Extract imports from original file before fix.File() clears them
+	var importPaths []string
```
briankane: What's changed that we need all this additional code?

pkg/definition/definition.go
```diff
@@ -652,12 +679,55 @@ func formatCUEString(cueString string) (string, error) {
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to parse file during format cue string")
 	}
+
+	// Extract imports before fix.File() clears them (same workaround as in ToCUEString)
```
briankane: If we are doing this twice, let's create a function for it

charts/vela-core/templates/defwithtemplate/apply-application-in-parallel.yaml
```diff
@@ -15,9 +15,7 @@ spec:
   schematic:
     cue:
       template: |
-        import (
-        	"vela/op"
-        )
+        import "vela/op"
```
briankane, line 18: Import block changes are mandatory?

charts/vela-core/templates/defwithtemplate/apply-terraform-provider.yaml
```diff
@@ -87,9 +84,9 @@ spec:
         	}
         }
         providerBasic: {
-        	accessKey: string
-        	secretKey: string
-        	region:    string
+        	accessKey!: string
```
briankane, line 87: Again, these changes are mandatory under the new version?
Can we link out to documentation to explain the changes?

roguepikachu: The changes are not exactly mandatory, it's more like a work around because with this new version of Cue. The embedding fields means, cue evaluates for concrete values. The options we have to tackle are: -
1. Mark the fields mandatory, which forces Cue to do only the schema evaluation 
2. Mark the fields optional, again same as former but it doesn't seem like the right use case here
3. Have some default value for the fields of embedding structs, which is not preferred

Here are some references on how the newer version of the Cue does schema evaluation. 
1. [Embedding means fields get combined, which CUE must evaluate](https://cuelang.org/docs/reference/spec/#:~:text=The%20successful%20unification,to%20bottom%2C%20recursively.)
2. [CUE allows incompleteness until a concrete value is needed](https://kubevela.io/docs/platform-engineers/cue/basic/?utm_source=chatgpt.com#:~:text=Export%20the%20result.%20cue%20export%20will%20export%20the%20result%20with%20final%20value.%20It%20will%20report%20an%20error%20if%20some%20variables%20are%20not%20definitive.)

references/docgen/parser_test.go
roguepikachu: This was addresses as part of the cubic ai comment, it could be an empty list or a list of strings. If we said "[]string", that would be misleading because an empty list [] is also valid.

charts/vela-core/templates/defwithtemplate/apply-application-in-parallel.yaml
roguepikachu: These are generated yamls from the Cue template, this wasn't manually modified.  Same in rest of the places

pkg/cue/convert.go
roguepikachu: .Label() is not deprecated - it still exists and is used in the codebase for getting labels from fields directly. However, in the newer version of Cue when working with iterators (from Fields() or similar methods), you need to use iter.Selector() to get the selector first, then extract the label from it. And the key issue is that Selector.Unquoted() can panic on certain selector types, which is why we need the safe utility function.

pkg/definition/definition.go
roguepikachu: The `fix.File()` function from the CUE library removes import declarations that it considers unused. This was causing imports to disappear from the output, which could break the generated CUE code.

The changes implement a workaround to preserve imports:
  1. Before the change: The code was calling fix.File() and then trying to extract imports from the result - but they were already gone!
  2. After the change:
    - Extract imports BEFORE calling fix.File() (while they still exist)
    - Store them separately
    - Call fix.File() (which removes the imports)
    - Manually reconstruct the import statements in the output

Why This Workaround is Needed
  The fix.File() function is important because it:
  - Fixes deprecated CUE syntax
  - Updates to newer CUE conventions
  - Cleans up the AST

But it has this side effect of removing imports it thinks are unused, even if they might be needed in the template context.

pkg/definition/definition.go
roguepikachu: addressed in newer commits

roguepikachu:
Adding to the references to the changes, the arithmetic operation have been deprecated from [Cue version 0.11](https://cuelang.org/docs/concept/faq/removing-list-arithmetic-operators-v0.11/) string concat using `+` ave been replaced with list concat.

briankane:
Merging to unblock some dependency issues. We'll continue looking at how to best manage the changes - can be backed out if we don't reach a conclusion.

PR #6902, opened 2025-09-12
Fix(addon): Correct path calculation and test location for GitLab reader

### LINKED ISSUE #6899: Bug: Incorrect path calculation in GitLab addon reader for root addons
---
name: "🐛 Bug report"
about: Create a report to help us improve
title: "Bug: Incorrect path calculation in GitLab addon reader for root addons"
labels: kind/bug
assignees: ''

---

**Describe the bug**
When a KubeVela addon is located at the root of a GitLab repository (not in a subdirectory), the `gitlabReader` incorrectly calculates the paths for the addon's files by trimming the first character from every file path. This leads to "file not found" errors when trying to enable such an addon.

**To Reproduce**
1. Configure a GitLab-based addon registry where the `path` in the registry configuration points to the root of the repository.
2. Place an addon's files (e.g., `metadata.yaml`) directly in that root.
3. Attempt to list or enable this addon via `vela addon` commands.
4. See error, which will be a "file not found" error because the system is looking for an incorrect path (e.g., `etadata.yaml` instead of `metadata.yaml`).

**Expected behavior**
The system should calculate the correct file path (`metadata.yaml`) and successfully read the addon files.

**Screenshots**
N/A

**KubeVela Version**
All versions with the GitLab addon provider are affected.

**Cluster information**
N/A. This is an issue in the addon reader logic, independent of cluster version.

**Additional context**
The bug is an off-by-one error in the `GetPath` function in `pkg/addon/reader_gitlab.go`.

The function uses `g.path[len(g.basePath)+1:]` to calculate the path. When the `basePath` is an empty string (for a repo root addon), its length is `0`, and the slice becomes `g.path[1:]`, which incorrectly removes the first character.

The fix is to add a check to handle the case where `basePath` is empty:
```diff
--- a/pkg/addon/reader_gitlab.go
+++ b/pkg/addon/reader_gitlab.go
@@ -45,5 +45,8 @@
 
 // GetPath get addon's sub item path
 func (g GitLabItem) GetPath() string {
+	if g.basePath == "" {
+		return g.path
+	}
 	return g.path[len(g.basePath)+1:]
 }
 
```

AshvinBambhaniya2003:
### Description

This PR fixes a bug in the GitLab addon reader where file paths were calculated incorrectly for addons located at the repository root. This was caused by an off-by-one error that truncated the first character of the path.

Additionally, this PR moves the `TestGitLabItem` unit test function from `pkg/addon/reader_local_test.go` to its correct location in `pkg/addon/reader_gitlab_test.go`.

I have:

### Related Issue
- Fixes #6899

The fix is validated by the `TestGitLabItem` unit test, which now passes. This test specifically covers the case of an empty base path that was causing the bug. All other existing tests continue to pass.

This PR includes two distinct changes: the bug fix itself and a minor test file cleanup (moving `TestGitLabItem` from `reader_local_test.go` to `reader_gitlab_test.go`).

### DIFF
--- a/pkg/addon/reader_gitlab.go
+++ b/pkg/addon/reader_gitlab.go
@@ -52,6 +52,9 @@ func (g GitLabItem) GetType() string {
 
 // GetPath get addon's sub item path
 func (g GitLabItem) GetPath() string {
+	if g.basePath == "" {
+		return g.path
+	}
 	return g.path[len(g.basePath)+1:]
 }
 
--- a/pkg/addon/reader_gitlab_test.go
+++ b/pkg/addon/reader_gitlab_test.go
@@ -111,3 +111,39 @@ func TestGitlabReader(t *testing.T) {
 	_, err := r.ReadFile("example/metadata.yaml")
 	assert.NoError(t, err)
 }
+
+func TestGitLabItem(t *testing.T) {
+	t.Run("Getters", func(t *testing.T) {
+		item := GitLabItem{
+			tp:   "blob",
+			name: "metadata.yaml",
+		}
+		assert.Equal(t, "blob", item.GetType())
+		assert.Equal(t, "metadata.yaml", item.GetName())
+	})
+
+	t.Run("GetPath", func(t *testing.T) {
+		testCases := map[string]struct {
+			basePath string
+			fullPath string
+			expected string
+		}{
+			"no base path": {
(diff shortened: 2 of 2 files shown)


AshvinBambhaniya2003:
@anoop2811 This PR is ready to merge.

PR #6915, opened 2025-09-19
Fix(references/appfile): Fix namespace check and Terraform output parsing

### LINKED ISSUE #6914: Bug: Incorrect namespace existence check in `generateSecretFromTerraformOutput`
**Describe the bug**
The function `generateSecretFromTerraformOutput` in `references/appfile/addon.go` contains a bug in its logic for checking if a namespace exists. It incorrectly uses a `Create` operation, which leads to contradictory behavior. If a namespace doesn't exist, the function creates it and then incorrectly returns an error stating the namespace *doesn't* exist.

**To Reproduce**
This bug was identified through code review and unit testing. It can be reproduced in a test environment by calling `generateSecretFromTerraformOutput` with a valid namespace that does not yet exist on the cluster.

**Expected behavior**
The function should correctly check for a namespace's existence using a `Get` call. It should only return a "namespace doesn't exist" error if the `Get` call fails with a `NotFound` error.

**KubeVela Version**
This bug was identified in the current `master` branch.

**Cluster information**
Not applicable, as this is a code-level logic bug independent of a specific cluster version.

**Additional context**

### Problematic Code

The current implementation attempts to create the namespace and returns a confusing error if the creation succeeds.

```go
// generateSecretFromTerraformOutput generates secret from Terraform output
func generateSecretFromTerraformOutput(k8sClient client.Client, outputList []string, name, namespace string) error {
	ctx := context.TODO()
	err := k8sClient.Create(ctx, &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
	if err == nil {
		return fmt.Errorf("namespace %s doesn't exist", namespace)
	}
    // ... rest of the function
}
```

### Suggested Fix

The logic should be changed to use `k8sClient.Get` and check for a `NotFound` error.

```go
import (
    "k8s.io/apimachinery/pkg/api/errors"
)

// generateSecretFromTerraformOutput generates secret from Terraform output
func generateSecretFromTerraformOutput(k8sClient client.Client, outputList []string, name, namespace string) error {
	ctx := context.TODO()
	// Check if namespace exists
	var ns v1.Namespace
	if err := k8sClient.Get(ctx, client.ObjectKey{Name: namespace}, &ns); err != nil {
		if errors.IsNotFound(err) {
			return fmt.Errorf("namespace %s doesn't exist", namespace)
		}
		return fmt.Errorf("failed to get namespace %s: %w", namespace, err)
	}
    // ... rest of the function
}
```

### LINKED ISSUE #6916: Bug: Incorrect parsing of Terraform output can corrupt secret values
**Describe the bug**
The function `generateSecretFromTerraformOutput` in `references/appfile/addon.go` incorrectly parses the output from the `terraform output` command.

The current implementation has two main flaws:
1.  It removes all spaces from the entire output string before parsing, which corrupts values that are supposed to contain spaces (e.g., `message = "Hello World"` becomes `message="HelloWorld"`).
2.  It splits each line by every `=` character, which fails when the value itself contains an `=` (e.g., a connection string like `conn_str = "user=admin;pass=123"`).

The problem is caused by two specific lines of code:

1.  In the `ApplyTerraform` function, all spaces are stripped from the entire output string, which corrupts values that should contain spaces.
    ```go
    // This line incorrectly removes all spaces from the output value
    outputList := strings.Split(strings.ReplaceAll(string(outputs), " ", ""), "\n")
    ```

2.  In the `generateSecretFromTerraformOutput` function, each line is split on *every* `=` character, which fails to parse values that legitimately contain an equals sign (like a URL or connection string).
    ```go
    // This line will incorrectly split a value that contains an equals sign
    line := strings.Split(i, "=")
    ```

**To Reproduce**
1. Create a Terraform component that defines an output variable with a value containing spaces or an equals sign.
   ```terraform
   output "my_message" {
     value = "This is a test message."
   }
   output "connection_string" {
     value = "user=admin;password=123"
   }
   ```
2. Deploy this component as part of a KubeVela application.
3. Inspect the secret generated from the Terraform output.
4. Observe that the `my_message` value is `Thisisatestmessage.` (corrupted) and the `connection_string` may cause a parsing error, preventing the secret from being created correctly.

**Expected behavior**
The parsing logic should correctly handle spaces and special characters in Terraform output values.

- For `my_message`, the secret value should be `This is a test message.`.
- For `connection_string`, the secret value should be `user=admin;password=123`.

The parser should only split the key/value pair on the first `=` in each line and should not globally remove spaces.

**KubeVela Version**

**Cluster information**

**Additional context**
The problematic code is located in `references/appfile/addon.go` within the `ApplyTerraform` and `generateSecretFromTerraformOutput` functions.

AshvinBambhaniya2003:
### Description

This PR addresses two distinct bugs in the `references/appfile/addon.go` file to improve the robustness of the addon system.

1.  **Incorrect Namespace Check:** Fixes a bug where the function to check for a namespace's existence was incorrectly using a `Create` call instead of a `Get` call. This corrects the logic to be idempotent and reliable.

2.  **Fragile Terraform Output Parsing:** Fixes a bug in the parsing of `terraform output`. The previous logic would corrupt values containing spaces and fail to parse values containing equals signs. This has been replaced with a robust parser that correctly handles these common cases.

I have:

### Issues Resolved
- Fixes #6914
- Fixes #6916

The unit tests in `references/appfile/addon_test.go` have been updated to align with the new function signatures and to verify the corrected behavior for both the namespace check and the robust Terraform output parsing.

This PR contains two main fixes:

1.  The namespace check now correctly uses `k8sClient.Get`.
2.  The Terraform output parser has been refactored to use `strings.SplitN`, making it significantly more resilient to special characters in Terraform output values.

### DIFF
--- a/references/appfile/addon.go
+++ b/references/appfile/addon.go
@@ -95,11 +95,7 @@ func ApplyTerraform(app *v1beta1.Application, k8sClient client.Client, ioStream 
 				return nil, err
 			}
 
-			outputList := strings.Split(strings.ReplaceAll(string(outputs), " ", ""), "\n")
-			if outputList[len(outputList)-1] == "" {
-				outputList = outputList[:len(outputList)-1]
-			}
-			if err := generateSecretFromTerraformOutput(k8sClient, outputList, name, namespace); err != nil {
+			if err := generateSecretFromTerraformOutput(k8sClient, string(outputs), name, namespace); err != nil {
 				return nil, err
 			}
 		default:
@@ -135,7 +131,7 @@ func callTerraform(tfJSONDir string) ([]byte, error) {
 }
 
 // generateSecretFromTerraformOutput generates secret from Terraform output
-func generateSecretFromTerraformOutput(k8sClient client.Client, outputList []string, name, namespace string) error {
+func generateSecretFromTerraformOutput(k8sClient client.Client, rawOutput string, name, namespace string) error {
 	ctx := context.TODO()
 
 	// Check if namespace exists
@@ -147,15 +143,20 @@ func generateSecretFromTerraformOutput(k8sClient client.Client, outputList []str
 		return fmt.Errorf("failed to get namespace %s: %w", namespace, err)
 	}
 
-	var cmData = make(map[string]string, len(outputList))
+	var cmData = make(map[string]string)
+	outputList := strings.Split(rawOutput, "\n")
 	for _, i := range outputList {
-		line := strings.Split(i, "=")
+		if strings.TrimSpace(i) == "" {
+			continue
+		}
+		line := strings.SplitN(i, "=", 2)
 		if len(line) != 2 {
-			return fmt.Errorf("terraform output isn't in the right format")
+			return fmt.Errorf("terraform output isn't in the right format: %q", i)
 		}
 		k := strings.TrimSpace(line[0])
-		v := strings.TrimSpace(line[1])
-		if k != "" && v != "" {
+		// Terraform string outputs are quoted, remove them.
+		v := strings.Trim(strings.TrimSpace(line[1]), "\"")
+		if k != "" {
 			cmData[k] = v
 		}
 	}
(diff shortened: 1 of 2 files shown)


vishal210893:
Changes look fine. Can be merged
cc @anoop2811

PR #6984, opened 2025-11-12
Feat: Improve Cue Error Reporting

briankane:

copilot:all

Component/Trait rendering only shows the first error - which is often only surfaced by output issues. For users this is confusing as the actual error is hidden and the one displayed is cryptic. 

This changes it (initially) to show all errors and group them by parameter fields and other template fields.

e.g.

Original:

```
Error: cannot generate manifests from components and traits: evaluate base template component=my-component app=my-app: failed to have the workload/trait unstructured: output.metadata.name: incomplete value string
```

New:
```
Error: cannot generate manifests from components and traits: evaluate base template app=my-app in namespace=default: validation failed for workload my-component:

Parameter errors:
  parameter.embedded.name: field is required but not present

Template errors:
  output.metadata.name: incomplete value string
```

This requires a few workarounds for the fact that KubeVela templates are not always Concrete cue and some validation issues (like missing fields, or optional fields which are unsafely referenced) slip through until we call Unstructured(). To account for this - we wait until we get failures from Unstructured() and attempt to validate with Concrete=true. Validating with Concrete any earlier can introduce breaking changes to end user templates. 

Fixes #

I have:

- Unit Tests

### DIFF
--- a/pkg/cue/definition/template_test.go
+++ b/pkg/cue/definition/template_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package definition
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


pkg/cue/definition/template_test.go
```diff
@@ -1377,6 +1378,112 @@ parameter: {
...
+			protocol: parameter.protocol
+		}]
+	}
+}`,
+			params: map[string]interface{}{
+				"port":     70000,
+				"protocol": "INVALID",
+			},
+			isWorkload: false,
+			wantErr:    "validation failed for trait my-trait:",
```
FogDong: can we have the full error message here?

Issue #6990, opened 2025-11-19
[Feature] Component Compositions


briankane:
**Is your feature request related to a problem? Please describe.**
Currently, KubeVela ComponentDefinitions cannot compose or reference other definitions, leading to duplicated template code. Teams building multi-layered abstractions must copy entire definition templates rather than reusing existing ones, or export and modify built-in KubeVela definitions. For large, complex orgs the built-in definitions are often too generic and permit too much control over what can be provisioned.

There's no way to create higher-level definitions that delegate to lower-level ones while preserving health checks and status evaluation.

**Describe the solution you'd like**
Enable ComponentDefinitions to compose and render other ComponentDefinitions with parameter transformation. This would support composition patterns such as:

  - Multi-cloud abstractions: A single "database" definition that delegates to cloud-specific definitions (aws-rds, azure-sql, gcp-cloudsql) based on parameters
  - Complexity Abstraction: Simple definitions with minimal parameters that internally compose complex definitions with smart defaults
  - Multi-component services: Definitions that compose multiple base components
  - Governance wrappers: Organization-specific definitions that wrap base definitions with security policies, resource limits, and compliance controls

Definitions should be able to access and aggregate health status from composed definitions, creating a proper dependency chain for status evaluation.

PR #6993, opened 2025-11-20
Feat(KEP): Nested Definition Rendering (Compositions)

briankane:

copilot:all

KEP for https://github.com/kubevela/kubevela/issues/6990

### DIFF
--- a/design/vela-core/nested-definition-rendering.md
+++ b/design/vela-core/nested-definition-rendering.md
@@ -48,6 +48,7 @@ Introduce a `def.#RenderComponent` CUE function that allows definitions to refer
 template: {
   _rendered: def.#RenderComponent & {
     definition: "base-component"  // Name of definition to render
+    version: "1.2.3"              // Version of the definition to render (should update to latest for consistency with app declarations)
     properties: {                 // Parameters to pass
       // Transform parameters to pass to the base definition
     }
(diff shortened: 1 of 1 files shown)


design/vela-core/nested-definition-rendering.md
```diff
@@ -0,0 +1,402 @@
...
+## Proposal
+
+### Core Design
+
+Introduce a `def.#RenderComponent` CUE function that allows definitions to reference and render other definitions:
+
+```cue
+template: {
+  _rendered: def.#RenderComponent & {
+    definition: "base-component"  // Name of definition to render
```
jguionnet, line 50: Component version combination? Should not a component version be specified here? Probably even mandate that the semantic main version is set: `v1.x.y` but not `vw.x.z` as major versions are more backward compatible

design/vela-core/nested-definition-rendering.md
```diff
@@ -0,0 +1,402 @@
...
+
+#### 1. CUE Provider Package
+Create a new `def` provider in `pkg/cue/def/` that:
+- Leverages the CueX provider framework recently made available to components/traits (previously only accessible to workflows)
+- Loads referenced ComponentDefinitions using `util.GetDefinition`
+- Evaluates templates with provided parameters
+- Returns rendered output as CUE values
+
+#### 2. Composition Context Tracking
+Track composition hierarchy through process context to enable:
```
jguionnet, line 98: Observability: It would be great to expose progress metrics (ie, which layers are rendered, in progress, or fail)

design/vela-core/nested-definition-rendering.md
```diff
@@ -0,0 +1,402 @@
...
+- **Health Status Aggregation**: Parent definitions can access `context.composition.{name}.status.isHealth`
+- **Status Message Propagation**: Custom status messages flow up through layers
+- **Metadata Access**: Components know their composition hierarchy
+
+**TODO**: Investigate exposing all rendered status fields (not just isHealth and message) to enable richer composition scenarios where parent definitions can access detailed status information from composed definitions - as compositions will not always do simple pass-throughs.
```
jguionnet: Do composed children inherit labels/annotations from parents automatically? If so, which ones?

design/vela-core/nested-definition-rendering.md
briankane: Yeah I didn't get versions in for the POC but it's on the todo list for productionising it. For the purposes of the KEP it would be clearer to flag it. Adding

design/vela-core/nested-definition-rendering.md
briankane: Added to this section for exposition of the composition context to the Application status and error reporting requirements.

design/vela-core/nested-definition-rendering.md
briankane: This will depend on how the called component itself handles it. 

The caller is just passing the available context through and getting the outputs for any given component. The developer can then add/remove/change labels as required in the caller.

briankane:
> Components can be namespace scoped. See https://github.com/kubevela/kubevela/blob/f196d66b/charts/vela-core/crds/core.oam.dev_componentdefinitions.yaml#L19
> 
> * How are namespaces and multi-tenant isolation handled when rendering cross-namespace definitions?
> 
> Consider adding other X-defintions example:
> 
> * Traits
> * Policy
> * Workflow
> 
> See also a few comments inline

We'd want to as closely follow the pre-existing logic from Applications. That is, the specified component would follow the 2-phase lookup logic (local has priority, then vela-system). We may need to look into whether we need an `alwaysUseSystemComponent` option or similar to force ignoring the local namespace. But we shouldn't allow cross namespace referencing as that breaks the isolation. 

I'm scoping this to components only right now. We can expand it down the line but I'm not sure or convinced of it's usefulness for traits/policies (as they should be much more atomically targeted to a specific need/modification). And for workflowsteps - we'd need to make the change into the workflow codebase which will be it's own follow-up undertaking if this feature gets adoption. 

[EDIT] Captured this into the KEP

Issue #6995, opened 2025-11-24
[Feature] Declarative Addon Support


briankane:
POC Code: https://github.com/kubevela/kubevela/pull/6997
KEP: https://github.com/kubevela/kubevela/pull/6996

### **Is your feature request related to a problem? Please describe.**

Currently, KubeVela addons face several management challenges:
- **No GitOps support**: Addons must be managed via CLI (`vela addon enable`) or UI, preventing declarative Git-based management
- **Manual lifecycle management**: Installation, upgrades, and cleanup require manual intervention
- **Poor orchestration**: No easy way to coordinate addon installation with dependent application resources
- **Inconsistent environments**: Difficult to ensure same addons across dev/staging/prod environments
- **Testing complexity**: Can't easily test applications with required addons in CI/CD pipelines

### **Describe the solution you'd like**

Enable addons to be used as first-class components within KubeVela Applications:
- **Component-based approach**: Deploy addons using `type: addon` in Application specs
- **Full addon features**: Support version constraints, registries, selective installation
- **GitOps compatible**: Manage addons declaratively alongside application resources
- **Automatic lifecycle**: Addons installed/upgraded/removed with application lifecycle
- **Performance optimized**: Built-in caching (100ms → 20μs) for repeated renders

### Example usage:
```yaml
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: my-app
spec:
  components:
    - name: prometheus
      type: addon
      properties:
        version: ">=2.0.0"
```

### **Describe alternatives you've considered**

1. **Create dedicated Addon CRD with controller**
   - Rejected: Massive code duplication, recreates existing Application controller logic
   - Would need separate reconciliation, status tracking, revision management

2. **Use workflow steps for addon installation**
   - Rejected: Not GitOps compatible, temporal execution without proper lifecycle
   - Still requires vela-workflow addon, lacks cleanup/deletion propagation

The component approach leverages existing KubeVela infrastructure, avoiding new controllers while providing full orchestration capabilities through existing traits, policies, and workflows.

PR #6996, opened 2025-11-24
Feat(KEP): Declarative Addon Support

briankane:

copilot:all

KEP for https://github.com/kubevela/kubevela/issues/6995

POC Code: https://github.com/kubevela/kubevela/pull/6989

### DIFF
--- a/design/addon-as-component.md
+++ b/design/addon-as-component.md
@@ -1,0 +1,365 @@
+# Addon as Component
+
+### Introduction
+
+This proposal introduces the capability to use KubeVela addons as first-class components within Applications. Rather than creating a new Addon CRD with a dedicated controller, this approach reuses KubeVela's existing Application and Component infrastructure to reduce the need for additional controllers and to reduce the maintenance burden. This enables users to deploy and manage addons using the same declarative approach as regular workloads, bringing addon management into the GitOps workflow while leveraging all existing KubeVela features like policies, workflows, and traits.
+
+### Background
+
+Currently, KubeVela addons are managed separately from applications through CLI commands (`vela addon enable`), the VelaUX UI, or WorkflowRuns. While WorkflowRun support exists, it requires the vela-workflow addon to be pre-installed, still depends on CLI operations within workflow steps, and lacks proper lifecycle management — installation and deletion are separate workflows without automatic cleanup via finalizers or deletion propagation. This separation creates several challenges:
+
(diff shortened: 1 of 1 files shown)


design/addon-as-component.md
```diff
@@ -0,0 +1,346 @@
...
+      type: addon
+      properties:
+        version: ">=2.0.0"
+    
+    - name: my-app
+      type: webservice
+      properties:
+        image: myapp:latest
+  
+  policies:
```
jguionnet, line 268: Add a comment indicating that it is an incomplete example, or add a `topology` prod policy and a workflow with a deploy step, associating the policy with the override setting.

jguionnet (approved):
One comment inline and I wonder about

Upgrade of the add-on version. Will it be a component that relies on semantic versioning? If so, we will need to run multiple versions of a K8s OOTB package and define a strategy for that.

Will it be discovered, as components, by VelaUX and VelaQ, and so they should not be impacted?

else LGTM

anoop2811:
Please share this WIP  KEP in the community meeting and also in our kubevela slack channel. Let us see what the community has to say

design/addon-as-component.md
briankane: Thanks, well spotted. Added the necessary to make this a working example.

briankane:
> One comment inline and I wonder about
> 
> Upgrade of the add-on version. Will it be a component that relies on semantic versioning? If so, we will need to run multiple versions of a K8s OOTB package and define a strategy for that.
> 
> Will it be discovered, as components, by VelaUX and VelaQ, and so they should not be impacted?
> 
> else LGTM

So currently upgrades are a CLI command `vela addon upgrade` which runs the rendering, creates an app and then it goes through the standard dispatching logic. 

This works much the same, except triggered by the Application manifest update. On next reconciliation, it will render the new version of the addon (in the same way the CLI currently does) and go through the standard dispatch logic - updating any changed resources. The only subtle difference is that the definitions, config-templates, views, etc. are actually added to the creating Application (the "addon installer app") and brought under it's control. Which I think is actually an improvement.

In terms of VelaUX - the addons remain a simple dynamically assembled Application - so all features that currently work will continue to. It's one of the benefits of reusing what's already built out - we get compatibility with everything for free.

briankane:
> Please share this WIP KEP in the community meeting and also in our kubevela slack channel. Let us see what the community has to say

Done

PR #7008, opened 2025-12-11
Feat: post dispatch output context

### LINKED ISSUE #6810: Unable to pass dynamic values to trait parameters
**Unable to pass any dynamic values to the trait parameter**
I'm trying to create a PV using trait and pass dynamic values to the volume handle which will be known only at the runtime (after the parent component - EFS is deployed). The trait is able to deploy the PV using static values but it is not recognising any dynamic values. 

**To Reproduce**
Create a trait with the below yaml file and try to attach it to a component. While deploying the application, try to pass any values that will be known at the runtime. In my case, the parent component will give me the `filesystemId` in it's status. 

```
"pv-using-trait": {
	alias: "pv-using-trait-v2"
	annotations: {}
	attributes: {
		appliesToWorkloads: []
		conflictsWith: []
		podDisruptive:   false
		workloadRefPath: ""
	}
	description: "deploy pv using trait"
	labels: {}
	type: "trait"
}

template: {
	outputs: persistentVolume: {
		apiVersion: "v1"
		kind:       "PersistentVolume"
		metadata: {
			name: parameter.volumeName
		}
		spec: {
			capacity: {
				storage: "10Gi"
			}
			storageClassName: "efs-csi-infra-sc"
			volumeMode:       "Filesystem"
			accessModes:      ["ReadWriteMany"]
			persistentVolumeReclaimPolicy: "Retain"
			csi: {
				driver:        "efs.csi.aws.com"
				volumeHandle:  "\(parameter.filesystemId)::\(parameter.accessPointId)"
				}
			}
		}
	parameter: {
		volumeName: string | *"atmos-efs-volume"
		filesystemId: string & {
		    if parameter.filesystemId == "" || parameter.filesystemId == null || parameter.filesystemId == _|_ {
		        errorMessage: "filesystemId must be specified"
		    }
		}
		accessPointId: string & {
		    if parameter.accessPointId == "" || parameter.accessPointId == null || parameter.accessPointId == _|_ {
		        errorMessage: "accessPointId must be specified"
		    }
		}
	}
}
```

```yaml
kind: Application
metadata:
  name: efs-kubevela-poc
  namespace: ajanta
spec:
  components:
    - name: provision-efs
      type: atmos-efs-v1
      properties:
        name: poc-minimal
        path: /data/poc-minimal
    - name: provision-webservice
      type: webservice
      properties:
        image: busybox:latest
        command:
          ["sh", "-c", "while true; do echo 'Hello from EFS'; sleep 5; done"]
        port: 80
      traits:
        - type: atmos-vol-v2
          properties:
            volumeName: tenant-ajanta-poc-minimal
            accessPointId: (provision-efs.outputs.ap-id)
            filesystemId: (provision-efs.outputs.fs-id)
  workflow:
    steps:
      - name: provision-efs
        type: apply-component
        properties:
          component: provision-efs
        outputs:
          - name: fs-id
            valueFrom: output.status.filesystemId
          - name: ap-id
            valueFrom: output.status.accessPointId
      - name: provision-webservice
        type: apply-component
        properties:
          component: provision-webservice
        dependsOn:
          - provision-efs
```
> I tried to attach in a single component but the kubevela is not waiting for the first component to deploy and fetch it's status. So, I tried this two component approach and still it fails

**Expected behavior**

Kubevela trait should be able to read the status from it's parent component and use the same dynamic value in it's template.  

**Screenshots**

Image

**KubeVela Version**

```
CLI Version: 1.10.3
Core Version: 1.10.2
GitRevision: ef9b6f3cc10a4b6871b5698ca41fea3f6b3bcaec
GolangVersion: go1.24.3
```

**Cluster information**

```
Client Version: v1.31.5
Kustomize Version: v5.4.2
Server Version: v1.31.5+k3s1
```

**Additional context**
I even tried to access it directly inside the cue template but that method also didn't work. Because, these data won't populated at the compile-time and kubevela is not waiting for that. 

```
"atmos-vol-v2": {
	alias: "atmos-efs-volume-v2"
	annotations: {}
	attributes: {
		appliesToWorkloads: []
		conflictsWith: []
		podDisruptive:   false
		workloadRefPath: ""
	}
	description: "deploy pv using trait"
	labels: {}
	type: "trait"
}

template: {
		outputs: persistentVolume: {
		apiVersion: "v1"
		kind:       "PersistentVolume"
		metadata: {
			name: parameter.volumeName
		}
		spec: {
			capacity: {
				storage: "10Gi"
			}
			storageClassName: "efs-csi-infra-sc"
			volumeMode:       "Filesystem"
			accessModes:      ["ReadWriteMany"]
			persistentVolumeReclaimPolicy: "Retain"
			csi: {
				driver:        "efs.csi.aws.com"
				volumeHandle:   context.output.status.filesystemId + "::" + context.output.status.accessPointId
				}
			}
		}

	parameter: {
        // +usage=Name of the volume to be created
		volumeName: string | *"atmos-efs-volume-v2"
	}
}
```

semmet95:

copilot:all

The PR makes the output/main workload (along with all its fields) of a component available to the PostDispatch type traits attached to it.
The changes include:
* Keeping the PostDispatch type traits attached to a workload on hold until the component workload is ready.
* Injecting the `output` into the context of these PostDispatch type traits while they are being rendered after the component workload is ready.

Fixes #6810

I have:

Added tests to verify that PostDispatch type traits have access to `output` fields like `output.status`, `output.status.availableReplicas` etc.

---
### Major issue encountered
**Fix 1: PostDispatch Traits Not Applied Mid-Workflow**
In multi-step workflows that consume PostDispatch trait outputs (e.g., `collect-service-endpoints -> export-data`), PostDispatch traits were only applied after the workflow completed successfully. During intermediate states (e.g., `suspending`, `executing`, `skipped`, or `requeues`), traits were never dispatched. As a result, outputs remained unavailable and downstream steps could not proceed.
[Solution](https://github.com/kubevela/kubevela/pull/7008/commits/595d31b548aa9c9e84f20488774eb5afb158bceb)
PostDispatch traits are now applied once components are healthy across all workflow states, not just at the end. A new `applyPostDispatchTraits` helper was added and is invoked in the `suspend`, `executing`, and `skipped` branches, and again before finalization. This updates `AppliedResources` when traits are dispatched.

**Fix 2: Deploy Step Stalls When PostDispatch Outputs Are Consumed**
In applications where a component exposes PostDispatch outputs consumed by another component (e.g., `outputs.service.spec.clusterIP` into a ConfigMap), the deploy step would hang indefinitely. Health checks never marked outputs as ready because traits weren’t rendered yet, causing unresolved downstream inputs and workflow stalls.
_Solution_
Track output readiness explicitly during health checks.
* If `generateOutput` fails due to missing `outputs.*`, mark outputs as not ready but do not fail.
* Aggregate output readiness into the overall workflow health state to allow requeues.
* Retain PostDispatch traits when other components consume their outputs.
* Label extracted outputs so they can be reliably looked up by name.
_Key Code Changes_
* [pkg/workflow/providers/multicluster/deploy.go](https://github.com/kubevela/kubevela/pull/7008/commits/9606e0adfc76fd80d43634874727e4d3a9e3099b)
```
type applyTaskResult struct {
    healthy     bool
    err         error
    task        *applyTask
    outputReady bool   // new: track output readiness
}

// Inside health check loop
if healthy {
    if errOutput := task.generateOutput(...); errOutput != nil {
        var notFound workflowerrors.LookUpNotFoundErr
        if errors.As(errOutput, &notFound) && strings.HasPrefix(string(notFound), "outputs.") {
            errOutput = nil
        }
        err = errOutput
    }
}

return &applyTaskResult{..., outputReady: true}
```
Output readiness is aggregated into the overall health status.
* [pkg/controller/core.oam.dev/v1beta1/application/generator.go](https://github.com/kubevela/kubevela/pull/7008/commits/06309a210d6ccb72349e3b3798d6add1a65c2fd8)
```
// Retain PostDispatch traits if consumed by other components
func componentOutputsConsumed(...) bool {
    ...
}
```
* `pkg/controller/core.oam.dev/v1beta1/application/apply.go`
```
// Label extracted outputs for lookup
for k, v := range templateContext["outputs"].(map[string]interface{}) {
    ...
    labels[oam.TraitResource] = k
    ...
}
```
_Result_
* Deploy health checks requeue until outputs are available, avoiding stalls.
* PostDispatch traits are preserved if outputs are consumed downstream.
* Outputs are labeled for discoverability, enabling reliable data flow.

**Fix 3: PostDispatch Traits Cause Admission Webhook Failures**
The admission webhook attempted to evaluate PostDispatch traits during validation. These traits rely on runtime-only data, causing evaluation to fail and requests to be rejected.
_Solution_
PostDispatch traits are now skipped during webhook-time validation. They are handled only at runtime when necessary data is available.
_Code Change_
* `pkg/appfile/validate.go`
```
for _, tr := range comp.Traits {
    if tr.CapabilityCategory != types.CUECategory {
        continue
    }
    if tr.FullTemplate != nil &&
        tr.FullTemplate.TraitDefinition.Spec.Stage == v1beta1.PostDispatch {
        continue // Skip PostDispatch traits
    }
    if err := tr.EvalContext(pCtx); err != nil {
        return errors.WithMessagef(err, "cannot evaluate trait %q", tr.Name)
    }
```

### DIFF
--- a/test/e2e-test/postdispatch_trait_test.go
+++ b/test/e2e-test/postdispatch_trait_test.go
@@ -506,7 +506,7 @@ if pod.metadata.annotations != _|_ {
 			}
 			Expect(k8sClient.Create(ctx, deploymentTrait)).Should(Succeed())
 
-			By("Creating PostDispatch configmap trait definition")
+			By("Creating PostDispatch configMap trait definition")
 			cmTrait := &v1beta1.TraitDefinition{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      cmTraitName,


pkg/appfile/validate.go
```diff
@@ -64,6 +66,10 @@ func (p *Parser) ValidateCUESchematicAppfile(a *Appfile) error {
 			if tr.CapabilityCategory != types.CUECategory {
 				continue
 			}
+			if tr.FullTemplate != nil &&
+				tr.FullTemplate.TraitDefinition.Spec.Stage == v1beta1.PostDispatch {
```
semmet95, line 70: PostDispatch traits might have references to fields generated during runtime so we need to delay this evaluation.

pkg/controller/core.oam.dev/v1beta1/application/generator.go
```diff
@@ -475,6 +522,31 @@ func renderComponentsAndTraits(manifest *types.ComponentManifest, appRev *v1beta
 	return readyWorkload, readyTraits, nil
 }
 
+// componentOutputsConsumed returns true if any other component depends on outputs produced
+// from PostDispatch traits (valueFrom starting with "outputs.").
+func componentOutputsConsumed(comp common.ApplicationComponent, components []common.ApplicationComponent) bool {
```
semmet95, line 530: Do we need this function? Is it not safe to assume that "output"/"outputs" is always needed?

pkg/controller/core.oam.dev/v1beta1/application/apply.go
```diff
@@ -430,9 +431,127 @@ func extractOutputAndOutputs(templateContext map[string]interface{}) (*unstructu
...
+			}
+			ctxData.ClusterVersion = multicluster.GetVersionInfoFromObject(
+				pkgmulticluster.WithCluster(ctx.GetContext(), types.ClusterLocalName),
+				h.Client,
+				ctxData.Cluster,
+			)
+
+			// Fetch live workload status for PostDispatch traits to use
+			tempCtx := appfile.NewBasicContext(*ctxData, wl.Params)
+			if err := wl.EvalContext(tempCtx); err != nil {
```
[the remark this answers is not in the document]
semmet95: updated

pkg/controller/core.oam.dev/v1beta1/application/apply.go
```diff
@@ -430,9 +431,127 @@ func extractOutputAndOutputs(templateContext map[string]interface{}) (*unstructu
...
+				return
+			}
+			base, _ := tempCtx.Output()
```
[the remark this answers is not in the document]
semmet95: what?? Output() does not return error.

pkg/controller/core.oam.dev/v1beta1/application/generator.go
```diff
@@ -444,6 +468,32 @@ func (h *AppHandler) prepareWorkloadAndManifests(ctx context.Context,
 		// cluster info are secrets stored in the control plane cluster
 		ctxData.ClusterVersion = multicluster.GetVersionInfoFromObject(pkgmulticluster.WithCluster(ctx, types.ClusterLocalName), h.Client, ctxData.Cluster)
 		ctxData.CompRevision, _ = ctrlutil.ComputeSpecHash(comp)
+
+		if utilfeature.DefaultMutableFeatureGate.Enabled(features.MultiStageComponentApply) {
+			// inject the main workfload output as "output" in the context
```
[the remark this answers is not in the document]
semmet95: updated

pkg/oam/util/helper.go
```diff
@@ -835,3 +838,42 @@ func (accessor *applicationResourceNamespaceAccessor) Namespace() string {
...
+		labels[oam.TraitResource] = outputsResource
+	}
+	if obj.GetName() != "" {
+		u, err := GetObjectGivenGVKAndName(ctx, client, obj.GroupVersionKind(), namespace, obj.GetName())
+		if err != nil {
+			return nil, err
+		}
+		return u.Object, nil
+	}
+	if ctxName := pctx.GetData(model.ContextName).(string); ctxName != "" {
```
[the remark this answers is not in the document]
semmet95: updated

pkg/controller/core.oam.dev/v1beta1/application/apply.go
```diff
@@ -430,9 +431,130 @@ func extractOutputAndOutputs(templateContext map[string]interface{}) (*unstructu
...
+			} else {
+				ctxData.Cluster = pkgmulticluster.Local
+			}
+			ctxData.ClusterVersion = multicluster.GetVersionInfoFromObject(
+				pkgmulticluster.WithCluster(ctx.GetContext(), types.ClusterLocalName),
+				h.Client,
+				ctxData.Cluster,
+			)
+
+			// Fetch live workload status for PostDispatch traits to use if its created on the cluster
```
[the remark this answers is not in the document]
semmet95: updated

pkg/controller/core.oam.dev/v1beta1/application/generator.go
vishal210893: outputs are basically required only when a component is returning something, which can be consumed by another component. e.g.,
``` yaml
kind: Application
metadata:
  name: app-deploy-io
spec:
  components:
    - name: podinfo
      outputs:
        - name: message
          valueFrom: output.status.conditions[0].message
        - name: ip
          valueFrom: outputs.service.spec.clusterIP

      properties:
        image: stefanprodan/podinfo:4.0.3
      type: webservice
      traits:
        - type: expose
          properties:
            port: [ 80 ]
```

briankane: So this will delay the dispatch of a component B if component A has outputs from a PostDispatch trait that is not yet rendered?

pkg/oam/util/helper.go
```diff
@@ -835,3 +838,42 @@ func (accessor *applicationResourceNamespaceAccessor) Namespace() string {
...
 }
+
+func WithCluster(ctx context.Context, o client.Object) context.Context {
+	if cluster := oam.GetCluster(o); cluster != "" {
+		return multicluster.WithCluster(ctx, cluster)
+	}
+	return ctx
+}
+
+func GetResourceFromObj(ctx context.Context, pctx process.Context, obj *unstructured.Unstructured, client client.Reader, namespace string, labels map[string]string, outputsResource string) (map[string]interface{}, error) {
```
briankane, line 849: Should be plural? GetResourcesFromObj?

pkg/features/controller_features.go
```diff
@@ -138,7 +138,7 @@ var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	GzipResourceTracker:                           {Default: false, PreRelease: featuregate.Alpha},
 	ZstdResourceTracker:                           {Default: false, PreRelease: featuregate.Alpha},
 	ApplyOnce:                                     {Default: false, PreRelease: featuregate.Alpha},
-	MultiStageComponentApply:                      {Default: false, PreRelease: featuregate.Alpha},
+	MultiStageComponentApply:                      {Default: true, PreRelease: featuregate.Alpha},
```
briankane, line 141: Should we be defaulting this to true?

pkg/cue/definition/template.go
```diff
@@ -239,10 +239,54 @@ func (td *traitDef) Complete(ctx process.Context, abstractTemplate string, param
 			buff += fmt.Sprintf("%s: %s\n", velaprocess.ParameterFieldName, string(bt))
 		}
 	}
+	var statusBytes []byte
```
briankane: I'm not really following this code? It looks like we are having to work around the existing data passing code?

[EDIT] I see there's an explanation in the description. Can we put the new code into a function that makes it more clear what is happening and why?

briankane: Also are a lot of these changes only applicable when the multi stage flag is true? Should this code be conditional on the flag to reduce the blast radius of the changes?

pkg/features/controller_features.go
semmet95: @briankane MultiStageComponentApply already [defaults](https://github.com/kubevela/kubevela/blob/master/charts/vela-core/values.yaml#L134) to `true` when Kubevela is installed via its helm chart. The change here is relevant to debugging when the controller runs in the IDE.

pkg/oam/util/helper.go
semmet95: This function is actually a copy of [this](https://github.com/kubevela/kubevela/blob/3cc668289e01f8198a18dbb91dfa28f6c6ff7850/pkg/cue/definition/template.go#L411) function with the same name. I needed to make it accessible across multiple packages so I made a copy in the `util` package. I think it's only returning a single object. Even if there are multiple matches based on the Kind, Namespace and labels, it returns the first object that matches the `outputsResource` value in its `trait.oam.dev/resource` label.

pkg/controller/core.oam.dev/v1beta1/application/generator.go
vishal210893: Yes, and Component A will be marked as ready even though the outputs are not yet populated, so that the post-dispatch trait can be applied (since the post-dispatch trait can only be executed when the corresponding component is healthy). Once the trait is applied, the outputs get populated.

The function basically checks whether there is any output to be consumed by another trait and provides the result to the calling function.

pkg/cue/definition/template.go
vishal210893: Updated the code by creating a new function, which is wrapped inside a multi-stage flag

pkg/features/controller_features.go
briankane: Got it, thanks. Strange that the chart defaults it to true if it's marked alpha. This increases the blast radius of the changes as it'll be rolled out near universally. 

Question; should the chart default this to false for safety?

pkg/controller/core.oam.dev/v1beta1/application/generator.go
briankane: Nice, I didn't consider that scenario.

FogDong (approved):

PR #7009, opened 2025-12-15
Docs(KEP): Go SDK for X-Definition Authoring (defkit)

anoop2811:
Introduces KEP proposal for defkit, a Go SDK that enables platform
  engineers to author X-Definitions using native Go code instead of CUE.

  Key proposed features:
  - Fluent builder API for Component, Trait, Policy, and WorkflowStep definitions
  - Transparent Go-to-CUE compilation
  - IDE support with autocomplete and type checking
  - Schema-agnostic resource construction
  - Collection operations (map, filter, dedupe)
  - Composable health and status expressions
  - Addon integration with godef/ folder support
  - Module dependencies for definition sharing via go get

### DIFF
--- a/design/vela-cli/kep-defkit.md
+++ b/design/vela-cli/kep-defkit.md
@@ -113,6 +113,8 @@ def.CustomStatus(defkit.DeploymentStatus().Build())
 
 ### When to Use What
 
+> **Note:** `RawCUE()` is an escape hatch for the rare patterns (~5% of cases) that cannot be expressed in the Go API. It allows embedding raw CUE code directly. See the [RawCUE() Escape Hatch](#rawcue-escape-hatch) section for details.
+
 ```
 ┌─────────────────────────────────────────────────────────────────┐
 │                     DECISION FRAMEWORK                          │


design/vela-cli/kep-defkit.md
```diff
@@ -0,0 +1,2058 @@
...
+## Motivation
+
+### Goals
+
+1. **Author X-Definitions in Go** - Write ComponentDefinition, TraitDefinition, PolicyDefinition, and WorkflowStepDefinition using fluent Go APIs
+2. **Transparent CUE compilation** - Go code compiles to CUE automatically; developers never see or write CUE
+3. **Full IDE support** - Autocomplete, type checking, and inline documentation
+4. **Testable definitions** - Unit test definitions using standard Go testing frameworks
+5. **Schema-agnostic resource construction** - Support any Kubernetes resource without coupling to K8s versions
+6. **Easy distribution via Go modules** - Share and version X-Definitions as standard Go packages, enabling `go get` for platform capabilities
```
FogDong, line 18: Could we also add security part in the goals to track?

sreejith-rajashekaran:
Should we also touch up on how existing CUE definitions can be migrated?

design/vela-cli/kep-defkit.md
anoop2811: @FogDong when you say security, do you mean golang security is more robust and easy or did you mean that the golang api guardrails could if written right make sure that cue code it generates is more secure?

anoop2811:
> Should we also touch up on how existing CUE definitions can be migrated?

Thats a little harder nut to crack and analyze backwards, so skipped it off this KEP. We dont need to migrate off, we will support both. But will be trying out the possibilities with existing code to see if possible. I want to only add it to the kep if we are likely to do it. We can edit the kep after if we find it easy to do

design/vela-cli/kep-defkit.md
FogDong: I was primarily thinking about code execution safety - running user-provided code internally could pose security risks like code injection. While executing code in the CLI should be relatively safer, we should add security measures to prevent malicious code execution.

anoop2811: Oh yeah thats good side effect of this!!! Thanks for pointing it out. Will update the doc

anoop2811: @FogDong I have pushed a section on the secure benefits. Also please take a look at a new section I added called Definition Placement and also some additinal information for addons.

design/vela-cli/kep-defkit.md
```diff
@@ -0,0 +1,2713 @@
...
+
+## Motivation
+
+### Goals
+
+1. **Author X-Definitions in Go** - Write ComponentDefinition, TraitDefinition, PolicyDefinition, and WorkflowStepDefinition using fluent Go APIs
+2. **Transparent CUE compilation** - Go code compiles to CUE automatically; developers never see or write CUE
+3. **Full IDE support** - Autocomplete, type checking, and inline documentation
+4. **Testable definitions** - Unit test definitions using standard Go testing frameworks
+5. **Schema-agnostic resource construction** - Support any Kubernetes resource without coupling to K8s versions
```
semmet95, line 17: Didn't get this part. Will this agnostic feature be present at the SDK level or CUE level?

anoop2811: @semmet95 : The schema-agnostic feature is at the SDK level (Go).

So, instead of using typed Go structs like appsv1.Deployment{} from k8s.io/api, defkit uses a path-based builder pattern:

  // Schema-agnostic - no import of k8s.io/api/apps/v1
  deploy := defkit.NewResource("apps/v1", "Deployment").
      Set("spec.replicas", replicas).
      Set("spec.template.spec.containers[0].image", image)
  This helps reduce the coupling where can against k8s modules and not be disrupted to upgrade the module versions. If the paths change, the users can update their definitions themselves

anoop2811: Its sort of like the cue definitions right now, there is not compile time validation, the validation of the schema happens at the runtime in the controller. Its the same concept here. Also this can support any custom resource definitions this way.

design/vela-cli/kep-defkit.md
anoop2811: Perhaps not in the kep, but we can possibly do the go to cue compilation in a docker image as well in an enhancement later. This kep is becoming too big, so we can do that later

design/vela-cli/kep-defkit.md
```diff
@@ -0,0 +1,2713 @@
...
+
+// Custom status using deployment helper
+def.CustomStatus(defkit.DeploymentStatus().Build())
+```
+
+### When to Use What
+
+```
+┌─────────────────────────────────────────────────────────────────┐
+│                     DECISION FRAMEWORK                          │
```
jguionnet, line 120: You need to introduce RawCue before this section as it uses RawCue but it was not mentioned before so the reader does not know what you are speaking about

design/vela-cli/kep-defkit.md
```diff
@@ -0,0 +1,2713 @@
+# KEP: Definition Kit (defkit) - Go SDK for X-Definition Authoring
+
+## Summary
+
+Definition Kit (defkit) is a Go SDK that enables platform engineers to author KubeVela X-Definitions using native Go code instead of CUE. The SDK compiles Go to CUE transparently, providing full IDE support while maintaining compatibility with the KubeVela controller.
```
jguionnet, line 5: General comments and Devil advocate
  - You're asking developers to "think declaratively but write imperatively"
  - Go is a compilation target to CUE, not the runtime execution language
  - Developers are still writing declarative specifications (just in Go syntax)
  - The Go code compiles away to CUE, which is the actual runtime engine
  - defkit's API is thick—it abstracts CUE patterns into Go idioms
  - This abstraction is good, but it can hide that "you're really writing declarative code"

design/vela-cli/kep-defkit.md
```diff
@@ -0,0 +1,2713 @@
...
+    cpu := defkit.String("cpu")
+
+    return defkit.NewComponent("webservice").
+        Description("A production-ready web service").
+        Workload("apps/v1", "Deployment").
+        Params(image, replicas, cpu).
+        Template(func(tpl *defkit.Template) {
+            vela := defkit.VelaCtx()
+
+            deploy := defkit.NewResource("apps/v1", "Deployment").
```
jguionnet, line 59: The web service example do not follow the best practices but that would make it more complex as traits are needed. Another alternative is to use other attributes 
- HPA is preferred over replica
- Trait is also preferred for limits

anoop2811: Yes, the idea was to do all those full fledged example in the docs and keep the KEP small enough to relay intent and structure

design/vela-cli/kep-defkit.md
anoop2811: We are calling it out that this will compile to CUE. I dont believe we need to write imperatively here till the point of building the fluent structure. You can use all of the power of go till the point of the definition. After that, since we are not adding support for golang at the controller level/runtime, we will have to keep it leveraged to cue patterns. I did think of support for it , unfortunately it becomes more complex at that point and also brings in security issues we have to consider when running 3rd party code and explore language specific nuances

design/vela-cli/kep-defkit.md
anoop2811: Good catch!

anoop2811: fixed with a forward reference

briankane (approved):

design/vela-cli/kep-defkit.md
```diff
@@ -0,0 +1,2715 @@
...
+
+Runtime context is accessed separately from template context, which focuses on output generation. Health policies and custom status use the built-in helpers:
+
+```go
+// Health check using deployment helper
+def.HealthPolicy(defkit.DeploymentHealth().Build())
+
+// Custom status using deployment helper
+def.CustomStatus(defkit.DeploymentStatus().Build())
+```
```
briankane, line 112: minor: we'll need to account for the new `status.details` field we introduced. Not KEP blocking, but one for implementation.

design/vela-cli/kep-defkit.md
```diff
@@ -0,0 +1,2715 @@
...
+
+### Common Patterns
+
+The SDK provides typed helpers for common definition patterns:
+
+| Pattern | Go API | Example |
+|---------|--------|---------|
+| Required parameter | `defkit.String("image").Required()` | `image` must be provided |
+| Default value | `defkit.Int("replicas").Default(3)` | `replicas` defaults to 3 |
+| Validation | `defkit.Int("replicas").Min(1).Max(100)` | `replicas` between 1-100 |
```
briankane, line 86: Not sure if regex patterns fall under this? Not needed for the KEP but including the pattern checks for strings will be needed for implementation.

design/vela-cli/kep-defkit.md
```diff
@@ -0,0 +1,2715 @@
...
+  # ... definition spec
+```
+
+### Future: Multi-Cluster Integration
+
+This design is forward-compatible with KubeVela's multi-cluster features:
+
+```bash
+# Future: Apply to specific clusters by label selector
+$ vela def apply-module ./aws-definitions --clusters "provider=aws"
```
briankane, line 2532: Not sure this is correct. In a multi cluster setup, the definitions are only deployed to the hub? They aren't required on the spokes.

briankane (approved):

design/vela-cli/kep-defkit.md
anoop2811: Understood, but this could be for the case if a hub is not present and the module is applied directly to clusters

design/vela-cli/kep-defkit.md
anoop2811: Yeah KEP is for the intent, we can add regex and others as we implement and retrofit to kep or better directly to docs

design/vela-cli/kep-defkit.md
anoop2811: Good catch, yes, lets do that as part of the implementation

PR #7021, opened 2026-01-09
Feat: Provider Registry

briankane:

copilot:all

Adds a minimal, interface-based provider registry to break import cycles that block development in the codebase.

## Why
The codebase has entrenched import cycles where full restructuring is impractical. This registry provides a fallback mechanism to unblock feature work while allowing refactoring to be planned appropriately.

## Philosophy
This is a **pragmatic fallback**, not a default pattern. New code should use constructor injection with well-defined package boundaries. The registry is for breaking existing cycles when immediate development is blocked and refactoring is difficult.

Fixes #

I have:

- Unit tests
- Will be utilised in upcoming features

### DIFF
--- a/cmd/core/app/bootstrap.go
+++ b/cmd/core/app/bootstrap.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2025 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


briankane:
> Do you think it makes sense if we move it to pkg repo? Since the providers interface is also inside of pkg.

Actually yeah that makes sense. The bulk of the cycles are in the core code but it wouldn't be any harm for it to be reusable across repos. Will look next week

[Edit] Gonna move forward for this to unblock other features. If the concept it sound and suitable for rolling out further; I'll move it into `pkg` and update vela-core. I don't think we have as many cycle and package boundary issues in the workflow and kube-trigger repos so this might not be as helpful there either.

PR #7048, opened 2026-02-24
Feat: defkit comp def discrepancies

roguepikachu:
copilot:all

This PR extends the defkit framework with several new capabilities needed to achieve CUE generation parity between Go-based component definitions and the original hand-written CUE templates. These are foundational framework features that the consumer repo (vela-go-definitions) depends on to correctly express complex definition patterns found in components like daemon, statefulset, and webservice.

**Parameter Directives**
Add `// +short=X` and `// +ignore` directive support to `StringParam`, `IntParam`, `BoolParam`, and `EnumParam` via new `.Short(s)` and `.Ignore()` fluent methods. CUE generator emits directives in correct order: `// +ignore` before `// +usage=`, `// +short=X` after.

**OneOf Parameter (Discriminated Union)**
Add `OneOfParam.Default(value)` to set a default variant. New `writeOneOfParam` CUE generator renders discriminator enum fields with conditionally-guarded variant blocks (e.g., `if type == "pvc" { claimName: string }`).

**Collection Operations**
- `MapVariant(discriminator, variantName, mappings)` — conditional field mappings within for comprehensions, for the deprecated volumes type-switch pattern
- `ConditionalOrFieldRef` — generates if/else fallback pattern (`if v.name != _|_ { ... } if v.name == _|_ { ... }`) via `FieldRef("name").OrConditional(fallback)`
- `CompoundOptionalField` — dual-condition guard (`if v.nodePort != _|_ if exposeType == "NodePort"`) via `OptionalFieldWithCond(field, cond)`

**Inline Array Values**
`InlineArray(map[string]Value)` — generates `[{containerPort: parameter.port}]` for deprecated single-port fallback patterns.

**Resource Directives**
`Resource.Directive(path, directive)` — attaches CUE directives like `// +patchKey=ip` to specific field paths in the output tree.

**ArrayBuilder Enhancements**
- `ForEachWithGuardedFiltered(guard, filter, source, fn)` — combines outer guard condition + item-level filter predicate with complex `ItemBuilder` per-item logic
- `HelperBuilder.FromArray(ab)` — use a pre-built `ArrayBuilder` as helper source, enabling patterns too complex for the From/Filter/Map pipeline

**CUE Generator Improvements**
- Multi-conditional leaf values (`condValues`) — same path set under different conditions renders as separate `if` blocks instead of overwriting
- Struct decomposition by condition — automatically lifts shared conditions from child nodes to produce cleaner `if cpu { resources: { limits: ..., requests: ... } }` blocks
- Fix: string params with defaults now correctly emit the optional `?` marker
- `CUEFunc.RequiredImports()` + recursive import detection through `PlusExpr` and `ArrayBuilder`/`ItemBuilder` ops

**HealthBuilder Auto-Parenthesization**
`HealthyWhen()` with multiple conditions now auto-wraps each in `()` when joining with `&&`, avoiding manual parenthesis tracking. Already-parenthesized conditions (from `StatusOr`) are not double-wrapped.

Fixes #

I have:

~54 new unit tests across 6 test files covering all new types, methods, and CUE generation paths:
- `param_test.go` — Short/Ignore directives, OneOf default (14 tests)
- `cuegen_test.go` — CUE generation for all new features: directives, OneOf, MapVariant, ConditionalOrFieldRef, CompoundOptionalField, InlineArray, condValues, struct decomposition (19 tests)
- `collections_test.go` — MapVariant, CompoundOptionalField, OrConditional (7 tests)
- `expr_test.go` — InlineArrayValue (3 tests)
- `resource_test.go` — DirectiveOp (3 tests)
- `status_test.go` — Auto-parenthesization, DeploymentHealth, DaemonSetHealth (8 tests)

All existing tests continue to pass (`go test ./pkg/definition/defkit/...`). Consumer-side validation: vela-go-definitions component tests for daemon and statefulset (107 tests) pass using these framework changes, with generated CUE verified against reference templates.

- This PR contains only framework changes (`pkg/definition/defkit/`). The consumer-side changes that use these features (daemon.go, statefulset.go rewrites) live in the vela-go-definitions repo and depend on this PR being merged first.
- All new features were driven by specific discrepancies found when comparing defkit-generated CUE against the original `vela-templates/definitions/internal/component/` CUE files.
- Remaining cosmetic differences (field ordering, `&&` vs chained `if`, column alignment, bracket vs dot notation for `parameter["x"]`) are acceptable and do not affect CUE evaluation semantics.

### DIFF
--- a/pkg/definition/defkit/collections_test.go
+++ b/pkg/definition/defkit/collections_test.go
@@ -1076,10 +1076,10 @@ var _ = Describe("Collections", func() {
 				})
 
 				items := []any{
-					map[string]any{"port": 80, "name": "http", "proto": "HTTP"},   // all primary fields present
-					map[string]any{"port": 443},                                    // name and proto nil → fallback
-					map[string]any{"port": 8080, "name": "", "proto": ""},          // name and proto empty → fallback
-					map[string]any{"port": 9090, "name": "grpc"},                   // name present, proto nil → mixed
+					map[string]any{"port": 80, "name": "http", "proto": "HTTP"}, // all primary fields present
+					map[string]any{"port": 443},                                 // name and proto nil → fallback
+					map[string]any{"port": 8080, "name": "", "proto": ""},       // name and proto empty → fallback
+					map[string]any{"port": 9090, "name": "grpc"},                // name present, proto nil → mixed
 				}
 
 				results := col.Collect(items)


pkg/definition/defkit/collections.go
```diff
@@ -376,6 +415,33 @@ func (m *mapOp) apply(items []any) []any {
...
+// When the iteration variable's discriminator field equals the variant name,
+// the variant's field mappings are included.
+// Generates: if v.discriminator == "variantName" { ...mappings... }
+type mapVariantOp struct {
+	discriminator string
+	variantName   string
+	mappings      FieldMap
+}
+
+func (m *mapVariantOp) apply(items []any) []any {
```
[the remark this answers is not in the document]
roguepikachu: fixed this....good catch

pkg/definition/defkit/collections_test.go
```diff
@@ -877,6 +966,58 @@ var _ = Describe("Collections", func() {
...
+					"displayName": defkit.FieldRef("name").OrConditional(defkit.Format("port-%v", defkit.FieldRef("port"))),
+				})
+
+				items := []any{
+					map[string]any{"port": 443},
+				}
+
+				results := col.Collect(items)
+				Expect(results).To(HaveLen(1))
+				Expect(results[0]["displayName"]).To(Equal("port-443"))
```
anoop2811: Dont you want to make sure we test with multiple items in the list?

anoop2811: Also probably test the whole structure of cue generated instead of just displayName??

pkg/definition/defkit/collections_test.go
```diff
@@ -877,6 +966,58 @@ var _ = Describe("Collections", func() {
 			})
 		})
 
+		Context("OrConditional (ConditionalOrFieldRef)", func() {
+			It("should create ConditionalOrFieldRef from FieldRef", func() {
+				ref := defkit.FieldRef("name").OrConditional(defkit.Format("port-%v", defkit.FieldRef("port")))
+				Expect(ref).NotTo(BeNil())
```
anoop2811: This test seems too shallow to just testing for not nil

pkg/definition/defkit/collections_test.go
roguepikachu: Updated to look for 3 fields (displayName, portNumber, protocol)..All 3 fields are asserted on every item.

pkg/definition/defkit/collections_test.go
roguepikachu: This now tests behaviour by constructing a pipeline and verifies the `ConditionalOrFieldRef` correctly resolves the primary field value

PR #7087, opened 2026-03-30
Docs: Cluster Infrastructure Abstraction — Declarative Lifecycle, Composition, and Fleet Rollouts

anoop2811:
This KEP introduces a layered cluster infrastructure abstraction for KubeVela, enabling platform teams to declaratively manage cluster lifecycle, infrastructure composition, and fleet-wide rollouts.

  - Adds a Cluster CRD with three modes (provision, adopt, connect) that delegates actual provisioning to pluggable backends like Crossplane, Terraform, or CAPI.
  - Infrastructure is modeled as composable ClusterPlane layers (networking, security, observability) that are versioned independently and assembled into ClusterBlueprint templates representing golden-path cluster
   configurations.
  - Shared planes with consumer tracking solve the multi-cluster shared infrastructure problem (e.g., a single VPC consumed by multiple EKS clusters via preCreate lifecycle phases).
  - Fleet rollouts are handled by ClusterRolloutStrategy with wave-based progression, maintenance windows, approval gates, and automatic rollback.
  - Rather than introducing new definition CRDs, existing KubeVela definitions (ComponentDefinition, TraitDefinition, etc.) are extended with a scope: cluster annotation.
  - Includes hierarchical health aggregation, drift detection against blueprint specs, and a decommissioning workflow.
  - All CRDs follow a pull-based reconciliation model where templates are only materialized when referenced by a Cluster, keeping the hub lightweight.


design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+- [API Reference](#api-reference)
+- [Implementation Plan](#implementation-plan)
+
+---
+
+## Introduction
+
+### The Problem
+
+KubeVela's `Application` CRD excels at deploying workloads, but it assumes clusters are already provisioned and configured. In practice, platform teams must manage a complex stack of infrastructure _beneath_ the application layer — networking (CNI, ingress, service mesh), security (OPA, cert-manager, network policies), observability (Prometheus, Grafana, logging), and the clusters themselves (VPC, EKS/GKE, node pools). Today, this is stitched together with a mix of Terraform, Helm charts, GitOps tools, and custom scripts, each solving one piece without a unifying model for:
```
jguionnet, line 82: OPA, Observability can be installed with add-on

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+| Wraps ASO resources    | ASO              | Call Azure APIs               |
+
+### The Approach
+
+This KEP extends OAM's abstraction model to cluster infrastructure by introducing a layered composition model:
+
+- A **Cluster** represents a managed Kubernetes cluster with its full lifecycle — provisioning, connecting, or adopting existing clusters. Each Cluster goes through optional lifecycle phases: shared cloud infrastructure can be set up before the cluster is created (`preCreate`), the cluster itself is provisioned and configured via a blueprint (`blueprintRef`), and post-creation tasks like deploying validation apps or running smoke tests can be triggered after the cluster is ready (`postCreate`).
+
+- A **ClusterPlane** is a composable infrastructure layer owned by a team — for example, a networking plane (Cilium, CoreDNS, ingress), a security plane (OPA, cert-manager), or an observability plane (Prometheus, Grafana). Each plane has its own components, versioning, health checks, and outputs. Planes can be scoped as `shared` (created once, consumed by many clusters) or `perCluster` (created for each cluster).
```
jguionnet, line 128: Shared for networking seems a stretch to me

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+
+- A **ClusterBlueprint** composes multiple ClusterPlanes into a complete cluster specification. It defines which planes a cluster needs, their dependency ordering, and version constraints. Blueprints are immutable once published — changes create new revisions.
```
jguionnet, line 130: I assumne then that a ClusterBlueprint will indicate which CP is shared vs perCluster

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+
+- A **ClusterRolloutStrategy** defines how blueprint changes propagate across a fleet of clusters — wave-based progression, maintenance windows, health gates, and automatic rollback on SLO breach.
```
jguionnet, line 132: ClusterRolloutStrategy for shared place is indepandant from the host cluster I assume and probaly happen after, requiring it to support multiple version

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+| **`PolicyDefinition`**       | Defines policy types for ClusterPlanes (e.g., `apply-order`, `health-check`)                     |
+| **`WorkflowStepDefinition`** | Defines workflow steps for cluster lifecycle (e.g., `apply-plane`, `validate-plane`, `suspend`)  |
+
+**Runtime CRDs:**
+
+| CRD                         | Description                                                       |
+| --------------------------- | ----------------------------------------------------------------- |
+| **`ObservabilityProvider`** | Instance of an observability provider with connection details     |
+| **`ClusterDriftReport`**    | Report of detected drift between desired and actual cluster state |
+| **`ClusterDriftException`** | Allowlist for expected drift that should not trigger alerts       |
```
jguionnet, line 176: ClusterDriftException: neither alert nor reconciliation, ie setup only at creation time? For slave clusters whose some properties are controlled by a "parent",  for the likes of vcluster

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+| ------------------------ | ------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- |
+| **Hub Cluster**          | A Kubernetes cluster that serves as the management/control plane                      | Platform team bootstraps using Terraform, eksctl, cloud console, CAPI, or any other tool |
+| **KubeVela Installed**   | vela-core and cluster-gateway installed on the hub cluster                            | `vela install` or Helm chart                                                             |
+| **Cloud Credentials**    | Credentials for cloud providers where spoke clusters will be provisioned or connected | Stored as Secrets in hub cluster                                                         |
+| **Network Connectivity** | Hub cluster must be able to reach spoke cluster API servers                           | VPC peering, transit gateway, or public endpoints                                        |
```
jguionnet, line 191: reach spoke cluster API servers  -> reach spoke cluster nodes. The hub is responsible to install K8s cluster on node and only after will have access to cluster API Server

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+| Deployed infrastructure                        | Spoke clusters                                                        | CNI, ingress, cert-manager, security policies, etc.                                                                                                                                                                                           |
+| Workload Applications                          | Spoke clusters                                                        | Your actual applications (managed via KubeVela `Application` CRD)                                                                                                                                                                             |
+
+> **The hub cluster is the single pane of glass** for managing your entire fleet. All declarative state (desired state) lives on the hub; spoke clusters receive the rendered infrastructure components.
```
jguionnet: Should not the spoke render its infra based on the definition sent by the hub?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+        type: deploy-plane
+        properties:
+          plane: vela-runtime
+      - name: deploy-vela-definitions
+        type: deploy-plane
+        properties:
+          plane: vela-definitions
+        dependsOn:
+          - deploy-vela-runtime # Definitions require vela-core to be running
+      - name: deploy-networking
```
jguionnet, line 429: Up to here networking = base networking like cilium, netpol, ingress, so I would epxect it to be installed before KV

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+        type: deploy-plane
+        properties:
+          plane: networking
+        dependsOn:
+          - deploy-vela-definitions # Infrastructure planes can depend on runtime
+```
+
+> **Note**: Spoke clusters that only receive raw Kubernetes resources (Deployments, Services, ConfigMaps) dispatched from the hub do NOT need `vela-runtime`. The hub's vela-core renders the resources and applies them directly via cluster-gateway. The `vela-runtime` plane is only needed when spokes independently process `Application` CRDs.
+
+### Definition Distribution to Spoke Clusters
```
jguionnet, line 439: Stopped here

design/vela-core/kep-cluster-infrastructure.md
anoop2811: The installation could be with addon or helm etc whichever a workflow can support, this kep is more about the orchestration and rollout across fleets and upgrades

design/vela-core/kep-cluster-infrastructure.md
anoop2811: So networking (CNI, ingress, DNS) is always perCluster. The shared scope applies to cloud infrastructure like a VPC, Transit Gateway, or shared IAM roles that multiple clusters consume. I'll update the ClusterPlane description to make the examples clearer.

design/vela-core/kep-cluster-infrastructure.md
anoop2811: The scope is defined on the ClusterPlane itself (spec.scope: shared | perCluster), not on the ClusterBlueprint. The Blueprint simply composes planes — it doesn't override their scope. This means the platform
 team that owns the plane decides whether it's shared or per-cluster, and blueprint composers just reference it. The later sections (Shared Infrastructure Planes) explain this in detail, but I can add a brief
 clarification here in the intro if that is not clear enough.

design/vela-core/kep-cluster-infrastructure.md
anoop2811: Correct — shared plane rollouts are independent. A shared plane (e.g., VPC) is reconciled once, and its outputs are consumed by all clusters referencing it. When a shared plane is updated.
For version compatibility: ClusterBlueprint pins plane versions via ref.revision or ref.versionConstraint, so different clusters can reference different revisions of the same shared plane. The Blueprint's dependsOn ordering ensures shared planes complete before per-cluster planes start.
 Good catch on the multi-version support — this is handled by the revision system (ClusterPlaneRevision), where each published version is an immutable snapshot.

design/vela-core/kep-cluster-infrastructure.md
anoop2811: Not sure I understand 100%, but assuming what you are saying is that for the spoke if they are host clusters for like vclusters what would the properties that change be handled. The assumption I am making is that the vcluster is a separate Cluster specification and we can set the driftexception in case there are properties for those clusters too.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+
+### VelaRuntime ClusterPlane
+
+If spoke clusters need to run KubeVela `Application` workloads (not just raw Kubernetes resources managed from the hub), they require the KubeVela runtime: `vela-core` controller and the relevant X-Definitions (`ComponentDefinition`, `TraitDefinition`, etc.).
+
+Rather than requiring manual installation, this is expressed declaratively as a **ClusterPlane component** using the existing `helm` ComponentDefinition:
+
+```yaml
+apiVersion: core.oam.dev/v1beta1
+kind: ClusterPlane
```
briankane, line 370: So I think there is a little bit of overreach here. 
ClusterPlanes look to be, by and large, a recreation of Applications.

This seems like the logical intersection point between current vs. proposed. Cluster + ClusterBlueprints (and their controllers) allow for fleet-wide deployment of prescribed KubeVela Applications across the hub & spoke landscape. 

The templating concept of ClusterPlanes is quite powerful even outside of this KEP. We've seen demos of users building Application blueprints / templates on top of KubeVela as complexity rises. I'd suggest we could extrapolate some of that need into core KubeVela functionality for an ApplicationDefinition/Template (which we've already been exploring) and then use that as references within the ClusterBlueprint. Under the hood, the Cluster system is reusing as much of KubeVela's bread & butter Application provisioning to build out it's infra. The Cluster controller can manage the abstraction of certain aspects of the Application/s. Within Cluster CRDs, we can still reference to ClusterPlanes as it's a sound logical concept but we offload the heavy work to the application controller; reducing the overhead of building / maintaining.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+metadata:
+  name: vela-runtime
+  namespace: vela-system
+spec:
+  description: "KubeVela application runtime for spoke clusters"
+  components:
+    - name: vela-core
+      type: helm
```
briankane, line 378: Minor but flagging it; Helm component is experimental and there are some known limitations with hooks etc. 

The expectation is it would be stable in advance of this work but worth a call out.

design/vela-core/kep-cluster-infrastructure.md
briankane: In current KubeVela multi-cluster Applications the hub does the rendering and dispatches output/s to the spokes. So this is consistent with the current workings

For the spoke to do the rendering would require KubeVela being installed to the spokes; which isn't currently a requirement. A KV hub can render and dispatch to a hub without a KV controller running. 

We have been having initial (mostly informal) conversations around whether a hub/agent architecture would be more scalable / secure etc. but for right now this makes sense and cluster-gateway would be the dispatcher for rendered resources.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+A blueprint for spoke clusters that run Application workloads includes this plane:
+
+```yaml
+apiVersion: core.oam.dev/v1beta1
+kind: ClusterBlueprint
+metadata:
+  name: production-standard
+spec:
+  planes:
+    - name: vela-runtime # Install KubeVela runtime first
```
briankane, line 404: This is installing KubeVela to the spoke cluster, presumably?

Is this mandatory or just an example of a setup where KubeVela runs on hub & spoke?

[Edit] Never mind - can see comment further down that this is an example for setups running KV on both.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+          - deploy-vela-definitions # Infrastructure planes can depend on runtime
+```
+
+> **Note**: Spoke clusters that only receive raw Kubernetes resources (Deployments, Services, ConfigMaps) dispatched from the hub do NOT need `vela-runtime`. The hub's vela-core renders the resources and applies them directly via cluster-gateway. The `vela-runtime` plane is only needed when spokes independently process `Application` CRDs.
+
+### Definition Distribution to Spoke Clusters
+
+When spoke clusters run their own vela-core (via the `vela-runtime` plane), they also need the X-Definitions (`ComponentDefinition`, `TraitDefinition`, etc.) that their Applications reference. There are two approaches:
+
+**Approach 1: Definitions as a ClusterPlane**
```
briankane, line 443: The more KubeVela-native way would be to package the definitions as an Addon and install the addon to the spokes.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+    targetNamespace: vela-system
+```
+
+---
+
+## Background
+
+### The Problem with Application-Centric Only
+
+KubeVela's `Application` CRD is designed for workload deployment. It assumes:
```
briankane, line 495: To play Devil's Advocate here - it is _possible_ to provision a cluster via an Application/s. 

But the abstraction isn't completely suitable - especially for managing fleets at scale.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+│  │   data:                                                         │   │
+│  │     endpoint: <base64>                                          │   │
+│  │     ca.crt: <base64>                                            │   │
+│  │     tls.crt: <base64>                                           │   │
+│  │     tls.key: <base64>                                           │   │
+│  └─────────────────────────────────────────────────────────────────┘   │
+│                              │                                         │
+│                              ▼                                         │
+│  ┌─────────────────────────────────────────────────────────────────┐   │
+│  │ VirtualCluster (cluster-gateway CRD)                            │   │
```
briankane, line 583: Just a note that VirtualCluster isn't a full fledged CRD afaik; just an API resource that delegates to labelled secrets. Limitations still all stand and are pertinent

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+
+**Key Architectural Decision:** The Cluster CRD manages connectivity **directly**. This ensures the core CRD is self-sufficient.
+
+```
+┌────────────────────────────────────────────────────────────────────────┐
+│           PROPOSED: Cluster CRD (Self-Sufficient Connectivity)         │
+├────────────────────────────────────────────────────────────────────────┤
+│                                                                        │
+│  ┌─────────────────────────────────────────────────────────────────┐   │
+│  │ Cluster CRD (core.oam.dev/v1beta1)                              │   │
```
briankane: I assume this would also hold the basics of what is currently in the Secrets? Labels etc.?

It sounds like we could implement a very basic version of this near-immediately, migrate the existing virtual clusters over and then iterate in the larger changes?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+│                         CORRECT UPDATE FLOW                                 │
+├─────────────────────────────────────────────────────────────────────────────┤
+│                                                                             │
+│  1. USER/GITOPS updates Cluster.spec.blueprintRef to new version            │
+│     │                                                                       │
+│     ▼                                                                       │
+│  2. ClusterController detects spec.blueprintRef != status.blueprint         │
+│     │                                                                       │
+│     ▼                                                                       │
+│  3. ClusterController queries ClusterRolloutController: "Can I update?"     │
```
briankane: Is there an intent issue here? Updating the version feels like it should reconcile immediately to fulfil it's contract. Are we missing a Fleet-level CRD for these purposes? Should we be updating a ClusterWave/FleetSet version which then owns the rollout strategy and versioning across many Clusters; and letting that propagate the version down to it's individual clusters?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+
+**Key ownership boundaries:**
+
+- `ClusterController` READS `spec.blueprintRef`, WRITES `status.blueprint`
+- `ClusterController` NEVER modifies `spec` or `ClusterBlueprint`
+- `ClusterRolloutController` gates timing via status fields
+
+#### Controller Responsibilities Matrix
+
+**Critical Design Principle:** ClusterPlane is ONLY reconciled when referenced by a Cluster. Creating a ClusterPlane CRD does NOT create infrastructure resources—the Cluster CRD is the reconciliation trigger.
```
briankane: > reconciled when referenced by a Cluster

This is referring to instantiating from a template? We don't actually reconcile the ClusterPlane; it's just acting as a definition?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+    provider: aws
+    tier: standard
+spec:
+  # Cluster access configuration
+  credential:
+    # Reference to kubeconfig secret
+    secretRef:
+      name: production-us-east-1-kubeconfig
+      namespace: vela-system
+    # Or inline (not recommended for production)
```
briankane: This is gonna through either `vela cluster join` which is gonna store into secrets, or via GitOps where you definitely aren't gonna clear text the creds. I don't think there is any value in allowing inline vs. ensuring security

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+    # kubeconfig: |
+    #   apiVersion: v1
+    #   kind: Config
+    #   ...
+
+  # Which blueprint this cluster should follow
+  blueprintRef:
+    name: production-standard
+    # Optional: pin to specific revision (otherwise uses latest)
+    revision: production-standard-v2.3.0
```
briankane, line 795: For consistency in KubeVela, this should probably be version. Revisions are auto-created incrementing snapshots, versions are persistent semvers

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+  patches:
+    # Override plane component properties for this cluster
+    - plane: networking
+      component: ingress-nginx
+      properties:
+        values:
+          controller:
+            replicaCount: 5 # This cluster needs more replicas
+
+  # Cluster metadata (synced from actual cluster)
```
briankane: What do we mean by synced from actual cluster here?

> These are auto-discovered

From where??

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+      # Stricter analysis thresholds for this cluster
+      analysis:
+        metrics:
+          - name: error-rate
+            thresholds:
+              - condition: "< 0.5%" # Stricter than strategy default
+      # Skip certain waves (useful for canary clusters)
+      # skipWaves: [non-critical, critical]
+
+  # Maintenance windows for this cluster
```
briankane: Going back to the fleet-level comment above; this feels like something you'd wanna manage at a higher level? We don't want discrepancies in a particular prod cluster etc.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+    # Optional: cluster-specific overrides
+    overrides:
+      # Stricter analysis thresholds for this cluster
+      analysis:
+        metrics:
+          - name: error-rate
+            thresholds:
+              - condition: "< 0.5%" # Stricter than strategy default
+      # Skip certain waves (useful for canary clusters)
+      # skipWaves: [non-critical, critical]
```
briankane: Things like this feel particularly difficult when waves are a virtual concept that live on each cluster

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+  # Cluster metadata (synced from actual cluster)
+  clusterInfo:
+    # These are auto-discovered but can be overridden
+    displayName: "Production US East 1"
+    description: "Primary production cluster for US East region"
+    contactEmail: "platform-us-east@example.com"
+
+  # Rollout strategy reference - determines how this cluster receives updates
+  # The cluster's labels determine which wave it belongs to
+  rolloutStrategyRef:
```
briankane, line 807: Would these be more flexible as workflows on an entity? Allowing users to publish their own checks etc. as workflow steps and handle orchestration concerns configurably?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+        schedule: "0 2 * * *"
+        retention: "30d"
+
+status:
+  # Connection status
+  connectionStatus: Connected # Connected, Disconnected, Unknown
+  lastProbeTime: "2024-12-24T10:00:00Z"
+  latency: "45ms"
+
+  # Cluster information (auto-discovered)
```
briankane, line 884: How will the auto discovery work?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+  #         image: cilium/cilium-cli:latest
+  #         command: ["cilium", "status", "--wait"]
+  #     - name: deploy-rest
+  #       type: deploy
+  #       dependsOn: [validate-cni]
+  #       properties:
+  #         components: [ingress-nginx, external-dns]
+
+  # Outputs exposed to other planes or blueprints
+  outputs:
```
briankane, line 1254: We've been drafting some concepts of component-level `exports` that would roll up to Application level. Might be useful here also if that goes ahead

briankane (commented):
Publishing comments so far but will continue with review tomorrow

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+kind: ClusterBlueprint
+metadata:
+  name: production-standard
+spec:
+  planes:
+    # Option 1: Pin to exact revision (recommended for production)
+    - name: networking
+      ref:
+        name: networking
+        revision: networking-v2.3.1 # Explicit revision
```
briankane, line 1628: The revision would just increment? It wouldn't have semver awareness?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+        name: storage
+        versionConstraint: ">=1.0.0 <2.0.0" # Any 1.x version
+```
+
+**CLI Commands for Versioning:**
+
+```bash
+# Publish a new version (sets publishVersion annotation)
+# This is equivalent to kubectl apply with plane.oam.dev/publishVersion annotation
+$ vela plane publish networking --version 2.4.0 --changelog "Added Gateway API support"
```
briankane, line 1654: Minor but I think `vela cluster` should be the top level command

`vela cluster plane ...` etc.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+**Scope Types:**
+
+| Scope        | Behavior                                                           | Use Case                                            |
+| ------------ | ------------------------------------------------------------------ | --------------------------------------------------- |
+| `perCluster` | Resources created for each cluster using the blueprint (default)   | EKS clusters, node groups, cluster-specific IAM     |
+| `shared`     | Resources created once, outputs available to all clusters in scope | VPCs, NAT Gateways, shared subnets, Transit Gateway |
+
+##### Shared Plane Ownership Model
+
+**The Critical Question:** Who "owns" a shared plane's resources? Since ClusterPlanes are only reconciled when referenced by a Cluster, we need clear ownership semantics.
```
briankane: Note: This is similar to how the shared-resource policy currently works on Applications

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+apiVersion: core.oam.dev/v1beta1
+kind: ClusterPlane
+metadata:
+  name: shared-vpc-us-east-1
+  namespace: vela-system
+spec:
+  scope: shared # Created once, not per-cluster
+
+  # Which clusters can consume this plane's outputs
+  sharedWith:
```
briankane, line 2164: This feels like another fleet-level concern leaking into the cluster. As fleet size grows, there is a growing maintenance burden to keeping all these access grants in sync

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+      clusterSelector:
+        matchLabels:
+          tier: staging
+      # Only proceed if previous wave succeeded
+      waitFor:
+        wave: canary
+        # How long canary must be healthy before proceeding
+        healthyDuration: "4h"
+        # What health checks must pass
+        healthChecks:
```
briankane, line 3472: What is this referencing? What are the type references here?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+
+- **Shared cloud infrastructure**: The `preCreate` blueprint runs on the hub against cloud APIs. Components use types like `terraform-module` or `crossplane-resource`. The spoke cluster does not yet exist, so no cluster-gateway connectivity is needed. If the same preCreate blueprint is referenced by another Cluster, its outputs are consumed without re-creation (shared plane semantics).
+- **Cluster provisioning & infrastructure**: The `blueprintRef` blueprint handles both cluster creation (against cloud APIs) and all cluster-level infrastructure (networking, security, observability planes). Provisioning outputs (cluster endpoint, CA cert, credentials) are used by the ClusterController to create a credential Secret, which registers the new cluster with cluster-gateway. Once connectivity is established, infrastructure planes (Cilium, cert-manager, ingress, monitoring) are dispatched to the spoke via cluster-gateway. This is the golden path — all cluster infrastructure is defined in the blueprint.
```
briankane: Should the Cloud API pieces not all be in the preCreate part?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+kind: Cluster
+metadata:
+  name: production-us-east-1
+  namespace: vela-system
+spec:
+  # MODE: Provision new cluster
+  mode: provision
+
+  # Cloud provider configuration - just credentials and region
+  provider:
```
briankane, line 3935: Is provisioning not being delegated to Crossplane / KRO / Terraform etc.?

What is this provider block required for? Credentials etc. should be a concern of the cluster components and underlying compositions?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+    #       values:
+    #         cidr: "10.100.0.0/16"  # Different CIDR for this cluster
+
+status:
+  mode: provision
+  phase: Provisioning # Pending, Provisioning, Ready, Failed
+
+  # Connection info (populated by aws-foundation plane outputs)
+  connection:
+    endpoint: "" # Populated when EKS is ready
```
briankane, line 3964: Again, how do we infer these details if provisioning is offloaded?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+
+When a Cluster is deleted, phases are cleaned up in reverse order:
+
+1. **postCreate** resources removed from spoke cluster
+2. **blueprintRef** resources cleaned up (cluster deprovisioned if `mode: provision`)
+3. **preCreate** consumer count decremented — shared resources remain if other consumers exist
+
+#### ClusterProviderDefinition
+
+To support multiple cloud providers, we introduce `ClusterProviderDefinition`:
```
briankane, line 4355: Is this not opposed to this statement at the top?

> **Positioning:**
>
> 1. KubeVela does **not** provision clusters directly
> 2. KubeVela orchestrates **composition + ordering + rollout + ownership** across clusters and infrastructure layers
> 3. Provisioning is delegated to CAPI / Crossplane / Terraform / KRO / cloud-native operators (pluggable)

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+
+```yaml
+apiVersion: core.oam.dev/v1beta1
+kind: ClusterProviderDefinition
+metadata:
+  name: aws-eks
+  namespace: vela-system
+spec:
```
briankane, line 4363: Can this use cue based schemas with parameter:{} etc.

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+        description: "AWS Secret Access Key"
+    optional:
+      - name: sessionToken
+        description: "AWS Session Token (for temporary credentials)"
+      - name: roleArn
+        description: "IAM Role ARN to assume"
+
+  # Provisioning schematic
+  schematic:
+    # Option 1: Use Crossplane
```
briankane, line 4384: Should the implementation not just be abstracted through components that offload to XP Compositions or TF modules etc.?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+  --region us-east-1 \
+  --blueprint production-standard
+
+# With options
+vela cluster create production-us-east-1 \
+  --provider aws \
+  --credentials aws-platform-creds \
+  --region us-east-1 \
+  --kubernetes-version 1.28 \
+  --node-type m5.xlarge \
```
briankane, line 4504: Will this work across all cloud platforms?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+        previousWaveHealthy: true
+```
+
+---
+
+### Definition Types
+
+#### Definition Scope Model
+
+The existing KubeVela definition CRDs are extended with a **scope annotation** to distinguish infrastructure definitions from application definitions. The definition system remains a single, unified type system — the scope annotation partitions it by usage context.
```
briankane, line 5077: Note: Some definitions already have `scope` present but I don't think it's enforced

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+          }]
+        }
+```
+
+#### Infrastructure WorkflowStepDefinition
+
+Defines workflow steps for cluster/plane operations. Uses the existing `WorkflowStepDefinition` CRD with `scope: cluster`.
+
+```yaml
+apiVersion: core.o
```
briankane, line 5603: This looks like it could be a ConfigTemplate rather than a full CRD

briankane, line 5642: How do these Providers know how to retrieve data from their respective APIs? It seems to only deal with connecting. Should these be cuex specified to encapsulate logic?

briankane, line 6207: Could this be generated on demand with the CLI vs. needing a CRD?

design/vela-core/kep-cluster-infrastructure.md
```diff
@@ -0,0 +1,8106 @@
...
+│           └────────────┴────────────┴────────────┴────────────┘             │
+│                                                                             │
+└─────────────────────────────────────────────────────────────────────────────┘
+```
+
+---
+
+### Relationship to Existing Multi-Cluster Architecture
+
+KubeVela currently uses **cluster-gateway** (`github.com/oam-dev/cluster-gateway`) for multi-cluster connectivity. It's important to understand how the proposed `Cluster` CRD relates to the existing architecture:
```
briankane, line 560: Note: cluster-gateway requires static credentials currently. It's not part of this but improving the security around that is an important pre-requisite to wider adoption.

design/vela-core/kep-cluster-infrastructure.md
anoop2811: Do you mean kubelet installations? or provisioning somethng like node groups?

design/vela-core/kep-cluster-infrastructure.md
anoop2811: Yes, you mean something like `ApplicationTemplates`. We should look at that, but I would not include it in scope for this implementation, but I think that merits a separate KEP in its own after we have reached some amount of implementation of the ClusterPlanes.

design/vela-core/kep-cluster-infrastructure.md
anoop2811: Yes that was the plan :). We can retrofic the specifics , in any case I dont think this entire KEP is going to be implemented in a hurry ;)

design/vela-core/kep-cluster-infrastructure.md
anoop2811: I think we should keep it open as a way to install. some teams may not want to have another abstraction of an Addon and may just want to install helm. But the abstraction should be able to handle these specific implementation through extensions

design/vela-core/kep-cluster-infrastructure.md
anoop2811: which comment are you mentioning? Yes, they could be managed and knob turned by the team managing it. The KEP allows any of the uses, we just provide the way to do it, but it would be left to the tolerances of the SRE team involved

design/vela-core/kep-cluster-infrastructure.md
anoop2811: This means cluster as a custom resource on the hub, not the actual k8s cluster

design/vela-core/kep-cluster-infrastructure.md
anoop2811: Awesome, could you raise a PR on this once this is merged?

design/vela-core/kep-cluster-infrastructure.md
anoop2811: The thing is the plane does have an independent existence, but no lifecycle without cluster... that said, not something I am going to fight, so will update

design/vela-core/kep-cluster-infrastructure.md
anoop2811: So what I was thinking is that this goes into a blueprint that would be used by multiple cluster provisioning during the preCreate phase. So this wont be a per cluster creation because if the resources exist, just uses that and we would create  it if not present

briankane (approved):

PR #7089, opened 2026-03-31
Feat: support per-application reconciliation interval override via annotation

officialasishkumar:

Add the `app.oam.dev/reconcile-interval` annotation that lets operators override the global `ApplicationReSyncPeriod` on a per-application basis.

Different applications have different drift-detection needs. A production app handling customer traffic probably wants reconciliation every minute or so. A dev/staging app can get away with 15m and save API server load. Right now the global `--application-re-sync-period` flag forces you to pick one number for everything.

**How it works:**
- When set, the controller uses the annotation value as `RequeueAfter` instead of the global default
- When absent, nothing changes — full backward compatibility
- Valid Go duration strings accepted (e.g. `"1m"`, `"15m"`, `"30s"`)
- Minimum floor of 10s to prevent accidental API server overload
- Invalid or below-minimum values log a warning and fall back to the global default
- When `ApplyOnce` sets the global interval to 10h, this annotation still takes precedence for opted-in apps

**Usage:**
```yaml
kind: Application
metadata:
  name: prod-checkout
  annotations:
    app.oam.dev/reconcile-interval: "1m"
spec:
  components:
    - name: checkout
      type: webservice
```

**Implementation details:**
- Added `AnnotationReconcileInterval` constant to `pkg/oam/labels.go` following existing annotation patterns
- Added a `forApp()` chain method on `reconcileResult` so only the return paths where the default resync period matters need annotation, leaving all explicit `requeue()` calls untouched
- Applied `.forApp(app)` to the normal completion path in `gcResourceTrackers` and to `endWithNegativeCondition`

Fixes #7076

I have:

- 15 unit tests covering all edge cases: valid durations, invalid values, below-minimum floor, negative durations, zero values, nil app, explicit requeue precedence, error paths, and annotation-absent fallback
- All existing application controller tests continue to pass
- Compilation verified across the full module

The `forApp()` method is intentionally opt-in — existing `r.result(err).ret()` call sites that don't chain `forApp()` continue using the global default. This minimizes the diff and makes it easy to extend to additional return paths if needed.

### DIFF
--- a/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
@@ -75,6 +75,10 @@ const (
 const (
 	// baseWorkflowBackoffWaitTime is the time to wait gc check
 	baseGCBackoffWaitTime = 3000 * time.Millisecond
+
+	// minPerAppResyncPeriod is the minimum reconciliation interval that can be
+	// set via the per-application annotation to prevent excessive API server load.
+	minPerAppResyncPeriod = 10 * time.Second
 )
 
 var (
@@ -408,7 +412,7 @@ func (r *Reconciler) gcResourceTrackers(logCtx monitorContext.Context, handler *
 		cond := condition.Deleting()
 		cond.Message = fmt.Sprintf("error encountered during garbage collection: %s", err.Error())
 		handler.app.Status.SetConditions(cond)
-		return r.result(statusUpdater(logCtx, handler.app, phase)).ret()
+		return r.result(statusUpdater(logCtx, handler.app, phase)).forApp(handler.app).ret()
 	}
 	if !finished {
 		logCtx.Info("GarbageCollecting resourcetrackers unfinished")
@@ -420,16 +424,33 @@ func (r *Reconciler) gcResourceTrackers(logCtx monitorContext.Context, handler *
 		return r.result(statusUpdater(logCtx, handler.app, phase)).requeue(baseGCBackoffWaitTime).ret()
 	}
 	logCtx.Info("GarbageCollected resourcetrackers")
-	return r.result(statusUpdater(logCtx, handler.app, phase)).ret()
+	return r.result(statusUpdater(logCtx, handler.app, phase)).forApp(handler.app).ret()
 }
 
 type reconcileResult struct {
 	time.Duration
-	err error
+	err           error
+	defaultResync time.Duration
 }
 
 func (r *reconcileResult) requeue(d time.Duration) *reconcileResult {
 	r.Duration = d
+	return r
+}
+
+// forApp overrides the default resync period with the per-application value
+// parsed from the AnnotationReconcileInterval annotation, if present and valid.
+func (r *reconcileResult) forApp(app *v1beta1.Application) *reconcileResult {
+	if app != nil && app.Annotations != nil {
+		if v, ok := app.Annotations[oam.AnnotationReconcileInterval]; ok {
+			if d, err := time.ParseDuration(v); err == nil && d >= minPerAppResyncPeriod {
+				r.defaultResync = d
+			} else {
+				klog.Warningf("ignoring invalid %s annotation %q on application %s/%s, using global default",
+					oam.AnnotationReconcileInterval, v, app.Namespace, app.Name)
+			}
+		}
+	}
 	return r
 }
 
@@ -439,7 +460,7 @@ func (r *reconcileResult) ret() (ctrl.Result, error) {
 	} else if r.err != nil {
 		return ctrl.Result{}, r.err
 	}
-	return ctrl.Result{RequeueAfter: common2.ApplicationReSyncPeriod}, nil
+	return ctrl.Result{RequeueAfter: r.defaultResync}, nil
 }
 
 func (r *reconcileResult) end(endReconcile bool) (bool, ctrl.Result, error) {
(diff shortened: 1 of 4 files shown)


roguepikachu:
Hi @officialasishkumar could you add some e2e test cases??

officialasishkumar:
Added a focused e2e case for the per-application reconcile interval override. It sets `app.oam.dev/reconcile-interval: 10s`, mutates the managed workload without forcing an app reconcile, and waits for scheduled state-keep repair.

Checks run:
- `go test -p=2 ./pkg/controller/core.oam.dev/v1beta1/application -run 'TestReconcileResult' -count=1`
- `go test -p=2 ./test/e2e-test -run '^$'`

roguepikachu:
Could you take a look at the failing E2E tests? If not related to your changes, you could re-run it by pushing an empty commit.

officialasishkumar:
Reran CI with empty commit 34f352d; the prior failed checks are passing now, with `e2e-tests (v1.31)` still pending.

roguepikachu:
The godoc on `AnnotationReconcileInterval` at `pkg/oam/labels.go:227` says invalid values are "silently ignored", but `forApp` logs a warning for them. The PR description mentions the warning behaviour, so this looks like a doc oversight. Just drop "silently":

```go
// global default.  Invalid values are logged as a warning and ignored.
```

roguepikachu:
The two `.forApp(app)` chains in `endWithNegativeCondition` (lines 525, 527) don't do anything. Both returns always have a non-nil `err`, so `ret()` exits at the `else if r.err != nil` branch and `defaultResync` is never read. The `forApp` call only ends up emitting a warning on a path where the requeue cadence isn't honoured anyway. Cleaner to drop them:

```go
func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.Application, condition condition.Condition, phase common.ApplicationPhase) (ctrl.Result, error) {
    app.SetConditions(condition)
    if err := r.patchStatus(ctx, app, phase); err != nil {
        return r.result(errors.WithMessage(err, "cannot update application status")).ret()
    }
    return r.result(fmt.Errorf("object level reconcile error, type: %q, msg: %q", string(condition.Type), condition.Message)).ret()
}
```

roguepikachu:
Setting `app.oam.dev/reconcile-interval: "5s"` produces `ignoring invalid ... annotation "5s" on application ...`, but `"5s"` is a perfectly valid Go duration. It's just below the 10s floor. Calling it "invalid" sends operators looking for a parse error that isn't there. Splitting the two failure modes makes the cause obvious in the log:

```go
func (r *reconcileResult) forApp(app *v1beta1.Application) *reconcileResult {
    if app == nil || app.Annotations == nil {
        return r
    }
    v, ok := app.Annotations[oam.AnnotationReconcileInterval]
    if !ok {
        return r
    }
    d, err := time.ParseDuration(v)
    if err != nil {
        klog.Warningf("ignoring %s annotation %q on application %s/%s: parse error: %v; using global default",
            oam.AnnotationReconcileInterval, v, app.Namespace, app.Name, err)
        return r
    }
    if d < minPerAppResyncPeriod {
        klog.Warningf("ignoring %s annotation %q on application %s/%s: below minimum (%s); using global default",
            oam.AnnotationReconcileInterval, v, app.Namespace, app.Name, minPerAppResyncPeriod)
        return r
    }
    r.defaultResync = d
    return r
}
```

Existing unit tests still pass — none of them check the warning text.

officialasishkumar:
Addressed the three latest review items in f7a8b8103: removed the stale "silently" wording, dropped the no-op `.forApp(app)` calls on error returns, and split parse errors from below-minimum interval warnings. `go test ./pkg/controller/core.oam.dev/v1beta1/application -run TestReconcileResult` passed.

roguepikachu:
Could you sign the DCO and address the linting failures.

officialasishkumar:
done

roguepikachu:
LGTM. Set up a local k3d cluster running this branch with `--application-re-sync-period=1m`, then deployed 15 Applications covering the annotation cases: six below the global default, four above it, one with no annotation as the control, and four invalid cases (below the 10s floor, parse error, empty string, negative duration). About observation, ~510 reconciles.

What I wanted to verify:
- each annotation value is actually honored as the requeue interval
- the new switch-case in `forApp` routes parse errors and below-floor values to their respective warning messages
- valid annotations don't trigger any warnings

| Scenario | Annotation | Expected | Result |
|---|---|---|---|
| Below default | `10s`, `12s`, `15s`, `20s`, `30s`, `45s` | requeue at the annotation value | median gap within 0.05s |
| Control | none | requeue at global default (1m) | 60.04s median |
| Above default | `1m30s`, `2m`, `3m`, `5m` | requeue at the annotation value | median gap within 0.06s |
| Below floor | `9s`, `-5m` | fallback to global with below-minimum warning | cadence at 60s |
| Parse error | `garbage-value`, `""` | fallback to global with invalid warning |  cadence at 60s |

roguepikachu (approved):

roguepikachu:
Hey @officialasishkumar, thank you for the contribution and the quick turnaround on the review feedback. Looking forward to more!

@anoop2811 @briankane @FogDong we're good to have this merged. Could you please have a look when you get a chance? Thanks!

roguepikachu:
Opened a docs tracking issue at [kubevela/kubevela.github.io#1431](https://github.com/kubevela/kubevela.github.io/issues/1431) so the new annotation is captured on kubevela.io once this lands. @officialasishkumar would you be able to help us document this feature there as a follow-up? Happy to review.

anoop2811 (approved):

anoop2811:
Thank you for your contribution!

PR #7107, opened 2026-04-30
Fix: scheduled workflow restart causes double Helm upgrade on helmchart+valuesFrom apps

Aman-Cool:

copilot:all

When a scheduled workflow restart fires (via the `app.oam.dev/restart-workflow` annotation) on an app that uses a `helmchart` component with `valuesFrom`, the AppRevision token written to `app.Status.Workflow` was set to the bare ApplicationRevision name (e.g. `myapp-v1`) without the `-vf-<fingerprint>` suffix. On the very next reconcile the revision-based gate computed `myapp-v1-vf-<fp>`, saw a mismatch, and triggered a second spurious restart; so every scheduled restart caused two full Helm upgrades instead of one.

The fix applies the same fingerprint-append logic that already exists in the revision-based restart path, just earlier, in the scheduled-restart branch.

---

### I have:

- Read and followed KubeVela's https://github.com/kubevela/kubevela/blob/master/contribute/create-pull-request.md.
- https://github.com/kubevela/kubevela.io updated properly. In a new feature or configuration option, an update to the documentation is necessary.
- Run `make reviewable` to ensure this PR is ready for review.
- Added backport `release-x.y` labels to auto-backport this PR if necessary.

---

Ran the full `valuesfrom_fingerprint` unit test suite (21 tests); all pass. Manually traced the reconcile loop for an app with a `helmchart+valuesFrom` component: after a `restart-workflow: "true"` annotation, `app.Status.Workflow.AppRevision` now carries the fingerprint suffix on the first reconcile, so the gate matches immediately and no second restart fires.

---

The change is 5 lines inside the scheduled-restart branch of `checkWorkflowRestart`. It mirrors logic that was already there for the revision-based path; no new behaviour, just making the two paths consistent. Apps without `helmchart+valuesFrom` components are unaffected (fingerprint returns `""` and the suffix is skipped).

### DIFF
--- a/pkg/controller/core.oam.dev/v1beta1/application/workflow.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/workflow.go
@@ -152,8 +152,14 @@ func (r *Reconciler) checkWorkflowRestart(ctx monitorContext.Context, app *v1bet
 			}
 		}
 		app.Status.Conditions = reservedConditions
+		appRevToken := handler.currentAppRev.Name
+		if !metav1.HasAnnotation(app.ObjectMeta, oam.AnnotationPublishVersion) {
+			if vfFp, err := computeValuesFromContentFingerprint(ctx, app); err == nil && vfFp != "" {
+				appRevToken = appRevToken + valuesFromSuffixSeparator + vfFp[:valuesFromSuffixHexLen]
+			}
+		}
 		app.Status.Workflow = &common.WorkflowStatus{
-			AppRevision: handler.currentAppRev.Name,
+			AppRevision: appRevToken,
 		}
 		return
 	}
--- a/pkg/controller/core.oam.dev/v1beta1/application/workflow_vf_restart_test.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/workflow_vf_restart_test.go
@@ -1,0 +1,193 @@
+/*
+Copyright 2026 The KubeVela Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


Aman-Cool:
Hey @anoop2811, found a bug where using the `restart-workflow` annotation on a `helmchart` app with `valuesFrom` was triggering the workflow twice every single time. First run would drop the fingerprint suffix from `AppRevision`, next reconcile would see a mismatch and kick off another restart. Simple 5-line fix to carry the fingerprint through the scheduled restart path the same way the revision-based path already does. All existing tests pass.

Also I'm pretty new to the org and don't have access to any of the communication channels yet, so I couldn't reach out before opening this. If you prefer contributors to discuss changes somewhere first or have a specific workflow you'd like me to follow, happy to do that going forward; just let me know!

Aman-Cool:
Hey @anoop2811, just wanted to flag that both failing checks (`Go / lint` and `Go / check-diff`) are pre-existing issues in the repo and not introduced by this PR.

Verified locally using `golangci-lint` v1.60.1 (same version CI uses); zero non-typecheck errors on both master and this branch. The lint action is exiting non-zero because of typecheck errors caused by a Go 1.23 module format incompatibility with `golangci-lint` v1.60.1, but those are filtered out by the wrapper script in `make lint`. The action calls `golangci-lint` directly without the wrapper, so it sees the non-zero exit and marks it as failed.

The `check-diff` failure cascades from the same root cause; `make reviewable` calls `make lint` internally, which triggers the same typecheck issue on the CI runner.

Happy to dig in further if helpful, but from what I can tell these failures predate this PR and would show up on any PR against master right now.

Reetika2509:
Thanks @Aman-Cool — the fix itself is correct (traced it through; the scheduled branch now writes the fingerprinted token so the gate matches on the next reconcile). Just the two red checks to clear:
  
  lint and check-diff — same single cause: workflow.go:104:1: cyclomatic complexity 36 of func (*Reconciler).checkWorkflowRestart is high (> 35) (gocyclo)(check-diff runs make reviewable → lint, so it fails for the same reason.) The two new fingerprint branches tipped the function from 35 → 36.

Cleanest fix: extract the fingerprint-append into a small helper shared by both the scheduled branch and the revision gate. That moves the branches out (complexity back under 35 → clears both checks) and keeps the two paths in sync — preferable to a //nolint:gocyclo.

PR #7108, opened 2026-05-01
Fix: storage trait gracefully patches env vars that conflict with component definition

### LINKED ISSUE #6841: Built-in trait type storage should gracefully patch resources.
**Describe the bug**
The storage trait does not patch the environment variable correctly using "mountToEnv" if it already exists.

**Steps To Reproduce**
1. Apply the sample Component Definition file:
```
test: {
	alias: ""
	annotations: {}
	attributes: workload: definition: {
		apiVersion: "<change me> apps/v1"
		kind:       "<change me> Deployment"
	}
	description: ""
	labels: {}
	type: "component"
}

template: {
	output: {
		apiVersion: "apps/v1"
		kind:       "Deployment"
		metadata: {
			labels: app: "nginx"
			name: "nginx-deployment"
		}
		spec: {
			replicas: 2
			selector: matchLabels: app: "nginx"
			template: {
				metadata: labels: app: "nginx"
				spec: containers: [{
					env: [{
						name:  "ENVIRONMENT"
						value: "production"
					}, {
						name:  "API_ENDPOINT"
						value: "https://api.example"
					}]
					image: "nginx:1.25"
					name:  "nginx"
					ports: [{
						containerPort: 80
					}]
				}]
			}
		}
	}
	outputs: {}
	parameter: {}
}
```
2. Apply the following application
```apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: example-webservice-app
spec:
  components:
    - name: my-webservice
      type: test
      traits:
        - type: storage
          properties:
            secret:
              - name: "test"
                mountToEnv:
                  envName: "API_ENDPOINT"
                  secretKey: "http://example.com/api"
```

Resulting yaml
```
---
# Application(example-webservice-app) -- Component(my-webservice) 
---

apiVersion: apps/v1
kind: Deployment
metadata:
  annotations: {}
  labels:
    app: nginx
    app.oam.dev/appRevision: ""
    app.oam.dev/component: my-webservice
    app.oam.dev/name: example-webservice-app
    app.oam.dev/namespace: default
    app.oam.dev/resourceType: WORKLOAD
    workload.oam.dev/type: test
  name: nginx-deployment
  namespace: default
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - env:
        - name: ENVIRONMENT
          value: production
        - name: API_ENDPOINT
          value: https://api.example # invalid
          valueFrom:
            secretKeyRef:
              key: http://example.com/api
              name: test
        image: nginx:1.25
        name: nginx
        ports:
        - containerPort: 80
        volumeDevices: []
        volumeMounts: []
      volumes: []

---
## From the trait storage 
apiVersion: v1
kind: Secret
metadata:
  annotations: {}
  labels:
    app.oam.dev/appRevision: ""
    app.oam.dev/component: my-webservice
    app.oam.dev/name: example-webservice-app
    app.oam.dev/namespace: default
    app.oam.dev/resourceType: TRAIT
    trait.oam.dev/resource: secret-test
    trait.oam.dev/type: storage
  name: test
  namespace: default

---
```

**Expected behavior**
If the environment variable already exists, the patch should first remove the field `value` and add `valueFrom` details.

Version details:
CLI Version: 1.10.2
Core Version: 1.10.2
GitRevision: 424e433963551eac070b4369829c4c674fce7ff1
GolangVersion: go1.24.0

Bbn08:
###  Summary

When a component defines an env var with a plain `value` (e.g. `API_KEY: hardcoded`) and the `storage` trait targets the same name via `mountToEnv`/`mountToEnvs`, the previous `+patchKey=name` strategy merged fields rather than replacing the matched entry. The resulting Deployment contained both `value` and `valueFrom` for the same env var, which is invalid and rejected by Kubernetes.

**Root cause:** `// +patchKey=name` on `env: envList` performs a field-level merge on matched items. Since the storage trait entry only has `valueFrom` and the base entry only has `value`, the merge keeps both fields.

**Fix:** Read the base container env from `context.output`, drop entries whose name conflicts with a storage-provided var, and apply `// +patchStrategy=replace` so the full list is replaced atomically. A `len(envList) > 0` guard prevents wiping the component env when no `mountToEnv`/`mountToEnvs` is configured.

Fixes #6841

Added an integration test in `application_controller_test.go`:
- Component declares `env: [{name: "API_KEY", value: "hardcoded"}]`
- Storage trait wires `API_KEY` from a Secret via `mountToEnv`
- Test reconciles the app and fetches the resulting Deployment
- Asserts `API_KEY` has `valueFrom.secretKeyRef` set correctly and no plain `value`

### DIFF
--- a/charts/vela-core/templates/defwithtemplate/storage.yaml
+++ b/charts/vela-core/templates/defwithtemplate/storage.yaml
@@ -115,7 +115,8 @@ spec:
         envList: [
         	if parameter.configMap != _|_ for v in parameter.configMap if v.mountToEnv != _|_ {
         		{
-        			name: v.mountToEnv.envName
+        			name:  v.mountToEnv.envName
+        			value: ""
         			valueFrom: configMapKeyRef: {
         				name: v.name
         				key:  v.mountToEnv.configMapKey
@@ -124,7 +125,8 @@ spec:
         	},
         	if parameter.configMap != _|_ for v in parameter.configMap if v.mountToEnvs != _|_ for k in v.mountToEnvs {
         		{
-        			name: k.envName
+        			name:  k.envName
+        			value: ""
         			valueFrom: configMapKeyRef: {
         				name: v.name
         				key:  k.configMapKey
@@ -133,7 +135,8 @@ spec:
         	},
         	if parameter.secret != _|_ for v in parameter.secret if v.mountToEnv != _|_ {
         		{
-        			name: v.mountToEnv.envName
+        			name:  v.mountToEnv.envName
+        			value: ""
         			valueFrom: secretKeyRef: {
         				name: v.name
         				key:  v.mountToEnv.secretKey
@@ -142,7 +145,8 @@ spec:
         	},
         	if parameter.secret != _|_ for v in parameter.secret if v.mountToEnvs != _|_ for k in v.mountToEnvs {
         		{
-        			name: k.envName
+        			name:  k.envName
+        			value: ""
         			valueFrom: secretKeyRef: {
         				name: v.name
         				key:  k.secretKey
@@ -176,23 +180,13 @@ spec:
         	},
         ]
 
-        _baseContainerEnv: *[] | [...]
-        if context.output.spec.template.spec.containers != _|_ {
-        	if context.output.spec.template.spec.containers[0].env != _|_ {
-        		_baseContainerEnv: context.output.spec.template.spec.containers[0].env
-        	}
-        }
-        _storageEnvNames: {for e in envList {(e.name): true}}
-
         patch: spec: template: spec: {
         	// +patchKey=name
         	volumes: deDupVolumesArray
 
         	containers: [{
-        		if len(envList) > 0 {
-        			// +patchStrategy=replace
-        			env: [for e in _baseContainerEnv if _storageEnvNames[e.name] == _|_ {e}] + envList
-        		}
+        		// +patchKey=name
+        		env: envList
         		// +patchKey=name
         		volumeDevices: volumeDevicesList
         		// +patchKey=name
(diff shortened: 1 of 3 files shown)


Bbn08:
The initial commit used `+patchStrategy=replace` which rebuilt the env list from `context.output` before applying the patch. This broke stacked storage traits when two storage traits are applied to the same component, `context.output` always reflects the original component state, not the accumulated result of previous trait patches. So the second storage trait's replace would wipe out env vars added by the first.

Reverted to `+patchStrategy=replace` → `+patchKey=name` (the original merg strategy) and instead added `value: ""` to every envList entry. When the patch engine merges by name and finds a conflict with a component-defined env var that has `value: "something"`, the patch's `value: ""` wins. Kubernetesonly rejects `value` + `valueFrom` when `len(value) > 0`, so an empty string alongside `valueFrom` is valid and accepted by the API server.This preserves correct accumulation across stacked storage traits whilefixing the original conflict.

vaibhav0096:
@Bbn08 Thanks for digging into this.
  
The fallback approach (+patchKey=name + value: "" on every envList entry) doesn't actually fix the original bug, and I can reproduce it failing on the exact scenario from the integration test.
The problem: +patchKey=name is CUE unification, not strategic-merge
The reasoning in the PR — "when the patch engine merges by name and finds a conflict ... the patch's value: "" wins" — describes strategic-merge / JSON-merge-patch semantics (last-writer-wins), which is how kubectl apply behaves. But KubeVela's // +patchKey=name doesn't merge that way; it performs CUE unification keyed by name. Under unification, two different concrete strings don't override each other — they conflict:
value: "" & value: "hardcoded"   →   _|_ (conflict)

 So value: "" never "wins"; the evaluation errors out before any object reaches the API server. The len(value) > 0 API-server rule is correct, but it's only reached if the render produces value: "" in the first place — and in the conflict
 case it doesn't.

PR #7121, opened 2026-05-04
Fix: http builtin configurable timeout

### LINKED ISSUE #7119: [Feature]: Allow configurable timeout for http builtin task
**Problem:**

The `http` builtin task hardcodes a 3-second client timeout with no way to override it:

```go
client = &http.Client{
    Transport: http.DefaultTransport,
    Timeout:   time.Second * 3,
}
```

Any workflow HTTP call to an endpoint that takes longer than 3 seconds fails with `context deadline exceeded (Client.Timeout exceeded while awaiting headers)`, with no indication the timeout is hardcoded and unconfigurable.

**Solution:**

An optional `timeout` field (in seconds) on the `http` task, defaulting to 3 for backward compatibility:

```cue
http.Do & {
  $params: {
    method:  "GET"
    url:     "https://slow-api.example.com/data"
    timeout: 30
  }
}
```

**Alternatives Considered:**

No viable workaround exists at the CUE level - the timeout is set inside the Go runner and cannot be influenced by task parameters currently.

**Additional context:**

Reproducible on current master. The fix is backward compatible - requests without `timeout` set continue to use the 3-second default.

Bbn08:
## About:

The `http` builtin task had a hardcoded 3-second client timeout with no way to override it. Slow external services or large payloads always failed with a deadline exceeded error.

## Changes:

**`pkg/builtin/http/http.go`**
- Read optional `timeout` field from the CUE task value
- Accepts Go duration string (`"30s"`, `"2m"`, `"500ms"`)
- Returns a descriptive error on invalid or non-positive values — no silent fallback
- Defaults to `3s` when absent (no behaviour change for existing workflows)

**`pkg/builtin/http/http_test.go`**
- `TestHTTPCmdRunWithTimeout`: covers valid timeout, invalid duration string, and non-positive duration

## Usage:

```cue
myStep: op.#HTTPDo & {
  method:  "GET"
  url:     "https://slow-api.example.com/data"
  timeout: "30s"
}
```

## Testing:

```
go test ./pkg/builtin/http/... -v
```

All tests pass. Closes #7119

### DIFF
--- a/.github/actions/env-setup/action.yaml
+++ b/.github/actions/env-setup/action.yaml
@@ -5,7 +5,7 @@ inputs:
   go-version:
     description: 'Go version to use for testing'
     required: false
-    default: '1.23.8'
+    default: '1.24.4'
   install-ginkgo:
     description: 'Install Ginkgo testing framework'
     required: false
@@ -112,6 +112,8 @@ runs:
     - name: Install kustomize
       if: ${{ inputs.install-kustomize == 'true' }}
       shell: bash
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
       run: |
         echo "Installing kustomize version ${{ inputs.kustomize-version }}..."
         mkdir -p ./bin
--- a/.github/actions/env-setup/action.yaml
+++ b/.github/actions/env-setup/action.yaml
@@ -114,6 +114,7 @@ runs:
       shell: bash
       run: |
         VERSION="${{ inputs.kustomize-version }}"
+        VERSION="${VERSION#v}"
         ARCH=$(uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')
         echo "Installing kustomize v${VERSION} for ${ARCH}..."
         mkdir -p ./bin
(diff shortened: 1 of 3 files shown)


Reetika2509:
Thanks for this, @Bbn08 — clean fix for #7119, two things before merge:
1. Split out the unrelated CI change. This PR also edits .github/actions/env-setup/action.yaml (Go 1.23.8 → 1.24.4 + kustomize install rewrite) — unrelated to the timeout fix and looks like a rebase artifact. A Go-version bump especially deserves its own PR.
 2. Use the existing request.timeout field instead of a new top-level one. I checked the #HTTPDo schemas in kubevela/workflow (modern pkg/providers/http and legacy pkg/providers/legacy/http) — both already declare request.timeout?: string as the documented timeout field, and the structs are open so CUE accepts extra fields either way. The bug is real (this legacy builtin hardcodes 3s and ignores request.timeout), but reading a new top-level timeout adds a second, inconsistent way to set the same thing. Honoring request.timeout matches the existing #HTTPDo
contract and keeps one consistent field.

Thanks again — once those two are sorted this should be in good shape

PR #7144, opened 2026-05-19
Feat: helm chart auth

roguepikachu:

copilot:all

Adds Secret-referenced authentication to the native `helmchart` component so it can pull charts from private registries. Today the component only works against public sources; this PR makes it work for OCI registries, HTTPS Helm repositories, and direct chart URLs that require credentials.

Approach:

- A new optional `chart.auth.secretRef` block on the `helmchart` component points at a Kubernetes Secret. The resolver fetches the Secret at chart-fetch time, derives an `HTTPOption` (basic creds, bearer token, TLS material, plain-HTTP opt-in), and threads it into the Helm SDK fetcher for the matching source type.
- Four Secret types are accepted, with per-key detection so the user does not have to declare the auth scheme separately:
  - `kubernetes.io/basic-auth` (`username`, `password`)
  - `kubernetes.io/dockerconfigjson` (verbatim docker config used as the OCI temp credfile)
  - `kubernetes.io/tls` (`tls.crt`, `tls.key`, optional `ca.crt`)
  - `Opaque` with any combination of `username`/`password`, `token`, `caFile`/`certFile`/`keyFile`, `insecureSkipTLS`, and (OCI only) `insecurePlainHTTP`
- Standards followed: RFC 7617 Basic, RFC 6750 Bearer (rejects bearer over plain HTTP, rejects bearer mixed with basic credentials, rejects bearer plus `insecureSkipTLS`, rejects user-supplied bearer on OCI because the registry performs its own Basic to Bearer exchange per the OCI Distribution Spec), RFC 7519 (tokens are passed opaquely, never decoded or claim-validated), and RFC 2119 wording on the user-facing fields.
- Cross-namespace policy mirrors `valuesFrom`: the Secret MUST live in the release namespace or the Application namespace. Anything else is rejected with a clear error.
- Resolver errors surface at admission time as well as install time: the validating webhook does a dry-run render that exercises the resolver, so misconfigured Secret references fail fast at Application creation rather than getting stuck mid-workflow.
- The chart cache cannot bypass auth. When the source declares `auth.secretRef`, the resolver runs on every fetch even on a cache hit, so a follow-up request that points at a missing or invalid Secret cannot ride chart bytes that an earlier authorized request pulled.

A small number of collateral fixes ship alongside the feature because the new test coverage exposed them:

- The `helmchart` audit ConfigMap moved the chart source from a label to an annotation. OCI and HTTPS chart URLs routinely exceed the 63 character Kubernetes label value limit, which was breaking dispatch for any non-trivial URL.
- The ComponentDefinition validating webhook now compiles CUE templates with the workload compiler that knows about internal provider packages (`vela/helm`, `vela/base64`, `vela/http`, `vela/kube`, `vela/util`) instead of the upstream default compiler. Without this, any update to a ComponentDefinition whose CUE imports one of those packages was denied at admission.

Fixes #

I have:

Two layers, both running in CI.

Unit tests for the helm provider package cover every Secret type, every validator, every dispatcher branch, the OCI temp-credfile writer (including its file-permission contract), the cross-namespace policy, and the cache plus auth interaction. The full provider suite runs in well under a second and stays at zero skips.

End-to-end tests stand up real registries inside the test cluster as part of `BeforeSuite` and exercise the controller against them, mirroring the existing `valuesFrom` e2e pattern:

- `zot` for OCI (plain HTTP)
- `chartmuseum` over HTTPS with basic auth
- nginx-fronted `chartmuseum` over HTTPS with bearer auth

Coverage across positive and negative paths:

- Positive: OCI plain HTTP via Opaque with `insecurePlainHTTP`; HTTPS Helm repo with typed `kubernetes.io/basic-auth`; HTTPS Helm repo with Opaque basic; HTTPS Helm repo with bearer; HTTPS Helm repo with Opaque basic plus `insecureSkipTLS`; direct .tgz URL with Opaque basic; secretRef namespace omitted (defaults to release namespace); secretRef namespace explicitly set to the Application namespace.
- Negative: missing Secret; wrong Secret type; cross-namespace Secret rejected; Opaque mixed credentials (basic plus token) rejected; bearer over plain HTTP rejected; bearer plus `insecureSkipTLS` rejected; user-supplied bearer on OCI rejected; username containing colon rejected; bearer token charset violation rejected.

The full `Helmchart` e2e suite (Self-Healing, Adoption and Takeover, State Integrity, Destructive and Chaos, Resource Ordering, Health Checks, Edge Cases, valuesFrom, Auth) was re-run after the changes and is at 80 of 80 passing. No prior test was modified to make the new code pass.

- The audit ConfigMap change (label to annotation) is technically a behavior change for any external consumer that was selecting the audit ConfigMap by `helm.oam.dev/chart` label. The data field is unchanged, so consumers reading the chart source from `.data.chartSource` are unaffected. Worth flagging because it is the kind of change that can show up in a downstream selector.
- The webhook compiler switch is the load-bearing fix that lets ComponentDefinitions whose CUE imports internal provider packages be updated after install. Without it the new `auth?` block could not be rolled out on an existing cluster.
- The cache plus auth interaction is the security-relevant fix. If a reviewer wants to confirm it, the new unit spec `re-runs the auth resolver on a cache hit when the source declares auth.secretRef` is the one to read; it pre-warms the cache with an unauthenticated fetch and then asserts that a follow-up request pointing at a missing Secret fails rather than returning the cached chart.
- OCI plain HTTP is gated behind an explicit Opaque key (`insecurePlainHTTP: "true"`). It is not inferrable from typed Secrets and is documented in the user-facing CUE schema as insecure by design.


roguepikachu:
### Code review

Found 1 issue:

1. `TestValidateCuexTemplate` resets `cuex.DefaultCompiler` (lines 292 and 295), but the PR switched `ValidateCuexTemplate` over to `velacuex.WorkloadCompiler`. The test fixture no longer governs the compiler the function actually uses, so the `withCuexPackageImports` case will not catch a future regression where `WorkloadCompiler` loses an internal provider package (`vela/helm`, etc.). The production switch is correct; the test that guards it is detached from the SUT.

https://github.com/kubevela/kubevela/blob/3e1b68d6a40280771e1fb6ab70e9e0f53de428e9/pkg/webhook/utils/utils.go#L77-L92

https://github.com/kubevela/kubevela/blob/3e1b68d6a40280771e1fb6ab70e9e0f53de428e9/pkg/webhook/utils/utils_test.go#L289-L302

PR #7149, opened 2026-05-21
Fix: propagate errors from GetStatus instead of swallowing CUE health evaluation failures

### LINKED ISSUE #7141: bug: GetStatus always returns nil error, silently swallowing CUE health evaluation failures
## Summary

`GetStatus` in `pkg/cue/definition/health/health.go` unconditionally returns `(result, nil)` even when the CUE health-policy template fails to evaluate. The error is logged as `klog.Warningf` and then discarded. Callers — including the application reconciler — cannot distinguish between:

- **Template evaluated; component is genuinely unhealthy** → `Healthy: false, err: nil`
- **Template failed to evaluate at all** → `Healthy: false, err: nil`

Both cases produce byte-for-byte identical output from `GetStatus`.

---

## Root Cause

```go
// pkg/cue/definition/health/health.go:87-107
healthy, healthErr := CheckHealth(templateContext, request.Health, request.Parameter)
if healthErr != nil {
    klog.Warningf("failed to check health: %v", healthErr)  // logged, then dropped
}
// ...
return &StatusResult{Healthy: healthy, ...}, nil  // ALWAYS nil — healthErr is gone
```

`CheckHealth` correctly returns an error (e.g. CUE compile error, type mismatch, missing `isHealth` field). `GetStatus` receives it, logs it, and throws it away. The same silent-discard pattern applies to `getStatusMap` errors (line 83–84) and `getStatusMessage` errors (lines 99–100).

---

## Reproduction (unit test)

```go
func TestGetStatus_SwallowsHealthEvalError(t *testing.T) {
    templateContext := map[string]interface{}{
        "output": map[string]interface{}{
            "spec":   map[string]interface{}{"replicas": int64(1)},
            "status": map[string]interface{}{"readyReplicas": int64(1)},
        },
    }

    // CUE type error: int + string — CheckHealth returns an error
    brokenPolicy := `isHealth: context.output.spec.replicas + "not-a-number" > 0`

    result, err := GetStatus(templateContext, &StatusRequest{Health: brokenPolicy})
    // err is nil  ← bug
    // result.Healthy is false — indistinguishable from a genuinely unhealthy component
}
```

Running this test produces:

```
E0517 health.go:64  CUE compilation error: isHealth: invalid operands 1 and "not-a-number" to '+' (type int and string)
W0517 health.go:89  failed to check health: compile health template: isHealth: invalid operands 1 and "not-a-number" to '+' (type int and string)

=== GetStatus with broken health template ===
  result.Healthy : false
  err            : <nil>
BUG CONFIRMED: GetStatus returned nil error despite health eval failure
  Healthy=false — operator sees 'unhealthy' with zero explanation
```

The controller log shows the error. `GetStatus` callers see nothing.

---

## Ambiguity demonstration

```go
// Case A: valid template, component genuinely not ready (1 of 3 replicas)
validPolicy  := `isHealth: context.output.status.readyReplicas == context.output.spec.replicas`
// Case B: broken template — CUE type error, evaluation fails entirely
brokenPolicy := `isHealth: context.output.spec.replicas + "not-a-number" > 0`

resultA, errA := GetStatus(ctx, &StatusRequest{Health: validPolicy})
resultB, errB := GetStatus(ctx, &StatusRequest{Health: brokenPolicy})

// resultA.Healthy == false, errA == nil
// resultB.Healthy == false, errB == nil  ← operator cannot distinguish these
```

---

## Impact

- An operator who writes a typo or type error in a ComponentDefinition `health.policy` CUE template will see **every component permanently stuck at `Healthy: false`** with no error condition set on any Kubernetes object, no controller event, and no reconciler warning — until they dig through pod logs.
- The only signal is a `klog.Warningf` at verbosity 0, not associated with any Kubernetes resource.
- Affects all ComponentDefinitions that define a `health.policy` field.

---

## Suggested Fix

Propagate the error out of `GetStatus` instead of discarding it:

```go
func GetStatus(templateContext map[string]interface{}, request *StatusRequest) (*StatusResult, error) {
    // ...
    healthy, healthErr := CheckHealth(templateContext, request.Health, request.Parameter)
    if healthErr != nil {
        return nil, fmt.Errorf("health policy evaluation failed: %w", healthErr)
    }
    // ...
}
```

At minimum, the function signature should surface the error so reconcilers can set an appropriate condition (e.g. `HealthCheckFailed`) on the Application or Component resource, giving operators an actionable signal via `kubectl get application`.

---

## Affected Code

- `pkg/cue/definition/health/health.go:87-90` — `GetStatus` discards `healthErr`
- `pkg/cue/definition/health/health.go:83-84` — `GetStatus` discards `mapErr`
- `pkg/cue/definition/health/health.go:99-100` — `GetStatus` discards `msgErr`

## Environment

Confirmed on KubeVela `master` branch (2026-05-17). Reproduced via unit test — no cluster required.

vansvan17:
Propagate errors from `getStatusMap`, `CheckHealth`, and `getStatusMessage`
through `GetStatus` instead of logging and discarding them.

Callers like the application reconciler could not distinguish between:
- Template evaluated; component is genuinely unhealthy (Healthy: false, err: nil)
- Template failed to evaluate at all (Healthy: false, err: nil)

Now errors are returned via `errors.Join` so callers can decide how to surface them.
The function still returns a best-effort status result alongside any errors, preserving
backwards compatibility for existing callers.

Fixes #7141

### DIFF
--- a/pkg/controller/core.oam.dev/v1beta1/application/apply.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/apply.go
@@ -25,9 +25,9 @@ import (
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/klog/v2"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	monitorContext "github.com/kubevela/pkg/monitor/context"
--- a/pkg/controller/core.oam.dev/v1beta1/application/apply.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/apply.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -302,7 +303,7 @@ func (h *AppHandler) collectWorkloadHealthStatus(ctx context.Context, comp *appf
 		}
 		statusResult, err := comp.EvalStatus(templateContext)
 		if err != nil {
-			return false, nil, nil, errors.WithMessagef(err, "app=%s, comp=%s, evaluate workload status message error", appName, comp.Name)
+			klog.Warningf("app=%s, comp=%s, evaluate workload status message error: %v, will continue with best-effort status", appName, comp.Name, err)
 		}
 		if statusResult != nil {
 			status.Healthy = statusResult.Healthy
--- a/pkg/cue/definition/health/health.go
+++ b/pkg/cue/definition/health/health.go
@@ -18,6 +18,7 @@ package health
 
 import (
 	"encoding/json"
+	stderrors "errors"
 	"slices"
 	"strings"
 
@@ -104,7 +105,7 @@ func GetStatus(templateContext map[string]interface{}, request *StatusRequest) (
 		Healthy: healthy,
 		Message: message,
 		Details: statusMap,
-	}, nil
+	}, stderrors.Join(mapErr, healthErr, msgErr)
 }
 
 func getStatusMessage(templateContext map[string]interface{}, customStatusTemplate string, parameter interface{}) (string, error) {
--- a/pkg/cue/definition/health/health_test.go
+++ b/pkg/cue/definition/health/health_test.go
@@ -460,6 +460,7 @@ func TestContextPassing(t *testing.T) {
 		request     StatusRequest
 		expMessage  string
 		expDetails  map[string]string
+		expErr      bool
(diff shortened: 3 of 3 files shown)


vansvan17:
The one failing check is E2E Test / e2e-tests (v1.31): it timed out at 58m. This is an environmental/flaky failure: the change only touches pure CUE evaluation in health.go with no Kubernetes API interaction, and every other check (lint, staticcheck, unit-tests, multi-cluster E2E) passes. Not related to this PR.

Reetika2509:
Thanks for tackling this, @vansvan17 — the direction is exactly right for #7141, and the component path + errors.Join approach (best-effort result plus surfaced error) is a clean way to keep backward compatibility. 
  
 unit-test failure: the fix updated the component path but not the trait path. There are two EvalStatus callers in
  apply.go:

  // collectWorkloadHealthStatus (component) — UPDATED in this PR 
  statusResult, err := comp.EvalStatus(templateContext)
  if err != nil {
      klog.Warningf("...evaluate workload status message error: %v, will continue...", err)  // log & continue
  }

  // collectTraitHealthStatus (trait) — apply.go:270, NOT updated 
  statusResult, err := tr.EvalStatus(templateContext)
  // ...
  return traitStatus, extractOutputs(templateContext), err   // still returned

  Before this PR, GetStatus always returned nil, so that trait err was always nil and harmless. Now it returns errors.Join(mapErr, healthErr, msgErr), so any trait whose status template surfaces a (previously-swallowed, often benign) sub-error makes collectTraitHealthStatus return hard error → collectHealthStatus propagates it → the app never transitions to running.

Mirror the component fix on the trait path. Update collectTraitHealthStatus (apply.go:270) to log-and-continue with the best-effort statusResult instead of returning the error, the same way you handled collectWorkloadHealthStatus. This should fix the unit-test failure.

PR #7162, opened 2026-05-25
Feat(helm): set User-Agent when downloading Helm repo index.yaml

goingforstudying-ctrl:
## What
Set a stable, identifiable User-Agent header when performing HTTP GET for Helm chart repository index.yaml via the Helm getter package.

## Why
Helm repository operators and HTTP gateways often rely on the User-Agent header to identify clients, apply rate limits, debug traffic, and distinguish automated pulls from browsers. Today, when KubeVela fetches a chart repository's index.yaml, there is an explicit , so requests may use a generic or empty agent string. That makes KubeVela traffic harder to recognize in registry logs and can complicate support when users report connectivity or throttling issues against a specific Helm backend.

## How
- Uses the existing Helm getter option 
- Constructs the user-agent string as , matching the pattern already used in  for the Kubernetes client config
- Removes the  comment

## Testing
- Verified syntax with gofmt
- The change is a straightforward getter option addition; existing integration tests in  cover the load path

Closes #7134

PR #7165, opened 2026-05-25
Feat: set stable User-Agent header when downloading Helm repo index.yaml

### LINKED ISSUE #7134: [Feature] : (helm) set User-Agent when downloading Helm repo index.yaml
Helm repository operators and HTTP gateways often rely on the User-Agent header to identify clients, apply rate limits, debug traffic, and distinguish automated pulls from browsers. Today, when KubeVela fetches a chart repository’s index.yaml via LoadRepoIndex / loadData in pkg/utils/helm/repo_index.go, there is an explicit // TODO add user-agent, so requests may use a generic or empty agent string. That makes KubeVela traffic harder to recognize in registry logs and can complicate support when users report connectivity or throttling issues against a specific Helm backend.

Set a stable, identifiable User-Agent when performing the HTTP GET for index.yaml (in the code path that uses getter.NewHTTPGetter() and g.Get(...) in repo_index.go). The value should follow the same spirit as the Kubernetes client configuration in cmd/core/app/server.go (KubeVela name plus version/revision where available), so operators see consistent “KubeVela” branding across different outbound calls where practical.

If the Helm getter package exposes a supported option for custom headers or user agent (for example a getter.With* option), use that rather than ad-hoc transport patching, so behavior stays aligned with Helm’s own HTTP stack.

Add a small unit test that verifies the user agent is applied (for example by asserting the getter is invoked with the expected option, or by using an httptest server that records the incoming User-Agent header, depending on what is easiest without brittle coupling to Helm internals).

Leave as-is: Minimal code change, but registry visibility and supportability stay poor.
Generic agent only (e.g. kubevela/helm-index): Simple but less useful than including a revision or version string for correlating with a specific release or build.
Configurable per-repo: More flexible but likely unnecessary for a first change; could be a follow-up if maintainers want per-registry overrides.

**Additional context**
Goal is outbound identification only; no change to chart install semantics beyond the index fetch client metadata.

Hi @Chaitanyareddy0702 @vishal210893 @roguepikachu can you please review this, Thanks for your time :)

manoj-1407:
While looking at how Helm repo index files are fetched, I noticed
HTTPGetResponse in pkg/utils/common/common.go never sets a User-Agent,
so every request goes out as Go-http-client/1.1. Some registries
rate-limit or log by User-Agent so this makes KubeVela traffic
unidentifiable in registry logs.

Added a UserAgent field to HTTPOption so it can be overridden per-call,
and set kubevela/<version> as the default in HTTPGetResponse. Backward
compatible — if UserAgent is empty it falls through to the default.

Added TestHTTPGetResponse_UserAgent covering nil opts, empty UserAgent,
and a custom override. All three pass.

Closes #7134

### DIFF
--- a/pkg/utils/common/common.go
+++ b/pkg/utils/common/common.go
@@ -69,6 +69,8 @@ import (
 	velacue "github.com/oam-dev/kubevela/pkg/cue"
 	"github.com/oam-dev/kubevela/pkg/cue/process"
 	"github.com/oam-dev/kubevela/pkg/oam"
+
+	"github.com/oam-dev/kubevela/version"
 )
 
 var (
@@ -112,6 +114,8 @@ type HTTPOption struct {
 	// HTTP rather than TLS. Honored only on the OCI fetch path. Insecure
 	// by design; users opt in via the Opaque Secret key insecurePlainHTTP.
 	PlainHTTP bool `json:"plainHTTP,omitempty"`
+	// UserAgent overrides the HTTP User-Agent header. If empty, defaults to "kubevela/<version>".
+	UserAgent string `json:"userAgent,omitempty"`
 }
 
 // InitBaseRestConfig will return reset config for create controller runtime client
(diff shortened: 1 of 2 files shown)


roguepikachu:
This is resolved in #7162

Issue #7164, opened 2026-05-25
[Bug] Application Ready condition stays True/ReconcileSuccess when Parsed=False


roguepikachu:
## Summary

When an Application fails to parse (e.g. its referenced `ComponentDefinition` or `DefinitionRevision` is missing), `status.conditions[Parsed]` correctly flips to `False / ReconcileError` and `status.status` flips to `rendering`. But `status.conditions[Ready]` stays at `True / ReconcileSuccess` indefinitely, frozen from the last successful reconcile. Anyone polling `Ready` as the health signal misses the failure.

Discovered while investigating #7163 (scenarios 3 and 4 in [this comment](https://github.com/kubevela/kubevela/issues/7163#issuecomment-4536192599)).

## Reproduction

Tested on master `a24d3a9c6` with `vela-core` built from source and `--definition-revision-limit=2`.

1. Apply a `ComponentDefinition` with auto-increment naming (no `spec.version`).
2. Create an Application that pins to `@v1` (e.g. `drift-probe@v1`).
3. Apply 3+ subsequent CD spec edits to trigger GC of `drift-probe-v1`.

The reconcile then fails with `WorkloadDefinition.core.oam.dev "drift-probe@v1" not found`.

## Expected

`status.conditions[Ready]` flips to `False` with a reason reflecting the failure (e.g. `ReconcileError`), so health checkers polling `Ready` can see it.

## Actual

```
Parsed=False  (ReconcileError)   WorkloadDefinition ... not found
Revision=True (Available)
Policy=True   (Available)
Render=True   (Available)
Workflow=True (Available)
Ready=True    (ReconcileSuccess)   <-- stale, never flipped
```

`status.status` does correctly flip to `rendering` and a `Warning FailedParse` event is emitted, so other surfaces work. Only `Ready` is wrong.

## Environment

- KubeVela master `a24d3a9c6`
- k3d, k3s `v1.31.5+k3s1`

## Notes

The failure mode isn't specific to the `@vN` + GC trigger I reproduced with; any reconcile path that ends in `Parsed=False` without flipping `Ready` would exhibit the same symptom. Worth checking whether other `ReconcileError` reasons hit the same gap.

Anuragp22:
I'd like to take this. Plan: have the Ready condition aggregate Parsed and the other sub-conditions in the application status controller, so it flips to False when any of them fail instead of holding the last successful reconcile.

PR #7167, opened 2026-05-27
fix: bare-type components track latest definition without autoUpdate …

### LINKED ISSUE #7163: [Bug] Bare-type components are pinned to first-render snapshot when   `app.oam.dev/autoUpdate` is absent
## Summary

An Application that references a ComponentDefinition by bare name (no `@vN`
suffix) freezes its rendered output at the first apprev's snapshot.
Subsequent edits to the ComponentDefinition do not generate a new
ApplicationRevision and do not re-render the workloads, even though the
Application reports `Ready / ReconcileSuccess`.

The only way to get the documented "bare type = latest" behavior is to add
`app.oam.dev/autoUpdate: "true"` to the Application. As of the semver PR
(#6648, 2025-02-03), that annotation is documented to mean "semver-promote
my `@vN` pins" — making it a prerequisite for bare-type updates is a
leftover from an unrelated 2022 PR.

## Evidence

Single component, bare type `aws-s3-v1`, no annotation. Application has
`generation=1` / `observedGeneration=1` / Ready. The only apprev embeds a
ComponentDefinition snapshot whose `Spec` hashes (via
`ComputeSpecHash` / `hashstructure/v2 v2.0.2`) to `cc1c3c2a461268e4` —
matching `DefinitionRevision aws-s3-v1-v12`. `aws-s3-v1-v13` was
published 23 days after the apprev and exists in `vela-system`, but the
Application never produced a new apprev.

## Root cause

`pkg/controller/core.oam.dev/v1beta1/application/revision.go:369-372` in
`currentAppRevIsNew`:

```go
isLatestRev := deepEqualAppInRevision(h.latestAppRev, h.currentAppRev)
if metav1.HasAnnotation(h.app.ObjectMeta, oam.AnnotationAutoUpdate) {
    isLatestRev = h.app.Status.LatestRevision.RevisionHash == h.currentRevHash && DeepEqualRevision(h.latestAppRev, h.currentAppRev)
}
```

`deepEqualAppInRevision` diffs `Application.Spec` + `Policies` + `Workflow`
only. `DeepEqualRevision` also diffs each embedded ComponentDefinition /
TraitDefinition / WorkloadDefinition `.Spec`. Without the annotation the
controller never sees that the referenced definition has changed, so it
reuses the existing apprev.

Introduced in commit `d7261cd965` ("Fix: disable cochange for apprev when
def changes", #3217, 2022-02-10), three years before the semver PR
overloaded the same annotation with its current `fetchDefinitionRevision`
semver-promotion job.

harshdwivediiiii:
Fixes #7163

## What changed
- `revision.go`: added `hasBareTypeComponents()` helper and updated `currentAppRevIsNew` to use `DeepEqualRevision` for bare-type components
- `baretype_test.go`: added unit tests covering the bug and fix

## Why
Bare-type components (e.g. `aws-s3-v1`, no `@vN`) were frozen at their
first ApplicationRevision snapshot even after the ComponentDefinition
was updated. The controller was using a shallow comparison that only
checked App.Spec — not the definition specs.

## Fix
If any component uses a bare type, the deep comparison is used regardless
of the `autoUpdate` annotation. The annotation keeps its existing meaning
for `@vN` pinned components.

## Tests
All 8 unit tests pass.
Screenshot From 2026-05-27 18-27-07

### DIFF
--- a/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
@@ -527,12 +527,19 @@ func (r *Reconciler) handleFinalizers(ctx monitorContext.Context, app *v1beta1.A
 	return r.result(nil).end(false)
 }
 
-func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.Application, condition condition.Condition, phase common.ApplicationPhase) (ctrl.Result, error) {
-	app.SetConditions(condition)
+func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.Application, cond condition.Condition, phase common.ApplicationPhase) (ctrl.Result, error) {
+	app.SetConditions(cond)
+	app.Status.SetConditions(condition.Condition{
+		Type:               condition.ConditionType(common.ReadyCondition.String()),
+		Status:             corev1.ConditionFalse,
+		LastTransitionTime: metav1.Now(),
+		Reason:             condition.ReasonReconcileError,
+		Message:            cond.Message,
+	})
 	if err := r.patchStatus(ctx, app, phase); err != nil {
 		return r.result(errors.WithMessage(err, "cannot update application status")).ret()
 	}
-	return r.result(fmt.Errorf("object level reconcile error, type: %q, msg: %q", string(condition.Type), condition.Message)).ret()
+	return r.result(fmt.Errorf("object level reconcile error, type: %q, msg: %q", string(cond.Type), cond.Message)).ret()
 }
 
 // Application status can be updated by two methods: patch and update.
--- a/pkg/controller/core.oam.dev/v1beta1/application/baretype_test.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/baretype_test.go
@@ -1,0 +1,88 @@
+package application
+
+import (
+	"testing"
+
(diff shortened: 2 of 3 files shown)


harshdwivediiiii:
Also Fixes #7164

## What changed
`endWithNegativeCondition` in `application_controller.go` now also sets
`Ready=False` whenever a reconcile failure occurs.

## Why
When parse or any other reconcile step failed, only the specific condition
(e.g. `Parsed=False`) was set. `Ready` stayed `True` from the last
successful reconcile — misleading any health checker polling `Ready`.

## Fix
Added `Ready=False` with reason `ReconcileError` inside
`endWithNegativeCondition` so all failure paths flip `Ready` correctly.

## Tests
All existing tests pass.
Screenshot From 2026-05-27 19-43-30

roguepikachu:
Hi @harshdwivediiiii,

The issue is subject to community discussion as this might introduce regression. Currently it is assigned to me, and we are discussing internally within our team on how to handle this. Closing this PR for now. There are plenty of issues available and open for contributions. Please get yourself assigned before raising a PR, this helps us coordinate efforts and ensures we're all working toward the same goals. Thanks for your understanding.

vansvan17:
Thanks for the detailed explanation. The \strings.LastIndex\ approach for version matching and decoupling the GC root cause from the visibility fix both sound like the right call. Happy to re-review once those updates land.

PR #7168, opened 2026-05-27
Fix(controller): set Ready condition to False on reconcile failure

### LINKED ISSUE #7164: [Bug] Application Ready condition stays True/ReconcileSuccess when Parsed=False

Anuragp22:
endWithNegativeCondition set only the failing sub-condition (e.g. Parsed=False/ReconcileError) but left the rollup Ready condition unchanged, so a reconcile failure left Ready at True/ReconcileSuccess from the last successful reconcile. Health checkers polling Ready as the application health signal missed the failure.

Also set Ready=False/ReconcileError alongside the sub-condition with the same message, so the rollup reflects the failure on every path that ends through endWithNegativeCondition.

Fixes #7164

copilot:all

Fixes #

I have:

### DIFF
--- a/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
@@ -527,12 +527,20 @@ func (r *Reconciler) handleFinalizers(ctx monitorContext.Context, app *v1beta1.A
 	return r.result(nil).end(false)
 }
 
-func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.Application, condition condition.Condition, phase common.ApplicationPhase) (ctrl.Result, error) {
-	app.SetConditions(condition)
+func (r *Reconciler) endWithNegativeCondition(ctx context.Context, app *v1beta1.Application, cond condition.Condition, phase common.ApplicationPhase) (ctrl.Result, error) {
+	// Flip the rollup Ready condition alongside the failing sub-condition so health checkers
+	// polling Ready see the failure instead of the last successful reconcile (#7164).
+	app.SetConditions(cond, condition.Condition{
+		Type:               condition.ConditionType(common.ReadyCondition.String()),
+		Status:             corev1.ConditionFalse,
+		LastTransitionTime: metav1.Now(),
+		Reason:             condition.ReasonReconcileError,
+		Message:            cond.Message,
+	})
 	if err := r.patchStatus(ctx, app, phase); err != nil {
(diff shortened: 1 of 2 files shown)


Reetika2509 (approved):
The change is correctly scoped and the regression test asserts the behavior directly. Also confirmed: `SetConditions` only rewrites a condition when Type/Status/Reason/Message differ, so the hardcoded `LastTransitionTime` won't cause status churn. Builds, vets, and gofmt are clean. LGTM

anoop2811 (approved):

anoop2811:
Thank you for your contribution!

PR #7188, opened 2026-06-08
Test: add table-driven tests for CUEGenerator enum evaluation edge cases

sakirr05:
## What

Add table-driven `DescribeTable`/`Entry` tests for CUEGenerator enum evaluation in `pkg/definition/defkit/cuegen_test.go`.

## Why

PR #7058 added enum generation logic across `writeStringParam`, `writeEnumParam`, and `writeStructFieldForHelper`. The existing tests cover happy-path cases via individual `It()` blocks but do not exercise edge cases such as empty `Values()`, single-value enums, default-not-in-values, and `OpenEnum` combinations. This PR closes that gap using the same table-driven pattern already established in `param_constraints_test.go`.

## Test coverage added
- `StringParam` with empty `Values()` → falls back to plain `string`
- Single-value enum, required and optional
- Default value not present in `Values()` list
- `OpenEnum` with and without values
- `EnumParam` required vs optional

PR #7189, opened 2026-06-08
Fix: add pod readiness guard and FlakeAttempts to Helm adoption E2E test contexts that were missing them

sakirr05:
## What

Add Deployment readiness guards and `FlakeAttempts(2)` to Helm adoption E2E test contexts that were missing them.

## Why

While working on PR #7188, the CI run ([#27159484350](https://github.com/kubevela/kubevela/actions/runs/27159484350/job/80170890418)) failed on the "adopt an existing vanilla helm Release" test, the diagnostics showed podss stuck in `ImagePullBackOff` they existed but were never actually ready,

Looking at the test code, the precondition check at line 648-654 only verified `len(podList.Items) >= 2`. That's a count check, not a readiness check. Pods in `ImagePullBackOff` satisfy that condition while never becoming functional, so when the test later checks `countSurvivingPods`, it records UIDs of broken pods and the assertion fails automatcally

I noticed the other adoption contexts in the exact same file already do this correctly:
- "adopt release with different values" (line 699): waits for `d.Status.ReadyReplicas == 1`
- "adoption with valuesFrom" (line 2037): waits for `d.Status.ReadyReplicas == 1`

 

The vanilla adoption and re-adopt contexts were simply written with a weaker precondition his pr aligns them with the samee pattern their siblings already uses it

## Approach

- Replaced the `len(podList.Items) >= 2` check with `Deployment.Status.ReadyReplicas == 2` in the vanilla adoption context
- Added the same readiness wait to the "re-adopt after application deletions" context (both of its `helm install` steps had no readiness check at all before calling `h.deployApp()`)
- Added `flakeAttempts(2)` to all three adoption contexts so that genuine transient infra issues (registry rate limits, KinD network blips) get a single retry rather than failing the entire CI run
- Increased timeout from 60s to 120s to match other adoption waits in the file
- Removed the unused `isPodReady` helper that was initially added but turned out unnecessary since `ReadyReplicas` on the deployment object covers the check more cleanly

## Verification

- `go build ./test/e2e-test/...` ---- passes
- `go vet ./test/e2e-test/...` ---- passes  
- `go fmt` ---- clean

### DIFF
--- a/test/e2e-test/helmchart_test.go
+++ b/test/e2e-test/helmchart_test.go
@@ -632,7 +632,7 @@ var _ = Describe("Helmchart Self-Healing", func() {
 
 var _ = Describe("Helmchart Adoption & Takeover", func() {
 
-	Context("Adopt an Existing Vanilla Helm Release", Ordered, func() {
+	Context("Adopt an Existing Vanilla Helm Release", FlakeAttempts(2), Ordered, func() {
 		h := newHelmTestContext()
 		BeforeAll(func() { h.createNamespace() })
 		AfterAll(func() { h.cleanup() })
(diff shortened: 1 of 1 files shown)


roguepikachu (approved):

anoop2811 (approved):

anoop2811:
Thank you for your contribution @sakirr05 !

PR #7191, opened 2026-06-18
[Backport release-1.10] Fix: prevent unbounded read in Terraform remote configuration loader (GHSA-fmgp-q6jx-gg3x)

roguepikachu:
Backports the Terraform remote loader DoS fix (GHSA-fmgp-q6jx-gg3x) to release-1.10.

- Prevents an unbounded read in the Terraform remote configuration loader: bounded clone (shallow depth 1, 2-minute fetch timeout, 64 MiB and file-count caps), regular-file and in-cache-path validation, symlink rejection, cache invalidation on URL change, and cleanup on rejected reads. Cherry-pick of f6a64398b from master.
- Resolves the gosec G304 lint failure in the module cache check. Cherry-pick of #7190.

A user with create/update on ComponentDefinition could point a Terraform remote schematic at a repository whose variables.tf symlinks to /dev/zero, making vela-core read an unbounded stream and get OOM killed (control-plane DoS). The fix validates the target is a regular file, keeps the resolved path inside the clone cache, and caps the file size before reading.

Verified locally: the pkg/controller/utils security tests pass on this branch (symlink-escape rejection, size-cap boundaries, lexical traversal rejection, cache invalidation, cleanup on failure). The envtest-based Ginkgo suite requires KUBEBUILDER_ASSETS and was not run locally.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
@@ -269,8 +269,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	// Rebuild appliedResources from the ResourceTracker now that the workflow has finished
-	// dispatching. The RT is the authoritative source
-	app.Status.AppliedResources = handler.resourceKeeper.GetAppliedResources()
+	// dispatching. The RT is the authoritative source of which resources exist; the creator
+	// attribution, which the RT does not persist, is re-attached from the dispatch records.
+	app.Status.AppliedResources = handler.appliedResourcesWithCreator()
 
 	var phase = common.ApplicationRunning
 	isHealthy := evalStatus(logCtx, handler, appFile, appParser)
--- a/pkg/controller/core.oam.dev/v1beta1/application/apply.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/apply.go
@@ -129,6 +129,35 @@ func (h *AppHandler) Delete(ctx context.Context, _ client.Client, cluster string
 	return nil
 }
 
+// appliedResourcesWithCreator rebuilds the applied resources from the ResourceTracker,
+// which is the authoritative source of which resources currently exist, and re-attaches
+// the creator recorded while dispatching them. The ResourceTracker does not persist the
+// creator, so rebuilding from it alone drops the creator attribution from the status.
+// reconcile runs single threaded, matching addAppliedResource, so no locking is needed.
+func (h *AppHandler) appliedResourcesWithCreator() []common.ClusterObjectReference {
+	rebuilt := h.resourceKeeper.GetAppliedResources()
+	for i := range rebuilt {
+		if rebuilt[i].Creator != "" {
+			continue
+		}
+		for _, dispatched := range h.appliedResources {
+			if dispatched.Creator != "" && sameResource(rebuilt[i], dispatched) {
+				rebuilt[i].Creator = dispatched.Creator
+				break
+			}
+		}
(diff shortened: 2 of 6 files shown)


roguepikachu:
A heads-up on the two commits here beyond the security backport, since neither is a straight cherry-pick from master.

**1. Restore `creator` on `appliedResources`**

release-1.10's rebuild of `status.appliedResources` from the ResourceTracker ([#7086](https://github.com/kubevela/kubevela/pull/7086)) drops the per-resource `creator`, because the ResourceTracker does not persist that field. That has been failing the `test application applied resource in workflow step status` test on this branch since #7086 merged, independent of this PR.

master does not hit this because of its larger post-dispatch rework ([#7008](https://github.com/kubevela/kubevela/pull/7008), [#7030](https://github.com/kubevela/kubevela/pull/7030) and related), which keeps the creator populated through the rebuild. That rework was never backported to release-1.10, and pulling it into a security backport felt too risky.

So this adds a small, branch-local fix instead: [`appliedResourcesWithCreator()`](https://github.com/kubevela/kubevela/blob/4e9ba461b8b979186b24f1a7f50ffe93c2753de2/pkg/controller/core.oam.dev/v1beta1/application/apply.go#L137) rebuilds from the ResourceTracker (keeping it authoritative for which resources exist) and re-attaches the creator recorded during dispatch, matched on object identity by [`sameResource()`](https://github.com/kubevela/kubevela/blob/4e9ba461b8b979186b24f1a7f50ffe93c2753de2/pkg/controller/core.oam.dev/v1beta1/application/apply.go#L156). The [rebuild call site](https://github.com/kubevela/kubevela/blob/4e9ba461b8b979186b24f1a7f50ffe93c2753de2/pkg/controller/core.oam.dev/v1beta1/application/application_controller.go#L274) uses it in place of `GetAppliedResources()`. This helper does not exist on master, it is intentional and scoped to this branch.

**2. Do not log or persist credentials in Git URLs**

The cache-reuse logic logged the raw remote URL and wrote it to the `.remote-url` marker, which can leak credentials embedded in authenticated Git URLs. [`redactURLCredentials()`](https://github.com/kubevela/kubevela/blob/4e9ba461b8b979186b24f1a7f50ffe93c2753de2/pkg/controller/utils/capability.go#L296) strips the userinfo before [logging](https://github.com/kubevela/kubevela/blob/4e9ba461b8b979186b24f1a7f50ffe93c2753de2/pkg/controller/utils/capability.go#L226) and before [persisting the marker](https://github.com/kubevela/kubevela/blob/4e9ba461b8b979186b24f1a7f50ffe93c2753de2/pkg/controller/utils/capability.go#L286), and the [reuse check](https://github.com/kubevela/kubevela/blob/4e9ba461b8b979186b24f1a7f50ffe93c2753de2/pkg/controller/utils/capability.go#L279) compares the same stripped form so cache behavior is unchanged.

On the other branches: the credential-redaction fix (2) also applies to master and the release-1.9 backport, which carry the same loader code, and I can follow up on those separately. The `appliedResources` fix (1) is specific to release-1.10: master is unaffected (the post-dispatch rework above keeps the creator populated), and release-1.9 does not do the ResourceTracker rebuild at all.

briankane (approved):

PR #7202, opened 2026-06-27
Fix: Update default registry definition.

### LINKED ISSUE #6835: [Feature] New default definition registry to store up-to-date definition list
Currently `oss://registry.kubevela.net` is [set](https://github.com/kubevela/kubevela/blob/master/references/cli/registry.go#L280) as the default registry, and this domain has expired certificates.

* Fix the [automation](https://github.com/kubevela/kubevela/blob/master/vela-templates/gen_definitions.sh) to update the definition every time master is updated.
* Leverage the already [existing path](https://github.com/kubevela/kubevela/tree/master/vela-templates/registry/auto-gen) that stores Kubevela definitions 
* Then point to this path as the default registry

kash2104:

This PR updates the default registry URL from `oss://registry.kubevela.net/` to `https://github.com/kubevela/kubevela/tree/master/vela-templates/registry/auto-gen/` since the certificate has been expired for the older one. Along with this, a new job has been introduced in the `registry` workflow which generates the registry definitions.

Fixes #6835 

I have:

- Built the modified vela CLI locally and ran `./bin/vela registry ls`.
- Verified that the updated default registry `https://github.com/kubevela/kubevela/tree/master/vela-templates/registry/auto-gen/` was added to the registry configuration, confirming that the new default registry is being picked up by the CLI.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7202?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/registry.yml
+++ b/.github/workflows/registry.yml
@@ -244,3 +244,42 @@ jobs:
     secrets:
       registry-username: ${{ secrets.DOCKER_USERNAME }}
       registry-password: ${{ secrets.DOCKER_PASSWORD }}
+
+  generate-definitions:
+    name: Generate Definitions
+    if: github.ref == 'refs/heads/master'
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491
+        with:
+          go-version: "1.23.8"
+
+      - name: Build vela CLI
+        run: |
+          make vela-cli
+
+      - name: Render registry definitions with vela binary
+        run: |
+          ./bin/vela def render vela-templates/definitions/registry -o vela-templates/registry/auto-gen --message "Definition source cue file: vela-templates/definitions/registry/{{INPUT_FILENAME}}"
+
+      - name: Commit and push generated definitions
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config --global user.email "135009839+kubevela[bot]@users.noreply.github.com"
+          git config --global user.name "kubevela[bot]"
+          git add vela-templates/registry/auto-gen
+          if git diff --quiet --cached; then
+            echo "No changes to commit"
+            exit 0
+          fi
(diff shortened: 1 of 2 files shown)


Reetika2509:
Hi @kash2104 , thanks for the contribution! Please take care of the AI review comment as its a valid concern.

kash2104:
@Reetika2509 Have taken into consideration the reviews and addressed them.

vaibhav0096:
Hi @kash2104 
Nice work on this.
One gap though: this fixes brand-new installs cleanly, but not upgrades on a machine that's already run vela before.

Here's the scenario. ListRegistryConfig() in references/cli/registry.go only writes the new default URL when ~/.vela/centers/config.yaml doesn't exist yet. If someone's already run any vela command in the past even just vela registry ls that file is sitting there with the old entry baked in: {Name: "default", URL: "oss://registry.kubevela.net/"}. Your change fixes the comparison from URL-based to name-based, which is correct and stops it from creating a duplicate "default" entry. But it still only checks whether a "default" entry exists, not whether its URL is actually current. So on any machine with a pre-existing config file, the default stays pointed at the dead OSS URL even after upgrading the CLI. That's basically the exact bug this PR is supposed to close, just showing up for existing users instead of new ones.

To close that gap, ListRegistryConfig() needs one more check: if the existing "default" entry's URL still equals the old literal (oss://registry.kubevela.net/), overwrite it with the new URL and persist it back via StoreRepos. I'd keep the comparison narrow only replace it when the URL matches that exact old string, so we don't stomp on anyone who's deliberately pointed "default" at something else (some setups do repoint it to an internal mirror).

Could you test this locally with a config.yaml that already has the old URL in it, to simulate an existing install, and confirm it gets corrected once the change is in? it'd help to run through all of these to make sure nothing regressed:
- vela registry ls
- vela registry config <name> <url>
- vela registry remove <name>
- vela comp --discover
- vela comp get <name>
- vela trait --discover
- vela trait get <name>
Just want to make sure the upgrade path actually gets covered, not just fresh clusters.

kash2104:
Hey @vaibhav0096 , Have made the changes to update the old URL in the existing ones and have checked with a config file and the listed cli commands as well.

PR #7205, opened 2026-06-30
fix: podDisruptive has no controller effect in TraitDefinition

### LINKED ISSUE #7204: TraitDefinition podDisruptive field is metadata-only, controller never acts on it
`podDisruptive: true` in a `TraitDefinition` has no effect. The field is stored in the spec but the controller reconciliation path never reads it — there is no code between the field and any pod restart logic. Pods only restart if the trait's CUE template happens to patch `spec.template` directly. The flag itself does nothing.

## Docs contradiction

The [x-definition page](https://kubevela.io/docs/platform-engineers/oam/x-definition) says:

> If the field is `true`, then it will cause the pod to disruptive and restart when the trait is added or updated.

The [patch-trait page](https://kubevela.io/docs/platform-engineers/traits/patch-trait) says:

> make it to be true to **tell users** that applying this trait will cause the pod to restart.

One page says it causes restarts. The other says it informs users. The code confirms neither.

## Reproduction

A trait with `podDisruptive: true` that creates a ConfigMap via `outputs:` without a `patch:` block. Updating the ConfigMap data in the Application does not restart the pod — the pod UID is unchanged before and after.

## Workaround

Put a content hash in the ConfigMap name (same pattern as Kustomize's `configMapGenerator`). The volume's `configMap.name` is part of the pod template, so when data changes the hash changes, the name changes, and Kubernetes issues a rolling update automatically. The old ConfigMap is GC'd by KubeVela's ResourceTracker since its output key also changes.

```cue
import (
    "strings"
    "encoding/json"
    "crypto/sha256"
    "encoding/hex"
)

_hash: {
    for i, _ in parameter.files {
        "\(i)": strings.SliceRunes(
            hex.Encode(sha256.Sum256(json.Marshal(_data["\(i)"]))),
            0, 8,
        )
    }
}

_cmName: {
    for i, _ in parameter.files {
        "\(i)": "\(_baseRef["\(i)"])-\(_hash["\(i)"])"
    }
}

// volume name stays stable; only configMap.name is versioned
_volumesPatch: [for i, _ in parameter.files {
    name: "\(_baseRef["\(i)"])"
    configMap: name: "\(_cmName["\(i)"])"
}]
```

## Fix options

1. **Implement it** — when a trait with `podDisruptive: true` renders a changed output, patch `spec.template.metadata.annotations` with a revision hash to trigger a rolling update.
2. **Document it as informational** — correct the x-definition page, update the field comment in `core_types.go`, and document the content-hash pattern as the recommended workaround.

## Environment

- KubeVela master (`52aaa5583`), k3d 5.8.3, Kubernetes v1.31.5+k3s1

Smitbafna:

This PR clarifies the behavior of the podDisruptive field in TraitDefinition.

Currently, podDisruptive: true is treated as metadata only and has no effect on the controller reconciliation logic. The controller does not read or act on this field, and therefore it does not trigger pod restarts.

This PR:

Documents the actual runtime behavior of podDisruptive
Resolves the mismatch between documentation claims and implementation
Clarifies that pod restarts only occur when the rendered workload template changes (e.g., via spec.template updates), not due to this field

Fixes #7204 

I have:

Verified by reproducing the behavior using a TraitDefinition with podDisruptive: true that modifies ConfigMap data without affecting spec.template. Pod UID remains unchanged before and after updates, confirming no restart is triggered by this field.

The current implementation does not include any controller logic tied to podDisruptive. This PR focuses on aligning documentation with actual behavior and highlighting the discrepancy.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7205?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apis/core.oam.dev/v1beta1/core_types.go
+++ b/apis/core.oam.dev/v1beta1/core_types.go
@@ -115,7 +115,13 @@ type TraitDefinitionSpec struct {
 	// +optional
 	WorkloadRefPath string `json:"workloadRefPath,omitempty"`
 
-	// PodDisruptive specifies whether using the trait will cause the pod to restart or not.
+	// PodDisruptive specifies whether applying this trait will cause the pod to restart.
+	// When set to true, the controller computes a hash of the trait's rendered output and
+	// injects it into the workload's spec.template.metadata.annotations. If the trait output
+	// changes on a subsequent reconcile, the hash changes and Kubernetes performs a rolling
+	// update of the pods. This works for any workload kind that has a spec.template path
+	// (e.g. Deployment, StatefulSet, DaemonSet). The field is ignored if the workload does
+	// not have a spec.template path.
 	// +optional
 	PodDisruptive bool `json:"podDisruptive,omitempty"`
 
--- a/pkg/controller/core.oam.dev/v1beta1/application/generator.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/generator.go
@@ -18,6 +18,8 @@ package application
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"hash/fnv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


roguepikachu:
Closing this PR in favor of assigning the issue to someone else. Going forward, please make sure you're assigned to an issue before starting work on it. Thanks!

PR #7207, opened 2026-06-30
Fix: nil pointer panic in getMatchingDefinitionRevision on malformed revision version

anxkhn:

`getMatchingDefinitionRevision` (`pkg/oam/util/helper.go`) dereferenced the
parsed semver result before checking the parse error:

```go
v, err := semver.NewVersion(version)
orignalVersions[v.String()] = version   // v.String() runs before err is checked
if err != nil {
    return "", err
}
```

When a `DefinitionRevision` name carries the expected prefix (`<name>-`) but the
version segment is not valid semver, `semver.NewVersion` returns `(nil, err)`, so
`v.String()` panics with a nil pointer dereference instead of returning the error.
This path runs whenever `GetLatestDefinitionRevisionName` resolves a definition
revision (the `autoUpdate` / `@vX` definition-resolution lane), so a single
malformed revision name crashes the caller rather than surfacing an error.

The fix moves the existing `if err != nil { return "", err }` check ahead of the
map write. Valid input is unaffected; the only change is that a bad version now
returns the error as the code already intended.

I have:

(This is an internal robustness fix with no new feature or configuration option,
so no docs change is needed. I ran the targeted checks below rather than the full
`make reviewable` chain.)

Added a table case to `TestGetLatestDefinitionRevisionName` in
`pkg/oam/util/helper_test.go` with a revision named `configmap-component-v1.bad`
(prefix matches, version segment is invalid semver). Before the fix this case
panics at `semver.(*Version).String` (`helper.go:357`); after it, the call returns
`error finding definition revision for Name: configmap-component, Type: Component`
(the generic error is expected because `GetLatestDefinitionRevisionName` swallows
the inner semver error and falls through both namespaces, matching the existing
trait/component cases).

```
go test ./pkg/oam/util/... -count=1                                  # PASS (mock client, no cluster/envtest)
go vet ./pkg/oam/util/                                               # clean
gofmt -l pkg/oam/util/helper.go pkg/oam/util/helper_test.go          # empty
goimports -local github.com/oam-dev/kubevela -l <changed files>      # empty
```

Single-line reorder plus one test. No signature, behavior, or error-string change
for valid input; the only difference is that a malformed version segment now
returns the error instead of triggering a nil pointer dereference.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/oam/util/helper.go
+++ b/pkg/oam/util/helper.go
@@ -354,10 +354,10 @@ func getMatchingDefinitionRevision(exactRevisionName, definitionName string, rev
 		if strings.HasPrefix(revision.Name, revisionPrefix) {
 			version := strings.Split(revision.Name, definitionName+"-")[1]
 			v, err := semver.NewVersion(version)
-			orignalVersions[v.String()] = version
 			if err != nil {
 				return "", err
 			}
+			orignalVersions[v.String()] = version
 			definitionVersions = append(definitionVersions, v)
 		}
 	}
--- a/pkg/oam/util/helper_test.go
+++ b/pkg/oam/util/helper_test.go
@@ -992,6 +992,11 @@ func TestGetLatestDefinitionRevisionName(t *testing.T) {
 		defRevisionList.DeepCopyInto(list.(*v1beta1.DefinitionRevisionList))
 		return nil
 	}}
+	malformedListCli := test.MockClient{MockList: func(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+		defRevisionList := getMalformedComponentDefRevisionList()
+		defRevisionList.DeepCopyInto(list.(*v1beta1.DefinitionRevisionList))
+		return nil
+	}}
 
 	testcases := []struct {
 		name                    string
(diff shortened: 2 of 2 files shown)


kunalworldwide (commented):
The fix is correct — `v.String()` must not run until `err != nil` is checked. The test case with the malformed revision name covers the exact panic path.

One minor edge case to consider: what happens if the version segment is empty (e.g., a `DefinitionRevision` named exactly `mydef-`)? `semver.NewVersion("")` will return an error, so the current fix handles it, but worth adding to the test if you want to be exhaustive.

LGTM otherwise.

anxkhn:
Good catch, I chased this one down. The empty-version-segment case is already safe, for two reasons.

First, a name like `mydef-` never reaches the parser: to enter the parsing branch the revision has to satisfy `strings.HasPrefix(revision.Name, exactRevisionName+".")`, and a name whose segment after the last hyphen is empty has no `.` to match that prefix, so it is filtered out and the lookup returns the normal "definition revision not found" error.

Second, even if it did reach the parser, this file imports `github.com/Masterminds/semver` v1.5.0, whose `NewVersion` regex requires at least one digit for the major version. I verified against that exact module version that `semver.NewVersion("")` returns `(nil, "Invalid Semantic Version")`, so the same `err != nil` guard from this fix rejects it before any dereference. No panic on either path.

To lock this in, I added a regression case (`configmap-component-`, empty segment) alongside the malformed one; it asserts the clean error and no panic. `go test ./pkg/oam/util/` passes.

kunalworldwide:
@anxkhn thanks for chasing that down. The prefix-filter explanation makes sense — `mydef-` never reaches the parser because it lacks the `.` after the revision name. And the semver v1.5.0 regex guard for non-empty major version is a good backstop even if it did. Adding the empty-segment regression case is the right call.

Reetika2509:
Approved. Reverted the fix locally and reproduced the exact crash it describes, then confirmed the fix resolves it cleanly. Also double-checked the second commit's added test case and confirmed its reasoning holds up.
 
 Ready to be merged.

briankane (approved):

Issue #7105, opened 2026-04-28
[LFX Mentorship] Introduce CRD-based Config Management (`config.oam.dev/v1alpha1`) with server-side validation


Chaitanyareddy0702:
## What would you like to be added?

Introduce a proper Kubernetes API group, `config.oam.dev/v1alpha1`, with two new CRDs — `ConfigTemplate` and `Config` — backed by controllers and validating webhooks. This replaces the current pattern of storing templates and configs as labeled ConfigMaps and Secrets, while preserving full backward compatibility.

```yaml
# Today (still works after this change)
kind: ConfigMap
metadata:
  name: config-template-image-registry
  labels:
    config.oam.dev/catalog: velacore-config

# After this change
apiVersion: config.oam.dev/v1alpha1
kind: ConfigTemplate
metadata:
  name: image-registry
spec:
  template: |
    # CUE template
    ...
---
apiVersion: config.oam.dev/v1alpha1
kind: Config
metadata:
  name: my-registry
spec:
  templateRef:
    name: image-registry
  propertiesFrom:
    secretRef:
      name: my-registry-credentials
      key: properties
```

## Why is this needed?

KubeVela's Config Management today stores `ConfigTemplate`s as labeled ConfigMaps (`config-template-*` prefix) and `Config`s as labeled Secrets. This has three real problems:

1. **No proper Kubernetes API types.** `kubectl get configs` returns nothing meaningful. There's no schema for users, no discoverability, no `kubectl explain`, no integration with kubectl plugins or GitOps tooling.

2. **All validation runs client-side.** CUE schema checks and `template.validation.$returns` run in the `vela` CLI. Configs created through any other path — workflow steps, VelaUX, direct `kubectl apply` of a Secret — bypass validation entirely. The same template enforces different rules depending on entry point.

3. **No native sensitive-property handling.** The current model stores property values directly in Secrets, which conflates "the resolved/materialized output" with "the user-supplied input."

### 1. New API group `config.oam.dev/v1alpha1`

Two CRDs, defined in `apis/config.oam.dev/v1alpha1/`:

```go
type ConfigTemplateSpec struct {
    Template    string `json:"template"`             // CUE template
    Scope       string `json:"scope,omitempty"`      // "system" | "namespace"
    Sensitive   bool   `json:"sensitive,omitempty"`
    Alias       string `json:"alias,omitempty"`
    Description string `json:"description,omitempty"`
}

type ConfigSpec struct {
[...]

PR #7211, opened 2026-07-02
Feat: Introduce configtemplate and config crds along with controllers

anishbista60:
-  PR for #7105 

Assited by: Claude <noreply@anthropic.com>

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7211?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Reetika2509:
Thanks @anishbista60 — this is a substantial addition (new CRDs + controllers + webhooks). A few things I'd like to mention:
 1. **Missing design proposal**:  This is a big feature, and changes like this (new controllers / project-level features) are expected to come with a design proposal first. Could you add one under design/vela-core/ using design/TEMPLATE.md. A short write-up of the motivation, goals/non-goals, and how Config/ConfigTemplate are meant to work would make it much easier to review the implementation.
 2. **No CR examples in docs**:  There's no example manifest for the new Config / ConfigTemplate CRs. The repo already has a docs/examples/config/ directory — could you add worked examples there (a sample ConfigTemplate and a Config that consumes it) so users have a clear reference to get started.
3. **API group placement**:  Today all the APIs live under apis/core.oam.dev/, and this PR adds a new group apis/config.oam.dev/. Was that a deliberate choice? A separate group is more to maintain, so unless there's a specific reason to keep it apart, it might be worth grouping these types under the existing core.oam.dev alongside the other APIs.

anishbista60:
> Thanks @anishbista60 — this is a substantial addition (new CRDs + controllers + webhooks). A few things I'd like to mention:
> 
> 1. **Missing design proposal**:  This is a big feature, and changes like this (new controllers / project-level features) are expected to come with a design proposal first. Could you add one under design/vela-core/ using design/TEMPLATE.md. A short write-up of the motivation, goals/non-goals, and how Config/ConfigTemplate are meant to work would make it much easier to review the implementation.
> 2. **No CR examples in docs**:  There's no example manifest for the new Config / ConfigTemplate CRs. The repo already has a docs/examples/config/ directory — could you add worked examples there (a sample ConfigTemplate and a Config that consumes it) so users have a clear reference to get started.
> 3. **API group placement**:  Today all the APIs live under apis/core.oam.dev/, and this PR adds a new group apis/config.oam.dev/. Was that a deliberate choice? A separate group is more to maintain, so unless there's a specific reason to keep it apart, it might be worth grouping these types under the existing core.oam.dev alongside the other APIs.

Hi, Thanks for suggestion. 

All the things has been communicated internally in slack with my mentors @jerrinfrancis @Chaitanyareddy0702. 

This is part of LFX mentorship. 


But happy to share all the things in this github comment.

anishbista60:
Hi, @Reetika2509. Please find the KEP for this PR.

https://github.com/kubevela/kubevela/pull/7308

anishbista60:
# How I tested this PR  for each part taking different scenario

## CLI testing 

Create the test template file first:

```
cat > /tmp/registry-template.cue <<'EOF'
metadata: {
	name:        "test-registry"
	alias:       "Test Registry"
	description: "test config template"
	sensitive:   false
}
template: {
	parameter: {
		username: string
		password: string
	}
	output: {
		apiVersion: "v1"
		kind:       "Secret"
		stringData: {
			username: parameter.username
			password: parameter.password
		}
	}
}
EOF
```

### Template lifecycle

Apply the CUE template, then verify:

```
./bin/vela config-template apply -f /tmp/registry-template.cue --name test-registry -n vela-system
# the config template test-registry applied successfully

./bin/vela config-template list -A
./bin/vela config-template show test-registry -n vela-system   # renders parameter docs
./bin/vela config-template delete test-registry -n vela-system # prompts unless -y
```

### Config with a template

```
./bin/vela config create my-config --template=test-registry username=foo password=bar -n vela-system
# the config my-config applied successfully

./bin/vela config list -A
./bin/vela config create preview --template=test-registry username=foo password=bar --dry-run

./bin/vela config delete my-config -n vela-system
```

### Config without a template (raw properties, no CUE rendering)

```
cat > /tmp/props.yaml <<'EOF'
username: foo
password: bar
EOF
./bin/vela config create my-plain-config -f /tmp/props.yaml -n vela-system
```

In this case `spec.templateRef` is nil and the properties land directly in `spec.properties`.

## Webhook testing (examples and expected behavior)

ConfigTemplate with invalid CUE is denied:

```yaml
apiVersion: config.oam.dev/v1alpha1
kind: ConfigTemplate
metadata:
  name: ct-bad
  namespace: vela-system
spec:
  template: "this is not valid cue {{{"
```

Expected: `denied the request: invalid template: <cue error> (requestUID=...)`. Valid CUE is allowed.

Config with properties and propertiesFrom together is always denied:

```yaml
apiVersion: config.oam.dev/v1alpha1
kind: Config
metadata:
  name: cfg-bad
  namespace: vela-system
spec:
  properties:
    username: alice
  propertiesFrom:
    secretRef:
      name: any-secret
```

Expected: `spec.properties and spec.propertiesFrom are mutually exclusive (requestUID=...)`

Config with a schema mismatch (requires an Available `test-registry` template expecting string username/password):

```yaml
apiVersion: config.oam.dev/v1alpha1
kind: Config
metadata:
  name: cfg-schema-bad
  namespace: vela-system
spec:
  templateRef:
    name: test-registry
    namespace: vela-system
  properties:
    username: 12345   # wrong type
```

Expected: `properties do not match template schema: <err> (requestUID=...)`

## Controller scenarios

**Scenario 1 - inline properties**

```yaml
apiVersion: config.oam.dev/v1alpha1
kind: ConfigTemplate
metadata:
  name: username-template
  namespace: default
spec:
  scope: namespace
  sensitive: false
  alias: "Username Secret Template"
  description: "renders a Secret with a username field"
  template: |
    template: {
      parameter: {
        username: string
      }
      output: {
        apiVersion: "v1"
        kind:       "Secret"
        stringData: {
          username: parameter.username
        }
      }
    }
---
apiVersion: config.oam.dev/v1alpha1
kind: Config
metadata:
  name: my-config
  namespace: default
spec:
  templateRef:
    name: username-template
    namespace: default
  properties:
    username: alice
```

Verify:

```
kubectl get configtemplate username-template -n default -o jsonpath='{.status.phase}'
# Available

kubectl get config my-config -n default -o jsonpath='{.status.phase}{" "}{.status.secretRef.name}'
# Available my-config

kubectl get secret my-config -n default -o jsonpath='{.data.username}' | base64 -d
# alice

kubectl get secret my-config -n default -o yaml
# ownerReference -> Config, controller:true
```

**Scenario 2 - propertiesFrom a Secret**

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: props-src
  namespace: default
stringData:
  properties: '{"username":"bob"}'
---
apiVersion: config.oam.dev/v1alpha1
kind: Config
metadata:
  name: my-config-from-secret
  namespace: default
spec:
  templateRef:
    name: username-template
    namespace: default
  propertiesFrom:
    secretRef:
      name: props-src
      # key omitted -> defaults to "properties"
```

Verify: Secret `my-config-from-secret` has `username: bob`.

**Scenario 3 - invalid CUE, ConfigTemplate errors**

```yaml
apiVersion: config.oam.dev/v1alpha1
kind: ConfigTemplate
metadata:
  name: broken-template
  namespace: default
spec:
  template: "this is not valid cue {{{"
```

Expected: `status.phase: Error`, with a `status.conditions` reason describing the parse failure.

**Scenario 4 - mutually exclusive properties + propertiesFrom**

```yaml
apiVersion: config.oam.dev/v1alpha1
kind: Config
metadata:
  name: bad-config-both
  namespace: default
spec:
  templateRef:
    name: username-template
  properties:
    username: alice
  propertiesFrom:
    secretRef:
      name: props-src
```

Expected: `status.phase: Error`, condition message containing "mutually exclusive". Accepted by the API server (no webhook in play), rejected only by the controller.

**Scenario 5 - templateRef points at a template that isn't ready / doesn't exist**

```yaml
apiVersion: config.oam.dev/v1alpha1
kind: Config
metadata:
  name: config-no-template
  namespace: default
spec:
  templateRef:
    name: does-not-exist
  properties:
    username: alice
```

Expected: `status.phase: Error`, message like "...not Available yet" (or "not found"). Then create the referenced ConfigTemplate afterward and confirm the Config gets re-reconciled automatically (via the Watches mapping) and flips to Available without touching the Config again.

**Scenario 6 - legacy fallback via ConfigMap (no ConfigTemplate CR)**

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-template-legacy-bar
  namespace: default
  labels:
    config.oam.dev/catalog: velacore-config
data:
  template: |
    template: {
      parameter: { username: string }
      output: {
        apiVersion: "v1"
        kind:       "Secret"
        stringData: { username: parameter.username }
      }
    }
---
apiVersion: config.oam.dev/v1alpha1
kind: Config
metadata:
  name: legacy-config
  namespace: default
spec:
  templateRef:
    name: legacy-bar   # prefix "config-template-" is stripped/added automatically
  properties:
    username: dave
```

Expected: Resolves via the ConfigMap fallback path; Secret has `username: dave`.

**Scenario 7 - deleting a Config garbage-collects its Secret**

```
kubectl delete config my-config -n default
kubectl get secret my-config -n default
# should be NotFound (owned + GC'd, no finalizer needed)
```

**Scenario 8 - sensitive: true + required scope enum validation**

```yaml
apiVersion: config.oam.dev/v1alpha1
kind: ConfigTemplate
metadata:
  name: sensitive-template
  namespace: default
spec:
  scope: system   # only "system" | "namespace" allowed — try "cluster" to see apiserver-level rejection
  sensitive: true
  template: |
    template: {
      parameter: { password: string }
      output: {
        apiVersion: "v1"
        kind:       "Secret"
        stringData: { password: parameter.password }
      }
    }
```

### Questions for the community on implementation choices

The first three questions relate to Scenario 6 (legacy ConfigMap fallback):

1. Right now my code only resolves the legacy ConfigMap fallback when it lives in `vela-system` — I matched this to how the CLI sets things up. Should this stay hardcoded to `vela-system`, or should I make it look up the ConfigMap in the same namespace as the Config instead?

2. Currently, if the Config is created before the ConfigTemplate exists (or the template moves namespaces), the Config's status shows "template doesn't exist" and doesn't self-heal — the user has to delete and reapply the Config to pick up the change. I didn't wire up a watch/re-reconcile for this legacy ConfigMap path the way I did for the native ConfigTemplate CR in Scenario 5. Should I add that reconciliation here too, or is matching the CLI's current (non-self-healing) behavior acceptable for this path?

3. My controller currently does not delete the Config/Secret when the underlying ConfigMap template is deleted — I left that as a no-op. Should I add cleanup logic here, or is it fine to leave the Config/Secret orphaned in that case?

4. In the output Secret, I'm currently writing both the raw properties blob and the individual resolved field, e.g.:

   ```
   properties: '{"username":"bob"}'
   username: "bob"
   ```

   Should I drop the `properties` key from the rendered Secret and only emit the resolved fields (`username`), or is keeping both intentional/useful?

anishbista60:
> Thanks @anishbista60 — this is a substantial addition (new CRDs + controllers + webhooks). A few things I'd like to mention:
> 
> 1. **Missing design proposal**:  This is a big feature, and changes like this (new controllers / project-level features) are expected to come with a design proposal first. Could you add one under design/vela-core/ using design/TEMPLATE.md. A short write-up of the motivation, goals/non-goals, and how Config/ConfigTemplate are meant to work would make it much easier to review the implementation.

Done, https://github.com/kubevela/kubevela/pull/7308

> 2. **No CR examples in docs**:  There's no example manifest for the new Config / ConfigTemplate CRs. The repo already has a docs/examples/config/ directory — could you add worked examples there (a sample ConfigTemplate and a Config that consumes it) so users have a clear reference to get started.

All the example are in the Github comment. Will follow up and add it in next PR. Current one is already bloated. 


> 3. **API group placement**:  Today all the APIs live under apis/core.oam.dev/, and this PR adds a new group apis/config.oam.dev/. Was that a deliberate choice? A separate group is more to maintain, so unless there's a specific reason to keep it apart, it might be worth grouping these types under the existing core.oam.dev alongside the other APIs.

It wasn't deliberate, I  followed the `apiVersion: config.oam.dev/v1alpha1` naming from the original issue proposal.

jerrinfrancis:
@anishbista60 
int parameters can't be rendered through the Config CR

  This ConfigTemplate:

  ```
apiVersion: config.oam.dev/v1alpha1
  kind: ConfigTemplate
  metadata:
    name: int-demo
  spec:
    scope: namespace
    template: |
      context: {
          name:      string
          namespace: string
      }
      template: {
          parameter: {
              count: int
              port:  *5432 | int
          }
          output: {
              apiVersion: "v1"
              kind:       "Secret"
              metadata: {
                  name:      context.name
                  namespace: context.namespace
              }
              stringData: {
                  count: "\(parameter.count)"
                  port:  "\(parameter.port)"
              }
          }
      }
```

  with a Config carrying properties: {count: 3} never reaches Available, and no Secret
  is created:

  failed to render config template: Field: count Message: conflicting values int and 3
  (mismatched types int and float)

  Cause. resolveProperties decodes the properties with json.Unmarshal into
  map[string]interface{}, so every JSON number arrives as float64 and CUE rejects it
  against count: int:

  - pkg/controller/config.oam.dev/v1alpha1/config/config_controller.go:178 (inline
    spec.properties) and :196 (spec.propertiesFrom)
  - the same logic is duplicated in
    pkg/webhook/config.oam.dev/v1alpha1/config/validating_handler.go:153 / :171, so
    with the webhook installed the Config is denied at admission with the same message
    rather than reaching Error. Both copies need the fix, which is an argument for
    sharing one helper.

  Scope. This isn't purely a regression — vela config create -f props.yaml hits it too,
  since references/cli/config.go:314 uses sigs.k8s.io/yaml, which also yields float64.
  Only key=value args work, because helm's strvals produces int64 (:322). What makes
  it blocking for this PR is that the CR has no key=value equivalent: in --config-mode 
  crd the CLI marshals the strvals result into spec.properties and the controller
  decodes it again, so there is no input form that renders an int parameter on the
  declarative path.

**Comment by jerrinfrancis** [comment]
Hi @anishbista60  I took some time to do extensive testing on this PR. Here are the
important issues that need to be addressed before we can merge. These are the
ones I've verified manually and that we've already gone through separately; I'm
collecting them here so nothing gets lost between now and the fix. Happy to go
deeper on any of them — I have a reproduction for each.

Line references are against the PR head.

### 1. Admission contract

The issue asks for these to be rejected at admission. Today they are accepted and surface later, or not at all.

- [ ] **Sensitive templates accept inline credentials** — P0, security. With `sensitive: true` on the template, a Config carrying inline `spec.properties` is admitted and the credential is stored in the CR as plaintext. `vela config create` already does the right thing and routes it to a companion Secret (`references/cli/config_crd.go:147-158`), but nothing enforces that for `kubectl` or GitOps — the webhook never looks at sensitivity. Suggest denying at admission, and repeating the check in `Reconcile` for the case where the template isn't `Available` yet.

- [ ] **`properties` and `propertiesFrom` are admitted together** — P1. `validating_handler.go:67-69` explicitly returns allowed for this combination, and that early return also short-circuits schema validation and output decoding for those objects. The reconciler then fails them to `Error`, so the first feedback is a status field. A CEL rule on `ConfigSpec` would enforce it even where the webhook isn't installed.

- [ ] **`spec.templateRef` is mutable** — P1. Patching it re-renders in place and rewrites the `config.oam.dev/type` label on the same Secret, silently re-classifying a live config — same name, same uid, different meaning. The imperative path refuses this (`ErrChangeTemplate`, `pkg/config/factory.go:661-665`) and has a test for it (`factory_test.go:300`). Suggest rejecting the change on update and keeping delete-and-recreate as the supported path.

### 2. Rendering and feature gaps

- [ ] **`template.outputs` is silently dropped** — P0 feature gap. `renderSecret` reads only `template.output` (`config_controller.go:225`). A template that also emits extra objects reaches `Available` having created just the Secret: no warning, no event, no `objects-reference` key. The CLI's fallback to legacy storage then can't be managed, because `list` and `delete` are CRD-only. If this won't be implemented now, please fail the Config rather than reporting success.

- [ ] **`int` parameters can't be satisfied** — P1. `resolveProperties` decodes with `json.Unmarshal` into `map[string]interface{}`, so JSON numbers arrive as `float64` and CUE rejects them against `count: int` (`config_controller.go:178`, duplicated in the webhook at `:153`). Worth flagging that `UseNumber()` on its own makes it worse — it needs a follow-up int64/float64 conversion pass, and the two copies should share one helper.

### 3. Lifecycle and cleanup

- [ ] **The distribution Application is orphaned** — P1. Recall is implemented only in the CLI (`references/cli/config.go:684-687`). `kubectl delete config` removes the CR and its owned Secret but leaves `distribute-<name>` behind, keeping the distributed copies alive in the target namespaces. GitOps deletes the CR directly, so this is now the normal path rather than an unusual one. Suggest owning the Application from the Config, or moving the targets into the CR so the reconciler owns it outright.

---

Thanks for the work on this — the core reconcile path behaves well in testing: CRD-first template lookup with the legacy ConfigMap fallback, Secret ownership, drift repair, dependency watches, and recovery once a missing dependency appears. The items above are about the contract at the edges and parity with the imperative path.

File: apis/config.oam.dev/v1alpha1/config_types.go [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// ConfigTemplateReference references a ConfigTemplate, falling back to a legacy
+// config-template-<name> ConfigMap if no CRD with this name exists.
+type ConfigTemplateReference struct {
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// Namespace defaults to "vela-system" if empty.
+	Namespace string `json:"namespace,omitempty"`
```
briankane, line 34: I think we need to secure against cross-namespace usage (not done in current solution afaik)
Probably `scope: system` should be restricted to `vela-system` namespace (validate on admission).  Doesn't look like `scope` ever did anything in the old solution. 

If scope == system, it should be universally available. If scope != system, then the user should need get permissions to the associated ConfigTemplate. 

Not sure how to address this in the workflow steps - one to think through

Happy to address this separately and not block the PR.

apis/config.oam.dev/v1alpha1/configtemplate_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// +kubebuilder:validation:Required
+	Template string `json:"template"`
+
+	// +kubebuilder:validation:Enum=system;namespace
+	// +kubebuilder:default=namespace
+	Scope ConfigTemplateScope `json:"scope,omitempty"`
+
+	// Sensitive marks Configs created from this template as not safe to read back
+	// through the API (e.g. CLI/workflow read operations).
+	Sensitive bool `json:"sensitive,omitempty"`
```
briankane, line 47: If we use the secretRef, should we automatically mark it sensitive (or enforce it?)

We should also validate that once marked sensitive, it cannot be changed.

briankane (approved):

briankane:
LGTM. Will defer to @jerrinfrancis for his review before we merge.
I've dropped a few comments - they can be addressed as follow-ups if we are happy to coordinate it that way.

The only other point is to test this with the addon code. I suspect contents in the `config-template` directory might break.

briankane:
Work so far lgtm. I've added some comments but we can follow these up in further PRs; provided it's done ahead of release. Please work with @jerrinfrancis 

**Also - merge is dependent on understanding with @jerrinfrancis and @anishbista60 that migration / upgrade work is to follow and ensure full compatibility with the existing system.**

Right now it looks like:
- `vela config[-template] list` will not see the legacy config[-template]s
- `vela config[-template] delete` will fail for legacy config[-template]s
- VelaUX will need some work (this is the case for a lot of stuff though)
- Need a way to migrate/adopt resources into new CRs from legacy (and potentially automated on startup)
- Nacos support doesn't seem to work via the controller - only the CLI path (will break GitOps)

I'd target some further testing around scenarios of creating existing configs on current controller -> update controller -> see the impacts

jerrinfrancis (approved):

briankane:
Merging - items to be follow up

PR #7221, opened 2026-07-06
Fix: assert the List error in multicluster e2e teardown

anxkhn:
copilot:all

In the multicluster e2e `AfterSuite` teardown, the final `Eventually` block
lists all `Application`s and then calls `g.Expect(err, nil)` on the `List`
error. That line asserts nothing: Gomega's `Expect(actual, extra...)` returns
an `Assertion` and only validates the extra arguments once a matcher
(`.Should`/`.To`/etc.) is invoked, so with no matcher chained the `List` error
is silently discarded. If the `List` fails, `apps` is left empty/stale and the
very next `Expect(len(apps.Items)).Should(Equal(0))` can pass for the wrong
reason, masking a real teardown failure.

This replaces `g.Expect(err, nil)` with `g.Expect(err).Should(Succeed())`,
matching the assertion style already used throughout this file (for example the
`k8sClient.List(...).Should(Succeed())` at the top of the same block and the
`g.Expect(err).Should(Succeed())` a few lines below). No product code is
touched and there is no behavior change on the happy path.

Fixes #

I have:

- [ ] Related Docs updated properly. (N/A - test-only change, no user-facing
      behavior or configuration.)
      
<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7221?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/e2e-multicluster-test/suite_test.go
+++ b/test/e2e-multicluster-test/suite_test.go
@@ -112,7 +112,7 @@ var _ = AfterSuite(func() {
 			g.Expect(k8sClient.Get(context.Background(), client.ObjectKey{Name: addon, Namespace: types.DefaultKubeVelaNS}, app)).Should(SatisfyAny(Succeed(), oamutil.NotFoundMatcher{}))
 		}
 		err := k8sClient.List(context.Background(), apps)
-		g.Expect(err, nil)
+		g.Expect(err).Should(Succeed())
 		g.Expect(len(apps.Items)).Should(Equal(0))
 	}, 5*time.Minute).Should(Succeed())
 	Eventually(func(g Gomega) {


Reetika2509:
Approving — good catch, and the fix is correct.

One thing that'll block the merge though: the DCO check, looks like the commit is missing a Signed-off-by trailer. Please run `git commit --amend -s` and a force-push before this can merge.

roguepikachu:
Hi @anxkhn, could you sign the DCO so we can get this merged? Thanks!

anxkhn:
updated the branch on current master and rewrote the commit with a verified signature and Signed-off-by trailer while preserving the one-line diff. Local compile-only test and lint pass. @Reetika2509 @roguepikachu, could you please take another look when CI finishes?

PR #7239, opened 2026-07-16
Fix: support configurable timeout via request.timeout in http builtin

### LINKED ISSUE #7119: [Feature]: Allow configurable timeout for http builtin task

Paranitharan1110:
## About

The `http` builtin task has a hardcoded 3-second client timeout with no way to override it. Slow external services or large payloads always fail with a `context deadline exceeded` error.

This PR adds support for the `request.timeout` field to override the default timeout.

## Why this approach over #7121

Per the review feedback on #7121:
- Uses `request.timeout` instead of top-level `timeout` (matches existing `#HTTPDo` schema in `kubevela/workflow`)
- No unrelated CI/Go version changes
- Silent fallback to default 3s for backward compatibility (invalid values don't break existing workflows)

## Changes

**pkg/builtin/http/http.go**
- Read optional `request.timeout` field from CUE task value
- Accepts Go duration string (`"30s"`, `"2m"`, `"500ms"`)
- Falls back to 3s default when absent or invalid (backward compatible)

**pkg/builtin/http/http_test.go**
- `TestHTTPCmdRunWithCustomTimeout`: valid timeout, short timeout (fails as expected)
- `TestHTTPCmdRunWithInvalidTimeout`: invalid duration string silently uses default

## Usage

```cue
myStep: op.#HTTPDo & {
  method: "GET"
  url: "https://slow-api.example.com/data"
  request: {
    timeout: "30s"
  }
}
```

## Testing

```bash
go test ./pkg/builtin/http/... -v
```

All tests pass.

Closes #7119
Supersedes #7121

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7239?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/builtin/http/http.go
+++ b/pkg/builtin/http/http.go
@@ -59,6 +59,16 @@ func (c *HTTPCmd) Run(meta *registry.Meta) (res interface{}, err error) {
 		}
 	)
 	if obj := meta.Obj.LookupPath(value.FieldPath("request")); obj.Exists() {
+		// Allow CUE authors to override the default 3s timeout via request.timeout
+		// (e.g., timeout: "30s"). Expects a string duration parseable by time.ParseDuration.
+		// Only positive durations are accepted; zero or negative values are ignored.
+		if v := obj.LookupPath(value.FieldPath("timeout")); v.Exists() {
+			if timeoutStr, parseErr := v.String(); parseErr == nil {
+				if d, parseErr := time.ParseDuration(timeoutStr); parseErr == nil && d > 0 {
+					client.Timeout = d
+				}
+			}
+		}
 		if v := obj.LookupPath(value.FieldPath("body")); v.Exists() {
 			r, err = v.Reader()
 			if err != nil {
--- a/pkg/builtin/http/http_test.go
+++ b/pkg/builtin/http/http_test.go
@@ -26,6 +26,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"cuelang.org/go/cue"
 	"cuelang.org/go/cue/cuecontext"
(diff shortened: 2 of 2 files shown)


roguepikachu (approved):

roguepikachu:
lgtm

briankane (approved):

briankane:
Should we follow this up and expose the setting to the builtin workflow steps?

roguepikachu:
@briankane, makes sense; I think it would be a `good-first-issue` candidate. I have created the issue to track it.

Issue #240, opened 2020-09-02
[Feature] Support dependency story in vela


wonderflow:
OAM has already support [dependency](https://github.com/crossplane/oam-kubernetes-runtime/blob/master/design/resource-dependency.md), we need to support this in vela.

wonderflow:
will be supported by the design #1642

PR #7246, opened 2026-07-19
Refactor: Use updated StepStatusCache

### LINKED ISSUE #7245: [Refactor] Use a generic cache package for  executor.StepStatusCache
The current `executor.StepStatusCache` in the `kubevela/workflow` repository is a tightly coupled implementation specific to workflow step status tracking. This limits reusability and makes it harder to maintain consistent caching patterns across the KubeVela ecosystem.

**Describe the solution you'd like:**
1. Introduce a generic [cache package](https://github.com/kubevela/pkg/issues/130) in `kubevela/pkg` repository that provides reusable caching utilities
2. Refactor [`executor.StepStatusCache`](https://github.com/kubevela/workflow/issues/239) in `kubevela/workflow` repository to use the newly introduced cache package from `kubevela/pkg`
3. Update `kubevela/kubevela` repository to use the refactored `StepStatusCache` from `kubevela/workflow`

This approach ensures:
- A single, well-maintained cache implementation across all KubeVela repositories
- Better testability and maintainability
- Consistent caching patterns for future features

**Additional context:**
This refactoring should be done in the order specified (pkg -> workflow -> kubevela) to maintain backward compatibility.

kash2104:

This PR updates the usage of `StepStatusCache` as per the methods introduced in the [pkg](https://github.com/kubevela/pkg/pull/131).

Fixes #7245 
Linked PR: https://github.com/kubevela/workflow/pull/240

I have:

Since the PR for the generic cache hasn't been merged yet, used workspaces to test this features.

1. Create a go workspace
```
go work init
go work use <workspace-folder-name>
go work use <kubevela-folder-name>
go work use <pkg-folder-name>
```
2. Build the project to ensure the workspace dependency is used.
```
go build ./...
```
Note: Avoid running go mod tidy in any repo, as it updates the module graph and throws error saying that `package cache doesn't exist in kubevela/pkg`

3. Update the sync-crds target in the KubeVela `dependency.mk` to disable the workspace when downloading the pinned module version for CRDs.
```
sync-crds: ## Copy CRD from pinned module version in go.mod
	@moddir=$$(GOWORK=off go mod download -json $(PKG_MODULE) 2>/dev/null | grep '"Dir"' | cut -d'"' -f4); \
	mkdir -p config/crd/base; \
	for file in $(COMMON_CRD_FILES); do \
		src="$$moddir/crds/$$file"; \
		cp -f "$$src" "config/crd/base/"; \
	done
```
This ensures that the CRDs are copied from the module version pinned in go.mod rather than from the local workspace checkout.

4. Build and install the updated KubeVela controller.
5. Deploy a sample application:
```
vela up -f https://kubevela.io/example/applications/first-app.yaml
```
6. Verify that the application is successfully reconciled and deployed, confirming that the local workspace version of `github.com/kubevela/pkg` and `github.com/kubevela/workflow` is being used and that the changes work as expected.

The CI pipeline is supposed to faile since this change depends on the generic cache implementation introduced in the corresponding github.com/kubevela/pkg https://github.com/kubevela/pkg/pull/131, which has not been merged yet. As a result, the cache package is not available in the published module version.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7246?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


roguepikachu:
## Phase 1 change request: finish the migration and deprecate `utils.MemoryCacheStore`

This PR migrates `StepStatusCache`, which is good, but Phase 1 is not complete until `github.com/oam-dev/kubevela/pkg/utils.MemoryCacheStore` is fully deprecated and deleted, with everything moved to `github.com/kubevela/pkg/cache.Cache[string]`. Right now the in-repo cache still exists and is still used, so we would end Phase 1 with two caches. That is the thing we are trying to avoid.

I checked every remaining consumer of the old cache (grep across the repo, excluding vendor). There are exactly two:

1. `pkg/cue/cuex/providers/helm/provider.go` (the native Helm provider, the Phase 2 subject)
2. `pkg/utils/helm/helm_helper.go` (used by apiserver)

Both must move to `pkg/cache` before `pkg/utils/cache.go` can be deleted.

### I tried the Helm provider migration locally and it works

I did the provider migration on a scratch branch (with a `replace` to a local pkg checkout), built the full controller, ran the helm unit tests, and validated it live in a k3d cluster with the auth-test ChartMuseum. Cache miss, then Put, then hit across resyncs, the `ttl: "0"` bypass path, and the auth-bound cache key all behaved exactly as before. Controller came up with zero restarts and no nil-cache panic. So the swap is mechanical. Partial references (not full code):

**Provider field + constructors** (`provider.go`):

```go
// import: drop "github.com/oam-dev/kubevela/pkg/utils", add "github.com/kubevela/pkg/cache"
type Provider struct {
    cache cache.Cache[string]   // was: *utils.MemoryCacheStore
    ...
}
// both NewProvider() and NewProviderWithConfig():
cache: cache.NewMemoryCacheStore[string](context.Background()),   // was: utils.NewMemoryCacheStore(...)
```

**The one semantic change** (`chart_fetch.go`, the cache-hit path). The old `Get` returned a single value (nil if absent or expired). The new `Get` returns `(value, expired)`:

```go
if cached, expired := p.cache.Get(cacheKey); !expired && cached != nil {
    if ch, ok := cached.(*chart.Chart); ok {
        ...
    }
}
```

(Same trap as #240: the second bool is `expired`, not `found`. This is why the pkg `Get` contract should change.)

### helm_helper.go needs the same treatment

`pkg/utils/helm/helm_helper.go` keeps its `utils` import (it also uses `utils.Sanitize`, `utils.IsValidURL`, `utils.JoinURL`), so only the cache parts change:

- field `cache *utils.MemoryCacheStore` becomes `cache cache.Cache[string]`
- constructor `utils.NewMemoryCacheStore(...)` becomes `cache.NewMemoryCacheStore[string](...)`
- the two `Get` sites (the repo-index read and the values read) become the two-value `if v, expired := ...; !expired && v != nil` form

Its cache keys are already strings (`fmt.Sprintf(repoPatten, ...)` / `valuesPatten`), so `Cache[string]` fits with no key-type change.

### Then delete the old cache

Once both consumers are off it, delete `pkg/utils/cache.go` and `pkg/utils/cache_test.go`. After that, `grep -rn MemoryCacheStore` should return nothing outside vendor, and Phase 1 is done: one cache in pkg, consumed by kubevela (Helm provider + helm_helper) and workflow (StepStatusCache).

### Minor: double init

Separately, `Setup` calls `executor.InitStepStatusCache(context.Background())` directly and also registers `stepStatusCacheInitializer` which calls it again on manager start. It is idempotent so it is harmless, but redundant. Keep the Runnable (it uses the manager context) and drop the eager call.

pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
```diff
@@ -743,6 +743,13 @@ func Setup(mgr ctrl.Manager, args core.Args) error {
 	// Register application status metrics after feature gates are initialized
 	metrics.RegisterApplicationStatusMetrics()
 
+	// Initialize the workflow StepStatusCache after manager starts
+	// This ensures that the cache is ready before any workflow execution occurs
+	if err := mgr.Add(&stepStatusCacheInitializer{}); err != nil {
+		return err
+	}
+	executor.InitStepStatusCache(context.Background())
```
roguepikachu: This eager call is redundant with the `stepStatusCacheInitializer` Runnable added just above (line 748), which calls `InitStepStatusCache` again on manager start. It is idempotent so harmless, but drop this line and keep the Runnable, which uses the manager's context instead of `context.Background()`.

pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
```diff
@@ -581,7 +581,7 @@ func (r *Reconciler) writeStatusByMethod(ctx context.Context, method method, app
 		panic("unknown method")
 	}
 	if err := f(); err != nil {
-		executor.StepStatusCache.Store(fmt.Sprintf("%s-%s", app.Name, app.Namespace), -1)
+		executor.StepStatusCache.Put(fmt.Sprintf("%s-%s", app.Name, app.Namespace), -1, time.Minute * 5)
```
roguepikachu: This is the only Application-controller consumer of `executor.StepStatusCache`, and it works because workflow now backs it with `pkg/cache`. Note for Phase 1 scope: the native Helm provider (`pkg/cue/cuex/providers/helm/provider.go`) and `pkg/utils/helm/helm_helper.go` still use the in-repo `utils.MemoryCacheStore`. Those are the two remaining consumers that need migrating to `pkg/cache.Cache[string]` before `pkg/utils/cache.go` can be deleted. Detail in the PR comment above.

pkg/cue/cuex/providers/helm/chart_fetch.go
```diff
@@ -143,7 +143,7 @@ func (p *Provider) fetchChart(ctx context.Context, params *ChartSourceParams, op
 	// a missing or malformed Secret immediately, with the same RFC-cited
 	// errors the cache-miss path would surface, instead of returning a
 	// confusing cache-hit chart for a misconfigured request.
-	if cached := p.cache.Get(cacheKey); cached != nil {
+	if cached, found := p.cache.Get(cacheKey); found && cached != nil {
```
roguepikachu, line 146: Retested this found-based lookup against the latest pkg and Workflow PR heads. The core controller builds, the affected Helm cache-hit and cache-store specs pass, and live k3d validation produced one miss, one 24-hour store, repeated hits across resyncs, and consistent bypass with `ttl: "0"`. Both test Applications reached `status: running` with healthy workloads.

roguepikachu (approved):
The KubeVela migration itself passes local builds, targeted tests, and live k3d validation. Workflow #240 still has a separate cache-initialization blocker that must be resolved before the dependency chain is ready.

pkg/controller/core.oam.dev/v1beta1/application/application_controller.go
```diff
@@ -743,6 +743,12 @@ func Setup(mgr ctrl.Manager, args core.Args) error {
 	// Register application status metrics after feature gates are initialized
 	metrics.RegisterApplicationStatusMetrics()
 
+	// Initialize the workflow StepStatusCache after manager starts
+	// This ensures that the cache is ready before any workflow execution occurs
+	if err := mgr.Add(&stepStatusCacheInitializer{}); err != nil {
```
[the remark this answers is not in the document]
kash2104: Need to bind the context of StepStatusCache with the manager to avoid go routine leaks and also the cache needs to be initialized before the workflow execution which will happen in this case.

roguepikachu:
Hey @kash2104, can we get the Codecov CI to pass and fix the linting and formatting issues? Thanks!

roguepikachu:
LGTM, great work @kash2104.

Thanks for driving this migration end to end, from wiring the updated `StepStatusCache` behavior to completing the Helm cache move cleanly. I just revalidated the live k3d helm path on this PR head, including cache miss and cache hit behavior, and it looks solid.

Really appreciate the steady follow-through and thoughtful fixes across pkg, workflow, and kubevela.

roguepikachu (approved):
Approved. Confirmed final changes and revalidated Helm behavior on a live k3d cluster.

briankane (approved):

Issue #6630, opened 2024-09-29
Chore: Upgrade go version and docker versions to mitigate security vulnerability


anoop2811:

sakirr05:
hey @anoop2811 :)
The issue doesn't list specific CVE IDs. Without that, there's a risk of upgrading versions that don't close any real attack surface while potentially introducing breaking changes. Running `govulncheck ./...` against the current tree would give me a concrete list to work from

my approach:
1. Run `govulncheck ./...` → produce a list of affected symbols + CVE IDs
2. Separate the work into two scopes:
   - **Go toolchain** (`go` directive in `go.mod` + builder image in `Dockerfile` / `Dockerfile.cli`) - patch bump only (e.g. `1.23.x`), no minor jump unless a CVE strictly requires it
   - **Docker library deps** (`docker/docker`, `docker/cli` in the `replace` block of `go.mod`) - these have their own CVE surface and should be updated independently
3. The base image `alpine:3.18` in both Dockerfiles is also a surface worth flagging `3.21` is current stable
4. After changes, re-run `govulncheck` to confirm the flagged symbols are resolved

**Worth considering:** would adding Dependabot or Renovate for Go and Docker image pins prevent this class of issue from going stale silently in the future?
Happy to pick this up if we agree on the scope

anoop2811:
Hi @sakirr05 , this is a floating target, so the aim here to keep doing the updates on a regular basis apart from just due to a critical cve and build the muscle as a community robust project. The acceptance should be based on  having enough tests and making sure we dont introduce any other regressions based on our existing acceptance tests.  I am removing golang upgrade from this ticket as we have some benchmarks we need to run before that to baseline, so this should be only about docker version.

cc: @gowthams316 @jerrinfrancis @roguepikachu

sakirr05:
Ran govulncheck on the current tree before bumping anything, and the docker situation is a bit odd.

There are 4 docker/docker (Moby) vulns that actually show up as reachable:

- GO-2026-5668 – docker cp symlink race
- GO-2026-5617 – docker cp bind-mount redirection
- GO-2026-4887 – AuthZ plugin bypass on oversized bodies
- GO-2026-4883 – off-by-one in plugin privilege validation

[...]

PR #7263, opened 2026-07-28
Chore: bump docker/docker to v28.5.2 and alpine base images to 3.21

sakirr05:
## What

- `docker/docker` replace directive 28.3.3 → 28.5.2
- alpine base images: `Dockerfile` 3.18 → 3.21, `Dockerfile.cli` and `Dockerfile.e2e` 3.15 → 3.21 (digests re-pinned)

## Why

Routine docker version bump, per @anoop2811's point about keeping these current on a regular basis rather than only reacting to a critical CVE.

Being straight about what this does and doesn't do: the 4 reachable docker/docker vulns (GO-2026-5668 / 5617 / 4887 / 4883) are still reported at 28.5.2, all still `Fixed in: N/A`. There's nothing upstream to close them with yet. Details are in the govulncheck writeup on #6630. This is hygiene, not a fix.

Picked up the alpine base images in the same pass since they're part of the same image build and were well out of date - Dockerfile.cli and Dockerfile.e2e
both still on 3.15, which is past EOL. Happy to split those into a separate PR if you'd rather keep this strictly to the go.mod side.

Left the `docker/cli` v24.0.9 downpin alone. It's pinned down from the v2 graph asks for, which looks deliberate, and it reported 0 reachable vulns so there's no security reason to disturb it.

## Testing

- `go mod tidy` (idempotent), `go mod verify`, `go build ./...`, `go vet
- Unit test target: 91 packages pass. Only failure is `pkg/definition/gen_sdk`,  which needs a live Docker daemon to pull the openapi-generator image, a fails identically on a clean master checkout
- Verified the pinned digest resolves correctly  `alpine:3.21` -> `sha256:48b0309c…` → alpine 3.21.7, and the index covers amd64/arm64/386/arm/  ppc64le/riscv64/s390x for the multi-arch builds
- Built the new base layer locally to confirm the `apk add` step still works on
  3.21: ca-certificates 20260413-r0, bash 5.2.37, expat 2.8.2 all install

Worth flagging that 3.15 -> 3.21 is a six-version jump for the cli/e2e ima (musl, openssl, busybox all move), so the e2e jobs are the real gate here rather
than the unit tests.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7263?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
 # You can replace distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 # Overwrite `BASE_IMAGE` by passing `--build-arg=BASE_IMAGE=gcr.io/distroless/static:nonroot`
-FROM ${BASE_IMAGE:-alpine:3.18}
+FROM ${BASE_IMAGE:-alpine:3.21}
 # This is required by daemon connecting with cri
 RUN apk add --no-cache ca-certificates bash expat
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
 # You can replace distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 # Overwrite `BASE_IMAGE` by passing `--build-arg=BASE_IMAGE=gcr.io/distroless/static:nonroot`
-FROM ${BASE_IMAGE:-alpine:3.21}
+FROM ${BASE_IMAGE:-alpine:3.21@sha256:48b0309ca019d89d40f670aa1bc06e426dc0931948452e8491e3d65087abc07d}
 # This is required by daemon connecting with cri
 RUN apk add --no-cache ca-certificates bash expat
 
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -32,7 +32,7 @@ RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH:-amd64} \
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 # Overwrite `BASE_IMAGE` by passing `--build-arg=BASE_IMAGE=gcr.io/distroless/static:nonroot`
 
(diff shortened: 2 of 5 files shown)


sakirr05:
hey maintainers:
e2e failed twice here but the failures are unrelated to this change and differ each run first time it was the scaler/ls assertion, on retrigger that passed and instead two unrelated specs timed out waiting on reconciliation (app_resourcetracker_test.go:119, application_test.go:135). No TLS/DNS/cert errors, which is what you'd expect if the alpine bump had actually broken
something, and both image build jobs pass.  specs green on the rerun.

roguepikachu:
Hi @sakirr05 could you push an empty commit to re-trigger the CI for the intermittent failure? Thanks

sakirr05:
> Hi @sakirr05 could you push an empty commit to re-trigger the CI for the intermittent failure? Thanks

just pushed it @roguepikachu

Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
 # You can replace distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 # Overwrite `BASE_IMAGE` by passing `--build-arg=BASE_IMAGE=gcr.io/distroless/static:nonroot`
-FROM ${BASE_IMAGE:-alpine:3.18}
+FROM ${BASE_IMAGE:-alpine:3.21}
```
roguepikachu: Pin the digest to match `Dockerfile.cli` / `Dockerfile.e2e`.

```suggestion
FROM ${BASE_IMAGE:-alpine:3.21@sha256:48b0309ca019d89d40f670aa1bc06e426dc0931948452e8491e3d65087abc07d}
```

roguepikachu (changes requested):
One small follow-up before merge: pin the controller `Dockerfile` alpine digest the same way `Dockerfile.cli` and `Dockerfile.e2e` already do.

Unpinned `alpine:3.21` can float to a newer patch without a go.mod/Dockerfile diff, which is the opposite of what we want from this bump. Local smoke on k3d with alpine 3.21.7 looked fine (webservice + scaler), so this is just making the pin consistent.

Suggested change is on `Dockerfile` line 37.

sakirr05:
@roguepikachu Done, pushed. Line 37 now uses the same digest Dockerfile.cli and Dockerfile.e2e already reference, so all three match:

`FROM ${BASE_IMAGE:-alpine:3.21@sha256:48b0309ca019d89d40f670aa1bc06e426dc0931948452e8491e3d65087abc07d}`

PR #7269, opened 2026-07-29
Fix(sdk): preserve equals in gen sdk args

### LINKED ISSUE #7268: [sdk] gen_sdk language args should preserve values containing '='
### Environment

* KubeVela master
* SDK generator: `pkg/definition/gen_sdk`

### Reproduction

Pass a language argument whose value contains `=`:

```go
langArgs := []string{GoProxy=https://proxy.example.com?a=b}
args, err := NewLanguageArgs(go, langArgs)
```

### Current behavior

The parser rejects the argument because it splits on every `=` and expects exactly two parts.

### Expected behavior

The parser should treat the first `=` as the separator and preserve the rest of the value, so the example above is parsed as:

```text
GoProxy -> https://proxy.example.com?a=b
```

### Why this matters

The current implementation makes it impossible to pass through language-argument values that naturally contain `=` characters.

hcthakur2004:
## Summary

- preserve `=` characters in language-argument values by splitting on the first separator only
- add a regression test for a Go proxy value containing `=`

Fixes #7268

## Verification

- `git diff --check`
- source-level review of the parser change and test

I could not run `go test` in this environment because the bundled toolchain does not include `go`/`gofmt` on PATH.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7269?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/definition/gen_sdk/gen_sdk.go
+++ b/pkg/definition/gen_sdk/gen_sdk.go
@@ -125,7 +125,7 @@ func NewLanguageArgs(lang string, langArgs []string) (LanguageArgs, error) {
 	availableArgs := LangArgsRegistry[lang]
 	res := languageArgs{}
 	for _, arg := range langArgs {
-		parts := strings.Split(arg, "=")
+		parts := strings.SplitN(arg, "=", 2)
 		if len(parts) != 2 {
 			return nil, errors.Errorf("argument %s is not in the format of key=value", arg)
 		}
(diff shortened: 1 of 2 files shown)


roguepikachu:
Hi @hcthakur2004, could you push an empty commit to re-trigger the webhook validation CI? Thanks!

briankane (approved):

PR #7274, opened 2026-07-30
Fix: vela cluster join no longer prints false success on declined overwrite

### LINKED ISSUE #7273: vela cluster join prints a false success message when the overwrite prompt is declined
**Describe the bug**

`vela cluster join` prints a false success message when the user declines the "cluster already exists, overwrite?" prompt. The join is correctly aborted (the original cluster's credentials are untouched), but the CLI still prints `Successfully add cluster <name>, endpoint: <new endpoint>` right after "Terminated.", so the output looks like the overwrite happened when it did not.

Root cause: in `RegisterByVelaSecret` (`pkg/multicluster/cluster_management.go`), when `ClusterAlreadyExistCallback` returns `false` the function does `return nil`:

```go
if !clusterConfig.ClusterAlreadyExistCallback(clusterConfig.ClusterName) {
    return nil
}
```

The `join` command in `references/cli/cluster.go` only checks `if err != nil { return err }` after this call, then falls through to the same success path it uses for a real join. A declined overwrite and a real join both return `nil`, so the CLI cannot tell them apart.

**To Reproduce**

1. Join a cluster under some name:
   ```
   vela cluster join <kubeconfig-a> --name victim -y
   ```
   Output: `Successfully add cluster victim, endpoint: <endpoint-a>`

2. Join a different cluster under the same name, and answer `n` to the overwrite prompt:
   ```
   vela cluster join <kubeconfig-b> --name victim
   ```
   Output:
   ```
   Cluster victim already exists, do you want to overwrite it? (y/n)Terminated.
   Successfully add cluster victim, endpoint: <endpoint-b>
   ```

3. Check the actual state:
   ```
   vela cluster list
   ```
   `victim` still shows `<endpoint-a>`, confirming the overwrite never happened, even though step 2 printed a success line naming `<endpoint-b>`.

**Expected behavior**

When the user declines the overwrite prompt, the command should exit without printing a success message, or should print something explicit like "Cancelled, cluster <name> was not changed." It should never print "Successfully add cluster" for an operation that did not run.

**KubeVela Version**

Reproduced on `master` at commit `410c2a832` (branch tip as of 2026-07-30). CLI built locally from source; kubectl client v1.36.2.

**Cluster information**

Reproduced against two k3d spoke clusters (k3s v1.35.5) joined to a k3d hub cluster. Not cluster-version-specific; the bug is in CLI/controller logic, not cluster behavior.

**Additional context**

Found while stress-testing an unrelated SpokeCluster (Cluster Connect KEP) change that also register()s cluster-gateway Secrets; while confirming `vela cluster join`'s collision-handling behavior for comparison, the false-success message showed up as a side finding. It is independent of that work.

The one-line fix is to have the `join` command in `references/cli/cluster.go` distinguish "declined" from "succeeded" (for example, have `ClusterAlreadyExistCallback`'s caller check a sentinel/bool returned alongside `nil`, or have `RegisterByVelaSecret` return a typed/sentinel error for the decline case) rather than relying on the shared `nil` return.

pujitha24:
Motivation:
When `vela cluster join` is used to re-join a cluster name that already
exists and the user answers "n" to the "already exists, overwrite?"
prompt, RegisterByVelaSecret (pkg/multicluster/cluster_management.go)
returns nil for both a declined overwrite and a real successful join.
The join command in references/cli/cluster.go only checks `err != nil`,
so it falls through to the same success path and prints
"Successfully add cluster <name>, endpoint: <new endpoint>" even though
the overwrite never happened and the original cluster's credentials are
untouched. This is a false/misleading CLI message, not data loss: the
existing cluster secret is never modified when the callback declines,
since RegisterByVelaSecret returns before calling
createOrUpdateClusterSecret.

Approach:
- Add a sentinel error, ErrClusterAlreadyExistDeclined, returned by
  RegisterByVelaSecret when ClusterAlreadyExistCallback declines the
  overwrite, instead of returning nil.
- In the `join` command, check errors.Is(err, ErrClusterAlreadyExistDeclined)
  and print "Cancelled, cluster <name> was not changed." instead of
  falling through to the success message, label-adding, and
  topology-policy-update steps meant for an actual join.
- The OCM engine path is unaffected: RegisterClusterManagedByOCM never
  reads ClusterAlreadyExistCallback and already hard-errors on a name
  collision, so it never had this bug.

Validation:
Ran `go build ./...` (passes) and
`go test ./pkg/multicluster/... -run TestRegisterByVelaSecret -v`
(passes), including a new table-driven case that reproduces the bug:
before this change the case failed because RegisterByVelaSecret
returned nil on a declined overwrite; after this change it returns
ErrClusterAlreadyExistDeclined and the pre-existing cluster secret is
verified to be untouched. `go vet ./pkg/multicluster/... ./references/cli/...`
is clean.

Fixes #7273

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7274?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/multicluster/cluster_management.go
+++ b/pkg/multicluster/cluster_management.go
@@ -55,6 +55,11 @@ type ContextKey string
 // KubeConfigContext marks the kubeConfig object in context
 const KubeConfigContext ContextKey = "kubeConfig"
 
+// ErrClusterAlreadyExistDeclined is returned by RegisterByVelaSecret when the
+// cluster already exists and ClusterAlreadyExistCallback declines the overwrite.
+// Callers must not treat this as a successful registration.
+var ErrClusterAlreadyExistDeclined = errors.New("declined to overwrite existing cluster")
+
 // KubeClusterConfig info for cluster management
 type KubeClusterConfig struct {
 	FilePath        string
@@ -176,7 +181,7 @@ func (clusterConfig *KubeClusterConfig) RegisterByVelaSecret(ctx context.Context
 			return fmt.Errorf("cluster %s already exists", cluster.Name)
 		}
 		if !clusterConfig.ClusterAlreadyExistCallback(clusterConfig.ClusterName) {
-			return nil
+			return ErrClusterAlreadyExistDeclined
 		}
 		if cluster.Spec.CredentialType == clusterv1alpha1.CredentialTypeInternal || cluster.Spec.CredentialType == clusterv1alpha1.CredentialTypeOCMManagedCluster {
 			return fmt.Errorf("cannot override %s typed cluster", cluster.Spec.CredentialType)
--- a/pkg/multicluster/cluster_management_test.go
+++ b/pkg/multicluster/cluster_management_test.go
@@ -203,11 +203,12 @@ func TestRegisterByVelaSecret(t *testing.T) {
 	t.Cleanup(func() { ClusterGatewaySecretNamespace = oldNS })
 
 	testCases := []struct {
-		name      string
-		cfg       *KubeClusterConfig
-		cli       client.Client
-		expectErr bool
-		verify    func(t *testing.T, cli client.Client, cfg *KubeClusterConfig)
+		name        string
+		cfg         *KubeClusterConfig
+		cli         client.Client
+		expectErr   bool
+		expectErrIs error
+		verify      func(t *testing.T, cli client.Client, cfg *KubeClusterConfig)
 	}{
 		{
 			name: "Token and endpoint",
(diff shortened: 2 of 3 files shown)


pujitha24:
The failing checks here are not caused by this diff — flagging in case that is what is holding it up.

The `webhook-upgrade-check` job fails at "Install KubeVela CLI" with `curl: (6) Could not resolve host: kubevela.io` — a DNS failure in the runner, which then cascades into `vela: command not found` in the next step. Pre-existing workflow behaviour, unrelated to this diff.

Happy to rebase to re-trigger CI if that would help, or to dig further if you think I have this wrong.

PR #7279, opened 2026-08-04
Fix(cli): guard top info helpers against nil rest.Config

sakirr05:

copilot:all

K8SVersion, velaCorePodUsage and velaCLusterGatewayPodUsage in references/cli/top/model/info.go pass *rest.Config straight into kubernetes.NewForConfig / metrics.NewForConfig without a nil check. client-go panics on a nil config rather than returning an error, so any caller that reaches these with nil crashes instead of getting the Unknown / N/A fallback the rest of those functions already return on failure.

To be clear about scope, this is not reachable from the CLI today. launchUI resolves the client and config first and returns early if either fails, so vela top with no kubeconfig exits with "invalid configuration: no configuration has been provided" long before App.Init runs. The panic only shows up if something calls App.Init or InfoBoard.Init with a nil *rest.Config directly, which is what the new tests do.

So this is defensive hardening plus test coverage for that path, not a user-visible bug fix. Happy to close it if you'd rather not carry guards for a case no caller can hit right now.

Added three unit tests in references/cli/top/model/info_test.go covering nil config for K8SVersion, VelaCoreRatio and CLusterGatewayRatio. Reverted the guards locally to confirm each one panics with the client-go nil pointer dereference, and passes with them in place. Ran the references/cli/top/... suite, no other changes in behavior.

I have:

- [ ] [Related Docs](https://github.com/kubevela/kubevela.io) updated properly. Not applicable, no user facing change.

### DIFF
--- a/references/cli/top/model/info.go
+++ b/references/cli/top/model/info.go
@@ -86,6 +86,9 @@ func (i *Info) ClusterNum() string {
 
 // K8SVersion return k8s version info
 func K8SVersion(cfg *rest.Config) string {
+	if cfg == nil {
+		return Unknown
+	}
 	c, err := kubernetes.NewForConfig(cfg)
 	if err != nil {
 		return Unknown
@@ -141,6 +144,9 @@ func ApplicationRunningNum(cfg *rest.Config) string {
 }
 
 func velaCorePodUsage(cfg *rest.Config) (*v1beta1.PodMetrics, error) {
+	if cfg == nil {
+		return nil, errors.New("no rest.Config available")
+	}
 	ctx := context.Background()
 	c, err := metrics.NewForConfig(cfg)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


roguepikachu:
Hi @sakirr05, Thanks for the PR. The nil guards look correct, and the unit tests show the panic clearly. `vela top` with no kubeconfig does not reach this code. launchUI fails earlier in GetClient / GetConfig with:

```bash
Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

So this is not a vela top crash for users. The panic only happens if something calls App.Init / InfoBoard.Init with a nil `*rest.Config` (as the new tests do). 

Could you reword the description to match that

sakirr05:
@roguepikachu changed the title and description both!

roguepikachu (approved):

briankane (approved):

PR #7297, opened 2026-08-06
Fix: mock OSS registry server serves file bytes directly instead of through html/template

### LINKED ISSUE #7296: Bug: mock OSS registry server in e2e/addon serves binary addon packages through html/template, silently corruptible and double-writes on parse error
**Describe the bug**

`ossHandler` in `e2e/addon/mock/vela_addon_mock_server.go` is the mock OSS registry server backing the default "KubeVela" registry in the addon e2e suite (`e2e/addon/mock/utils/utils.go:41`, `oss.end_point` points at this server). Both branches have the same broken pattern:

```go
t, err := template.New("").Parse(string(data))
if err != nil {
    t.Execute(rw, error)   // only meant to run on error
}
t.Execute(rw, data)         // runs unconditionally either way
```

Two problems here.

1. The if err != nil branch doesn't stop execution. On a parse error it writes the error render, then falls straight through into the unconditional t.Execute(rw, data) right after. So it writes a second, different render into the same response body.
2. The file serving branch (ossHandler, non prefix case) parses raw file bytes, including .tgz addon package archives read via testData.ReadFile(...), as a Go template (template.New("").Parse(string(file))) before serving them. Neither template has a {{.Field}} placeholder anywhere, so none of this templating actually does anything. But binary/compressed archive content can coincidentally contain byte sequences that look like {{...}} template syntax, which html/template will either fail to parse (hitting bug #1) or interpret as a template action and corrupt the served package.

To Reproduce

1. Run the e2e/addon suite (make e2e-addon-test or ginkgo -v ./test/e2e-addon-test) against a cluster with the mock addon server running (go run ./e2e/addon/mock &).
2. Any test installing an addon whose packaged .tgz (embedded under e2e/addon/mock/testdata/) happens to contain the byte sequence {{ would hit this. None of the current fixtures trigger it, which is why it hasn't shown up as a CI failure yet. It's reproducible by adding a fixture whose compressed bytes contain that sequence, or by unit testing ossHandler directly with input that fails Parse.

Expected behavior

The handler should just write the bytes to the response (rw.Write(data) / rw.Write(file)) instead of round tripping static, possibly binary content through html/template.Parse/Execute with no placeholders to fill. Drop the templating and the broken error branch goes away with it.

KubeVela Version

master, current. This is test only infrastructure (e2e/addon/mock/vela_addon_mock_server.go), not the controller or CLI.

Cluster information

N/A, applies to any cluster running the e2e/addon suite locally or in CI.

sakirr05:

ossHandler parsed static file content as a Go template before serving it, with no placeholders actually used anywhere. On a parse failure the error branch didn't stop execution and fell through to the success render right after. Turns out worse than that: Parse returns a nil *Template on failure, so Execute on it panics.

Drops the templating, writes bytes directly. Added tests, including a fixture that reproduces the parse failure, confirmed it panics on the old code and passes with the fix.

Fixes #7296

I have:

Added vela_addon_mock_server_test.go with 4 cases, including one fixture with an unbalanced `{{` to reproduce the parse failure directly. Reverted the fix locally and confirmed that test panics with the exact trace, then confirmed it passes clean with the fix in place.

### note for reviewer

Test-only infra, no production code touched. The nil pointer panic on parse failure is worse than what the linked issue originally described, current fixtures just don't happen to trigger it, which is why it never showed up in CI.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7297?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/e2e/addon/mock/testdata/template-edge-case/broken.txt
+++ b/e2e/addon/mock/testdata/template-edge-case/broken.txt
@@ -1,0 +1,1 @@
+plain content with an unbalanced action delimiter: {{ .Whatever
--- a/e2e/addon/mock/vela_addon_mock_server.go
+++ b/e2e/addon/mock/vela_addon_mock_server.go
@@ -20,7 +20,6 @@ import (
 	"embed"
 	"encoding/xml"
 	"fmt"
-	"html/template"
 	"io/fs"
 	"log"
 	"net/http"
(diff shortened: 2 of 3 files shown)


roguepikachu:
Hi @sakirr05, the checks/tests haven't run on this PR. Could you retrigger by pushing an empty commit? Thanks!

sakirr05:
@roguepikachu pushed the empty commit

roguepikachu (approved):

roguepikachu:
LGTM. Thank you for your contribution!

briankane (approved):

PR #7299, opened 2026-08-09
Fix: defkit health conditions erroring when status is absent

### LINKED ISSUE #7284: Defkit health conditions fail when status or conditions are absent
**Describe the bug**

A health policy built with `HealthPolicyExpr` returns a CUE evaluation error while the resource still has no status. That is the normal state for the first few reconciles after a Crossplane claim is applied.

`ConditionExpr.Preamble()` emits a comprehension over `context.output.status.conditions` up front. `Exists("status")` only adds a boolean term inside `isHealth`, so it cannot guard that preamble, and CUE propagates bottom through `&&` rather than short-circuiting. The comprehension also reads `c.type` without checking whether it exists.

- [`ConditionExpr.Preamble` dereferences `status.conditions` unconditionally](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/health_expr.go#L86-L106)
- [`Exists` emits only a bottom comparison](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/health_expr.go#L248-L266)

**To Reproduce**

```go
h := defkit.Health()

defkit.HealthPolicy(h.And(
    h.Exists("status"),
    h.Exists("status.conditions"),
    h.AllTrue("Ready", "Synced"),
))
```

generates:

```cue
_readyCond: [for c in context.output.status.conditions if c.type == "Ready" {c}]
_syncedCond: [for c in context.output.status.conditions if c.type == "Synced" {c}]
isHealth: (context.output.status != _|_) &&
    (context.output.status.conditions != _|_) &&
    (len(_readyCond) > 0 && _readyCond[0].status == "True") &&
    (len(_syncedCond) > 0 && _syncedCond[0].status == "True")
```

Evaluated against five `context.output` fixtures:

| Fixture | Result |
|---|---|
| `{}` | evaluation error |
| `{status: {}}` | evaluation error |
| `Ready=True`, `Synced=True` | `true` |
| `Ready=False`, `Synced=True` | `false` |
| healthy conditions plus an entry with no `type` | evaluation error |

Put the generated fields beside `context: output: {}` and run `cue eval -c -e isHealth policy.cue`. Both comprehensions report `undefined field: status`.

**Expected behavior**

Missing status means unhealthy, not a broken policy. The first two fixtures should be `false`, and a condition entry without `type` should be skipped.

One shape that would work: default `isHealth` to false, build the condition helpers inside nested `if` blocks that run only once `status` and `status.conditions` exist, and filter on `c.type != _|_`. The layout matters less than the outcome, which is that no fixture produces bottom.

Nobody should have to work out a particular ordering of `Exists` calls to get a safe policy.

**Screenshots**

Not applicable.

**KubeVela Version**

Reproduced against `master` at [`410c2a8`](https://github.com/kubevela/kubevela/commit/410c2a832163ad598ecefdb8718dbc2b2bba4de0), the head at the time of writing. The generated policy above is what that commit produces. Also present in v1.11.0.

**Cluster information**

Not cluster-specific. It reproduces by evaluating the generated CUE.

**Additional context**

A hand-written, null-guarded `HealthPolicy(string)` avoids this, which is what we ship. The problem is confined to the expression builder.

The practical cost: when the health check errors, `GetStatus` logs a warning and returns `(false, err)`, so the resource reads as not-ready and every reconcile logs `failed to check health` for the whole creation window.

pujitha24:

copilot:all

Fixes #

`defkit.ConditionExpr` (used by `HealthPolicyExpr`/`AllTrue`/`AnyTrue`) generates a CUE
comprehension over `context.output.status.conditions` to build health policies. When
`status` or `status.conditions` doesn't exist yet — e.g. during the first few reconciles
after a Crossplane claim is created — that comprehension source is CUE bottom (`_|_`).
CUE's `&&` does not short-circuit, so the bottom propagates through the whole `isHealth`
expression even past `Exists()` guards meant to protect it. A condition entry without a
`type` field hit the same problem. Separately, `ToCUE()` indexed the filtered list at
`[0]`, which also errors on an empty list regardless of a preceding `len(...) > 0` check,
for the same non-short-circuit reason.

The fix:
- `Preamble()` now sources the comprehension from `*context.output.status.conditions | []`.
  CUE disjunctions discard any operand that evaluates to bottom, so this falls back to an
  empty list instead of erroring when status/conditions are absent, while still ranging
  over the real list when it exists. `if c.type != _|_` skips condition entries missing
  a `type` field instead of letting the comparison go bottom.
- `ToCUE()` now checks status/reason by filtering the (always-concrete) condition list
  with `len([for c in ... if ...]) > 0` instead of indexing `[0]`, so there's never an
  out-of-range index to evaluate.

I have:

- `go build ./...` passes.
- `go test ./pkg/definition/defkit/... ./pkg/cue/definition/health/...` passes.
- Added `TestConditionExprHandlesMissingStatus` in `pkg/definition/defkit/health_expr_test.go`,
  which builds the exact policy shape from the issue repro
  (`h.And(h.Exists("status"), h.Exists("status.conditions"), h.AllTrue("Ready","Synced"))`)
  and runs it through the real production evaluation path (`health.CheckHealth`, the same
  function `GetStatus` calls) against the issue's 5 fixtures: no status, empty status,
  Ready+Synced true, Ready false, and a condition entry missing `type`.
  I verified this is a genuine regression test, not a tautology: stashing only the
  `health_expr.go` change (keeping the new test) reproduces the exact errors from the
  issue (`_readyCond: undefined field: status`, `...: undefined field: conditions`,
  `...: undefined field: type`) on 3 of the 5 subtests; unstashing makes all 5 pass.
- No live cluster / e2e run: this is pure CUE-string-generation logic with no Kubernetes
  API interaction, so the unit tests above exercise the actual `cuecontext` CUE evaluator
  end to end and are the appropriate validation tier for this change.

Impact is precise, not dramatic: before this fix, `GetStatus` (`pkg/cue/definition/health/health.go`)
already returned `healthy=false` on this error path — it just also logged
`"failed to check health"` on every reconcile during the affected window, and the error
was propagated up as a warning rather than a clean `false`. This is a recovered/logged
evaluation error that produced log spam and an imprecise error path, not an incorrect
final health verdict, a panic, or an outage. The end-user-visible behavior (resource
reads as not-ready while status is absent) is unchanged; the fix removes the spurious
error and makes the "absent status = unhealthy" case go through the normal, quiet path.

```release-note
Fixed `defkit` health policies built with `ConditionExpr`/`AllTrue`/`AnyTrue` erroring
instead of evaluating to unhealthy when the target resource has no `status` or
`status.conditions` yet.
```

Fixes #7284

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7299?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/definition/defkit/health_expr_test.go
+++ b/pkg/definition/defkit/health_expr_test.go
@@ -19,6 +19,8 @@ package defkit
 import (
 	"strings"
 	"testing"
+
+	"github.com/oam-dev/kubevela/pkg/cue/definition/health"
 )
 
 func TestConditionIsTrue(t *testing.T) {
@@ -336,6 +338,74 @@ func TestComplexComposition(t *testing.T) {
 	}
 	if !strings.Contains(policy, "!(") {
 		t.Errorf("Missing Not() expression")
+	}
+}
+
+// TestConditionExprHandlesMissingStatus reproduces the fixtures from the
+// GitHub issue: a health policy built with AllTrue must evaluate to false
+// (not a CUE evaluation error) when status or status.conditions is absent,
+// and must skip condition entries that have no type instead of erroring.
+func TestConditionExprHandlesMissingStatus(t *testing.T) {
+	h := Health()
+	policy := h.Policy(h.And(
+		h.Exists("status"),
+		h.Exists("status.conditions"),
+		h.AllTrue("Ready", "Synced"),
+	))
+
+	cases := []struct {
+		name   string
+		output map[string]interface{}
+		want   bool
+	}{
+		{
+			name:   "no status",
+			output: map[string]interface{}{},
+			want:   false,
+		},
+		{
+			name:   "empty status",
+			output: map[string]interface{}{"status": map[string]interface{}{}},
+			want:   false,
+		},
+		{
+			name: "ready and synced",
+			output: map[string]interface{}{"status": map[string]interface{}{"conditions": []interface{}{
+				map[string]interface{}{"type": "Ready", "status": "True"},
+				map[string]interface{}{"type": "Synced", "status": "True"},
+			}}},
+			want: true,
+		},
+		{
+			name: "ready false",
+			output: map[string]interface{}{"status": map[string]interface{}{"conditions": []interface{}{
+				map[string]interface{}{"type": "Ready", "status": "False"},
+				map[string]interface{}{"type": "Synced", "status": "True"},
+			}}},
+			want: false,
+		},
+		{
+			name: "condition entry without type",
+			output: map[string]interface{}{"status": map[string]interface{}{"conditions": []interface{}{
+				map[string]interface{}{"type": "Ready", "status": "True"},
+				map[string]interface{}{"type": "Synced", "status": "True"},
+				map[string]interface{}{"status": "True"},
+			}}},
+			want: true,
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			templateContext := map[string]interface{}{"output": c.output}
+			got, err := health.CheckHealth(templateContext, policy, nil)
+			if err != nil {
+				t.Fatalf("CheckHealth returned an error instead of a health verdict: %v", err)
+			}
+			if got != c.want {
+				t.Errorf("expected healthy=%v, got %v", c.want, got)
+			}
+		})
 	}
 }
 
--- a/pkg/definition/defkit/health_expr_test.go
+++ b/pkg/definition/defkit/health_expr_test.go
@@ -393,6 +393,14 @@ func TestConditionExprHandlesMissingStatus(t *testing.T) {
 			}}},
 			want: true,
 		},
+		{
+			name: "matched condition without status",
+			output: map[string]interface{}{"status": map[string]interface{}{"conditions": []interface{}{
+				map[string]interface{}{"type": "Ready"},
+				map[string]interface{}{"type": "Synced", "status": "True"},
+			}}},
+			want: false,
+		},
 	}
 
 	for _, c := range cases {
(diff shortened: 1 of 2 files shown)


pkg/definition/defkit/health_expr_test.go
krishnankm, line 1: Nice fix, 
Quick question, `TestConditionExprHandlesMissingStatus` is the first table-style test in this file (`for` loop + `t.Run`), all the other 27 are flat, one assertion per func. 
Since the package already uses Ginkgo’s `DescribeTable`/`Entry` (e.g. `cuegen_test.go`), that seemed like a natural fit for this “named fixture → expected bool” case.

Just curious, was using plain `testing.T` intentional to match the rest of the file, or simply what you reached for here? Not blocking either way. 👀

roguepikachu: I think since the test file is in plain tex this is fine. The changes LGTM.

roguepikachu (approved):

pujitha24:
It was intentional, actually — `health_expr_test.go` itself is 100% plain `testing.T` (all 27 other tests here use that style), so I matched the convention of this specific file rather than the package-wide Ginkgo `DescribeTable`/`Entry` style used in `cuegen_test.go` and others. I didn't have a strong opinion beyond "match the file it's in." Happy to convert it to `DescribeTable` if you'd rather have this test line up with the rest of the package instead.

briankane (approved):

PR #7303, opened 2026-08-12
Fix: enforce TraitDefinition conflictsWith at admission

junnhwan:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

`TraitDefinition.spec.conflictsWith` was declared on the API and emitted by defkit, but never enforced. Applications could attach conflicting traits and still be admitted.

This PR adds admission-time validation in the Application validating webhook:

- Resolve each attached trait with `GetCapabilityDefinition` (app namespace first, then `vela-system`, including `@revision` types)
- Keep the check outside `ValidateComponents` so it still runs under sharding
- Match unidirectionally across the documented rule forms: definition name, CRD name, `*.group`, and `labelSelector:`

**Which issue(s) this PR fixes**:
Fixes #7294

**Bug evidence (required for bug-related PRs)**:

Reproduction from #7294: two TraitDefinitions that list each other in `conflictsWith`, attached to the same component, are admitted and reconcile normally because nothing reads the field. Docs claim KubeVela rejects such Applications.

Relevant prior state:
- Field declaration: `apis/core.oam.dev/v1beta1/core_types.go` (`ConflictsWith`)
- No webhook/controller/CLI consumer existed before this change

**Special notes for your reviewer**:

Followed the guidance on #7294:

- Use `GetCapabilityDefinition` instead of the permission-check lookup path
- Keep validation outside `ValidateComponents`
- Unidirectional matching; CRD/group rules only apply when `spec.reference` is set

Fake-client unit tests cover rule matching and the admission check.

This PR was written in part with the assistance of generative AI.

**Does this PR introduce a user-facing change?**:
```release-note
Validate TraitDefinition.spec.conflictsWith when admitting Applications so conflicting traits on the same component are rejected.
```

Made with [Cursor](https://cursor.com)

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7303?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #7322, opened 2026-08-13
Fix: reject invalid conflictsWith label selectors

### LINKED ISSUE #7315: invalid labelSelector: in conflictsWith is ignored
**Describe the bug**

`TraitDefinition.spec.conflictsWith` now rejects conflicting traits at Application admission (#7303, follow-up to #7294). One leftover: if a `labelSelector:` rule cannot be parsed, the matcher returns false and the Application is still admitted.

A typo in the selector string makes that rule a silent no-op. The unit test currently encodes this:

```go
Entry("invalid label selector", "labelSelector:@@@", cueTrait, false)
```

in `pkg/webhook/core.oam.dev/v1beta1/application/validation_conflicts_test.go`.

**To Reproduce**

1. A TraitDefinition with a broken selector, plus a second trait that would match if the selector were valid:

```yaml
kind: TraitDefinition
metadata:
  name: label-conflict
  namespace: vela-system
spec:
  appliesToWorkloads: ["*"]
  conflictsWith: ["labelSelector:@@@"]
  schematic:
    cue:
      template: |
        patch: spec: template: metadata: annotations: "label-conflict": "applied"
        parameter: {}
---
kind: TraitDefinition
metadata:
  name: expose-labeled
  namespace: vela-system
  labels:
    feature: expose
spec:
  appliesToWorkloads: ["*"]
  schematic:
    cue:
      template: |
        patch: spec: template: metadata: annotations: "expose-labeled": "applied"
        parameter: {}
```

2. An Application attaching both:

```yaml
kind: Application
metadata:
  name: bad-selector-demo
  namespace: default
spec:
  components:
    - name: web
      type: webservice
      properties:
        image: nginx:1.25
      traits:
        - type: label-conflict
          properties: {}
        - type: expose-labeled
          properties: {}
```

3. `kubectl apply -f traits.yaml && kubectl apply -f app.yaml`

4. No error. The Application is admitted. With a valid selector such as `labelSelector:feature=expose`, the same pair is rejected.

**Expected behavior**

A `labelSelector:` value that fails `labels.Parse` should be rejected when the TraitDefinition is created or updated, with a clear admission error naming the bad rule. It should not wait until Application apply and then silently not conflict.

**Screenshots**

N/A.

**KubeVela Version**

master, after #7303.

**Cluster information**

N/A.

**Additional context**

- **Where it appears.** `traitConflictRuleMatches` in `pkg/webhook/core.oam.dev/v1beta1/application/validation.go`:

```go
case strings.HasPrefix(rule, "labelSelector:"):
    selector, err := labels.Parse(strings.TrimPrefix(rule, "labelSelector:"))
    if err != nil {
        return false
    }
```

- **Suggested fix.** Validate `labelSelector:` rules in the TraitDefinition validating webhook (`pkg/webhook/core.oam.dev/v1beta1/traitdefinition/`), next to the existing CUE / version checks. `labels.Parse` already tells you if the expression is valid.

  Optionally also stop returning false from `traitConflictRuleMatches` on parse error (fail closed on Application admission) so a bad rule that somehow got stored cannot bypass conflict checks. Definition-time rejection is the better user-facing fix.

- **Tests.** TraitDefinition webhook: admit `labelSelector:feature=expose`, deny `labelSelector:@@@`. Update or replace the Application matcher test that currently expects false for the invalid selector.

This is leftover from review of #7303. It is not a merge blocker for that PR.

HurrairaBaloch:

Reject malformed `labelSelector:` entries in `TraitDefinition.spec.conflictsWith` during both create and update admission. The validation error identifies the invalid rule, while Application conflict matching now fails closed for malformed selectors in legacy stored definitions.

Adds focused validator, webhook admission, and conflict-matching coverage.

Fixes #7315

I have:

- [x] [Related Docs](https://github.com/kubevela/kubevela.io) updated properly. No docs update is needed because the existing documented selector syntax and API are unchanged.
- [ ] Run `make reviewable` to ensure this PR is ready for review. The focused formatting, vet, and tests below were run locally; full repository checks are delegated to CI because the local environment is Windows.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary. No backport is requested.

- `go test ./pkg/webhook/core.oam.dev/v1beta1/traitdefinition -run '^TestValidateConflictsWith$' -count=1`
- Focused TraitDefinition Ginkgo admission specs: valid create plus invalid create/update (3/3 passed)
- Focused Application trait-conflict specs (19/19 passed; the Windows-only envtest teardown subsequently reports unsupported process signaling)
- `go vet ./pkg/webhook/core.oam.dev/v1beta1/traitdefinition ./pkg/webhook/core.oam.dev/v1beta1/application`
- `gofmt` and `git diff --check`

This is the follow-up identified during review of #7303. New definitions are rejected at admission; the matcher fallback protects clusters that already contain a malformed definition.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7322?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/webhook/core.oam.dev/v1beta1/application/validation.go
+++ b/pkg/webhook/core.oam.dev/v1beta1/application/validation.go
@@ -548,7 +548,9 @@ func traitConflictRuleMatches(rule string, target *v1beta1.TraitDefinition) bool
 	case strings.HasPrefix(rule, "labelSelector:"):
 		selector, err := labels.Parse(strings.TrimPrefix(rule, "labelSelector:"))
 		if err != nil {
-			return false
+			// TraitDefinition admission rejects malformed selectors. Fail closed
+			// here in case a legacy invalid definition is already stored.
+			return true
 		}
 		return selector.Matches(labels.Set(target.GetLabels()))
 	case rule == target.Name:
--- a/pkg/webhook/core.oam.dev/v1beta1/application/validation_conflicts_test.go
+++ b/pkg/webhook/core.oam.dev/v1beta1/application/validation_conflicts_test.go
@@ -66,7 +66,7 @@ var _ = Describe("Trait conflict validation", func() {
 			Entry("group wildcard ignored for empty reference", "*.k8s.io", cueTrait, false),
 			Entry("label selector match", "labelSelector:team=platform", cueTrait, true),
 			Entry("label selector miss", "labelSelector:team=edge", cueTrait, false),
-			Entry("invalid label selector", "labelSelector:@@@", cueTrait, false),
+			Entry("invalid label selector fails closed", "labelSelector:@@@", cueTrait, true),
 		)
 	})
 
--- a/pkg/webhook/core.oam.dev/v1beta1/traitdefinition/trait_definition_validating_handler.go
+++ b/pkg/webhook/core.oam.dev/v1beta1/traitdefinition/trait_definition_validating_handler.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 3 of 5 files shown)


Reetika2509:
Hey @HurrairaBaloch , we tried testing your changes on a local k3d cluster using the sample manifests provided in the associated issue. It looks like the TraitDefinition where the parsing error is expected to be caught is still getting admitted successfully.

Could you please test this scenario locally as well and verify the behavior?

Issue #7287, opened 2026-08-05
[Enhancement] Defkit: support structured value schemas for maps with dynamic keys


krishnankm:
**Which existing feature or component would you like to improve?**

`MapParam`.

**What is the current limitation or pain point?**

Two map shapes work today: dynamic keys with scalar values, and a fixed object via `WithFields`. A dynamic-key map whose values have a typed schema does not:

```cue
accessPoints?: [string]: {
    path: string
    ownerUID: *1000 | int
    ownerGID: *1000 | int
    permissions: *"0755" | string
}
```

The nearby APIs all mean something else. `Of` takes a `ParamType`. `WithFields` puts fixed fields on the map itself. `WithSchemaRef` applies to the whole parameter, so you get `accessPoints: #Type` rather than `accessPoints: [string]: #Type`.

Checked against `master` at
[`410c2a8`](https://github.com/kubevela/kubevela/commit/410c2a832163ad598ecefdb8718dbc2b2bba4de0).
`Map("ap").WithFields(...)` generates a fixed struct:

```cue
ap: {
    path: string
    uid: int
}
```

and both `Map("ap").Of(ParamTypeString)` and `StringKeyMap("ap")` generate `ap: [string]: string`.

- [`MapParam.Of` takes a ParamType](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/param.go#L859-L863)
- [`MapParam.WithFields` stores fields on the parameter](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/param.go#L896-L905)

That leaves `WithSchema` with a raw string, which also means hand-writing the descriptions, defaults and optional markers the builders would normally handle.

**Describe the improvement you'd like**

A builder for the value schema, separate from the map's own schema:

```go
defkit.Map("accessPoints").
    OfObject(
        defkit.String("path").Required(),
        defkit.Int("ownerUID").Default(1000),
        defkit.String("permissions").Default("0755"),
    ).
    Optional()
```

and something for reusable schemas:

```go
defkit.Map("accessPoints").OfSchemaRef("AccessPointConfig").Optional()
```

giving `accessPoints?: [string]: #AccessPointConfig`.

[...]

PR #7330, opened 2026-08-17
Feat: support map-to-list comprehensions in ArrayBuilder

### LINKED ISSUE #7288: [Enhancement] Defkit: support guarded map-to-list comprehensions in ArrayBuilder

`ArrayBuilder` and its `ForEach*` methods.

Array to array works. Map to map works. Map to list, with the key still available, does not.

The input is a name-keyed map, and the template needs a positional list that keeps the name:

```cue
_accessPointList: [
    if parameter.accessPoints != _|_
    for name, spec in parameter.accessPoints {
        name:        name
        path:        spec.path
        permissions: spec.permissions
    }
]
```

`ForEachWithVar` and its variants generate a single iteration variable, so they can reach the value but never the key. `ForEachMap` has both variables but always emits a struct comprehension.

- [`ForEachWithVar` accepts one iteration variable](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/array_builder.go#L114-L131)
- [`ForEachMap` generates `{for k, v in source {(k): v}}`](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/expr.go#L614-L657)

A map-entry variant on `ArrayBuilder` that exposes both:

```go
defkit.NewArray().ForEachMapWithGuarded(
    accessPoints.IsSet(),
    accessPoints,
    func(entry *defkit.MapEntryBuilder) {
        entry.Set("name", entry.Key())
        entry.Set("path", entry.Value().Field("path"))
        entry.Set("permissions", entry.Value().Field("permissions"))
    },
)
```

Names are a suggestion; matching the existing `ArrayBuilder` and `ItemBuilder` conventions matters more. It should take any map-valued `Value`, expose the key and value without raw strings, accept a guard for an unset source, and leave `ForEachMap` behaving as it does now.

**Additional context**

Raw CUE covers this, but it puts the source path, the guard, both iteration variables and every field mapping into a string nothing checks.

While looking at this, we also found that `ForEachMap().WithBody(...)` is ignored by the generator. Filed separately as #7283.

krishnankm:

Adds guarded and unguarded map-to-list comprehensions to `ArrayBuilder`.

`ForEachMapWith` and `ForEachMapWithGuarded` expose a `MapEntryBuilder`, allowing the callback to access both the map key and value without raw CUE references:

```go
items := defkit.NewArray().ForEachMapWithGuarded(
    accessPoints.IsSet(),
    accessPoints,
    func(entry *defkit.MapEntryBuilder) {
        entry.Set("name", entry.Key())
        entry.Set("path", entry.Value().Field("path"))
    },
)
 ```

The generated list comprehension binds both variables:

 ```cue
if parameter["accessPoints"] != _|_
for k, v in parameter.accessPoints {
    name: k
    path: v.path
}
 ```

MapEntryBuilder delegates the existing item operations for field assignments, conditional blocks, bindings, and defaults. The generator reuses the existing item-operation renderer and import collector.

Existing ForEachMap behavior is unchanged.

Fixes #7288

I have:

The generated CUE was compiled and evaluated against absent, empty, single-entry, and multi-entry map inputs. Tests also cover key/value references, guarded generation, existing item operations, and import discovery.

This accepts any map-valued Value and does not depend on the structured MapParam schema work in #7287. Tests use the existing WithSchema escape hatch.

Map iteration order is intentionally not asserted.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7330?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/definition/defkit/array_builder.go
+++ b/pkg/definition/defkit/array_builder.go
@@ -20,20 +20,23 @@ package defkit
 type entryKind int
 
 const (
-	entryStatic      entryKind = iota // always-present item
-	entryConditional                  // conditional item (if cond { item })
-	entryForEach                      // iterated item (for v in source { item })
+	entryStatic         entryKind = iota // always-present item
+	entryConditional                     // conditional item (if cond { item })
+	entryForEach                         // iterated item (for v in source { item })
+	entryForEachWith                     // iterated item using an ItemBuilder
+	entryForEachMapWith                  // map-to-list item using a MapEntryBuilder
 )
 
 // arrayEntry is a single entry in an ArrayBuilder.
 type arrayEntry struct {
-	kind        entryKind
-	element     *ArrayElement // the item fields (for static, conditional, forEach)
-	cond        Condition     // for conditional entries
-	source      Value         // for forEach entries (iteration source)
-	guard       Condition     // for forEach entries (optional guard: if source != _|_)
-	filter      Predicate     // for forEach entries (optional filter: if v.field == value)
-	itemBuilder *ItemBuilder  // for forEachWith entries (complex per-item logic)
+	kind            entryKind
+	element         *ArrayElement    // the item fields (for static, conditional, forEach)
+	cond            Condition        // for conditional entries
+	source          Value            // for forEach entries (iteration source)
+	guard           Condition        // for forEach entries (optional guard: if source != _|_)
+	filter          Predicate        // for forEach entries (optional filter: if v.field == value)
+	itemBuilder     *ItemBuilder     // for forEachWith entries (complex per-item logic)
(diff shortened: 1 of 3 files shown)


vishal210893:
Verified locally. We can approve this

briankane (approved):

PR #7333, opened 2026-08-17
Fix: vela install shows a clear error for non-semver CLI versions

### LINKED ISSUE #7331: `vela install` fails with "Malformed version: master" when built from source without setting VELA_VERSION
**What happened?**

Building `vela` CLI from source using the default `make build` command produces a binary with `VELA_VERSION=master` baked in (from `makefiles/const.mk`, line 27: `VELA_VERSION ?= master`). Running `./bin/vela install` with this binary fails immediately:

**Expected behavior**

Either:
1. The build/development documentation (CONTRIBUTING.md) should mention that `VELA_VERSION` needs to be set explicitly when building from source for local testing, e.g. `make build VELA_VERSION=v1.10.7`, or
2. The install command's version check should handle a non-semver CLI version more gracefully (e.g. skip the check or show a clearer message pointing to the root cause).

**How to reproduce**

```bash
git clone https://github.com/kubevela/kubevela.git
cd kubevela
make build
./bin/vela install
```

**Workaround found**

```bash
make build VELA_VERSION=v1.10.7
./bin/vela install
```

This resolves the issue — installation completes successfully once VELA_VERSION is set to a valid semver.

**Environment**

- Built from `master` branch
- Go version: go1.26.0
- OS: WSL2 (Ubuntu) on Windows

**Additional context**

I searched CONTRIBUTING.md and found no mention of `VELA_VERSION` or guidance for this scenario. This is likely to affect other first-time contributors building from source.

pujitha24:

copilot:all

When `vela` is built from source with the default `make build` (no `VELA_VERSION` set), the binary bakes in `VELA_VERSION=master` (see `makefiles/const.mk`). Running `./bin/vela install` then fails immediately with a bare, confusing error from the underlying semver parser (e.g. `Malformed version: master`), giving no hint about the actual cause.

This PR makes the `vela install` version check fail with a clear, actionable message instead of the raw parser error, pointing the user at the root cause (`VELA_VERSION` not set when building from source) and the two ways to fix it (`make build VELA_VERSION=v1.10.7` or `vela install --version <ver>`).

Fixes #7331

I have:

- `go build ./references/cli/...` — passes.
- Added `TestInstallCommandRejectsNonSemverVersion` in `references/cli/install_test.go`, which drives `vela install --version master` through the real `NewInstallCommand`/`RunE` code path (no cluster/kube-client access happens before the version check, so no envtest/live cluster is needed). Verified it fails before the fix (`"Malformed version: master" does not contain "VELA_VERSION"`) and passes after: `go test ./references/cli/... -run TestInstallCommandRejectsNonSemverVersion -v` → PASS.
- `go vet ./references/cli/... ./version/...` — clean.
- `gofmt -l references/cli/install.go references/cli/install_test.go` — no output (already formatted).
- Ran the repo's `hack/utils/golangci-lint-wrapper.sh` (golangci-lint v1.60.1, matching `makefiles/dependency.mk`'s pinned version) — exits 0, no new issues.
- Checked `gh run list --repo kubevela/kubevela --branch master --workflow unit-test.yml` / `--workflow go.yml` — both green on current `master`, so this isn't riding on a pre-existing red base.

This is purely a clearer error message for an already-invalid state; it does not change behavior for valid `--version` values, and does not attempt to auto-detect or skip the version check for non-release builds (kept minimal/surgical, per the issue's second suggested option).

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7333?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/references/cli/install.go
+++ b/references/cli/install.go
@@ -111,7 +111,7 @@ func NewInstallCommand(c common.Args, order string, ioStreams util.IOStreams) *c
 		RunE: func(cmd *cobra.Command, args []string) error {
 			v, err := version.NewVersion(installArgs.Version)
 			if err != nil {
-				return err
+				return fmt.Errorf("invalid KubeVela version %q: %w\nIf you built the vela CLI from source, the binary was likely built without setting VELA_VERSION to a release version, e.g. run `make build VELA_VERSION=v1.10.7` or pass an explicit `--version` flag", installArgs.Version, err)
 			}
 			// Step1: Download Helm Chart
 			ioStreams.Info("Installing KubeVela Core ...")
--- a/references/cli/install.go
+++ b/references/cli/install.go
@@ -111,6 +111,9 @@ func NewInstallCommand(c common.Args, order string, ioStreams util.IOStreams) *c
 		RunE: func(cmd *cobra.Command, args []string) error {
 			v, err := version.NewVersion(installArgs.Version)
 			if err != nil {
+				if cmd.Flags().Changed("version") {
+					return fmt.Errorf("invalid KubeVela version %q: %w", installArgs.Version, err)
+				}
 				return fmt.Errorf("invalid KubeVela version %q: %w\nIf you built the vela CLI from source, the binary was likely built without setting VELA_VERSION to a release version, e.g. run `make build VELA_VERSION=v1.10.7` or pass an explicit `--version` flag", installArgs.Version, err)
 			}
 			// Step1: Download Helm Chart
--- a/references/cli/install_test.go
+++ b/references/cli/install_test.go
@@ -23,6 +23,9 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/stretchr/testify/assert"
+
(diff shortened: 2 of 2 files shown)


pujitha24:
Good catch. I've made the source-build hint conditional on `cmd.Flags().Changed("version")`, so it only shows up when the user didn't explicitly pass `--version`. If they did pass an explicit (bad) value, they now just get `invalid KubeVela version %q: %w` without the misleading build guidance. Split the existing test into two subtests covering both paths (default build-time version vs. explicit invalid flag).

roguepikachu:
Hi @pujitha24 Im closing this PR in favour of the issue assignee. Please comment against the issue you'd like to pick up and work on before raising a PR. Thanks

PR #7335, opened 2026-08-17
Fix: validate PolicyDefinition CUE with the cuex compiler

### LINKED ISSUE #7334: PolicyDefinition CUE templates are validated without the cuex package set
**Describe the bug**

`ComponentDefiniion` and `TraitDefinition` validate their CUE templates with `ValidateCuexTemplate`, which resolves the cuex package set. `PolicyDefinition` uses the plain `ValidateCueTemplate`, which does not:

```go
// componentdefinition/component_definition_validating_handler.go:110
webhookutils.ValidateCuexTemplate(ctx, cueTemplate)

// traitdefinition/trait_definition_validating_handler.go:128
webhookutils.ValidateCuexTemplate(ctx, cueTemplate)

// policydefinition/policy_definition_validating_handler.go:96
webhookutils.ValidateCueTemplate(cueTemplate)
```

So a policy importing an internal provider package is rejected at admission, though the render path compiles it fine. Nothing explains the difference; both helpers sit together in [`pkg/webhook/utils/utils.go`](https://github.com/kubevela/kubevela/blob/5cf3e3a80ade9d156b60e0691f2c198c3de004ee/pkg/webhook/utils/utils.go#L68-L97).

There are two call sites, not one. The handler runs a second validator later in the same request, and it has the same problem:

```go
// application/policy_validation.go:92, reached from the handler at L131
if err := webhookutils.ValidateCueTemplate(cueTemplate); err != nil {

// application/policy_validation.go:160, same template, global policies only
ctx := cuecontext.New()
value := ctx.CompileString(cueTemplate)
```

Fixing the handler alone leaves the request denied by the second one.

**To Reproduce**

```yaml
kind: PolicyDefinition
metadata:
  name: repro-policy
  namespace: vela-system
spec:
  schematic:
    cue:
      template: |
        import "vela/base64"

        encoded: base64.#Encode & {
          $params: "hello"
        }
```

Run through the validating handler, observed by adding it as a spec in the `policydefinition` webhook suite:

```text
builtin package "vela/base64" undefined (requestUID=)
```

After patching only the handler, the same request fails at the second val

```text
invalid PolicyDefinition: [builtin package "vela/base64" undefined] (requestUID=)
```

**Expected behavior**

`PolicyDefinition` validates like the other two kinds, or the difference documented.

`WorkloadCompiler` is the compiler to match, not the workflow one: policies are parsed into a `Component` and evaluated by the workload abstract engine, compiles through it.

```go
// appfile/parser.go:550
engine: definition.NewWorkloadAbstractEngine(name),

// cue/definition/template.go:139
val, err := velacuex.WorkloadCompiler.Get().CompileString(ctx.GetCtx(), .
```

That is what `ValidateCuexTemplate` already uses, so admission would check what render actually runs.

Worth a test that submits a policy importing a provider package. The exis specs use import-free templates, so they pass either way.

**KubeVela Version**

Checked against `master` at
[`5cf3e3a8`](https://github.com/kubevela/kubevela/commit/5cf3e3a80ade9d15
the head at the time of writing.

**Cluster information**

Not applicable. Admission-time validation, reproduces under envtest.

**Additional context**

Low impact today: none of the nine bundled policy definitions in `vela-templates/definitions/internal/policy/` import anything. A parity g rather than an outage.

One snag for whoever picks it up. `ValidateCuexTemplate` needs a `context`, and the second validator has none, so its signature and callers have to chang

```go
func ValidatePolicyDefinition(policy *v1beta1.PolicyDefinition) *PolicyValidationResult

### PR DESCRIPTION
**Pull request by sakirr05**

PolicyDefinition validated its CUE template with `ValidateCueTemplate`, which doesn't know about the cuex package set. The other two definition kinds already use `ValidateCuexTemplate`, so a policy importing something like `vela/base64` got rejected at admission even though render compiles it fine.

```go
// before
webhookutils.ValidateCueTemplate(cueTemplate)

// after
webhookutils.ValidateCuexTemplate(ctx, cueTemplate)
```

Two call sites had this, not one. `ValidatePolicyDefinition` runs a second check
later in the same request, so fixing just the handler still left it denied.

Fixes #7334

I have:

Added a spec with a policy that imports `vela/base64`. Fails without the fix, passes with it.

Validation now reaches the client singletons, so the test suite needed its envtest config wired in, or it dies on `GetConfigOrDie` with no kubeconfig.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/controller/core.oam.dev/v1beta1/application/policy_validation.go
+++ b/pkg/controller/core.oam.dev/v1beta1/application/policy_validation.go
@@ -17,7 +17,6 @@ limitations under the License.
 package application
 
 import (
-	"context"
 	"fmt"
 
 	"cuelang.org/go/cue"
@@ -28,7 +27,6 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 
 	"github.com/oam-dev/kubevela/apis/core.oam.dev/v1beta1"
-	"github.com/oam-dev/kubevela/pkg/cue/upgrade"
 	"github.com/oam-dev/kubevela/pkg/features"
 	webhookutils "github.com/oam-dev/kubevela/pkg/webhook/utils"
 )
@@ -44,9 +42,14 @@ func (r *PolicyValidationResult) IsValid() bool {
 	return len(r.Errors) == 0
 }
 
-// ValidatePolicyDefinition validates a PolicyDefinition
+// ValidatePolicyDefinition validates a PolicyDefinition.
+// cueTemplate and val are the effective (post-upgrade) template and its
+// compiled cue.Value, computed once by the caller: this function and
+// validateNoRequiredParameters both check against them rather than each
+// compiling the template again. Neither is read when the policy has no CUE
+// schematic, since that case is rejected before either is touched.
 // Returns validation result with errors (blocking) and warnings (informational)
-func ValidatePolicyDefinition(ctx context.Context, policy *v1beta1.PolicyDefinition) *PolicyValidationResult {
+func ValidatePolicyDefinition(policy *v1beta1.PolicyDefinition, cueTemplate string, val cue.Value) *PolicyValidationResult {
 	result := &PolicyValidationResult{
 		Errors:   []string{},
 		Warnings: []string{},
@@ -61,7 +64,7 @@ func ValidatePolicyDefinition(ctx context.Context, policy *v1beta1.PolicyDefinit
 	// Validate global policies have specific requirements
 	if policy.Spec.Global {
 		// No required parameters
-		if err := validateNoRequiredParameters(ctx, policy); err != nil {
+		if err := validateNoRequiredParameters(policy, val); err != nil {
 			result.Errors = append(result.Errors, err.Error())
 		}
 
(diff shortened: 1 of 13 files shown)


vishal210893:
Hi @sakirr05 
Can you look into cubic comments and failing e2e test cases?
Also, could we look into avoiding repeated CUE compilation in the PolicyDefinition admission path?
Unlike ComponentDefinition and TraitDefinition, which compile their templates once during validation, PolicyDefinition currently compiles once in the webhook and again in ValidatePolicyDefinition. Global policies compile it a third time in validateNoRequiredParameters.
Could we compile the effective template once and reuse the resulting cue.Value for the policy-specific and global-parameter checks? This would avoid duplicate work and ensure all checks use the same compiler, options, and upgraded template.

sakirr05:
Hi @vishal210893, all three don

Provider resolutio: fixed in 883f77eb with disable resolve provider function,
 also covers ComponentDefinition and TraitDefinition, which shared the same gap. Was worse than reported, a concrete kube # gots crashes the process via GetConfigOrDie, not just readness of the cluster.

Kubeconfig risk in the test suite: fixed in a665eff7, singleton.KubeConfig.Set(cfg) added in the webhook suite

Compile once: done in 5ec5797a, one cue. value sharesacross all three checks.

e2e: failing on a different unrelated spec each run, on a visibly degraded cluster each time fails on master itself too. Not a required check, lint and check-diff are green.

PR #7339, opened 2026-08-20
Fix: render ForEachMap body operations

### LINKED ISSUE #7283: Defkit: ForEachMap.WithBody is ignored by the CUE generator
**Describe the bug**

`ForEachMap().WithBody(...)` stores the operations you give it, and `Body()` hands them back, but the generator never reads them. You always get the default `(key): value` comprehension. The method looks like it works and does nothing.

- [`WithBody` appends to the body field](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/expr.go#L746-L750)
- [`forEachMapOpToCUE` never inspects `op.Body()`](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/cuegen.go#L2315-L2339)

**To Reproduce**

```go
body := defkit.NewResource("", "").
    Set("name", defkit.Reference("k")).
    Set("value", defkit.Reference("v")).
    Ops()

content := defkit.ForEachMap().
    Over("parameter.labels").
    WithVars("k", "v").
    WithBody(body...)

tpl.AddLetBinding("content", content)
```

Generated:

```cue
let content = {for k, v in parameter.labels {(k): v}}
```

`name` and `value` are gone.

**Expected behavior**

Either honour the body, or remove the method so nobody builds on a no-op. Honouring it looks like the smaller change, since the field and the accessor already exist. Output should stay as it is when no body is passed.

A test that greps for `for k, v` will pass either way, so this needs a test that compares the generated body.

**Screenshots**

Not applicable.

**KubeVela Version**

Checked against `master` at [`410c2a8`](https://github.com/kubevela/kubevela/commit/410c2a832163ad598ecefdb8718dbc2b2bba4de0), the head at the time of writing. `forEachMapOpToCUE` returns a fixed `{for k, v in src { (k): v }}`, and `Body()` is not referenced anywhere in `cuegen.go`.

**Cluster information**

Not applicable. This happens during generation.

**Additional context**

Found while looking into map-to-list comprehensions, which is a separate request (filed as #7288 ). Fixing the body would make map-to-map work as described, but it does not give you a map-to-list builder.

krishnankm:

`ForEachMap.WithBody` now renders its stored resource operations as a structured value under each generated map key.

The empty-body path is unchanged and continues to generate:

```cue
{for k, v in source { (k): v }}
 ```

When body operations are supplied, the generated value is built under the computed key:

 ```cue
{for key, val in source {
    (key): {
        name: key
        value: val
    }
}}
 ```

The implementation reuses Defkit's existing field-tree renderer, keeping nested field paths and resource-operation rendering consistent with Resource and PatchResource.

Import discovery also traverses ForEachMap body operations so values that require CUE standard-library imports are registered when the expression is inspected by the existing import collector.

Fixes #7283

I have:

Added regression coverage for:

 - Preserving the existing no-body map comprehension.
 - Rendering multiple body operations.
 - Rendering nested field paths.
 - Using custom key and value variable names.
 - Discovering imports required by body values.
 - Compiling the complete generated CUE.
 - Evaluating a generated map with two input entries and decoding the result.

The evaluation verifies that each source entry produces an independent structured value under its original key.

Body operations replace the normal valExpr for each generated key. When no body is supplied, the existing keyExpr and valExpr behavior and generated output remain unchanged.

No public API was added or changed. The existing WithBody and Body API is now honored by the CUE generator.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7339?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/definition/defkit/cuegen.go
+++ b/pkg/definition/defkit/cuegen.go
@@ -282,15 +282,7 @@ func (g *CUEGenerator) collectImportsFromOps(ops []ResourceOp) {
 			g.collectImportsFromValue(o.Cond())
 		case *IfBlock:
 			g.collectImportsFromValue(o.Cond())
-			for _, innerOp := range o.Ops() {
-				switch inner := innerOp.(type) {
-				case *SetOp:
-					g.collectImportsFromValue(inner.Value())
-				case *SetIfOp:
-					g.collectImportsFromValue(inner.Value())
-					g.collectImportsFromValue(inner.Cond())
-				}
-			}
+			g.collectImportsFromOps(o.Ops())
 		case *PatchKeyOp:
 			for _, elem := range o.Elements() {
 				g.collectImportsFromValue(elem)
--- a/pkg/definition/defkit/cuegen.go
+++ b/pkg/definition/defkit/cuegen.go
@@ -208,6 +208,8 @@ func (g *CUEGenerator) collectImportsFromValue(v interface{}) {
 				g.collectImportsFromItemOps(entry.itemBuilder.Ops())
 			}
 		}
+	case *ForEachMapOp:
+		g.collectImportsFromOps(val.Body())
 	case *PlusExpr:
 		for _, part := range val.Parts() {
 			g.collectImportsFromValue(part)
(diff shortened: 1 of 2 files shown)


vaibhav0096:
The Changes looks good. @krishnankm can you please resolve the conflict. Thanks

roguepikachu (approved):

briankane (approved):

PR #7348, opened 2026-08-26
Feat(defkit): support expression-based validator messages

### LINKED ISSUE #7289: [Enhancement] Defkit: support expression-based validator messages

The validator API, and validators attached to array parameters.

`Validate` takes a Go string, and the generator writes that same quoted string as the field label in both branches. A message can describe the rule but not the value that broke it.

We wanted:

```text
replicas[2].region 'us-east-1' must not match the primary region
```

and settled for:

```text
replica regions must not match the primary region
```

- [`Validate` takes a string](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/validator.go#L19-L42)
- [`writeValidator` quotes it in both branches](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/cuegen.go#L537-L576)

CUE allows a computed field label, so including a local value does not need a raw `error(...)`:

```cue
let message = "region '\(region)' must not match the primary region"
_validateRegion: {
    (message): true
    if region == parameter.primaryRegion {
        (message): false
    }
}
```

A validator that takes a `Value` for the message, alongside the existing string form:

```go
defkit.ValidateValue(
    defkit.Interpolation(
        defkit.Lit("region '"),
        defkit.LocalField("region"),
        defkit.Lit("' must not match the primary region"),
    ),
).
    FailWhen(defkit.Eq(defkit.LocalField("region"), primaryRegion)).
    WithName("_validateReplicaRegion")
```

Both branches need the identical computed key, and a message that will not resolve to a concrete string should fail with something readable.

The array index is a separate problem. `ArrayParam.Validators` emits `[...{...}]`, so there is no iterator in scope and no way to reach `[2]`. Getting the index in needs a different validation context, and it seems worth splitting out rather than holding up the message change.

**Additional context**

Fixed messages are fine for simple rules and should stay the default. Dynamic ones earn their keep when a list has several similar entries and you need to find which one failed.

sakirr05:

Validate takes a Go string, and writeValidator writes that same quoted string as the field label in both branches. That means a message can describe the rule but not the value that broke it, so on a list of similar entries you can say "replica regions must not match the primary region" but not which replica.

This adds ValidateValue, which takes a Value instead of a string:

```defkit.ValidateValue(
    defkit.Interpolation(
        defkit.Lit("region '"),
        defkit.LocalField("region"),
        defkit.Lit("' must not match the primary region"),
    ),
).
    FailWhen(defkit.Eq(defkit.LocalField("region"), defkit.Reference("parameter.primaryRegion"))).
    WithName("_validateReplicaRegion")
```

which generates:

```cue
_validateReplicaRegion: {
      let _message = "region '\(region)' must not match the primary region"
      (_message): true
      if region == parameter.primaryRegion {
              (_message): false
      }
}
```
The message goes through a let rather than being repeated in both branches. The two labels have to unify into one field for the validator to conflict at all, and reaching for the same binding twice guarantees that in a way repeating an expression does not. It also keeps a long message out of the output twice over.

Two smaller decisions worth flagging:

- Passing a literal string Value skips the let and emits the same quoted label Validate would, so the two forms stay interchangeable and simple rules do not pay for the indirection.
- A message the generator has no rendering for becomes _|_ with a comment, instead of a label CUE never resolves. A validator that silently never fires seemed like the one outcome worth ruling out.

Validate is untouched and generates byte-identical CUE. Fixed messages aris is for when a list has several similar entries and the message has tosay which one failed.

Fixes #7289

I have:

- [x] Read and followed KubeVela's pull request process (https://kubevela.io/docs/contributor/code-contribute#create-a-pull-request).
- [ ] Related Docs (https://github.com/kubevela/kubevela.io) updated properly.
- [x] Run make reviewable to ensure this PR is ready for review.
- [ ] Added backport release-x.y labels to auto-backport this PR if necessary.

How has this code been tested

Nine specs in validator_test.go and validator_internal_test.go covering the let binding and shared key, array elements, two validators in one struct, guarded validators, the
literal-string equivalence, and the unrenderable case.

One of them compiles the generated CUE and asserts on the error, since thlure says:

parameter.replicas.1._validateReplicaRegion."region 'us-west-2' must not nflicting values true and false

I also checked these actually bite: disabling only the new rendering path stay green are pure accessor tests that never reach the generator). Fullsuite is 1262 specs green, and make reviewable is clean.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/definition/defkit/cuegen.go
+++ b/pkg/definition/defkit/cuegen.go
@@ -19,6 +19,7 @@ package defkit
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 )
 
@@ -149,6 +150,12 @@ func (g *CUEGenerator) detectRequiredImports(c *ComponentDefinition) {
 		}
 	}
 
+	// Check validator message expressions, at the top level and inside params.
+	g.collectImportsFromValidators(c.GetValidators())
+	for _, param := range c.GetParams() {
+		g.collectImportsFromParamValidators(param)
+	}
+
 	// Check resource operations in output
 	if output := tpl.GetOutput(); output != nil {
 		g.collectImportsFromResource(output)
@@ -157,6 +164,42 @@ func (g *CUEGenerator) detectRequiredImports(c *ComponentDefinition) {
 	// Check resource operations in outputs
 	for _, res := range tpl.GetOutputs() {
 		g.collectImportsFromResource(res)
+	}
+}
+
+// collectImportsFromValidators adds the imports needed by validator message
+// expressions.
+//
+// A message can hold any Value, including a CUE standard library call such as
+// strings.ToLower, and the reference is emitted whether or not the import is
+// there. Only the message is walked: a validator built with Validate carries no
+// expression and contributes nothing, so this cannot change what any existing
+// definition generates.
+func (g *CUEGenerator) collectImportsFromValidators(validators []*Validator) {
+	for _, v := range validators {
+		if v == nil {
+			continue
+		}
+		if msg := v.MessageValue(); msg != nil {
+			g.collectImportsFromValue(msg)
+		}
+	}
+}
+
+// collectImportsFromParamValidators walks a parameter for validators, including
+// the ones attached to nested fields of maps and array elements.
+func (g *CUEGenerator) collectImportsFromParamValidators(param Param) {
+	switch p := param.(type) {
+	case *MapParam:
+		g.collectImportsFromValidators(p.GetValidators())
+		for _, field := range p.GetFields() {
+			g.collectImportsFromParamValidators(field)
+		}
+	case *ArrayParam:
+		g.collectImportsFromValidators(p.GetValidators())
+		for _, field := range p.GetFields() {
+			g.collectImportsFromParamValidators(field)
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


krishnankm:
@sakirr05 Could you address the review comments ?

vishal210893:
changes verifies successfully locally so LGTM
Parameter errors:
  parameter.replicas.0._validateReplicaRegion."replica region 'us-east-1' must not match primaryRegion": conflicting values true and false. (requestUID=34429bcd-5b84-46a2-9bb3-a9a75ff2deb5)

sakirr05:
Thanks both. @vishal210893 good to see it verified on a cluster, and that output is a nice illustration of the index point: replicas.0 comes from CUE's error path for free
All three cubic findings were real, fixed in the latest push

briankane (approved):

PR #7359, opened 2026-09-04
Fix: watch external Package CRs in WorkloadCompiler

### LINKED ISSUE #7358: [Bug] WorkloadCompiler missing external Package watch

- [x] I have searched the existing issues

### Affected area

Application controller

### KubeVela version

1.11.0

1.31.0

### Describe the bug

When external package watching is enabled, new Package CRs are picked up automatically by the workflow and policy compilers but not by the compiler used for component and trait definitions. That compiler only loads Packages once at startup and never watches for changes, so any Package added after startup is ignored until vela-core is restarted.

### To reproduce

1. Start vela-core with --enable-external-package-for-default-compiler=true --enable-external-package-watch-for-default-compiler=true
2. Create a test Package CR after vela-core is running:
```
kubectl apply -f - <<'EOF'
apiVersion: cue.oam.dev/v1alpha1
kind: Package
metadata:
  name: test-repro
  namespace: vela-system
spec:
  path: test/repro
  provider:
    protocol: http
    endpoint: http://localhost:9999
  templates:
    repro.cue: |
      package repro
      greeting: "hello"
EOF
```

3. Try to create a ComponentDefinition that imports it:

```
kubectl apply -f - <<'EOF'
kind: ComponentDefinition
metadata:
  name: test-repro-import
  namespace: vela-system
spec:
  workload:
    type: autodetects.core.oam.dev
  schematic:
    cue:
      template: |
        import "test/repro"
        output: {
          apiVersion: "v1"
          kind: "ConfigMap"
          metadata: name: context.name
          data: value: repro.greeting
        }
        parameter: {}
EOF
```

Webhook rejects with builtin package "test/repro" undefined. WorkloadCompiler doesn't have the Package watch, so it never loaded the Package CR created after init.

4. Restart vela-core and reapply step 3. Succeeds, WorkloadCompiler loaded the Package during restart.

### Expected behavior

Step 3 should succeed without a restart. The enableExternalPackageWatchForDefaultCompiler flag enables the Package watch for providers.DefaultCompiler and cuex.DefaultCompiler, but the WorkloadCompiler (used for component/trait rendering and webhook validation) is not covered. All three compilers should pick up Package CRs dynamically.

### Screenshots or logs

_No response_

### Additional context

_No response_

pujitha24:

copilot:all

`WorkloadCompiler` (used for component/trait definition rendering and webhook
validation) only loaded external `Package` CRs once at startup via
`LoadExternalPackages`. Unlike `providers.DefaultCompiler` and `cuex.DefaultCompiler`,
it never started a watch, so `Package` CRs created after startup were ignored
until `vela-core` was restarted, even with
`--enable-external-package-watch-for-default-compiler=true`.

This adds the same `go compiler.ListenExternalPackages(nil)` call (gated behind
`cuex.EnableExternalPackageWatchForDefaultCompiler`) that `providers.DefaultCompiler`
already uses in `pkg/workflow/providers/compiler.go`, so all three compilers now
watch for `Package` changes consistently.

Fixes #

I have:

- `go build ./pkg/cue/cuex/...` — succeeds.
- `go vet ./pkg/cue/cuex/...` and `gofmt -l pkg/cue/cuex/compiler.go` — clean.
- `golangci-lint` (repo-pinned v1.60.1, via `./hack/utils/golangci-lint-wrapper.sh`) — exit 0, no findings.
- `go test ./pkg/cue/cuex/ -run TestWorkloadCompiler -v` (with `KUBEBUILDER_ASSETS` pointed
  at a local envtest toolchain) — `TestWorkloadCompiler` passes, including the
  "cuex enabled with external packages" subcase which loads a real `Package` CR via
  envtest and confirms `WorkloadCompiler` still resolves external packages correctly
  after this change.

Note: in this local sandbox (no real/default kubeconfig available), the `cue/cuex`
test *package* reports an overall `FAIL` after `TestWorkloadCompiler` passes. This is
pre-existing and unrelated to this change — verified it reproduces identically via
`git stash` on the unmodified baseline. Root cause: `TestMain` in
`pkg/cue/cuex/compiler_test.go` calls `singleton.KubeConfig.Reload()` after
`m.Run()`, which invokes controller-runtime's `config.GetConfigOrDie()` and requires
a real, reachable kubeconfig (`KUBECONFIG` / `~/.kube/config` / in-cluster) rather
than the in-process envtest config — none of which exist in this sandbox.
`.github/workflows/unit-test.yml` sets up a real KinD cluster via
`./.github/actions/setup-kind-cluster` before running unit tests, which is why this
passes in CI. I could not run against a live KinD cluster here, but the change
itself (a 3-line addition mirroring an already-working, identical pattern) is
covered by the assertions above.

The added code path (`ListenExternalPackages`) is not exercised by the existing
test suite, since `EnableExternalPackageWatchForDefaultCompiler` defaults to `false`
and no test sets it to `true`; it uses the exact same `*cuex.Compiler` method,
called the same way, as the already-merged and working `providers.DefaultCompiler`.

This mirrors the identical, already-working pattern in
`pkg/workflow/providers/compiler.go`'s `DefaultCompiler` initializer
(`go c.ListenExternalPackages(nil)` gated on the same flag). No other compiler
in this codebase needed a change — `cuex.DefaultCompiler` (from the
`github.com/kubevela/pkg` module) already does this internally.

Fixes #7358

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7359?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pkg/cue/cuex/compiler.go
+++ b/pkg/cue/cuex/compiler.go
@@ -54,5 +54,8 @@ var WorkloadCompiler = singleton.NewSingleton[*cuex.Compiler](func() *cuex.Compi
 			klog.Errorf("failed to load external packages for workload compiler: %v", err.Error())
 		}
 	}
+	if cuex.EnableExternalPackageWatchForDefaultCompiler {
+		go compiler.ListenExternalPackages(nil)
+	}
 	return compiler
 })
--- a/pkg/cue/cuex/compiler_test.go
+++ b/pkg/cue/cuex/compiler_test.go
@@ -124,7 +124,6 @@ func TestMain(m *testing.M) {
 func TestWorkloadCompiler(t *testing.T) {
 	testCases := map[string]struct {
 		cuexEnabled            bool
-		watchEnabled           bool
 		workloadTemplate       string
 		params                 map[string]interface{}
 		expectedObj            runtime.Object
@@ -167,20 +166,10 @@ func TestWorkloadCompiler(t *testing.T) {
 			expectedAdditionalObjs: make(map[string]runtime.Object),
 			hasCompileErr:          false,
 		},
-		"cuex enabled with external packages and watch enabled": {
-			cuexEnabled:            true,
-			watchEnabled:           true,
-			workloadTemplate:       getWorkloadTemplate(true),
-			params:                 make(map[string]interface{}),
-			expectedObj:            getExpectedObj(true),
-			expectedAdditionalObjs: make(map[string]runtime.Object),
-			hasCompileErr:          false,
-		},
 	}
 
 	for _, tc := range testCases {
 		cuex.EnableExternalPackageForDefaultCompiler = tc.cuexEnabled
-		cuex.EnableExternalPackageWatchForDefaultCompiler = tc.watchEnabled
 		velacuex.WorkloadCompiler.Reload()
 
 		ctx := process.NewContext(process.ContextData{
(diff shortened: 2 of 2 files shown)


roguepikachu (approved):
LGTM.

Verified locally on a k3d cluster running vela-core built from master with both external package flags enabled. Before the change, a ComponentDefinition importing a Package created after the compiler had initialized was rejected at admission with `builtin package "test/repro" undefined`, and the identical manifest was accepted after a restart. With the fix applied, the same sequence succeeds with no restart, deleting the Package makes the import fail again (so the informer's delete handler is working too, not just the add handler), and a definition importing a package that never existed is still correctly rejected.

I also confirmed the new test is a genuine regression test: reverting only the three lines in `pkg/cue/cuex/compiler.go` and keeping the test makes `TestWorkloadCompilerExternalPackageWatch` fail at the full 10s poll timeout with `builtin package "cuex/ext-watch" undefined`. Good call moving the Package creation to after `Reload()` in the third commit. Since `Reload()` runs the loader eagerly, a version that created the Package first would have passed against the unfixed code and been worthless as a guard.

roguepikachu:
@pujitha24 one process request for next time: please get yourself assigned to the issue before opening a PR against it, so we do not end up with duplicate work on the same bug. Treating this one as an exception given the fix is already done and verified.

roguepikachu (approved):

briankane (approved):

pujitha24:
Understood, thanks for the exception here and for the thorough review. Will get assigned to the issue first next time before opening a PR against it.

Issue #7356, opened 2026-09-02
[Bug] vela addon upgrade constructs invalid K8s resource name when using registry prefix


oanasc:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Affected area

CLI (vela)

### KubeVela version

1.11.0 , 1.10.8

### Kubernetes version and distribution

1.31.11

### Describe the bug

When upgrading an addon installed from a named registry using the registry/name syntax, vela addon upgrade constructs an invalid Kubernetes resource name ie addon-myregistry/fluxcd. The / character is not allowed in K8s resource names.

### To reproduce

1. Add a custom Helm registry:
vela addon registry add myregistry --type helm --endpoint=http://localhost:9090
2. Enable an addon from that registry:
vela addon enable myregistry/fluxcd
This succeeds
3. Upgrade the same addon:
vela addon upgrade myregistry/fluxcd
Error: cannot fetch addon related addon vortexa/fluxcd: invalid resource name "addon-myregistry/fluxcd": [may not contain '/']

### Expected behavior

The upgrade succeeds, using just the addon name (fluxcd) for the K8s resource name addon-fluxcd.

dchaudhari7177:
This looks like the same defect as #7363, which reports the identical `invalid resource name "addon-<registry>/<addon>": [may not contain '/']` from `vela addon upgrade`.

The cause is that `enable` splits the registry-qualified name through `splitSpecifyRegistry` before using it, while `upgrade` passes `args[0]` straight to `pkgaddon.FetchAddonRelatedApp`, so the prefix ends up in the Application name. #7366 fixes it by having the upgrade path call the same helper.

Flagging in case the two are worth linking or closing as duplicates.

roguepikachu:
I hit this same bug independently and filed #7363 before I spotted this report. Closing mine as a duplicate of this one and moving the extra detail here so everything stays in one place.

### Root cause

`enable` splits a registry-qualified name with `splitSpecifyRegistry` before using it (`references/cli/addon.go:204`), but the `upgrade` path passes `args[0]` straight into `pkgaddon.FetchAddonRelatedApp`, which builds the Application name as `addon-<name>` through `addonutil.Addon2AppName`. The registry prefix therefore ends up inside a Kubernetes resource name, which cannot contain `/`.

### No installed addon is needed to reproduce

[...]

PR #7366, opened 2026-09-08
Fix: vela addon upgrade accepts the <registry>/<addon> form

### LINKED ISSUE #7363: [Bug] vela addon upgrade rejects the <registry>/<addon> form that vela addon enable accepts

Searched the open issues, nothing matching.

### Affected area

CLI (vela) / Addon

### KubeVela version

CLI built from master at `e6bbce3fc` (`go build ./references/cmd/cli/main.go`). Controller v1.10.0.

v1.31.5+k3s1, k3d

### Describe the bug

`vela addon enable` accepts a registry-qualified addon name (`<registry>/<addon>`) and splits it, but `vela addon upgrade` does not. The registry prefix is carried straight into the Application name, which then fails Kubernetes name validation.

The upgrade command reads `args[0]` into `addonOrDir` and passes it to `pkgaddon.FetchAddonRelatedApp` unchanged, so the prefix ends up in `addon-<registry>/<addon>`:

https://github.com/kubevela/kubevela/blob/e6bbce3fc/references/cli/addon.go#L329

`enableAddon` calls `splitSpecifyRegistry(name)` before using the name, which is what `upgrade` is missing.

### To reproduce

1. Have any addon registry configured. The community `KubeVela` registry that the chart pre-registers is enough.
2. Run an upgrade using the registry-qualified form that `enable` accepts:

```console
$ vela addon upgrade KubeVela/velaux
Error: cannot fetch addon related addon KubeVela/velaux: invalid resource name "addon-KubeVela/velaux": [may not contain '/']
```

The same form works with `enable`:

```console
$ vela addon enable KubeVela/velaux
Addon velaux enabled successfully.
```

### Expected behavior

`vela addon upgrade <registry>/<addon>` should behave like `vela addon enable <registry>/<addon>`: split the registry from the addon name, scope the upgrade to that registry, and use the bare addon name when resolving the Application.

At minimum the error should say the registry-qualified form is not supported here, rather than surfacing an invalid Kubernetes resource name.

### Screenshots or logs

Covered above.

### Additional context

Workaround: drop the prefix and use the bare addon name, which resolves across all configured registries.

```console
$ vela addon upgrade velaux
```

Suggested labels for triage: `type/bug`, `area/cli`, `area/addon`, `good first issue`, `effort/small`.

dchaudhari7177:
Fixes #7363.

## Description of your changes

`vela addon enable` splits a registry-qualified addon name before using it; `vela addon upgrade` passed `args[0]` through unchanged. The prefix then reached `pkgaddon.FetchAddonRelatedApp` and ended up in the Application name:

```console
$ vela addon upgrade KubeVela/velaux
Error: cannot fetch addon related addon KubeVela/velaux: invalid resource name "addon-KubeVela/velaux": [may not contain '/']

$ vela addon enable KubeVela/velaux
Addon velaux enabled successfully.
```

The upgrade path now calls `splitSpecifyRegistry` — the same helper `enableAddon` uses — and resolves the Application and the merged install args from the **bare** addon name.

`enableAddon` still receives the **qualified** form, because it splits the registry itself (`addon.go:576`) and needs the prefix to scope the lookup to that registry. That keeps `upgrade KubeVela/velaux` scoped to the `KubeVela` registry, matching `enable`.

A side effect worth noting: an invalid multi-slash name such as `a/b/c` now reports `failed to split addonName and addonRegistry: invalid addon name, ...` instead of surfacing a Kubernetes name-validation error, which the issue also asked for as a minimum.

Behaviour for a bare addon name is unchanged — `splitSpecifyRegistry` returns `("", name)` for it.

## Verification

- `go build ./references/cli/` — passes
- `go vet ./references/cli/` — clean
- `go test ./references/cli/ -run 'TestAddonUpgradeCmdWithErrLocalPath|TestCheckSpecifyRegistry' -v` — both PASS

I did not add a unit test for the registry-qualified path: the branch runs after `c.GetClient()`, so reaching it needs a live cluster, and the existing `TestAddonUpgradeCmdWithErrLocalPath` cases stop at the local-directory guard before that point. Happy to add an envtest-based case if you would like one.

Not run here: an end-to-end `vela addon upgrade` against a cluster.

## I have:

- [x] Read and followed the [CONTRIBUTING](https://github.com/kubevela/kubevela/blob/master/CONTRIBUTING.md) guide.
- [x] Signed off my commits (DCO).
- [x] Run `go build` and `go vet`.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7366?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/references/cli/addon.go
+++ b/references/cli/addon.go
@@ -323,15 +323,24 @@ non-empty new arg
 				if filepath.IsAbs(addonOrDir) || strings.HasPrefix(addonOrDir, ".") || strings.HasSuffix(addonOrDir, "/") {
 					return fmt.Errorf("addon directory %s not found in local", addonOrDir)
 				}
-				name = addonOrDir
-				_, err = pkgaddon.FetchAddonRelatedApp(context.Background(), k8sClient, addonOrDir)
+				// `enable` accepts <registry>/<addon> and splits it before resolving the
+				// Application; do the same here. Without this the registry prefix reaches
+				// FetchAddonRelatedApp and lands in the Application name, which then fails
+				// Kubernetes name validation with `may not contain '/'`.
+				_, addonName, err := splitSpecifyRegistry(addonOrDir)
 				if err != nil {
-					return errors.Wrapf(err, "cannot fetch addon related addon %s", addonOrDir)
+					return fmt.Errorf("failed to split addonName and addonRegistry: %w", err)
+				}
+				name = addonName
+				_, err = pkgaddon.FetchAddonRelatedApp(context.Background(), k8sClient, name)
+				if err != nil {
+					return errors.Wrapf(err, "cannot fetch addon related addon %s", name)
 				}
 				addonArgs, err := pkgaddon.MergeAddonInstallArgs(ctx, k8sClient, name, addonInputArgs)
 				if err != nil {
 					return err
 				}
+				// enableAddon splits the registry off itself, so it keeps the qualified form.
 				additionalInfo, err = enableAddon(ctx, k8sClient, dc, config, addonOrDir, addonVersion, addonArgs)
 				if err != nil {
 					return err


roguepikachu:
Thanks for the fix, and thanks especially for spotting that #7356 and #7363 were the same bug and flagging it. That was helpful triage.

I am closing this one in favour of #7357, which addresses the same bug. That PR was opened on Sep 3, five days before this one, its CI is fully green, and it includes an envtest regression test that was verified to fail before the change and pass after. This PR has no test coverage for the fixed path, which you noted yourself in the description.

To be fair on CI, the `unit-tests` failure here does not look like your code. It fails during KinD cluster setup (`0/1 nodes are available: 1 node(s) had untolerated taint node.kubernetes.io/not-ready`, with coredns unable to schedule) rather than in any Go test, so it reads as infrastructure flake and would probably pass on a re-run. The `e2e-tests (v1.31)` job also failed. Either way, the deciding factor is that #7357 got there first with a regression test, not the CI state.

### There is real work left here if you want it

Neither PR fixes the whole bug. `upgrade` is only one of three affected commands. `splitSpecifyRegistry` is called only on the enable path, so `status` and `disable` pass `args[0]` through unsplit into the same `FetchAddonRelatedApp` and fail identically:

```console
$ vela addon status KubeVela/velaux
Error: invalid resource name "addon-KubeVela/velaux": [may not contain '/']

$ vela addon disable KubeVela/velaux
Error: invalid resource name "addon-KubeVela/velaux": [may not contain '/']
```

I have asked on #7357 for that PR to be extended to cover both. If it is not picked up there, this is genuinely useful follow-up work and you would be welcome to take it.

### One process request

Please comment on the issue and get yourself assigned before opening a PR against it. That is what would have avoided the duplicate effort here, since #7356 already had an open PR when this was raised. Also worth pointing the `Fixes` line at the original issue rather than a duplicate: this PR referenced #7363, which I have now closed as a duplicate of #7356, so merging it would have left the original open.

PR #7367, opened 2026-09-08
Feat(module): install a module as an Application component

roguepikachu:

Adds KubeVela modules: versioned packages, published as OCI Helm-chart artifacts, that carry definitions and their auxiliary objects so a module author can ship and version them independently of the platform release.

A module is installed as an Application component. The new `type: module` ComponentDefinition fetches the module and renders the Application it owns, so installing one is an ordinary Application apply rather than a separate lifecycle.

This builds directly on the addon-as-component work from #7360 and follows the same shape, including the feature gate.

**What is added**

- `pkg/module`: parsing, validation, packaging, publish and fetch for a module.
- `vela module init`, `publish`, `deploy` and `registry`. Publish refuses to overwrite an already published tag, and deploy reports per-tier install status, timing out against the tier that is actually stuck rather than the whole deploy.
- The `type: module` ComponentDefinition and its `vela/module` CueX provider. Parameters are module name (defaulting to the component name), registry, namespace and version, where an empty registry or version selects the configured default and the latest published version.
- Type resolution, so an Application may name a module definition as `apiVersion/name` or `module/apiVersion/name` and have it resolved to the installed definition name.

**Feature gate**

The render service is registered only when `EnableModuleComponent` is on, and the gate is Alpha and off by default. The ComponentDefinition and the `vela/module` provider package always ship, because the definition cannot compile without the package. With the gate off, an Application using `type: module` fails at render with an actionable message rather than a CUE import error. This is the same arrangement `EnableAddonComponent` already uses.

**Refactor: one shared registry and OCI layer**

A module and an addon are fetched from the same kinds of place, so the last commit moves the shared half into `pkg/registry/component`: the registry model and its ConfigMap-backed store, the source types and their credential handling, the git, gitee, gitlab, OSS, local and memory readers, and the OCI chart transport.

`pkg/addon` keeps what is genuinely addon-specific, which is interpreting fetched files as an addon (`GetUIData`, `ListUIData`, `GetInstallPackage`, `ListAddonInfo`, `ClassifyItemByPattern`), the versioned-registry backends and catalog listing. It reaches the shared half through `component_alias.go`, so spellings such as `addon.Registry` keep working and callers outside the package are unchanged. Functions that were methods on `Registry` are now free functions taking one, since a package cannot define methods on a type it does not own.

The OCI transport moves rather than being copied, which is the main reason for the refactor. The addon path and the module path had each grown their own version, and they disagreed: one parsed only `oci://`, the other also accepted `http://` and bare registry hosts such as an ECR endpoint. The same registry record could therefore resolve differently depending on which path reached it. The surviving implementation keeps both behaviours that mattered, the wider URL handling and the client cache, per-call timeout and context-aware `AwaitOCICall` that releases a cancelled caller without leaking the connection.

Module registries are stored in their own ConfigMap and secret prefix, so module entries and credentials stay separate from the addon registry while sharing one implementation. `NewRegistryDataStore` keeps the addon defaults, so addon behaviour is unchanged.

Fixes #

I have:

- [ ] All commits are signed off (DCO).

- Unit tests for `pkg/module` and `pkg/registry/component`, including the reader and source tests that moved with the code.
- `pkg/addon`'s own mock-backed OCI, push, catalog and versioned-registry tests pass unchanged after the extraction, which is the check that matters most for the transport move.
- `references/cli` and `pkg/appfile` tests pass.
- `go vet` is clean across every changed package, tests included.
- The publish and fetch round trips in `test/e2e-test` run against a real registry and skip when their environment variable is unset, so a default e2e run never reaches one. They are opt-in verification rather than a gate.

Not yet run here, and worth a maintainer's attention: the `pkg/addon` envtest suites, which need `KUBEBUILDER_ASSETS` and fail identically on an unmodified master in this environment, and a full `make reviewable`.

The last commit is the shared-layer extraction and is reviewable on its own. Everything before it is the module feature; nothing outside `pkg/addon` and `pkg/registry` changes behaviour for existing users.

Worth looking at closely:

- The extraction boundary. The split is between fetching a package and interpreting it as an addon. If a symbol looks misplaced, that is the line it was drawn on.
- `AwaitOCICall` and `loadTokenFromSecret` carry the two fixes that came out of review on #7360, moved into their new homes so they are not lost.
- In the application validating webhook, permission is checked before existence on purpose. Resolution has to run first so the check uses the installed name rather than the `apiVersion/name` spelling, but reporting existence to a caller with no permission on the definition would hand them an existence oracle they did not have before type resolution existed.
- The workload-type label now uses the resolved definition name, because a module type string containing slashes would otherwise produce an invalid label value.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7367?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


roguepikachu:
@cubic-ai-dev review

Reetika2509:
@cubic-ai-dev review

PR #7368, opened 2026-09-08
Fix: reject addon registry endpoints the reader cannot use

### LINKED ISSUE #7364: [Bug] vela addon registry add --type git accepts an unusable endpoint, then vela addon list panics

Searched the open issues, nothing matching.

### Affected area

CLI (vela) / Addon

### KubeVela version

CLI built from master at `e6bbce3fc` (`go build ./references/cmd/cli/main.go`). Controller v1.10.0.

v1.31.5+k3s1, k3d

### Describe the bug

`vela addon registry add --type git` accepts any endpoint without validating it, and a subsequent `vela addon list` then crashes the CLI with a nil pointer dereference.

Two things combine here.

`addAddonRegistry` only performs a reachability check when the record has a Helm source, so `git`, `gitlab`, `gitee` and `OSS` registries are stored unverified:

https://github.com/kubevela/kubevela/blob/e6bbce3fc/references/cli/addon-registry.go#L84

`--type git` registries are read through the GitHub REST API. `createGitHelper` builds a `gitHelper` from whatever `utils.Parse` returned, and for a non-GitHub URL that leaves `Meta.GithubContent` nil. `readRepo` then dereferences it:

https://github.com/kubevela/kubevela/blob/e6bbce3fc/pkg/addon/addon.go#L547

So any endpoint that is not a GitHub URL is accepted at add time and segfaults at read time. A `git://` endpoint is the obvious case, since the protocol looks like it ought to work for a git registry, but it is never used: the reader always talks to `api.github.com`.

### To reproduce

1. Add a git registry with a `git://` endpoint. Note that it is accepted:

```console
$ vela addon registry add repro-git --type git \
    --endpoint=git://127.0.0.1:9418/poc.git --path addons
Successfully add an addon registry repro-git
```

2. List addons:

```console
$ vela addon list
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x411e5f4]

goroutine 1 [running]:
github.com/oam-dev/kubevela/pkg/addon.(*gitHelper).readRepo(0x203b7781fe80, {0x0?, 0x203b7656a670?})
	/kubevela/pkg/addon/addon.go:547 +0x34
github.com/oam-dev/kubevela/pkg/addon.(*gitReader).ListAddonMeta(0x203b7680a478)
	/kubevela/pkg/addon/reader_github.go:45 +0x3c
github.com/oam-dev/kubevela/pkg/addon.(*Registry).ListAddonMeta(0x203b7739f478?)
	/kubevela/pkg/addon/source.go:415 +0x44
github.com/oam-dev/kubevela/references/cli.listAddons({0x7fddcb0, 0x82c04e0}, {0x7ff93c8, 0x203b77506a80}, {0x0, 0x0})
	/kubevela/references/cli/addon.go:1002 +0x200
```

Any non-GitHub HTTP endpoint reaches the same place. The registry is persisted, so every later `vela addon list` crashes until it is deleted with `vela addon registry delete repro-git`.

### Expected behavior

Two separate fixes, either of which removes the crash:

1. `vela addon registry add` should validate a `git` endpoint the way it validates Helm and OCI ones, and reject an endpoint the GitHub reader cannot use. Ideally with a message saying `--type git` expects a GitHub URL and that the git protocol is not supported.
2. `createGitHelper` should return an error when `utils.Parse` produced no `GithubContent`, and `readRepo` should not dereference it unconditionally. A stored registry from an older version, or a hand-edited ConfigMap, should surface an error rather than crash the CLI.

A panic is also a poor failure mode for one bad registry among several: `listAddons` deliberately tolerates a registry it cannot read, and that intent is defeated here.

### Screenshots or logs

Stack trace above, taken from a CLI built at `e6bbce3fc`. Paths have been shortened.

### Additional context

Recovering from the state needs the registry name, which `vela addon registry list` can still print because it does not read the registries:

```console
$ vela addon registry delete repro-git
```

Suggested labels for triage: `type/bug`, `area/cli`, `area/addon`, `effort/small`.

dchaudhari7177:
Fixes #7364.

## Description of your changes

`vela addon registry add --type git` accepted any endpoint, and a subsequent `vela addon list` crashed the CLI with a nil pointer dereference.

**The panic.** `utils.Parse` reports an address it does not recognise as `(TypeUnknown, nil, nil)` — no error, no content. Only `http` and `https` carry a host switch, so a `git://` or `ssh://` endpoint falls through to that return. `NewAsyncReader` passed the nil straight into the helper, and `gitHelper.readRepo` then dereferenced `h.Meta.GithubContent.Owner`.

Reproduced as a test against the current tree, with the guard removed:

```
NewAsyncReader -> reader=true err=<nil>
PANIC from ReadFile: runtime error: invalid memory address or nil pointer dereference
```

With the guard in place the same call is refused, with the endpoint quoted so the operator can tell which of their registries is the bad one:

```
NewAsyncReader -> reader=false err=unsupported endpoint for a git addon registry:
it is read through the GitHub API, so it must be an http(s) github.com or
api.github.com URL: "git://127.0.0.1:9418/poc.git"
```

The gitee reader has the identical shape, so it gets the same guard and its own sentinel.

**The unvalidated add.** The second half of the issue: `addAddonRegistry` only checked reachability for a Helm source, so `git`, `gitee`, `gitlab` and `OSS` registries were stored unverified. Building a git or gitee reader is offline construction with no request — and it is the same call `vela addon list` makes — so `add` and `update` now do it and refuse a registry that would fail at read time.

`gitlab` and `OSS` are deliberately left alone: building those readers dials out, and adding a network call to `add` is a bigger behaviour change than this issue calls for.

`update` is included because without it the same bad endpoint walks in through `vela addon registry update`.

## Related issue

#7364

## Verification

- `go build ./pkg/addon/... ./references/cli/...`, `go vet` on both: clean.
- `go test ./pkg/addon/ -run TestNewAsyncReader`: ok. Covers `git://`, `ssh://` for git, `git://` for gitee, and pins that `https://github.com/...`, the `/tree/<branch>/` form, `https://api.github.com/repos/...` and a `.git` suffix all still build a reader.
- `go test ./references/cli/ -run TestValidateReadableEndpoint`: ok. Seven cases, including that Helm and OSS registries are deliberately not checked on this path.
- `go test ./pkg/utils/ -run TestParse`: ok — `utils.Parse` itself is unchanged.

Two pre-existing failures, present on a clean `origin/master` checkout here and unrelated to this change: `pkg/addon` `TestGetAddonData` and the `references/cli` package suite. I verified both fail identically without this patch applied.

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dchaudhari7177:
Pushed `07df935`: both commits now carry a DCO sign-off, and the three review findings are addressed.

- **Parsed type, not just nil content (P2).** Confirmed: `utils.Parse` returns content for `oss://`, `file://` and `gitee.com` addresses, so a git registry pointing at any of them passed the nil check and queried owner `""` / repo `""`. The git case now requires `utils.TypeGithub` and the gitee case `utils.TypeGitee`. New rejection cases: `oss://`, `file://`, a `gitee.com` URL on a git registry, and a `github.com` URL on a gitee registry. With the old nil-only guard restored, the `oss://` case fails.
- **Endpoint in the error (P2).** `validateReadableEndpoint` wraps parse failures (for example another HTTP host) as `endpoint "<url>": …`, with a test. The unsupported-endpoint errors already quote the URL, so they're returned as they are.
- **Redundant table cases (P3).** Removed from `TestNewAsyncReader`; the dedicated test is the single source for those.

`go test ./pkg/addon/ -run TestNewAsyncReader` and `go test ./references/cli/ -run TestValidateReadableEndpoint` pass, and gofmt is clean.

Reetika2509:
Thanks for your contribution @dchaudhari7177. Verified the changes locally and it covers all the issues ask. Nice addition that the guard goes further than the issue asked. Rejecting a non-GitHub addrType, not just nil content.

Can you add a test confirming that one invalid registry does not panic or prevent other registries from being processed as mentioned in the issue.

dchaudhari7177:
Thanks for verifying it. Added in `08a6dc6`: `TestListAddonsSkipsAnUnreadableGitRegistry` in `references/cli/addon_registry_endpoint_test.go`.

It uses a fake client and writes two registries through the data store, so the endpoint is stored without the add-time check, as an older or hand-edited ConfigMap would be:

- `a-broken-git`: `--type git` with `git://127.0.0.1:9418/poc.git`, the endpoint from the issue. Registries are listed by name, so this one is read first.
- `b-healthy-helm`: a Helm registry served by `httptest` with a one-addon `index.yaml`.

Two subtests:

1. `listAddons(ctx, clt, "")` does not panic, returns no error, and still lists `fluxcd` from `b-healthy-helm`. The broken registry is skipped.
2. `listAddons(ctx, clt, "a-broken-git")` does not panic and returns `ErrUnsupportedGitEndpoint`.

Red before green: with only the `addrType != utils.TypeGithub || content == nil` guard removed from `NewAsyncReader`, both subtests fail with `runtime error: invalid memory address or nil pointer dereference`, the panic from the issue. With the guard, both pass. `go test ./pkg/addon/ -run TestNewAsyncReader` still passes and gofmt is clean.

dchaudhari7177:
Thanks for the retest @Reetika2509. It came back with the same single failure,
and I don't think it belongs to this PR:

```
[FAIL] Test application of the specified definition version [BeforeEach]
       Test tries to deploy component which has both spec.version and
       revision name annotation
Ran 174 of 174 Specs -- 173 Passed | 1 Failed
```

That same spec fails on `master` with no PR involved — run
[35375193979](https://github.com/kubevela/kubevela/actions/runs/35375193979)
(E2E Test, 2026-09-18) fails on it, plus a second one (`Application Normal
tests / test app with custom status fields`). So the branch is actually one
failure *better* than master right now.

It's also outside this PR's reach: the diff only touches
`pkg/addon/source.go` and `references/cli/addon-registry.go` (addon registry
endpoint validation), with no path to definition-version resolution or
revision annotations. The other 27 checks pass, including unit tests.

Happy to rebase onto master once the E2E flake is sorted if that makes the
signal cleaner — just say the word.

jerrinfrancis:
@dchaudhari7177 there has been few other merges to master. Could you please resolve the conflicts.

dchaudhari7177:
@jerrinfrancis done. I merged `master` into the branch, and it's signed off for DCO.

There were two conflicts, both with the `WithRef` / `ReaderOption` change:
- In `NewAsyncReader`, the git and gitee cases now reject an unusable endpoint first and then apply `cfg.ref`, so an override only ever lands on parsed GitHub or Gitee content.
- In `source_test.go`, both sides added new test functions, and I kept all of them.

Locally with go 1.26: `TestNewAsyncReader*`, `TestWithRefOverridesTheRegistryRevision`, `TestBuildReaderForwardsOptions`, and the three `references/cli` endpoint tests all pass. `go vet ./pkg/addon/ ./references/cli/` is clean.

Issue #6947, opened 2025-10-16
Enhance Application Controller with MDC-Style Logging for Better Traceability and Monitoring


roguepikachu:
# Enhance Application Controller with MDC-Style Logging for Better Traceability and Monitoring

## Summary

Implement comprehensive Mapped Diagnostic Context (MDC) style logging in the Application controller to improve traceability, debugging, and observability across the entire application reconciliation lifecycle.

## Current State

The Application controller currently uses a hybrid logging approach:

**Logging Framework:**
- Primary: `github.com/kubevela/pkg/monitor/context` (monitorContext)
- Legacy: `k8s.io/klog/v2` direct calls
- Location: `pkg/controller/core.oam.dev/v1beta1/application/`

**Current Context Tracking (from `application_controller.go`):**
```go
logCtx := monitorContext.NewTraceContext(ctx, "").AddTag("application", req.String(), "controller", "application")
logCtx.AddTag("resource_version", app.ResourceVersion)
logCtx.AddTag("generation", app.Generation)
logCtx.AddTag("publish_version", app.GetAnnotations()[oam.AnnotationPublishVersion])
```

**Issues Identified:**

1. **Inconsistent Logging Patterns**: Mix of structured (monitorContext) and unstructured (klog) logging
   - Structured example: `logCtx.Info("Successfully prepare current app revision", "revisionName", handler.currentAppRev.Name)`
   - Unstructured example: `klog.Infof("garbage collecting application revisions for application %s/%s", app.Namespace, app.Name)`

2. **Missing Correlation IDs**: No trace/correlation ID propagated across the entire reconciliation flow
   - Hard to correlate logs across workflow execution, policy application, and resource dispatch
   - Difficult to trace a single reconciliation request through multiple components

3. **Incomplete Context Propagation**: Context not consistently passed through all sub-operations
   - Files like `apply.go`, `dispatcher.go`, `generator.go`, `revision.go` don't always receive full context
   - Context is forked but not always enriched with operation-specific metadata

4. **Limited Observability**: Missing key contextual fields
   - No request ID or operation ID
   - Reconciliation trigger reason not logged
   - Parent/child operation relationships not tracked
   - No consistent fields for log aggregation tools (Loki, Elasticsearch, CloudWatch)

### 1. Standardize MDC Context Fields

Add standard fields to all log statements throughout the reconciliation lifecycle:

```go
type ApplicationLogContext struct {
    // Request Identification
[...]

PR #7370, opened 2026-09-09
Docs: correct the unknown-flag claim in the IDE debugging guide

### LINKED ISSUE #7369: [Bug] docs/development/ide-debugging.md documents the opposite of the controller's unknown-flag behavior

- [x] I have searched the existing issues

### Affected area

Other / not sure

### KubeVela version

master at e6bbce3f

Na, This is flag parsing plus a docs claim, reproduced without a cluster

### Describe the bug

`docs/development/ide-debugging.md`, added in #7354, tells contributors that an unregistered controller flag is rejected at startup:

> There's no `--webhook-timeout` flag on the controller side, don't add one, `pflag` rejects unregistered flags and the process won't start at all.

The controller does the opposite. `NewCoreCommand` explicitly opts out of that check:

```go
// cmd/core/app/server.go:86-89
FParseErrWhitelist: cobra.FParseErrWhitelist{
    // Allow unknown flags for backward-compatibility.
    UnknownFlags: true,
},
```

An unregistered flag is accepted, ignored, and never reported.

The doc warns about the harmless failure and misses the costly one. A rejected flag tells you immediately. A silently ignored one does not: the controller starts, looks healthy, and behaves as though you never passed it. This lands on the IDE debugging page specifically, where the whole workflow is "put flags in a launch configuration and press F5", so a mistyped flag there is exactly the case that burns an afternoon.

### To reproduce

```markdown
Checked against `master` at `e6bbce3f`, the head at the time of writing.

Drop one letter from `--use-webhook`, which the controller logs at startup:

$ go build -o vela-core ./cmd/core

$ ./vela-core --use-webhook=true 2>&1 | grep "Starting KubeVela"
"Starting KubeVela core controller" context="initialization" leaderElection=false webhookEnabled=true

$ ./vela-core --use-webook=true 2>&1 | grep "Starting KubeVela"
"Starting KubeVela core controller" context="initialization" leaderElection=false webhookEnabled=false

The webhook server never starts, and nothing says why.

For contrast, a bad *value* on a real flag is rejected loudly, which is presumably the behavior the doc had in mind:

$ ./vela-core --webhook-port=notanumber
Error: invalid argument "notanumber" for "--webhook-port" flag: strconv.ParseInt: parsing "notanumber": invalid syntax
```

### Expected behavior

The paragraph should describe what actually happens: unregistered flags are silently accepted and ignored, so a flag should be checked against `vela-core --help` rather than trusted to fail loudli
`--webhook-timeout` genuinely does not exist, and the webhook call timeout genuinely is `timeoutSeconds` on the webhook configuration. Only the stated reason is inverted

### Screenshots or logs

_No response_

### Additional context

Whether `UnknownFlags: true` should stay is a separate question about the controller's behavior rather than the docs, and I'll file it separately. This issue is only about the documented claim, which is wrong under either outcome: if the whitelist stays, the docs should describe silent acceptance; if it is removed, the docs should be updated after the behavior actually changes

- [The claim in ide-debugging.md](https://github.com/kubevela/kubevela/blob/e6bbce3fcc241a4857731553e6dc50ba3d88922a/docs/development/ide-debugging.md#L77-L78)
- [`NewCoreCommand` whitelists unknown flags](https://github.com/kubevela/kubevela/blob/e6bbce3fcc241a4857731553e6dc50ba3d88922a/cmd/core/app/server.go#L86-L89)

sakirr05:

copilot:all

`ide-debugging.md` says pflag rejects unregistered flags and the process won't start. It's the other way round: `NewCoreCommand` sets `FParseErrWhitelist{UnknownFlags: true}`, so an unknown flag is accepted and silently ignored
Corrects the reason and notes the case that actually bites, a typo'd flag name
getting swallowed. The rest of the paragraph was already right and is unchanged

Fixes #7369 

I have:

Docs only. Checked against a build from this branch: `--use-webhook=true` logs `webhookEnabled=true`, `--use-webook=true` logs `webhookEnabled=false` and the webhook server never starts, with nothing in the output naming the flag. A bad value on a real flag (`--webhook-port=notanumber`) still errors, so it's only the name that fails silently.

Skipped `make reviewable`, nothing in it applies to markdown

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/ide-debugging.md
+++ b/docs/development/ide-debugging.md
@@ -74,11 +74,23 @@ common case for day-to-day reconciler work. The second additionally starts
 the webhook server on `:9445`; see [`webhook-debugging.md`](./webhook-debugging.md)
 for the certificate and cluster-side setup it needs.
 
-There's no `--webhook-timeout` flag on the controller side, don't add one,
-`pflag` rejects unregistered flags and the process won't start at all. The
-webhook call's timeout is a cluster-side setting: `timeoutSeconds` on each
-rule in the `ValidatingWebhookConfiguration`/`MutatingWebhookConfiguration`
-objects that `hack/debug-webhook-setup.sh` creates.
+There's no `--webhook-timeout` flag on the controller side, and adding one
+fails quietly rather than loudly. `NewCoreCommand` sets
+`FParseErrWhitelist{UnknownFlags: true}` (`cmd/core/app/server.go`) for
+backward compatibility, so an unregistered flag is accepted, ignored, and
+never reported. The controller starts normally and the setting you thought
+you applied does nothing.
+
+That applies to typos too, which is the case worth watching for here: drop a
+letter and `--use-webook=true` leaves the webhook server switched off, with
+no error to point at. Check a flag against `vela-core --help` before relying
+on it. A bad *value* on a real flag is still rejected loudly, so it's only
+the flag name that fails silently.
+
+The webhook call's timeout is a cluster-side setting anyway: `timeoutSeconds`
+on each rule in the
+`ValidatingWebhookConfiguration`/`MutatingWebhookConfiguration` objects that
+`hack/debug-webhook-setup.sh` creates.
 
 ## IntelliJ IDEA / GoLand
 


roguepikachu (approved):

sakirr05:
@roguepikachu found a second thing worth a separate issue

```go
// cmd/core/app/server.go
FParseErrWhitelist: cobra.FParseErrWhitelist{
    // Allow unknown flags for backward-compatibility.
    UnknownFlags: true,
},
```

This is why the typo in # 7369 fails silently in the first place. An unknown flag is accepted and dropped with nothing logged, which reads fine as backward compat for a flag we removed, but it also swallows every typo the same way 

my approach- keep the whitelist (don't want old deployments crash-looping on upgrade), but log each ignored flag once at warning level, something like

ignoring unrecognized flag" flag"=--use-webhook"

Does that approach sound right, or would you rather drop the whitelist outright and force a real deprecation cycle for removed flags?

briankane (approved):

roguepikachu:
I guess it could provide some additional observability, but I don’t see much value beyond that. I’d prefer improving the overall logging of the controller startup instead, as mentioned in https://github.com/kubevela/kubevela/issues/6947

PR #7371, opened 2026-09-09
Feat(sources): SourceDefinition

briankane:

Adds a CRD for fetching external data, a CEL expression syntax for reading it from
an Application, and admission validation that type-checks the two against each
other before the Application is stored.

```yaml
kind: Application
metadata:
  annotations: {app.oam.dev/cel-expressions: "true"}
spec:
  sources:
    - name: registry
      type: configmap-local
      properties: {name: platform-config}
  components:
    - name: web
      type: webservice
      properties:
        image: '$(source.registry.data["image"]):v1'
```

Design and rationale are in
[`design/vela-core/keps/2.16-source-definition`](design/vela-core/keps/2.16-source-definition/README.md),
which this PR also updates to match what shipped.

#### The definition

A `SourceDefinition` is a CUE template with four authored blocks and one generated:

| Block | Is |
|---|---|
| `schema` | the fields consumers may read, and their types. The public contract |
| `parameter` | what an Application supplies when binding it |
| `output` | what the fetch returns; unified against `schema` at resolution |
| `storage` | `storageTTL` and `onStaleFailure`. Optional |
| `$internal` | **generated**: the cache key and the context it varies by |

`schema` is lifted out of the template on its own so admission can type-check reads
without evaluating the fetch.

#### Resolution

Two compiles per binding:

1. `storage` and `$internal` are evaluated with providers disabled. This is pure
   interpolation over context and parameters, so it yields the cache key without
   performing the fetch the cache exists to avoid.
2. On a cache miss the template is compiled again with providers live, the fetch
   runs, and `output` is validated against `schema`.

Cache identity is `<readable prefix>-<hash>`. The hash covers the template
fingerprint, the binding's properties and the context values the template reads;
the prefix is cosmetic and exists for grepping. Because the template is in the
hash, editing a definition stops it serving values its previous version produced.

Two layers sit behind it: a process-level LRU, and a `Config` object per entry so
the value survives a controller restart and is shared by every Application
resolving to the same key. A GC sweep collects stale entries and orphaned
ConfigTemplates.

#### Expressions

`$( ... )` in a string property is parsed into literal and expression fragments and
compiled to CEL. `$$(` is the escape for a literal `$(`.

Two roots are readable: `source.<binding>.<path>` and `context.<field>`. A property
that is nothing but an expression keeps the expression's type, so an int stays an
int; one with text around it yields a string.

Substitution is wired into `workloadDef.Complete`, `traitDef.Complete`, the appfile
parse and the workflow-step generator, covering components, traits, workflow steps,
source chaining and resource-rendering policies. Built-in and Application-scoped
policies read `context` but not `source`: they are consumed before or outside any
render, so there is no resolver to reach a source through.

Which fields `context` offers per surface is declared once in
`pkg/definition/propexpr/context.cue` and read by Go, so admission and render cannot
disagree about what a field means or whether it exists.

#### Validation

Admission runs in two phases and does not attempt the second while the first has
findings. Soundness is cheap and side-effect free: expressions parse, their CEL
types fit the parameters they feed, every read path is declared in the source's
schema, an optional field feeding a required parameter carries a guard, bindings are
unique and referenced only after declaration, and the surface may read a source at
all. Rendering then generates the appfile and resolves for real.

A `SourceDefinition` is validated when applied: the template compiles, `schema` and
`storage` are present, and `$internal` is re-derived and compared, so a hand-edited
or stale key is refused rather than silently changing which Applications share an
entry.

#### Capability

Sources compile against a dedicated `SourceCompiler` carrying `kube`, `http`,
`base64`, `registry`, `velaconfig` and `cue`. 

#### Off by default

| Gate | Default | Controls |
|---|---|---|
| `EnableCelExpressions` | off | the expression mechanism |
| `RequireCelExpressionOptIn` | on | per-Application annotation as well |
| `EnableSourceAutoUpdate` | off | re-dispatch on a changed value |

The delimiter is Kubernetes' own: `$(VAR_NAME)` is the documented syntax for a
dependent environment variable and is ordinary in `env`, `command` and `args`. With
the pass running such a value is read as an expression and refused, so it does not
run until an operator asks, and then only for Applications carrying
`app.oam.dev/cel-expressions: "true"` until they drop that too. Three states, so a
platform can adopt it over time.

Named for CEL rather than sources because `$(context.appName)` needs no
SourceDefinition.

#### What ships

- The CRD, its controller, revisions and admission
- Nine sources: `configmap-local`, `git-file`, `http-get`,
  `vela-config`, `vela-app`, `vela-component`, `vela-addon`, `vela-env`
- `vela status --sources`, `vela def compat` support, a generated Go SDK
- A startup check naming any required CRD the cluster does not serve

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7371?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jerrinfrancis (approved):

anoop2811:
@briankane : Looks like there are conflicts to resolve

briankane:
@jerrinfrancis has been doing extensive testing of this. We are happy to merge

PR #7379, opened 2026-09-17
Fix(defkit): compile generated CUE during render

### LINKED ISSUE #7291: [Enhancement] Defkit: add generated-CUE validation and evaluation helpers

The in-process testing surface: `TestContext`, `Render`, `RenderAll` and `ToCue`.

`Render` walks the Go builder. It never compiles the CUE that `ToCue` produces.

It covers a fair amount: parameter values and defaults, the common context fields, `Set`, `SetIf`, `IfBlock` and several collection transforms. Raw header blocks and validators sit outside it, and a value it does not recognise comes back as a Go builder object rather than being evaluated.

- [`Render` runs the template and calls `renderResource`](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/render.go#L19-L66)
- [`resolveValue` returns unknown values unchanged](https://github.com/kubevela/kubevela/blob/410c2a832163ad598ecefdb8718dbc2b2bba4de0/pkg/definition/defkit/render.go#L164-L213)

That leaves two blind spots. You can assert on the string `ToCue` returns and never notice it is malformed, and anything whose behaviour only exists in CUE gets no coverage at all.

We hit the first one. A dotted path passed to `ItemBuilder.Set` generated an invalid label, the unit tests stayed green, and it surfaced on a cluster.

Two helpers, rather than growing the Go interpreter one case at a time:

```go
err := defkit.ValidateGeneratedCUE(component)

result, err := component.EvaluateCUE(
    defkit.TestContext().WithName("example").WithParam("replicas", replicas),
)
```

The first compiles the generated definition and reports syntax and type errors with positions. The second injects fixtures, evaluates with the CUE Go API, and hands back the concrete outputs plus any validation errors.

The compile-only one is worth doing first. It is much smaller, and it would already catch the class of bug above. `Render` can stay as it is for fast unit tests.

**Additional context**

Evaluating the real generated CUE avoids maintaining a second implementation of template semantics inside `render.go`.

Aksaykanthan:
Fixes #7291

### Description

This PR is focused specifically on `ComponentDefinition.Render` and `RenderAll`. They now compile the generated `ToCue()` output by default before performing the existing fast Go-side rendering. This catches invalid generated CUE in ordinary Render tests while preserving an explicit opt-out for tests that intentionally exercise only the partial Go renderer.

The compile check is self-contained in the Render path and does not add or duplicate the separate `EvaluateCUE` API. The generator label fix ensures nested and conditional fields such as `production-like` emit valid CUE.

### Example

```go
// Default: generated CUE is compiled before rendering.
rendered := component.Render(defkit.TestContext())

// Opt out only for an intentionally Go-only test case.
rendered := component.Render(
    defkit.TestContext().WithParam("ports", "not-an-array"),
    defkit.SkipCompile(),
)
```

The default compile check panics on invalid generated CUE, matching the existing Render API's non-error return shape. The variadic option keeps existing Render and RenderAll call sites source-compatible.


briankane:
Hey @Aksaykanthan 
DCO checks are failing. Can you sign the commit and repush? Thanks

PR #7383, opened 2026-09-18
Fix(helmchart): keep empty maps and lists in rendered manifests

briankane:

copilot:all

cleanResource() dropped empty maps and lists along with nil values. That removed CRD `subresources: {status: {}}` and `emptyDir: {}` volumes, so charts like cert-manager installed without a status subresource.

Only strip nil values now.

Fixes #

I have:

- Rewrote the two cleanResource unit tests that expected empty maps and lists to be removed, so they now expect them to be kept.
- Added unit tests for:
  - an empty map inside a map (subresources: {status: {}})
  - a map whose only fields were nil (emptyDir: {sizeLimit: null}), which should come out as {}
  - empty lists

<a href="https://cubic.dev/pr/kubevela/kubevela/pull/7383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


roguepikachu (approved):
I reproduced this locally on k3d. Packed a small chart with a CRD that has subresources.status: {} and a Deployment with emptyDir: {}, then installed it through a helmchart Application.

On master the CRD came up with no status subresource at all. emptyDir also got stripped from last-applied (the live Deployment still shows it because the apiserver fills it back in). Rebuilt the controller from this branch, same app, and both fields were kept.

```yaml
kind: Application
metadata:
  name: empty-maps-repro
spec:
  components:
    - name: empty-maps
      type: helmchart
      properties:
        chart:
          source: http://<host>/empty-maps-repro-0.1.0.tgz
        release:
          name: empty-maps
          namespace: default
        options:
          includeCRDs: true
```

LGTM.

